PR #64, opened 2022-03-31
feat: add snippet promise support

dnshi:
# Description

Add snippet promise support

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/analytics-browser/src/utils/snippet-helper.ts
+++ b/packages/analytics-browser/src/utils/snippet-helper.ts
@@ -18,7 +18,7 @@ export const convertProxyObjectToRealObject = <T>(instance: T, queue: QueueProxy
     const fn = instance && instance[name as keyof T];
     if (typeof fn === 'function') {
       const result = fn.apply(instance, args) as AmplitudeReturn<Result>;
-      if (typeof resolve === 'function') {
+      if (typeof resolve === 'function' && result) {
         resolve(result.promise);
       }
     }
--- a/packages/analytics-browser/src/utils/snippet-helper.ts
+++ b/packages/analytics-browser/src/utils/snippet-helper.ts
@@ -1,4 +1,5 @@
 import { QueueProxy, InstanceProxy } from '../typings/browser-snippet';
+import { AmplitudeReturn, Result } from '@amplitude/analytics-types';
 
 /**
  * Applies the proxied functions on the proxied amplitude snippet to an instance of the real object.
@@ -13,10 +14,13 @@ export const runQueuedFunctions = (instance: object, queue: QueueProxy) => {
  */
 export const convertProxyObjectToRealObject = <T>(instance: T, queue: QueueProxy): T => {
   for (let i = 0; i < queue.length; i++) {
-    const [functionName, ...args] = queue[i];
-    const fn = instance && instance[functionName as keyof T];
+    const { name, args, resolve } = queue[i];
+    const fn = instance && instance[name as keyof T];
     if (typeof fn === 'function') {
-      fn.apply(instance, args);
+      const result = fn.apply(instance, args) as AmplitudeReturn<Result>;
+      if (typeof resolve === 'function') {
+        resolve(result.promise);
+      }
     }
   }
   return instance;
--- a/packages/analytics-browser/src/utils/snippet-helper.ts
+++ b/packages/analytics-browser/src/utils/snippet-helper.ts
@@ -18,8 +18,8 @@ export const convertProxyObjectToRealObject = <T>(instance: T, queue: QueueProxy
     const fn = instance && instance[name as keyof T];
     if (typeof fn === 'function') {
       const result = fn.apply(instance, args) as AmplitudeReturn<Result>;
-      if (typeof resolve === 'function' && result) {
-        resolve(result.promise);
+      if (typeof resolve === 'function') {
+        resolve(result?.promise);
       }
     }
   }
--- a/packages/analytics-browser/template/browser-snippet.template.js
+++ b/packages/analytics-browser/template/browser-snippet.template.js
@@ -95,7 +95,7 @@ const snippet = (integrity, version) => `
                 args: Array.prototype.slice.call(arguments, 0),
                 resolve: resolve,
               });
-            });
+            }),
           }
           if (isPromise) return result;
         };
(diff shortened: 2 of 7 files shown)


packages/analytics-browser/template/browser-snippet.template.js
```diff
@@ -80,14 +74,37 @@ const snippet = (integrity, version) => `
...
         instance[fn] = function () {
-          instance._q.push([fn].concat(Array.prototype.slice.call(arguments, 0)));
+          var result = {
+            promise: new Promise((resolve) => {
+              instance._q.push({
+                name: fn,
+                args: Array.prototype.slice.call(arguments, 0),
+                resolve: resolve,
+              });
+            });
```
kevinpagtakhan, line 98:
```suggestion
            }),
```

dnshi: Good catch! My bad. I did not see it because they were same color in the vscode...

packages/analytics-browser/template/browser-snippet.template.js
```diff
@@ -80,14 +74,37 @@ const snippet = (integrity, version) => `
...
+          }
+          if (isPromise) return result;
         };
       }
       for (var k = 0; k < funcs.length; k++) {
-        proxyMain(funcs[k]);
+        proxyMain(funcs[k], false);
+      }
+      for (var l = 0; l < funcsWithPromise.length; l++) {
+        proxyMain(funcs[l], true);
```
kevinpagtakhan, line 107:
```suggestion
        proxyMain(funcsWithPromise[l], true);
```

dnshi: Good catch again!

packages/analytics-browser/src/utils/snippet-helper.ts
```diff
@@ -13,10 +14,13 @@ export const runQueuedFunctions = (instance: object, queue: QueueProxy) => {
...
   for (let i = 0; i < queue.length; i++) {
-    const [functionName, ...args] = queue[i];
-    const fn = instance && instance[functionName as keyof T];
+    const { name, args, resolve } = queue[i];
+    const fn = instance && instance[name as keyof T];
     if (typeof fn === 'function') {
-      fn.apply(instance, args);
+      const result = fn.apply(instance, args) as AmplitudeReturn<Result>;
+      if (typeof resolve === 'function') {
+        resolve(result.promise);
```
kevinpagtakhan, line 22: I'm seeing an error that points to this line:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'promise')
```

packages/analytics-browser/src/utils/snippet-helper.ts
```diff
@@ -13,10 +14,13 @@ export const runQueuedFunctions = (instance: object, queue: QueueProxy) => {
...
 export const convertProxyObjectToRealObject = <T>(instance: T, queue: QueueProxy): T => {
   for (let i = 0; i < queue.length; i++) {
-    const [functionName, ...args] = queue[i];
-    const fn = instance && instance[functionName as keyof T];
+    const { name, args, resolve } = queue[i];
+    const fn = instance && instance[name as keyof T];
     if (typeof fn === 'function') {
-      fn.apply(instance, args);
+      const result = fn.apply(instance, args) as AmplitudeReturn<Result>;
+      if (typeof resolve === 'function') {
```
kevinpagtakhan, line 21: looks like this should solve the error below

```suggestion
      if (typeof resolve === 'function' && result) {
```

kevinpagtakhan (commented):
found a couple of small issues but looking good otherwise 🙌

packages/analytics-browser/src/utils/snippet-helper.ts
dnshi: You are right. This solves the error.

kevinpagtakhan: Only concern here is some promises will remain pending, and never gets resolved

packages/analytics-browser/src/utils/snippet-helper.ts
dnshi: Right. We need to check if the result is valid.

packages/analytics-browser/src/utils/snippet-helper.ts
dnshi: Let me update it to resolve the pending issue.

kelvin-lu (commented):
Code overall lgtm.

How do `getSessionId`, `getUserId`, and `getDeviceId` work in this snippet? From what I've read in this pr, it seems like they return nothing until the full SDK loads? I think you could consider modifying their signatures to `Promise<OriginalType>` so that the snippet plays a little better (at the cost of promise-ifying their signature even after the SDK loads).

On an aside, I think the [explicit module boundary](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) rule may help with readability of some files :) and has a minor benefit of having the TS compile a little bit faster.

packages/analytics-browser/template/browser-snippet.template.js
```diff
@@ -80,14 +74,37 @@ const snippet = (integrity, version) => `
...
+          var result = {
+            promise: new Promise((resolve) => {
+              instance._q.push({
+                name: fn,
+                args: Array.prototype.slice.call(arguments, 0),
+                resolve: resolve,
+              });
+            }),
+          }
+          if (isPromise) return result;
```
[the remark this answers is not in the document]
dnshi: Because those methods return `{ promise: Promise }`, here the result is `{ promise: Promise }` as well.

dnshi:
> Code overall lgtm.
> 
> How do `getSessionId`, `getUserId`, and `getDeviceId` work in this snippet? From what I've read in this pr, it seems like they return nothing until the full SDK loads? I think you could consider modifying their signatures to `Promise<OriginalType>` so that the snippet plays a little better (at the cost of promise-ifying their signature even after the SDK loads). I think you could make the argument that **all** of the functions could be promises to allow the end-user to know when the function "actually" executes.
> 
> On an aside, I think the [explicit module boundary](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) rule may help with readability of some files :) and has a minor benefit of having the TS compile a little bit faster.

I tested out the JS SDK, and for any get method, it just returns undefined before the script is loaded. Right now I think we can keep current behavior and discuss later if we want to make all methods returning promise type.

PR #76, opened 2022-04-08
chore: edit github pr template

kevinpagtakhan:


.github/pull_request_template.md
```diff
@@ -1,14 +1,14 @@
-# Description
+<!---
+Thanks for contributing to the Amplitude JavaScript SDK! 🎉
```
yuhao900914, line 2: The same question about the pr template. 
Not limited to javascript sdk.

kevinpagtakhan: Yeah since there is only one, i think we should name this with something more generic

yuhao900914: Agree.

kevinpagtakhan: Alright i just said "Amplitude TypeScript repository"

dnshi (approved):

PR #80, opened 2022-04-09
build: setup integration testing

kevinpagtakhan:
### Summary

* Setup up new private package in workspace
  * Private packages are excluded from publish cycle
  * Integration tests are in separate package to ensure testing is black box
* Use [nock](https://www.npmjs.com/package/nock) to intercept http requests
* Covers the following scenario
  * case: event with custom config
  * case: event with event options
  * case: 400 error response
  * case: 413 error response
  * case: 429 error response
  * case: 500 error response
  * case: exhausting max retries
  * case: missing api key
  * case: client opt out

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,8 @@
     "lint:staged": "lint-staged",
     "postinstall": "husky install",
     "test": "lerna run test --stream",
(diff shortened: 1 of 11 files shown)


qingzhuozhen (approved):

justin-fiedler (approved):
🔥

PR #89, opened 2022-05-06
refactor: class design for client and config

kevinpagtakhan:
### Checklist


packages/analytics-core/src/core-client.ts
```diff
@@ -9,81 +8,84 @@ import {
 } from './utils/event-builder';
 import { deregister, push, register } from './timeline';
 import { buildResult } from './utils/result-builder';
+export class AmplitudeCore<T extends Config> {
```
dnshi, line 11: Interesting idea. Now both browser and NodeJS can extend the core class. 👍

dnshi (approved):
LGTM

Issue #88, opened 2022-05-02
Unable to set event-level groups


gesposito:
## Expected Behavior
[EventOptions](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-types/src/base-event.ts#L10) parameter in `logEvent` allows for  a [groups](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-types/src/base-event.ts#L7) key

## Current Behavior
`groups` (event-level groups) cannot be set with `logEvent`:
```ts
logEvent(eventType, eventProperties, { groups?: { [key: string]: any } })
```

## Steps to Reproduce
I'm trying to match the `amplitude-js` [logEventWithGroups](https://github.com/amplitude/Amplitude-JavaScript/blob/main/src/amplitude-client.js#L1611) API that has support for
```ts
logEventWithGroups(eventType, eventProperties, groups)
```
https://github.com/amplitude/Amplitude-JavaScript/blob/main/src/amplitude-client.js#L1406

## Environment
- JS SDK Version: 0.3.1
- Installation Method: npm

kevinpagtakhan:
@gesposito thanks for submitting an issue! Thanks for your patience as we continuously iterate on this new SDK. This is unintended behavior and we will work on addressing this issue.

PR #90, opened 2022-05-07
fix: allow event level groups tracking

kevinpagtakhan:
### Summary

Fixes https://github.com/amplitude/Amplitude-TypeScript/issues/88

### Checklist

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -34,7 +34,6 @@ describe('integration', () => {
         os_name: 'WebKit',
         os_version: '537.36',
         device_manufacturer: undefined,
-        device_model: undefined,
         language: 'en-US',
         ip: '$remote',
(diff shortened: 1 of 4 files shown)


packages/analytics-browser-test/test/index.test.ts
```diff
@@ -148,7 +190,6 @@ describe('integration', () => {
         os_name: 'WebKit',
         os_version: '537.36',
         device_manufacturer: undefined,
-        device_model: undefined,
```
yuhao900914, line 151: Why device_model got removed from this file?

kevinpagtakhan: the expected value does not have it explicitly and jest prefers that it's not specified on the received value. but regardless if i remove or not, the assertion is the "same" so this is mostly clean up.

yuhao900914 (approved):

PR #92, opened 2022-05-10
feat: introduce NodeJS package

dnshi:
### Summary

This PR is to introduce Node JS SDK package. It shares and reuses configs, components and plugins from the analytics core package. The changes include the following parts:
- NodeJS APIs
- Context plugin
- Http/Https
- Memory storage
- Unit tests

### Checklist


packages/analytics-node/src/config.ts
```diff
@@ -0,0 +1,40 @@
...
+  return eventsStorage;
+};
+
+export const createDeviceId = () => {
+  return UUID();
+};
+
+export const createSessionId = () => {
+  return Date.now();
+};
```
kevinpagtakhan, line 36: same question for session id?

packages/analytics-node/src/config.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const useNodeConfig = (apiKey: string, userId?: string, overrides?: NodeOptions): INodeConfig => {
+  return new NodeConfig(apiKey, userId, overrides);
+};
+
+export const createEventsStorage = () => {
+  const eventsStorage = new MemoryStorage<Event[]>();
+  return eventsStorage;
+};
```
kevinpagtakhan, line 28: i'd say we can skip createEventStorage and create Transport since they are one liner, then we can just specify default values and spread them in NodeConfig's super call

packages/analytics-node/src/plugins/context.ts
```diff
@@ -0,0 +1,38 @@
...
+    return Promise.resolve(undefined);
+  }
+
+  execute(context: Event): Promise<Event> {
+    return new Promise((resolve) => {
+      const time = new Date().getTime();
+
+      const contextEvent: Event = {
+        user_id: this.config.userId,
+        device_id: this.config.deviceId,
```
kevinpagtakhan, line 27: i believe these values should come from client device since events processed by node sdk come from different user and device

packages/analytics-node/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.3.1';
```
kevinpagtakhan, line 1: version here should be similar with of package.json

packages/analytics-types/src/config.ts
```diff
@@ -47,6 +47,8 @@ export interface BrowserConfig extends Config {
   trackingOptions: TrackingOptions;
 }
 
+export type NodeConfig = Config;
```
kevinpagtakhan, line 50: do we need this then?

packages/analytics-node/src/node-client.ts
```diff
@@ -0,0 +1,226 @@
+import { AmplitudeCore, Destination } from '@amplitude/analytics-core';
+import { NodeConfig, NodeOptions } from '@amplitude/analytics-types';
+import { Context } from './plugins/context';
+import { useNodeConfig } from './config';
+
+export class AmplitudeNode extends AmplitudeCore<NodeConfig> {
+  async init(apiKey: string, userId?: string, options?: NodeOptions) {
```
kevinpagtakhan, line 7: `userId` may not be a valuable argument here in init since it may be different for each event

packages/analytics-node/src/node-client.ts
```diff
@@ -0,0 +1,226 @@
...
+    this.config.deviceId = deviceId;
+  }
+
+  getSessionId() {
+    return this.config.sessionId;
+  }
+
+  setSessionId(sessionId: number) {
+    this.config.sessionId = sessionId;
+  }
```
kevinpagtakhan, line 43: are these apis valuable in node sdk?

packages/analytics-types/src/config.ts
kevinpagtakhan: actually can we base the config off of https://github.com/amplitude/Amplitude-Python/blob/main/src/amplitude/config.py#L15 and https://github.com/amplitude/Amplitude-Node/blob/main/packages/types/src/options.ts#L10. some config may not be needed and should just be moved to browser config

packages/analytics-node/src/config.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const createDeviceId = () => {
+  return UUID();
+};
```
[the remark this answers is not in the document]
dnshi: We can do a conditional check and if device id is not provided, then generate it.

packages/analytics-node/src/config.ts
dnshi: Yep, do a conditional check and generate session id if it is not provided.

packages/analytics-node/src/node-client.ts
dnshi: Sure, will move it.

packages/analytics-node/src/config.ts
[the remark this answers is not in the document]
kevinpagtakhan: probably not ideal since either we generate one for all events from different user/device combo, or a unique one for each event from the same user/device combo

packages/analytics-node/src/node-client.ts
dnshi: Just check other SDKs, these APIs can be removed.

kevinpagtakhan: generally we want it to look/feel like browser sdk but overall behavior should be pattered with other server-side sdks

kevinpagtakhan (changes requested):

packages/analytics-node/src/node-client.ts
```diff
@@ -0,0 +1,142 @@
+import { AmplitudeCore, Destination } from '@amplitude/analytics-core';
```
kevinpagtakhan, line 1: python has [flush](https://developers.amplitude.com/docs/python-beta#flush) and [shutdown](https://developers.amplitude.com/docs/python-beta#shutdown)

dnshi: Added flush, but shutdown might not work for node sdk

kevinpagtakhan: @bohan-amplitude any thoughts here?

packages/analytics-core/src/timeline.ts
```diff
@@ -93,3 +93,13 @@ export const apply = async () => {
 
   return;
 };
+
+export const flush = async (config: Config) => {
```
kevinpagtakhan, line 97: my understanding of flush is that it is typically called before server exits, and meant to send all unset events. i think this would only send events that are in the destination plugin's queue, which may not include all unsent events. Other unsent events are in [timeline queue](https://github.com/amplitude/Amplitude-TypeScript/blob/92325bc34cf0143f5f33ec4b0afd3e2d148c3d38/packages/analytics-core/src/timeline.ts#L15) (pending to be applied with before and enrichment plugins) and [delayed retryable events queue](https://github.com/amplitude/Amplitude-TypeScript/blob/f353367b8b264f86b6ea15b15f30385f8d5b8ad5/packages/analytics-core/src/plugins/destination.ts#L80) in amplitude destination plugin.

packages/analytics-node/src/plugins/context.ts
```diff
@@ -0,0 +1,35 @@
...
+export class Context implements BeforePlugin {
+  name = 'context';
+  type = PluginType.BEFORE as const;
+
+  // this.config is defined in setup() which will always be called first
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  config: NodeConfig;
+  eventId = 0;
+  library = `amplitude-ts/${VERSION}`;
```
qingzhuozhen, line 14: The library name should be different than browser one

kevinpagtakhan: I would suggest `amplitude-node-ts`

dnshi: Good idea. Will update it.

packages/analytics-types/src/plugin.ts
```diff
@@ -27,6 +27,7 @@ export interface DestinationPlugin {
   type: PluginType.DESTINATION;
   setup(config: Config): Promise<undefined>;
   execute(context: Event): Promise<Result>;
+  flush(): Promise<void>;
```
kevinpagtakhan, line 30: should be removed

packages/analytics-types/src/config.ts
```diff
@@ -12,28 +12,25 @@ export enum ServerZone {
...
   apiKey: string;
-  appVersion?: string;
-  deviceId?: string;
   flushIntervalMillis: number;
   flushMaxRetries: number;
   flushQueueSize: number;
   logLevel: LogLevel;
   loggerProvider: Logger;
   minIdLength?: number;
   optOut: boolean;
```
kevinpagtakhan, line 21: To be moved in browser config

packages/analytics-types/src/config.ts
```diff
@@ -12,28 +12,25 @@ export enum ServerZone {
...
-  partnerId?: string;
   plugins: Plugin[];
   saveEvents: boolean;
```
kevinpagtakhan, line 23: this may turn out not to be a node config but we can hold off

kevinpagtakhan (approved):
Providing approval in the condition that unresolved comment are to be address in future PRs. Let's keep future PRs small please.

PR #93, opened 2022-05-10
feat: add event bridge components

kevinpagtakhan:
### Summary

Add event bridge components

### Checklist

### DIFF
--- a/packages/analytics-core/src/event-bridge/event-bridge-channel.ts
+++ b/packages/analytics-core/src/event-bridge/event-bridge-channel.ts
@@ -1,0 +1,33 @@
+import { Event, EventBridgeReceiver } from '@amplitude/analytics-types';
+
+const QUEUE_CAPACITY = 512;
+
+export class EventBridgeChannel {
+  channel: string;
+  queue: Event[] = [];
+  receiver: EventBridgeReceiver | undefined;
+
+  constructor(channel: string) {
+    this.channel = channel;
+  }
+
+  sendEvent(event: Event) {
+    if (!this.receiver) {
+      this.queue = [...this.queue.slice(0, QUEUE_CAPACITY), event];
+      return;
+    }
+    this.receiver.receive(this.channel, event);
+  }
+
+  setReceiver(receiver: EventBridgeReceiver) {
+    if (this.receiver) {
+      return;
+    }
+    this.receiver = receiver;
+    const events = this.queue;
+    this.queue = [];
+    events.forEach((event) => {
+      (this.receiver as EventBridgeReceiver).receive(this.channel, event);
+    });
+  }
+}
--- a/packages/analytics-core/src/event-bridge/event-receiver.ts
+++ b/packages/analytics-core/src/event-bridge/event-receiver.ts
@@ -1,0 +1,24 @@
+import { Config, CoreClient, Event, EventBridgeReceiver } from '@amplitude/analytics-types';
+import { EventChannel } from './event-channel';
+
+export class AnalyticsEventReceiver implements EventBridgeReceiver {
+  constructor(private client: CoreClient<Config>) {}
+
+  receive(_channel: EventChannel, event: Event) {
+    this.client.config.loggerProvider.log(`Receive event from event bridge ${event.event_type}`);
+    this.client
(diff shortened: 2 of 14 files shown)


packages/analytics-core/src/event-bridge/event-bridge-channel.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  sendEvent(event: Event) {
+    if (!this.receiver) {
+      this.queue.push(event);
+      return;
+    }
+    this.receiver.receive(this.channel, event);
+  }
+
+  setReceiver(receiver: EventBridgeReceiver) {
```
yuhao900914, line 20: I know the logic is `Only allow one receiver per channel`. But curious about how can we change the receiver?

packages/analytics-core/test/event-bridge/event-bridge-container.test.ts
```diff
@@ -0,0 +1,39 @@
+import { EventBridgeContainer } from '../../src/event-bridge/event-bridge-container';
+import * as EventBridgeModule from '../../src/event-bridge/event-bridge';
+
+describe('EventBridgeContainer', () => {
+  beforeEach(() => {
+    EventBridgeContainer.instances = {};
+  });
+
+  describe('getInstance', () => {
+    test('should create new instance', () => {
```
yuhao900914, line 10: Feels this test has already been included in the next test case. Might be we can merge those two?

packages/analytics-core/src/event-bridge/event-bridge-channel.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (this.receiver) {
```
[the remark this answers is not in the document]
kevinpagtakhan: This is patterned after kotlin implementation. @qingzhuozhen mind chiming in here?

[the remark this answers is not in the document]
qingzhuozhen: @yuhao900914 What's the use case to change the receiver? For starting we are hoping it would a 1 to 1 for both side.

packages/analytics-core/src/event-bridge/event-receiver.ts
```diff
@@ -0,0 +1,24 @@
+import { Config, CoreClient, Event, EventBridgeReceiver } from '@amplitude/analytics-types';
+import { EventChannel } from './event-channel';
+
+export class AnalyticsEventReceiver implements EventBridgeReceiver {
```
qingzhuozhen, line 4: Is this now used anywhere? Or this is just about the module?

packages/analytics-core/src/event-bridge/event-bridge-channel.ts
```diff
@@ -0,0 +1,31 @@
+import { Event, EventBridgeReceiver } from '@amplitude/analytics-types';
+
+export class EventBridgeChannel {
+  channel: string;
+  queue: Event[] = [];
```
qingzhuozhen, line 5: Wonder if we need a limit in TS as well, just for the normal case and we don't have any receiver, we don't always hold them in memory.

kevinpagtakhan: @qingzhuozhen Good call, I will fix this. What's the limit in Kotlin?

packages/analytics-core/src/event-bridge/event-receiver.ts
kevinpagtakhan: Not used since yet. Since it's not being used yet, I'd prefer not to include it in the bundle yet. Maybe we can include in the lazy loading strategy we previously talked about.

packages/analytics-core/src/event-bridge/event-bridge-channel.ts
kevinpagtakhan: I set the limit to 512, similar to: https://github.com/amplitude/Amplitude-Kotlin/blob/main/event-bridge/src/main/java/com/amplitude/eventbridge/EventBridge.kt#L52

qingzhuozhen (approved):

PR #95, opened 2022-05-11
fix: fix early return if array element is a valid object

bohan-amplitude:
### Summary

fix early return if array element is a valid object

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ lib/
 *.tsbuildinfo
 coverage/
 docs/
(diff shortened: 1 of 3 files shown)


kevinpagtakhan:
@bohan-amplitude mind sharing what the initial issue we are seeing to discover this problem?

bohan-amplitude:
> haring what the initial issue we are seeing to discover this problem?

I was learning typescript and reading source code of this repo

PR #96, opened 2022-05-11
build: limit snippet regeneration when ready to release

kevinpagtakhan:
### Summary

`yarn build` runs twice in deployment, which is fine. the first one is for validation of its current state, the second one is to include the new version (as the upcoming release would be assigned a new version). however, the first occurrence generates a snippet, and we don't want that because lerna will find uncommitted changes. we only want it to run on the second occurrence since we need to include the new version in the final bundle

### Checklist

### DIFF
--- a/packages/analytics-browser/package.json
+++ b/packages/analytics-browser/package.json
@@ -33,7 +33,7 @@
     "publish": "node scripts/publish/upload-to-s3.js",
     "test": "jest",
     "typecheck": "tsc -p ./tsconfig.json",
-    "version": "GENERATE_SNIPPET=true && yarn build",
+    "version": "GENERATE_SNIPPET=true yarn build",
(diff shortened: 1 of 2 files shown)


packages/analytics-browser/package.json
```diff
@@ -33,7 +33,7 @@
     "publish": "node scripts/publish/upload-to-s3.js",
     "test": "jest",
     "typecheck": "tsc -p ./tsconfig.json",
-    "version": "yarn build",
+    "version": "GENERATE_SNIPPET=true yarn build",
```
kevinpagtakhan, line 36: this is the second call which is called via lerna version. the first one is `yarn build` called directly from cd.yml.

haoliu-amp (approved):

PR #117, opened 2022-06-17
feat: add Plan option to config

dnshi:
### Summary

This PR is to add Plan option to config when Amplitude SDK is initialized. 

### Checklist


packages/analytics-core/src/config.ts
```diff
@@ -40,6 +42,7 @@ export class Config implements IConfig {
   loggerProvider: ILogger;
   logLevel: LogLevel;
   minIdLength?: number;
+  plan: Plan | undefined;
```
kevinpagtakhan, line 45: nit
```suggestion
  plan?: Plan;
```

packages/analytics-core/src/config.ts
```diff
@@ -30,6 +31,7 @@ export const getDefaultConfig = () => ({
   serverUrl: AMPLITUDE_SERVER_URL,
   serverZone: ServerZone.US,
   useBatch: false,
+  plan: undefined,
```
kevinpagtakhan, line 34: if not passed, it is undefined by default so no need to specify here

```suggestion
```

packages/analytics-core/src/config.ts
```diff
@@ -58,6 +61,7 @@ export class Config implements IConfig {
     this.loggerProvider = options.loggerProvider || defaultConfig.loggerProvider;
     this.logLevel = options.logLevel ?? defaultConfig.logLevel;
     this.minIdLength = options.minIdLength;
+    this.plan = options.plan ?? defaultConfig.plan;
```
kevinpagtakhan, line 64: no need for else case since default is `undefined` anyway

```suggestion
    this.plan = options.plan;
```

kevinpagtakhan (changes requested):

kevinpagtakhan (approved):

Issue #120, opened 2022-06-21
Cannot find name 'Awaited'.


wangbaiwang:
## Expected Behavior
Should not be prompted `Cannot find name 'Awaited'.`

## Current Behavior
![image](https://user-images.githubusercontent.com/46243630/174748057-10da3523-2422-43f2-9860-d526a6cb8608.png)

## Possible Solution
I think that should be Typescript version is too old.

Compatible with 4.5 before Typescript version, please

## Steps to Reproduce
Installed

## Environment
- JS SDK Version: @amplitude/analytics-browser@0.5.1
- Typescript Version: 3.8.3
- Installation Method: Yarn
- Browser and Version: Chrome 102.0.5005.115 (x86_64)

kevinpagtakhan:
Hi @wangbaiwang, thanks for submitting this issue! We'll look into how to support older version of TS!

PR #121, opened 2022-06-22
fix: remove Awaited type to support older versions of typescript

kevinpagtakhan:
### Summary

Fixes https://github.com/amplitude/Amplitude-TypeScript/issues/120

`Awaited` is only available starting TypeScript v4.5. This PR removes the use of `Awaited` type and replaces it with something effectively similar to support older versions of TypeScript.

### Checklist

### DIFF
--- a/packages/analytics-types/src/amplitude-promise.ts
+++ b/packages/analytics-types/src/amplitude-promise.ts
@@ -1,3 +1,3 @@
 export interface AmplitudeReturn<T> {
-  promise: Promise<T>;
+  promise: T;
 }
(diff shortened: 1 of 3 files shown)


packages/analytics-types/src/amplitude-promise.ts
```diff
@@ -1,3 +1,3 @@
 export interface AmplitudeReturn<T> {
-  promise: Promise<T>;
+  promise: T;
```
justin-fiedler, line 2: Seems a little weird the promise is no longer a Promise. It is possible to keep Promise<T> on this interface?

kevinpagtakhan: I agree! I tried different variations and this is the best I could get to without using `Awaited` :\

justin-fiedler (approved):

PR #122, opened 2022-06-24
feat: make storage interface async to enable react-native

bgiori:
### Summary

* Make `Storage` interface async

### Checklist

* Does your PR have a breaking change?:   yes


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -224,33 +224,33 @@ export class Destination implements DestinationPlugin {
...
-    this.snapshot();
+    await this.snapshot();
   }
 
-  snapshot() {
+  async snapshot(): Promise<void> {
     if (!this.config.saveEvents) return;
     const events = Array.from(this.backup);
-    this.config.storageProvider.set(this.storageKey, events);
+    await this.config.storageProvider.set(this.storageKey, events);
```
kevinpagtakhan, line 254: probably dont need to await here. we only read this on the next `init()`

packages/analytics-browser/src/session-manager.ts
```diff
@@ -10,23 +10,39 @@ export class SessionManager implements ISessionManager {
...
+    this.ready = this.storage.get(this.storageKey).then((userSession) => {
+      // Strip existing but undefined fields from cache, otherwise spreading
+      // will overwrite a value with undefined.
+      Object.keys(this.cache).forEach((key) => {
+        if (typeof this.cache[key as keyof UserSession] === 'undefined') {
+          delete this.cache[key as keyof UserSession];
+        }
+      });
+      this.cache = { ...userSession, ...this.cache } ?? { optOut: false };
+    });
```
bgiori, line 31: This was a little weird, tests were failing due to the object spread keeping values explicitly set to `undefined` over values that were set with real values.

Happy to make changes here, not sure what the best course of action is.

kevinpagtakhan (approved):

qingzhuozhen (approved):
@kevinpagtakhan Is this going to be included in GA?

PR #124, opened 2022-06-25
fix: address regression caused by async methods in destination plugin

kevinpagtakhan:
### Summary

Opens a PR to @bgiori's PR to use async storage

### Checklist


packages/analytics-browser-test/test/index.test.ts
```diff
@@ -24,10 +24,10 @@ describe('integration', () => {
     test('should track event', async () => {
       const scope = nock(url).post(path).reply(200, success);
 
-      amplitude.init('API_KEY', undefined, {
+      await amplitude.init('API_KEY', undefined, {
         ...opts,
-      });
-      const response = await amplitude.track('test event', {
+      }).promise;
```
kevinpagtakhan, line 29: added await on init here if not, init has not finished yet before the first track event is processed

packages/analytics-browser/src/config.ts
```diff
@@ -140,12 +141,12 @@ export const useBrowserConfig = async (
   const cookieName = getCookieName(apiKey);
   const cookies = await cookieStorage.get(cookieName);
   const queryParams = getQueryParams();
-  const sessionTimeout = options?.sessionTimeout ?? defaultConfig.sessionTimeout;
+  const sessionManager = await new SessionManager(cookieStorage, apiKey).load();
```
kevinpagtakhan, line 144: session manager instantiation moved to `useBrowserConfig` so we can do async ops. also calling load() through chaining

packages/analytics-browser/src/session-manager.ts
```diff
@@ -1,34 +1,28 @@
...
-    this.ready = this.storage.get(this.storageKey).then((userSession) => {
-      // Strip existing but undefined fields from cache, otherwise spreading
-      // will overwrite a value with undefined.
-      Object.keys(this.cache).forEach((key) => {
-        if (typeof this.cache[key as keyof UserSession] === 'undefined') {
-          delete this.cache[key as keyof UserSession];
-        }
-      });
-      this.cache = { ...userSession, ...this.cache } ?? { optOut: false };
-    });
```
kevinpagtakhan, line 31: i believe this issue with overwriting with undefined were caused due to this.cache getting processed after config

packages/analytics-browser/src/session-manager.ts
```diff
@@ -1,34 +1,28 @@
...
+   * load() must be called immediately after instantation
+   *
+   * ```ts
+   * await new SessionManager(...).load();
+   * ```
+   */
+  async load() {
+    this.cache = (await this.storage.get(this.storageKey)) ?? {
+      optOut: false,
+    };
```
kevinpagtakhan, line 24: with session manager being loaded before config is created, no need to pluck explicit undefined values

packages/analytics-browser/src/session-manager.ts
```diff
@@ -37,15 +31,19 @@ export class SessionManager implements ISessionManager {
   }
 
   getSessionId() {
+    this.isSessionCacheValid = true;
     void this.storage.get(this.storageKey).then((userSession) => {
-      if (this.cache) {
+      // Checks if session id has been set since the last get
+      if (this.isSessionCacheValid) {
         this.cache.sessionId = userSession?.sessionId;
       }
```
kevinpagtakhan, line 39: cancels cache update if setSessionId is called in between sync and async "get"

packages/analytics-browser/test/attribution/campaign-parser.test.ts
```diff
@@ -37,6 +37,7 @@ describe('campaign-parser', () => {
   describe('getUtmParam', () => {
     test('should return utm param from query params', async () => {
       const parser = new CampaignParser();
+      jest.spyOn(parser.utmCookieStorage, 'isEnabled').mockReturnValueOnce(Promise.resolve(false));
```
kevinpagtakhan, line 40: added to fix coverage issue

packages/analytics-types/src/config.ts
```diff
@@ -92,7 +93,7 @@ export type BrowserOptions = Omit<
       transport: TransportType;
     }
   >,
-  'apiKey' | 'plugins'
+  'apiKey' | 'plugins' | 'sessionManager'
```
kevinpagtakhan, line 96: hides sessionManager from public config interface - so they dont think they can pass this on init

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -59,16 +59,16 @@ export class Destination implements DestinationPlugin {
     });
   }
 
-  async addToQueue(...list: Context[]): Promise<void> {
-    const tryable = list.filter(async (context) => {
+  addToQueue(...list: Context[]) {
+    const tryable = list.filter((context) => {
```
kevinpagtakhan, line 63: i reverted the async stuff here. the integration tests had inconsistent results when running single/multiple test and when reordered. i finally figured it was due to a test that exhausts maximum retry limit. the issue is when list.filter() is passed with an async callback, the callback always return a truthy value and therefore does not perform filter correctly. this lead to the test continuously retrying and ignoring the max retry limit as checked on the line after this

bgiori:
Makes sense, much cleaner LGTM

PR #125, opened 2022-06-28
feat: add flush() api to send all events immediately

kevinpagtakhan:
### Summary

* Adds `flush()` api
* Converts timeline from functional module to class module
* Consolidates event `queue` and `queueBuffer`  in destination plugin
* Removes `backup` in destination plugin

### Checklist


qingzhuozhen (approved):

dnshi (approved):
👍

dnshi:
One quick question: does the new `flush` work for Node SDK as well?

kevinpagtakhan:
> One quick question: does the new `flush` work for Node SDK as well?

Hmm theoretically, it should. While I have only tested browser, the flush logic is completely shared. The only changes to browser package here is exposing the flush method.

PR #138, opened 2022-07-13
fix: handle error gracefully when identify is set with null

### LINKED ISSUE #136: Calling Identify.set with a null value throws an unhandled exception
## Expected Behavior

```
const user = new Amplitude.Identify();
user.set('key', null) <-- crashes
```

## Current Behavior

Error: Cannot convert undefined or null to object

because: ```typeof null === 'object'```

## Possible Solution

adding || {} to make sure Object.keys doesn't throw
```
export var isValidObject = function (properties) {
    if (Object.keys(properties || {}).length > MAX_PROPERTY_KEYS) {
        return false;
    }
}
```

## Steps to Reproduce

```
const user = new Amplitude.Identify();
user.set('key', null) <-- crashes
```

## Environment
- JS SDK Version: amplitude-ts/1.0.0 
- Installation Method: yarn
- Browser and Version: 87.0.4280

kevinpagtakhan:
### Summary

When identify is assigned a `null` value, make it no-op. `null` values of identify is ignored by the API.

Fixes #136

### Checklist

### DIFF
--- a/packages/analytics-core/src/utils/valid-properties.ts
+++ b/packages/analytics-core/src/utils/valid-properties.ts
@@ -28,6 +28,8 @@ export const isValidProperties = (property: string, value: any): boolean => {
         return false;
       }
     }
+  } else if (value === null || value === undefined) {
+    return false;
   } else if (typeof value === 'object') {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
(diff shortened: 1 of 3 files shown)


justin-fiedler (approved):
LGTM

Issue #132, opened 2022-07-04
Option for tracking events before initializing


yejunlee-29cm:
## Summary

Is there any option for tracking events before init?
I am using `next.js` and I'm not 100% sure that the initializing is finished before rendering the app.

kevinpagtakhan:
@yejunlee-29cm thanks for reaching out! currently, init has to complete before any tracking can begin. How are you using amplitude in Next.js? It would be great if you are able to share a simplified code snippet. Not sure if this helps your specific case but `init()` can be awaited like

```ts
await init(API_KEY).promise;
```

yejunlee-29cm:
@kevinpagtakhan  You can see it [here](https://codesandbox.io/s/old-morning-k51ko6?file=/pages/index.js)!

If the app renders before initializing, some events cannot be tracked.

liuyang1520:
@yejunlee-29cm , thanks for providing the example! The issue should be fixed with https://github.com/amplitude/Amplitude-TypeScript/pull/144. We will publish a newer version soon, feel free to try it out then.

PR #144, opened 2022-07-19
fix: missing tracked events before init issue

liuyang1520:
### Summary
This PR addresses https://github.com/amplitude/Amplitude-TypeScript/issues/132, by:
- Elevate `timeline` initialization in the `AmplitudeCore` constructor
- When events are `track`ed before `init`, `push` the events to `timeline` without `apply`
- When `init` happens:
  - `reset` plugins in `timeline`
  - install plugins, this will make sure `init` also resets the plugins like the behavior before
  - `flush` all pushed events

### Test
To test this rare case, I use the following code to reproduce the issue and test the fix:
```
function App() {
  const buttonClickHandler = async () => {
    await track(`Button Click ${new Date().toISOString()}`, { name: 'App' })
    await init('api-key', 'marvin@amplitude.com')
  }

  const reinitButtonClickHandler = async () => {
    await init('api-key', 'marvin@amplitude.com')
    await track(`Reinit ${new Date().toISOString()}`, { name: 'App' })
  }

  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <h2>Amplitude Analytics Browser Example with React</h2>

        <button onClick={buttonClickHandler}>
          Track
        </button>

        <button onClick={reinitButtonClickHandler}>
          Re-init
        </button>
      </header>
    </div>
  );
}
```

### Checklist

No

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -50,10 +50,13 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
     await this.add(new Context());
     await this.add(new Destination());
 
-    // Step 5: Track attributions
+    // Step 5: Set timeline ready for processing events
+    this.timeline.isReady = true;
+
+    // Step 6: Track attributions
     await this.runAttributionStrategy(options?.attribution, isNewSession);
 
-    // Step 6: Flush existing events, which might be collected by track before init
+    // Step 7: Flush existing events, which might be collected by track before init
     // This flush needs to run after plugin installation to gain the correct attributes
     await this.timeline.flush();
   }
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -52,6 +52,10 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
 
     // Step 5: Track attributions
     await this.runAttributionStrategy(options?.attribution, isNewSession);
+
+    // Step 6: Flush existing events, which might be collected by track before init
+    // This flush needs to run after plugin installation to gain the correct attributes
+    await this.timeline.flush();
   }
 
   async runAttributionStrategy(attributionConfig?: AttributionBrowserOptions, isNewSession = false) {
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -51,14 +51,14 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
     await this.add(new Destination());
 
     // Step 5: Set timeline ready for processing events
+    // Send existing events, which might be collected by track before init
     this.timeline.isReady = true;
+    if (!this.config.optOut) {
+      this.timeline.scheduleApply(0);
+    }
 
     // Step 6: Track attributions
     await this.runAttributionStrategy(options?.attribution, isNewSession);
-
-    // Step 7: Flush existing events, which might be collected by track before init
-    // This flush needs to run after plugin installation to gain the correct attributes
-    await this.timeline.flush();
   }
 
   async runAttributionStrategy(attributionConfig?: AttributionBrowserOptions, isNewSession = false) {
--- a/packages/analytics-core/src/core-client.ts
+++ b/packages/analytics-core/src/core-client.ts
@@ -82,10 +82,6 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
       if (this.config?.optOut) {
         return buildResult(event, 0, OPT_OUT_MESSAGE);
       }
-      // if not init yet, only #push the events, #apply will be skipped in timeline
-      if (!this.config) {
-        return this.timeline.push(event);
-      }
       const result = await this.timeline.push(event);
       if (result.code === 200) {
         this.config.loggerProvider.log(result.message);
--- a/packages/analytics-core/src/core-client.ts
+++ b/packages/analytics-core/src/core-client.ts
@@ -27,13 +27,14 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
   timeline: Timeline;
 
   constructor(name = '$default') {
+    this.timeline = new Timeline();
     this.name = name;
   }
 
   // NOTE: Do not use `_apiKey` and `_userId` here
   init(_apiKey: string | undefined, _userId: string | undefined, config: T) {
     this.config = config;
-    this.timeline = new Timeline();
+    this.timeline.reset();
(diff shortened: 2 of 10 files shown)


bgiori (approved):
I'd defer to @kevinpagtakhan but lgtm.

packages/analytics-core/src/core-client.ts
```diff
@@ -75,6 +76,10 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
 
   async dispatch(event: Event) {
     try {
+      // if not init yet, only #push the events, #apply will be skipped in timeline
+      if (!this.config) {
+        return this.timeline.push(event, this.config);
```
dnshi, line 81: If config does not exist, do you need to push the event with `this.config`?

packages/analytics-node/src/node-client.ts
```diff
@@ -13,6 +13,10 @@ export class AmplitudeNode extends AmplitudeCore<NodeConfig> {
 
     await this.add(new Context());
     await this.add(new Destination());
+
+    // Flush existing events, which might be collected by track before init
+    // This flush needs to run after plugin installation to gain the correct attributes
+    await this.timeline.flush();
```
dnshi, line 19: Node might not need this because it is at backend and events should not be collected before init. Correct me if I miss any scenario.

packages/analytics-core/src/core-client.ts
kevinpagtakhan: I'm thinking that `this.config` is a reference the the Client instance property which will later be defined. However, i'm concerned if they instantiate config to `optOut: true` which may not catch.

packages/analytics-node/src/node-client.ts
liuyang1520: Thanks for reviewing! I think node and react-native might have the similar issue too. As we don't have a hard control on the order of `#init` and `#track`.

packages/analytics-core/src/timeline.ts
```diff
@@ -31,13 +31,20 @@ export class Timeline {
     return Promise.resolve();
   }
 
+  reset() {
+    this.plugins = [];
```
kevinpagtakhan, line 35: should we also reset `applying` and `flushing`? i think if there's no plugins those ops will be no-op

packages/analytics-core/src/timeline.ts
```diff
@@ -31,13 +31,20 @@ export class Timeline {
...
+  }
+
   push(event: Event, config: Config) {
     return new Promise<Result>((resolve) => {
-      if (config.optOut) {
+      if (config?.optOut) {
```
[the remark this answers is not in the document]
liuyang1520: Good point, I think we can move this to the `dispatch` in `core-client`, if `config` is ready, and it is `optOut`, we don't need to push to the timeline queue, just need to `buildResult` there.

packages/analytics-core/src/timeline.ts
liuyang1520: Good point! I don't see any usage of `flushing`, so I will remove this flag variable. For `applying`, I will reset it too.

packages/analytics-core/src/timeline.ts
```diff
@@ -106,7 +113,7 @@ export class Timeline {
     );
 
     const executeDestinations = destination.map((plugin) => {
-      return plugin.flush && plugin.flush();
+      return plugin.flush && plugin.flush(true);
```
liuyang1520, line 116: Add the `flush(true)` here as the logic in `Destination` actually use `useRetry` here: https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/plugins/destination.ts#L129. The response might be slightly different.

kevinpagtakhan: 🤔 this changes the behavior of `client.flush()` since that solves flushing events immediately before an app closes. I'm worried that if retry, for node it will wait for the promises to resolve/reject before the app closes. I think in client, instead of calling flush - we might need to call scheduleApply instead. What do you think?

packages/analytics-core/src/timeline.ts
```diff
@@ -31,19 +33,24 @@ export class Timeline {
...
+    this.isReady = false;
+    this.plugins = [];
+  }
+
+  push(event: Event, config: Config | null) {
     return new Promise<Result>((resolve) => {
-      if (config.optOut) {
-        resolve(buildResult(event, 0, OPT_OUT_MESSAGE));
+      this.queue.push([event, resolve]);
+      if (config === null) {
```
kevinpagtakhan, line 45: do we still need this if we are checking `this.isReady` in `scheduleApply`?

packages/analytics-core/src/core-client.ts
```diff
@@ -75,6 +78,14 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
 
   async dispatch(event: Event) {
     try {
+      // skip event processing if opt out
+      if (this.config?.optOut) {
+        return buildResult(event, 0, OPT_OUT_MESSAGE);
+      }
+      // if not init yet, only #push the events, #apply will be skipped in timeline
+      if (!this.config) {
+        return this.timeline.push(event, null);
```
kevinpagtakhan, line 87: we may not need to pass config here anymore?

packages/analytics-core/src/timeline.ts
liuyang1520: Good catch, will remove this.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -50,8 +50,15 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
...
-    // Step 5: Track attributions
+    // Step 5: Set timeline ready for processing events
+    this.timeline.isReady = true;
+
+    // Step 6: Track attributions
     await this.runAttributionStrategy(options?.attribution, isNewSession);
+
+    // Step 7: Flush existing events, which might be collected by track before init
+    // This flush needs to run after plugin installation to gain the correct attributes
+    await this.timeline.flush();
```
kevinpagtakhan, line 61:
```suggestion
    // Step 5: Set timeline ready for processing events
    this.timeline.isReady = true;
    if (!this.config.optOut) {
        await this.timeline.scheduleApply();
    }

    // Step 6: Track attributions
    await this.runAttributionStrategy(options?.attribution, isNewSession);
```

kevinpagtakhan: given the other suggestions i think we may need to do this here.

1. `config.optOut` check since we did not check this before init
2. Call scheduleApply instead to go through normal flow

kevinpagtakhan (approved):

PR #146, opened 2022-07-19
fix: allow undefined storage provider

kevinpagtakhan:
### Summary

Sets storage provider in core config as optional. Allows SDK to be configured with no storage provider, effectively turning off persisting unsent events. 

Before the fix, the default config for Node SDK uses in memory storage with `saveEvents` is set to `true`. In memory storage for Node doesn't provide any value to Node but only hurts it by taking memory resources.

This fix allows Node SDK to set storage provider to `undefined`, turning of saving unsent events and fixing the issue on memory usage. SDK can still be configured to have a custom storage provider.

### Checklist

### DIFF
--- a/packages/analytics-core/src/config.ts
+++ b/packages/analytics-core/src/config.ts
@@ -44,7 +44,7 @@ export class Config implements IConfig {
   serverUrl: string | undefined;
   serverZone?: ServerZone;
   transportProvider: Transport;
-  storageProvider: Storage<Event[]>;
+  storageProvider?: Storage<Event[]>;
   useBatch: boolean;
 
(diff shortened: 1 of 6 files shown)


dnshi (approved):
It looks good to me. Just curious how Kotlin and Python handle it. Do they allow undefined storage provider as well?

kevinpagtakhan:
> It looks good to me. Just curious how Kotlin and Python handle it. Do they allow undefined storage provider as well?

For python at least, the reads from storage before sending so it requires it. Kotlin and browser have them but I believe Node does not benefit from in memory storage @dnshi

PR #147, opened 2022-07-19
fix: removes saveEvents config

kevinpagtakhan:
### Summary

Removes `saveEvents` config. After this fix, devs can opt out of saving events by setting storageProvider to undefined explicitly.

```ts
// to use default storage (local storage)
amplitude.init(API_KEY, OPTIONAL_USER_ID, {});

// to opt out
amplitude.init(API_KEY, OPTIONAL_USER_ID, {
  storageProvider: undefined,
});

// to use custom storage provider
amplitude.init(API_KEY, OPTIONAL_USER_ID, {
  storageProvider: new MyCustomStorageProvider
});
```

### Checklist

* Does your PR have a breaking change?:  No 😉


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -244,10 +244,7 @@ export class Destination implements DestinationPlugin {
    * 2) response comes back for a request
    */
   saveEvents() {
-    if (!this.config.saveEvents) {
-      return;
-    }
     const events = Array.from(this.queue.map((context) => context.event));
-    void this.config.storageProvider.set(this.storageKey, events);
+    void this.config.storageProvider?.set(this.storageKey, events);
```
dnshi, line 248: This part works fine. I am thinking maybe we can check `this.config.storageProvider` first and if there is no storage provider, we can skip iterating the events.

dnshi (approved):

packages/analytics-core/src/plugins/destination.ts
kevinpagtakhan: That's better, I will make the change.

kevinpagtakhan: Updated, @dnshi

qingzhuozhen (approved):

PR #156, opened 2022-07-27
fix: implement integration with experiment sdk for browser

bgiori:
### Summary

implement analytics-connector for analytics-browser package to integrate with expeirment-js-client

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -16,6 +16,8 @@ import { Context } from './plugins/context';
 import { useBrowserConfig, createTransport, createDeviceId, createFlexibleStorage } from './config';
 import { parseOldCookies } from './cookie-migration';
 import { CampaignTracker } from './attribution/campaign-tracker';
+import { getAnalyticsConnector } from './utils/analytics-connector';
+import { IdentityEventSender } from './plugins/identity';
 
 export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
   async init(apiKey: string, userId?: string, options?: BrowserOptions & AdditionalBrowserOptions) {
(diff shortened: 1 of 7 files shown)


bgiori:
codecov is failing, I'll add specific unit tests here soon

packages/analytics-browser/src/browser-client.ts
```diff
@@ -45,9 +47,22 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
...
+    // identity store.
+    const connector = getAnalyticsConnector();
+    connector.eventBridge.setEventReceiver((event) => {
+      void this.track(event.eventType, event.eventProperties);
+    });
+    connector.identityStore.setIdentity({
+      userId: this.config.userId,
+      deviceId: this.config.deviceId,
+    });
+
```
dnshi, line 61: Hey @bgiori , is analytics connector optional or required for analytics browser? For the customers who do not use experiment SDK, do they need this analytics connector? cc: @kevinpagtakhan

kevinpagtakhan (approved):

packages/analytics-browser/src/browser-client.ts
kevinpagtakhan: with these changes, it will emit and listen for 'messages' regardless but experiment SDK may/may not be on the other end.

dnshi: Correct me if I am wrong. "@amplitude/analytics-connector" will not be tree shaking because the source file "./utils/analytics-connector" imports it. The TS browser bundle might include the "@amplitude/analytics-connector".

kevinpagtakhan: Correct, not treeshaken.

bgiori: Not treeshaken. I am not sure what the requirements are for treeshaking this. If you have suggestions about how to make this tree-shakable, I'm happy to implement.

dnshi: I think you can move `getAnalyticsConnector` and `IdentityEventSender` into a separate file and then try the following 
```js
if (this.config.enableExperiment) {
  const { getAnalyticsConnector, IdentityEventSender } = await import('FILE_NAME');
  const connector = getAnalyticsConnector();
  connector.eventBridge.setEventReceiver((event) => {
    void this.track(event.eventType, event.eventProperties);
  });
  connector.identityStore.setIdentity({
    userId: this.config.userId,
    deviceId: this.config.deviceId,
  });

  await this.add(new IdentityEventSender());
}
```

Does experiment SDK have timeout or need to initialize after core SDK initialization? With this change, core SDK might take longer time to complete the initialization.

bgiori: I want to avoid additional configuration on the analytics SDK, might be worth the tradeoff since the package is pretty small. 

We can always re-optimize later since this is all internal.

PR #157, opened 2022-07-27
feat: add reset method for resetting userId and deviceId

liuyang1520:
### Summary
This PR adds and exposes the `#reset` method for user to reset `userId` and `deviceId`.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/index.ts
+++ b/packages/analytics-browser/src/index.ts
@@ -9,6 +9,7 @@ export {
   init,
   logEvent,
   remove,
+  reset,
   revenue,
   setDeviceId,
   setGroup,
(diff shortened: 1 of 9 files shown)


packages/analytics-browser/src/index.ts
```diff
@@ -9,6 +9,7 @@ export {
   init,
   logEvent,
   remove,
+  reset,
```
dnshi, line 12: If `reset` is a public API, you might need to add it into the snippet: https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/scripts/templates/browser-snippet.template.js#L78-L89

liuyang1520: Thanks for reviewing! Good catch, updated!

dnshi (approved):

kevinpagtakhan (approved):

kevinpagtakhan:
Let's run docs workflow to update api docs.

PR #172, opened 2022-08-04
chore: GitHub template

yuhao900914:
### Summary
Make typescript Jira Create github action as a template.
Other SDK repos can reuse this github action flow directly.

### Checklist

### DIFF
--- a/.github/workflows/jira-issue-create.yml
+++ b/.github/workflows/jira-issue-create.yml
@@ -7,7 +7,6 @@ on:
   workflow_call:
     inputs:
       label:
-        required: true
         type: string
         default: 'TS'
     secrets:
--- a/.github/workflows/jira-issue-create.yml
+++ b/.github/workflows/jira-issue-create.yml
@@ -4,6 +4,13 @@ name: Jira Issue Creator
 on:
   issues:
     types: [opened]
+  workflow_call:
+    inputs:
+      label:
+        required: true
+        type: string
+env:
+  JIRA_TAG: TS
 
 jobs:
   build:
@@ -25,12 +32,12 @@ jobs:
           project: ${{ secrets.JIRA_PROJECT }}
           issuetype: Task
           summary: |
-            [SDK - TS] ${{ github.event.issue.title }}
+            [SDK - $JIRA_TAG] ${{ github.event.issue.title }}
           description: |
             ${{ github.event.issue.html_url }}
           fields: '{
             "customfield_11794": {"value": "Github"},
-            "labels": ["ts"]
+            "labels": ["$JIRA_TAG"]
           }' # origin, labels
 
       - name: Log created issue
--- a/.github/workflows/jira-issue-create.yml
+++ b/.github/workflows/jira-issue-create.yml
@@ -9,8 +9,20 @@ on:
       label:
         required: true
         type: string
-env:
-  JIRA_TAG: TS
+        default: 'TS'
+    secrets:
+      JIRA_BASE_URL:
+        description: 'Jira Base Url'
+        required: true
+      JIRA_USER_EMAIL:
+        description: 'Jira User Email'
+        required: true
+      JIRA_API_TOKEN:
+        description: 'Jira Token'
+        required: true
+      JIRA_PROJECT:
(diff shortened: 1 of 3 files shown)


.github/workflows/jira-issue-create.yml
```diff
@@ -4,6 +4,13 @@ name: Jira Issue Creator
 on:
   issues:
     types: [opened]
+  workflow_call:
+    inputs:
+      label:
+        required: true
+        type: string
```
qingzhuozhen, line 11: Curious how other repos can use this action then? And how ts can trigger this action?

yuhao900914: https://github.com/yuhao900914/Amplitude-Flutter-Example/issues
This is the GitHub action of my example app. 
Other repos need to call this template with the required input and label and secret. 
I need to update the Jira create GitHub action in other repos as well.

Defined the specific secrets works fine. But I'm still trying to make it use inherit secrets to avoid defining multiple secrets in other repo.

.github/workflows/jira-issue-create.yml
```diff
@@ -18,19 +36,29 @@ jobs:
...
 
+      - name: Convert Label String
+        id: string
+        uses: ASzc/change-string-case-action@v1
+        with:
+          string: ${{ inputs.label }}
+
+      - name: lowercase the label
+        run: |
+          echo "${{ steps.string.outputs.lowercase }}"
```
kevinpagtakhan, line 47: what is this for?

.github/workflows/jira-issue-create.yml
```diff
@@ -18,19 +36,29 @@ jobs:
           JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
 
+      - name: Convert Label String
+        id: string
+        uses: ASzc/change-string-case-action@v1
+        with:
+          string: ${{ inputs.label }}
```
kevinpagtakhan, line 43: How does this work for TS?

yuhao900914: Good point. 
Mark the label default as TS and not required.

.github/workflows/jira-issue-create.yml
yuhao900914: Forgot to remove the test.

.github/workflows/jira-issue-create.yml
liuyang1520: I saw the secrets in this repo Settings. As we don't have a mid-level group between `amplitude` and `Amplitude-TypeScript`, curious how does the inherit work.

.github/workflows/jira-issue-create.yml
kevinpagtakhan: I wonder if we can create this usable workflow in ts but also have something like https://github.com/yuhao900914/Amplitude-Flutter-Example/blob/main/.github/workflows/main.yml in ts

.github/workflows/jira-issue-create.yml
yuhao900914: amplitude is the organization. Amplitude-typescript is the repo. 
I think it should be if other repos like Amplitude-Javascript call this GitHub template without any secrets settings, it will use the secrets under this repo. 
https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows

`Workflows that call reusable workflows in the same organization or enterprise can use the inherit keyword to implicitly pass the secrets.`

qingzhuozhen (approved):

kevinpagtakhan (approved):

.github/workflows/jira-issue-create.yml
yuhao900914: That's a good point. I think in this case the label is always required, then we will not miss or assign the wrong label. I will try this on my end.

liuyang1520 (approved):

kevinpagtakhan (approved):

liuyang1520 (approved):

PR #174, opened 2022-08-04
docs(reset): Add `reset` documentation

BrodaNoel:
### Summary
Add doc

### Checklist

### DIFF
--- a/packages/analytics-react-native/README.md
+++ b/packages/analytics-react-native/README.md
@@ -148,3 +148,13 @@ track('Button Clicked').promise.then((result) => {
   result.message; // "Event tracked successfully" (The response message)
 });
 ```
+
+### User Log out
+
+Reset the user (usually called after the user logs out in the app). This will clear the `userId` and user properties (previously set using `Identify`)
+
+```typescript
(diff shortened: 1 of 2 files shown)


packages/analytics-react-native/README.md
```diff
@@ -148,3 +148,13 @@ track('Button Clicked').promise.then((result) => {
   result.message; // "Event tracked successfully" (The response message)
 });
 ```
+
+### User Log out
+
+Reset the user (usually called after the user logs out in the app). This will clear the `userId` and user properties (previously set using `Identify`)
```
kevinpagtakhan, line 154: `reset()` only updates user ID and device ID sent in event payload, but does not necessarily clear user properties set using Identify. The user property in identify calls remain in the user's identity to which the identify was called for. After calling `reset()` the succeeding events now belong to a new user identity.

BrodaNoel: Done. Updated.
Take a look now.

Feel free to edit the file and add any change you want.

kevinpagtakhan:
@yuhao900914 mind if we have you take a look at this quickly? thanks!

yuhao900914 (approved):

PR #181, opened 2022-08-11
fix: add callable queue when init is pending

kevinpagtakhan:
### Summary

Adds a queue to push callable functions that require `this.config` to be called. Once `this.config` is ready, invoke all functions in the queue.

### Checklist

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -22,7 +22,7 @@ describe('integration', () => {
   });
 
   // WARNING: This test has to run first
-  // It assumes tests under the assumption that amplitude has not be initiated at all
+  // This test is under the assumption that amplitude has not be initiated at all
   // To achieve this condition, it must run before any other tests
   describe('FIRST TEST: defer initialization', () => {
     test('should allow init to be called after other APIs', () => {
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -3,6 +3,7 @@ import { default as nock } from 'nock';
 import { success } from './responses';
 import 'isomorphic-fetch';
 import { path, SUCCESS_MESSAGE, url, uuidPattern } from './constants';
+import { PluginType } from '@amplitude/analytics-types';
 
 describe('integration', () => {
   const uuid: string = expect.stringMatching(uuidPattern) as string;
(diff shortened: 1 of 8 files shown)


packages/analytics-core/src/core-client.ts
```diff
@@ -67,11 +73,18 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
...
+      this.q.push(this.add.bind(this, plugin));
+      return;
+    }
+    return this.timeline.register(plugin, this.config);
   }
 
   async remove(pluginName: string) {
+    if (!this.config) {
+      this.q.push(this.remove.bind(this, pluginName));
+      return;
```
liuyang1520, line 86: Minor, do we want to return a similar resolved promise here, to be align with `this.timeline.deregister`? Same question for the early return in `add`.

dnshi (approved):

packages/analytics-core/src/core-client.ts
dnshi: `return undefined` is same as `return Promise.resolve()` in async function. So it should be fine.

kevinpagtakhan: That's correct. Unless @liuyang1520, you are referring to await until the plugins are actually added/removed?

liuyang1520: Got it, no worries, then, I forget that `async` will return a promise anyways.

liuyang1520 (approved):

packages/analytics-core/src/core-client.ts
```diff
@@ -34,15 +34,15 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
...
     this.timeline.reset();
     const queuedFunctions = this.q;
     this.q = [];
-    queuedFunctions.map(async (queuedFunction) => {
+    for (const queuedFunction of queuedFunctions) {
       await queuedFunction();
-    });
-    return Promise.resolve();
+    }
+    return undefined;
```
dnshi, line 45: This function will return `undefined` by default, so you might not need this line

liuyang1520 (approved):

packages/analytics-core/src/core-client.ts
kevinpagtakhan:
```suggestion
```

dnshi (approved):

packages/analytics-browser-test/test/index.test.ts
```diff
@@ -20,6 +21,66 @@ describe('integration', () => {
     document.cookie = 'AMP_API_KEY=null; expires=-1';
   });
 
+  // WARNING: This test has to run first
+  // It assumes tests under the assumption that amplitude has not be initiated at all
```
kevinpagtakhan, line 25:
```suggestion
  // This test is under the assumption that amplitude has not be initiated at all
```

Issue #177, opened 2022-08-09
Different `platform` values


BrodaNoel:
After migrating from `expo-analytics-amplitude` (which uses Amplitude-ReactNative) to this library, I'm seeing different values in the Platform:
<img width="877" alt="Screen Shot 2022-08-09 at 15 15 59" src="https://user-images.githubusercontent.com/1837969/183730580-8aff0ff9-f3a2-4552-9f9a-faf11c5a1370.png">

As you can see, there is an `Android` and an `android`, which breaks all the charts.

You may want to consider defining an standard between your libraries, in order to avoid this.

liuyang1520:
Hi, @BrodaNoel , thanks for reporting!

I took a look at our libs, and found that we are consistent about the OS name attribute, see list:
- This new RN lib: https://github.com/amplitude/Amplitude-TypeScript/blob/7d11fdbee2ece65b64cec873d2f2ab74ed5fecf3/packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AndroidContextProvider.kt#L404
- The new Android Kotlin SDK: https://github.com/amplitude/Amplitude-Kotlin/blob/8c46250e3ee957e22f97ad5d6e957e12ebe82272/common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt#L406
- Amplitude-Android SDK (used by `expo-analytics-amplitude`, not familiar with this, as this is not the official one, so not maintained by us, but I saw they mention this SDK in the readme. I don't think they are using `Amplitude-ReactNative`, see https://github.com/expo/expo-analytics-amplitude): https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/DeviceInfo.java#L31

Could you provide more information about
1. which even property
2. any customization of this value
3. ways to reproduce/setup
4. any sample raw event (remove personal information) would be helpful here

BrodaNoel:
That's weird... They say they wrap here https://github.com/expo/expo/tree/sdk-45/packages/expo-analytics-amplitude#expo-analytics-amplitude

> Provides access to Amplitude (https://amplitude.com/) mobile analytics. This module wraps Amplitude-iOS (https://github.com/amplitude/Amplitude-iOS) and Android (https://github.com/amplitude/Amplitude-Android) SDKs.

[...]

Issue #176, opened 2022-08-08
[docs] Missing docs for `trackingOptions`


BrodaNoel:
## Expected Behavior
There is no documentation about how to disable some trackings.

For example, in the previous library I was using for Amplitude, I had this function:
```
await setTrackingOptionsAsync({
  disableAdid: true,
});
```

Is there a way to disable these options in this library?

## Current Behavior
No docs about it

## Possible Solution
Adding doc (or the feature if missing)

liuyang1520:
Hi, @BrodaNoel , I cannot find any related information about `setTrackingOptionsAsync` method in current SDKs, do you mind sharing the `previous library` version/repo? And the runtime/version you are using for this library?

This library allows custom setup of tracking options, see: https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/#opt-users-out-of-tracking.

BrodaNoel:
My bad.
I migrated from `https://www.npmjs.com/package/expo-analytics-amplitude` to this component (because Expo SDK 46 doesn't support `expo-analytics-amplitude` anymore.
That library has the `setTrackingOptions` method, which was properly working, so, clearly it's supported by Amplitude

BrodaNoel:
In [the link you sent](https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/#optional-tracking), I don't see a way to disable the ADID tracking (the old `disableAdid`)

liuyang1520:
I see, thanks for the explanation! Just confirmed that our React-Native SDK (in this repo) doesn't support ADID tracking yet. We will add the ADID tracking in this React-Native SDK.

liuyang1520:
@BrodaNoel , please try @amplitude/analytics-react-native v0.2.3 for disabling the `adid`, I will update the doc there.

BrodaNoel:
@liuyang1520 my problem is that I don't even know how to do it.

I'm currenlty initilizing the library like this:

```js
await initAmplitude(apiKey).promise;
```

At that time, I don't know the user id. What should I do? Something like this?:

```js
await initAmplitude(apiKey, undefined, { adid: false }).promise;
```

kevinpagtakhan:
@BrodaNoel This is how you can disable `adid` tracking:

```ts
const trackingOptions = {
  adid: false,
};
await init(API_KEY, undefined, { trackingOptions }).promise;
```

We'll formally add this to our developer docs pretty soon.

[...]

PR #184, opened 2022-08-13
fix: correct platform and add adid

liuyang1520:
### Summary
- Update react-native readme to use `yarn` instead of `npm`
- Correct `platform` value
- Add `adid` option, see example:
```
trackingOptions: {
  adid: false
}
```

This PR addresses:
- https://github.com/amplitude/Amplitude-TypeScript/issues/177
- https://github.com/amplitude/Amplitude-TypeScript/issues/176
(@BrodaNoel)

### Checklist

### DIFF
--- a/packages/analytics-types/src/config.ts
+++ b/packages/analytics-types/src/config.ts
@@ -46,7 +46,9 @@ export interface BrowserConfig extends Config {
   userId?: string;
 }
 
-export type ReactNativeConfig = BrowserConfig;
+export type ReactNativeConfig = Omit<BrowserConfig, 'trackingOptions'> & {
+  trackingOptions: ReactNativeTrackingOptions;
+};
 
 export type NodeConfig = Config;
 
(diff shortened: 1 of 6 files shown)


haoliu-amp (approved):

packages/analytics-types/src/config.ts
```diff
@@ -73,6 +75,10 @@ export interface TrackingOptions {
   versionName?: boolean;
 }
 
+export interface ReactNativeTrackingOptions extends TrackingOptions {
+  adid?: boolean;
```
yuhao900914, line 79: feels we also need idfa/idfv for iOS?
But that's out of scope for this pr.

yuhao900914 (approved):

packages/analytics-types/src/config.ts
liuyang1520: Thanks for reviewing! Yes, I don't find the logic in iOS code here to generate idfa/idfv, probably this is missing. Think we can check with Brian next week, probably can be updated with the Kotlin/iOS import change (instead of duplicating some logics here).

PR #186, opened 2022-08-15
feat: add 'extra' to eventOptions

falconandy:
### Summary

Adds `extra` to eventOptions.

### Checklist

### DIFF
--- a/packages/analytics-core/test/timeline.test.ts
+++ b/packages/analytics-core/test/timeline.test.ts
@@ -128,6 +128,62 @@ describe('timeline', () => {
       await timeline.deregister(before.name);
       expect(beforeExecute).toHaveBeenCalledTimes(0);
     });
+
+    test("should pass event's extra to plugins", async () => {
+      const beforeExecute = jest.fn().mockImplementationOnce((event: Event) => {
+        expect(event.extra).toStrictEqual({ 'extra-key': 'extra-value' });
+        return Promise.resolve(event);
+      });
+      const before: Plugin = {
+        name: 'plugin:before',
+        type: PluginType.BEFORE,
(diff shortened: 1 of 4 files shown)


packages/analytics-core/test/timeline.test.ts
```diff
@@ -128,6 +128,62 @@ describe('timeline', () => {
       await timeline.deregister(before.name);
       expect(beforeExecute).toHaveBeenCalledTimes(0);
     });
+
+    test("should not pass event's extra to destination plugins", async () => {
```
justin-fiedler, line 132: We want to pass the extra to the destination plugins to allow users to access it, but do not want to include it in the payload sent to the server in the Amplitude Destination.

justin-fiedler: @kevinpagtakhan I noticed that the Amplitude Destination Plugin is named `Destination` do you think we could rename it to be more explicit e.g. `AmplitudeDestinationPlugin`?

justin-fiedler (changes requested):

packages/analytics-core/test/timeline.test.ts
kevinpagtakhan: Sure, that makes sense!

falconandy: > We want to pass the extra to the destination plugins to allow users to access it

Fixed

>  noticed that the Amplitude Destination Plugin is named Destination do you think we could rename it to be more explicit e.g. AmplitudeDestinationPlugin?

In this PR?

justin-fiedler: Thanks @falconandy we can rename the plugin in another PR. I just wanted to call it out.

justin-fiedler (approved):

PR #188, opened 2022-08-16
feat: adds create instance api

kevinpagtakhan:
### Summary

Adds support for multiple instance creation. Each instance has its own API_KEY, plugins, configurations.

```typescript
// this is the default instance
amplitude.init('asdf');
amplitude.track('event-1');

// this is the second instance
const amp2 = amplitude.createInstance();
amp2.init('qwerty', undefined, {
  serverUrl: 'https://domain.com',
});
amp2.track('event-2');
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -1,10 +1,11 @@
-import { AmplitudeCore, Destination, Identify, Revenue, returnWrapper, UUID } from '@amplitude/analytics-core';
+import { AmplitudeCore, Destination, Identify, Revenue, UUID, amplitudePromise } from '@amplitude/analytics-core';
 import {
   AdditionalBrowserOptions,
   AttributionBrowserOptions,
   BrowserConfig,
   BrowserOptions,
   Campaign,
+  CreateBrowserInstance,
   EventOptions,
   Identify as IIdentify,
   Result,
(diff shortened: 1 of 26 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -174,226 +175,29 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
...
- *
- * // alternatively, this tracking method is awaitable
- * await flush().promise;
- * ```
- */
-export const flush = returnWrapper(client.flush.bind(client));
+export const createInstance: CreateBrowserInstance = () => {
+  const client = new AmplitudeBrowser();
+  return {
+    init: amplitudePromise(client.init.bind(client)),
```
justin-fiedler, line 181: Will exporting the methods grouped like this affect tree-shaking?

packages/analytics-types/src/client/index.ts
```diff
@@ -0,0 +1,8 @@
+import { BaseClient } from './base-client';
+import { BrowserClient, ReactNativeClient } from './web-client';
+import { NodeClient } from './node-client';
+
+type CreateInstance<T extends BaseClient> = () => T;
+export type CreateBrowserInstance = CreateInstance<BrowserClient>;
+export type CreateReactNativeInstance = CreateInstance<ReactNativeClient>;
+export type CreateNodeInstance = CreateInstance<NodeClient>;
```
justin-fiedler, line 8: How do you see these CreateXInstance types being used? Do we need them?

Could we `export const createInstance: CreateInstance<XClient>` in the individual platforms to keep the SDK footprint smaller?

packages/analytics-core/src/utils/amplitude-promise.ts
```diff
@@ -0,0 +1,5 @@
+export const amplitudePromise =
```
justin-fiedler, line 1: Why the rename? I though return wrapper was ok

packages/analytics-browser/src/browser-client.ts
kevinpagtakhan: Yes! Although, importing `init` and `track` imports the ~90% SDK that's why I don't feel as bad now.

packages/analytics-core/src/utils/amplitude-promise.ts
kevinpagtakhan: Oh I was under the impression there was confusion on what's returning `.promise` and what's not so I thought of giving it a more obvious name. I can revert though

packages/analytics-types/src/client/index.ts
kevinpagtakhan: That's a good point. I'll give it a try!

kevinpagtakhan: I think I can just remove `CreateBrowserInstance` and annotate `createInstance` like

const createInstance = (): BrowserClient => {};

I think it makes it cleaner and easier to understand.

What do you think?

packages/analytics-core/src/utils/amplitude-promise.ts
justin-fiedler: Sorry before I was asking why we couldn't have `Promise<T>` in the `AmplitudeReturn.promise` type def. It looks like you fixed that in this PR 🙌 

I think `AmplitudeReturn` is more flexible and would allow us to return additional data if needed in the future, including synchronously.
```
export interface AmplitudeReturn<T> {
  promise: Promise<T>;
  someSyncronousReturnValue: string;
}
```

falconandy (commented):
It looks like different instances can share the same cookies if they use the same api key:
```
const cookieName = getCookieName(apiKey);
```
Is it expected?

kevinpagtakhan:
> It looks like different instances can share the same cookies if they use the same api key:
> 
> ```
> const cookieName = getCookieName(apiKey);
> ```
> 
> Is it expected?

@falconandy yes, they can share the same api key and cookies. Do customer normally create multiple instances with the same api key? I'm not familiar with this yet. Although, for web - opening multiple tabs for the same site "creates" multiple instances that share the same cookie

justin-fiedler:
> @falconandy yes, they can share the same api key and cookies. Do customer normally create multiple instances with the same api key? I'm not familiar with this yet. Although, for web - opening multiple tabs for the same site "creates" multiple instances that share the same cookie

I think this is ok. Afaik customer use multiple instances to support multiple keys, not multiple instances for the same key.

justin-fiedler (approved):

falconandy (approved):

PR #191, opened 2022-08-16
fix: prevent concurrent init calls

kevinpagtakhan:
### Summary

Prevents concurrent init calls for the same amplitude instance

Previously, the following code performs initialization multiple times. This leads to race conditions on 1) reading from storage; 2) resetting timeline state; and 3) installing plugins

```ts
amplitude.init(API_KEY); // performs full init
amplitude.init(API_KEY); // performs full init
amplitude.init(API_KEY); // performs full init
```

With this fix, succeeding `init` calls while one execution is pending exits early.

```ts
amplitude.init(API_KEY); // performs full init
amplitude.init(API_KEY); // returns early
amplitude.init(API_KEY); // returns early
```

However, serial `init` calls retains status quo behavior.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -21,6 +21,12 @@ import { IdentityEventSender } from './plugins/identity';
 
 export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
   async init(apiKey: string, userId?: string, options?: BrowserOptions & AdditionalBrowserOptions) {
+    // Step 0: Block concurrent initialization
+    if (this.initializing) {
+      return;
+    }
+    this.initializing = true;
+
     // Step 1: Read cookies stored by old SDK
     const oldCookies = await parseOldCookies(apiKey, options);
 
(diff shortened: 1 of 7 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -65,6 +71,8 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
     await this.add(new IdentityEventSender());
     await this.add(new Destination());
 
+    this.initializing = false;
```
kevinpagtakhan, line 74: Safe to allow another init after plugin installation. The next operations are sending track calls.

liuyang1520 (approved):

yuhao900914 (approved):

liuyang1520:
> I am curious after your change, the second init and third init function will be skipped right? It is okay for single instance, but it might introduce problem if we support multiple instances, right?

I think this PR will prevent concurrent `init`, if user wants multiple instances, probably can do:
```js
const ins1 = await amplitude.init(API_KEY);
const ins2 = await amplitude.init(API_KEY);
const ins3 = await amplitude.init(API_KEY);
```

kevinpagtakhan:
> I am curious after your change, the second init and third init function will be skipped right? It is okay for single instance, but it might introduce problem if we support multiple instances, right?

@dnshi I think it's fine to have init concurrently as long as it's happening in different instances. I am assuming they'd need another instance for a different project/API key. With this, storage and timeline and plugins are separate and the operations won't conflict with each other.

dnshi (approved):

PR #193, opened 2022-08-16
fix: removes unused tracking options

kevinpagtakhan:
### Summary

Removes unused tracking options from type definition

Docs update: https://github.com/Amplitude-Developer-Docs/amplitude-dev-center/pull/209

### Checklist


packages/analytics-browser/src/config.ts
```diff
@@ -21,6 +21,15 @@ import { SessionManager } from './session-manager';
 
 export const getDefaultConfig = () => {
   const cookieStorage = new MemoryStorage<UserSession>();
+  const trackingOptions: Required<TrackingOptions> = {
```
kevinpagtakhan, line 24: I created this to enforce all tracking options to have a explicit default value

packages/analytics-types/src/config.ts
```diff
@@ -60,23 +60,19 @@ export type InitOptions<T extends Config> =
...
   deviceModel?: boolean;
-  dma?: boolean;
   ipAddress?: boolean;
   language?: boolean;
   osName?: boolean;
   osVersion?: boolean;
   platform?: boolean;
-  region?: boolean;
-  versionName?: boolean;
+  [key: string]: boolean | undefined;
```
kevinpagtakhan, line 70: I added this to make the changes here backwards compatible for the removed properties

dnshi (approved):

packages/analytics-types/src/config.ts
qingzhuozhen: We don't track these in react native?

kevinpagtakhan: @qingzhuozhen region and app version being tracked but not controlled by tracking options as of the moment

PR #198, opened 2022-08-22
fix: add conditional check for window.location.search

dnshi:
### Summary

Add conditional check for `window.location.search`

### Checklist

### DIFF
--- a/packages/analytics-browser/src/utils/query-params.ts
+++ b/packages/analytics-browser/src/utils/query-params.ts
@@ -1,6 +1,6 @@
 export const getQueryParams = (): Record<string, string | undefined> => {
   /* istanbul ignore if */
-  if (typeof window === 'undefined' || !window.location?.search) {
+  if (typeof window === 'undefined' || !window.location || !window.location.search) {
     return {};
   }
   const pairs = window.location.search.substring(1).split('&').filter(Boolean);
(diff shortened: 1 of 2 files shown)


packages/analytics-browser/src/utils/query-params.ts
```diff
@@ -1,6 +1,6 @@
 export const getQueryParams = (): Record<string, string | undefined> => {
   /* istanbul ignore if */
-  if (typeof window === 'undefined') {
+  if (typeof window?.location?.search === 'undefined') {
```
kevinpagtakhan, line 3: I think accessing an `undefined` global variable throws an error. The `typeof` op above requires accessing `window`. Consider the following example:

```ts
if (typeof window?.location?.search) {
```

transpiles to:

```js
var _a;
if (typeof ((_a = window === null || window === void 0 ? void 0 : window.location) === null || _a === void 0 ? void 0 : _a.search)) {
```

It appears that the code accesses `window` object directly which will throw:

```
Uncaught ReferenceError: window is not defined
```

So I think it's probably best to check `typeof` window on its own.

```suggestion
  if (typeof window === 'undefined' || !window.location?.search) {
```

transpiles to:

```js
var _a;
if (typeof window === 'undefined' || !((_a = window.location) === null || _a === void 0 ? void 0 : _a.search)) {
```

kevinpagtakhan (changes requested):

packages/analytics-browser/src/utils/query-params.ts
dnshi: Good catch. It makes more sense. Thanks!

dnshi:
The test complains about the `?.`. I switch the implementation from `!window.location?.search` to `!window.locatioin || !window.location.search`

kevinpagtakhan (approved):

liuyang1520 (approved):

PR #201, opened 2022-08-23
fix: pass options to testStorage

bohan-amplitude:
### Summary

pass options to testStorage in CookieStorage.isEnabled()

### Checklist


kevinpagtakhan:
we probably need to do this for `packages/analytics-react-native/src/storage/cookie.ts` as well.

@dnshi this is an example of where the same logic is copied in browser and react-native pkgs

kevinpagtakhan (approved):

PR #208, opened 2022-08-30
feat: add support for auto page view tracking

dnshi:
### Summary

This PR is to add support for auto page view tracking:
- Migrate web attribution into a plugin
- Add page tracking plugin
- Update CampaignTracker
- Remove `runAttributionStrategy` function

#### Note to reviewers
The code base and test cases for both browser and react native are similar. Please review browser one primarily. 

### Checklist

### DIFF
--- a/packages/analytics-browser/src/attribution/campaign-tracker.ts
+++ b/packages/analytics-browser/src/attribution/campaign-tracker.ts
@@ -4,9 +4,8 @@ import {
   Campaign,
   CampaignParser as ICampaignParser,
   CampaignTracker as ICampaignTracker,
-  CampaignTrackFunction,
   CampaignTrackerOptions,
-  BaseEvent,
+  PageTrackingFilter,
 } from '@amplitude/analytics-types';
 import { getCookieName as getStorageKey } from '../utils/cookie-name';
 import { CampaignParser } from './campaign-parser';
@@ -16,30 +15,39 @@ export class CampaignTracker implements ICampaignTracker {
   storage: Storage<Campaign>;
   storageKey: string;
   parser: ICampaignParser;
-  track: CampaignTrackFunction;
-  onNewCampaign: (campaign: Campaign) => unknown;
 
   disabled: boolean;
-  trackNewCampaigns: boolean;
-  trackPageViews: boolean;
   excludeReferrers: string[];
   initialEmptyValue: string;
+  resetSessionOnNewCampaign: boolean;
+
+  private _currentCampaign!: Campaign;
+  private _isNewCampaign!: boolean;
 
   constructor(apiKey: string, options: CampaignTrackerOptions) {
     this.storage = options.storage;
     this.storageKey = getStorageKey(apiKey, MKTG);
     this.parser = new CampaignParser();
-    this.track = options.track;
-    this.onNewCampaign = options.onNewCampaign;
 
     this.disabled = Boolean(options.disabled);
-    this.trackNewCampaigns = Boolean(options.trackNewCampaigns);
-    this.trackPageViews = Boolean(options.trackPageViews);
     this.excludeReferrers = options.excludeReferrers ?? [];
+    this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
+    this.resetSessionOnNewCampaign = options.resetSessionOnNewCampaign ?? options.trackNewCampaigns === true;
+
     if (typeof location !== 'undefined') {
       this.excludeReferrers.unshift(location.hostname);
     }
-    this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
+
+    void this.refreshCampaignState();
+  }
+
+  private async refreshCampaignState() {
+    return Promise.all([this.parser.parse(), this.getCampaignFromStorage()]).then(
+      ([currentCampaign, previousCampaign]) => {
+        this._isNewCampaign = this.isNewCampaign(currentCampaign, previousCampaign);
+        this._currentCampaign = currentCampaign;
+      },
+    );
   }
 
   isNewCampaign(currentCampaign: Campaign, previousCampaign: Campaign) {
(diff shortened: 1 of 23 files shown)


kevinpagtakhan (changes requested):
pending design discussion with justin

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -16,30 +15,39 @@ export class CampaignTracker implements ICampaignTracker {
...
     this.parser = new CampaignParser();
-    this.track = options.track;
-    this.onNewCampaign = options.onNewCampaign;
 
     this.disabled = Boolean(options.disabled);
-    this.trackNewCampaigns = Boolean(options.trackNewCampaigns);
-    this.trackPageViews = Boolean(options.trackPageViews);
     this.excludeReferrers = options.excludeReferrers ?? [];
+    this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
+    this.resetSessionOnNewCampaign = options.resetSessionOnNewCampaign ?? options.trackNewCampaigns === true;
```
kevinpagtakhan, line 35:
```suggestion
    this.resetSessionOnNewCampaign = options.resetSessionOnNewCampaign ?? options.trackNewCampaigns;
```

packages/analytics-types/src/config.ts
```diff
@@ -77,16 +77,24 @@ export interface ReactNativeTrackingOptions extends TrackingOptions {
...
+  trackPageViews?: boolean | PageTrackingBrowserOptions;
 }
 
 export interface AttributionBrowserOptions {
   disabled?: boolean;
   excludeReferrers?: string[];
   initialEmptyValue?: string;
+  resetSessionOnNewCampaign?: boolean;
   trackNewCampaigns?: boolean;
   trackPageViews?: boolean;
```
AlecBoutin, line 89: Recommend we add some deprecation jsdoc comments here
```suggestion
  /** @deprecated New campaigns are now always tracked by default. Setting this to true also sets resetSessionOnNewCampaign to true. */
  trackNewCampaigns?: boolean;
  /** @deprecated Use the new top level trackPageViews configuration instead. */
  trackPageViews?: boolean;
```

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -80,33 +88,38 @@ export class CampaignTracker implements ICampaignTracker {
...
-  async send(isNewSession: boolean) {
-    if (this.disabled) {
-      return;
-    }
-    const currentCampaign = await this.parser.parse();
-    const previousCampaign = await this.getCampaignFromStorage();
-    if (!isNewSession) {
-      if (!this.trackNewCampaigns || !this.isNewCampaign(currentCampaign, previousCampaign)) {
-        return;
+  trackOn(filter: PageTrackingFilter = undefined, callback: (currentCampaing: Campaign) => unknown) {
```
AlecBoutin, line 94: typo - currentCampaing

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -16,30 +15,39 @@ export class CampaignTracker implements ICampaignTracker {
...
     this.excludeReferrers = options.excludeReferrers ?? [];
+    this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
+    this.resetSessionOnNewCampaign = options.resetSessionOnNewCampaign ?? options.trackNewCampaigns === true;
+
     if (typeof location !== 'undefined') {
       this.excludeReferrers.unshift(location.hostname);
     }
-    this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
+
+    void this.refreshCampaignState();
```
AlecBoutin, line 41: Race condition - Track completion of refreshCampaignState in a promise field so getCurrentState can await it below.

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -80,33 +88,38 @@ export class CampaignTracker implements ICampaignTracker {
...
+    await callback(currentState);
+
+    if (currentState.isNewCampaign) {
+      await this.saveCampaignToStorage(currentState.currentCampaign);
     }
-    await this.track(this.createCampaignEvent(currentCampaign));
-    await this.saveCampaignToStorage(currentCampaign);
+  }
+
+  private getCurrentState() {
```
AlecBoutin, line 119: Make this async, see above

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -80,33 +88,38 @@ export class CampaignTracker implements ICampaignTracker {
...
+      default: {
+        if (typeof filter === 'function') {
+          return filter() && callback(currentCampaign);
+        }
+        return callback(currentCampaign);
+      }
+    }
+  }
+
+  async onStateChange(callback: (state: { isNewCampaign: boolean; currentCampaign: Campaign }) => Promise<unknown>) {
```
AlecBoutin, line 110: What is onStateChange meant to do? Neither of the places its used use the currentState passed back to the callback. It's confusing that unrelated components are triggering saveCampaignToStorage.

packages/analytics-browser/src/attribution/campaign-tracker.ts
AlecBoutin: `trackOn` feels confusing. I think the places that call it currently would be simplified if they just called getCurrentState directly and did like `if (isNewCampaign){...`. The attribution plugin doesn't need to interact with code related to the pageview plugin that way.

AlecBoutin (changes requested):

packages/analytics-browser/src/attribution/campaign-tracker.ts
dnshi: Initially I added "onNewSession" event listener here to trigger this callback when there is a new session. Also we can support single page app in this function, so if a page is changed, we can trigger this callback to send out identity event or page view event. 

Both attribution plugin and pageTracking plugin link to this function. Each time the page state changes (session, campaign, or page view for SPA), we can run the MA strategy again.

packages/analytics-browser/src/attribution/campaign-tracker.ts
```diff
@@ -16,30 +15,39 @@ export class CampaignTracker implements ICampaignTracker {
...
 
   disabled: boolean;
-  trackNewCampaigns: boolean;
-  trackPageViews: boolean;
   excludeReferrers: string[];
   initialEmptyValue: string;
+  resetSessionOnNewCampaign: boolean;
+
+  private _currentCampaign!: Campaign;
+  private _isNewCampaign!: boolean;
```
kevinpagtakhan, line 25: why do we need `!`? i prefer that we do not bypass type rules

packages/analytics-browser/src/attribution/campaign-tracker.ts
AlecBoutin: Ok that makes sense, but onNewSession is a bit of a misnomer since the passed callback is always called, even if there's no new session. Maybe `onPageChange`?

dnshi:
Close it. Open a new PR #213 213

PR #212, opened 2022-09-01
feat: add ingestion_metadata field

liuyang1520:
### Summary
- feat: add ingestion_metadata field

Add this option in the configuration, which can be used in the library wrapper/code generation/dynamic loading cases, for setting the `ingestion_metadata` information.

### Checklist


packages/analytics-browser/src/plugins/context.ts
```diff
@@ -65,6 +65,14 @@ export class Context implements BeforePlugin {
...
       partner_id: this.config.partnerId,
       plan: this.config.plan,
+      ...(this.config.ingestionMetadata && {
+        ingestion_metadata: {
+          ...(this.config.ingestionMetadata.sourceName && { source_name: this.config.ingestionMetadata.sourceName }),
+          ...(this.config.ingestionMetadata.sourceVersion && {
+            source_version: this.config.ingestionMetadata.sourceVersion,
+          }),
+        },
+      }),
```
kevinpagtakhan, line 75: do we need the guards here? does it hurt if we send `ingestion_metadata: {}` vs `ingestion_metadata: undefined`?

```suggestion
      ingestion_metadata: {
        source_name: this.config.ingestionMetadata?.sourceName,
        source_version: this.config.ingestionMetadata?.sourceVersion,
      },
```

packages/analytics-browser-test/test/index.test.ts
```diff
@@ -180,6 +181,49 @@ describe('integration', () => {
       scope.done();
     });
 
+    test('should track event with optional ingestionMetadata option', async () => {
```
kevinpagtakhan, line 184: thank you for adding this

packages/analytics-browser/src/plugins/context.ts
liuyang1520: Thanks for reviewing! Good question, ingestion and pipeline accept the empty value cases for this field without issue. From SDK side, I mentioned we will try to not send this if that's easy to avoid. Also, it slightly improves the performance by reducing the event size, and help the pipeline side to skip some checking logic. If you think this guard is too much, I can remove the inner ones, I think keeping the `this.config.ingestionMetadata &&` is good in most cases. Let me know your thoughts or concerns on this!

kevinpagtakhan: yeah i like that

yuhao900914 (approved):

kevinpagtakhan (approved):

PR #213, opened 2022-09-02
feat: new marketing analytics plugin

dnshi:
### Summary

This PR is to add new marketing analytics plugin

### Checklist


packages/analytics-browser/src/index.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 import client from './browser-client';
-export { createInstance } from './browser-client';
+export { AmplitudeBrowser, createInstance } from './browser-client';
```
kevinpagtakhan, line 3: Why do we need to expose this?

packages/analytics-types/src/config.ts
```diff
@@ -77,16 +77,26 @@ export interface ReactNativeTrackingOptions extends TrackingOptions {
 
 export interface AdditionalBrowserOptions {
   attribution?: AttributionBrowserOptions;
+  trackPageViews?: boolean | PageTrackingBrowserOptions;
```
kevinpagtakhan, line 80: Would this add it to the existing conifg param in init which we are trying to avoid?

packages/analytics-types/src/config.ts
```diff
@@ -77,16 +77,26 @@ export interface ReactNativeTrackingOptions extends TrackingOptions {
...
 }
 
 export interface AttributionBrowserOptions {
   disabled?: boolean;
   excludeReferrers?: string[];
   initialEmptyValue?: string;
+  resetSessionOnNewCampaign?: boolean;
```
kevinpagtakhan, line 87: I think this one will affect core sdk

packages/analytics-browser/src/index.ts
dnshi: I am creating a "analytics-ma-borrowser" package, and it will be extended by MA browser. There is a few issue on ma browser package. I will commit that part later.

packages/analytics-types/src/config.ts
dnshi: I was thinking of adding all types in one place. But it makes sense to host this part in the plugin.

packages/analytics-types/src/config.ts
dnshi: Right. I will move it into plugin.

packages/plugin-ma-tracking/rollup.config.js
```diff
@@ -0,0 +1,93 @@
+import typescript from '@rollup/plugin-typescript';
+import commonjs from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import { terser } from 'rollup-plugin-terser';
+import gzip from 'rollup-plugin-gzip';
+import execute from 'rollup-plugin-execute';
+import { exec } from 'child_process';
+
+const amplitudeSnippet = () => {
```
kevinpagtakhan, line 9: Do we need this?

packages/plugin-ma-tracking/rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+  };
+};
+
+const umd = [
+  {
+    input: 'src/index.ts',
+    output: {
+      name: 'amplitude',
+      file: 'lib/scripts/amplitude-min.umd.js',
```
kevinpagtakhan, line 32: Same name as core sdk?

packages/plugin-ma-tracking/rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+    ],
+  },
+];
+
+const iife = [
+  {
+    input: 'src/snippet-index.ts',
+    output: {
+      name: 'amplitude',
+      file: 'lib/scripts/amplitude-min.js',
```
kevinpagtakhan, line 57: Also here

packages/plugin-ma-tracking/rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+  {
+    output: {
+      name: 'amplitude',
+      file: 'lib/scripts/amplitude-snippet-min.js',
+      format: 'iife',
+    },
+    plugins: [
+      amplitudeSnippet(),
+      terser(),
+      execute('node scripts/version/create-snippet-instructions.js'),
```
kevinpagtakhan, line 87: Do we need these?

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
```diff
@@ -0,0 +1,43 @@
+import { Campaign, CampaignTrackerOptions } from '@amplitude/analytics-types';
+import { CampaignTracker } from '@amplitude/analytics-client-common';
+
+export class AdvancedCampaignTracker extends CampaignTracker {
+  resetSessionOnNewCampaign: boolean;
+
+  private _currentCampaign!: Campaign;
```
kevinpagtakhan, line 7: No ! please

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
```diff
@@ -0,0 +1,43 @@
...
+  private _isNewCampaign = true;
+  ready: Promise<unknown>;
+
+  constructor(apiKey: string, options: CampaignTrackerOptions) {
+    super(apiKey, options);
+
+    this.resetSessionOnNewCampaign = options.resetSessionOnNewCampaign ?? Boolean(options.trackNewCampaigns);
+    this.ready = new Promise((resolve) => resolve(this.refreshCampaignState()));
```
kevinpagtakhan, line 15: Where will this be used?

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+  BrowserConfig,
+  Campaign,
+  Event,
+  PluginType,
+  Storage,
+} from '@amplitude/analytics-types';
+import type { AmplitudeBrowser } from '@amplitude/analytics-browser';
+import { AdvancedCampaignTracker } from './advanced-campaign-tracker';
+
+export class MATrackingPlugin implements BeforePlugin {
```
kevinpagtakhan, line 14: Im thinking if we should just go functions vs classes. Wanted to update other plugins and examples too

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+    const attributionConfig = this.options.attribution ?? {};
+    const pageTrackingConfig =
+      this.options.trackPageViews ??
+      (attributionConfig.trackPageViews && {
+        filter: 'onAttribution',
+      });
+
+    const instance = this.instance;
+    // Share cookie storage with user session storage
+    const storage = config.cookieStorage as unknown as Storage<Campaign>;
```
kevinpagtakhan, line 30: As unknown?

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+    const campaignTracker = new AdvancedCampaignTracker(config.apiKey, {
+      ...attributionConfig,
+      trackPageViews: false,
+      track: /* istanbul ignore next */ () => Promise.resolve(),
+      onNewCampaign: /* istanbul ignore next */ () => () => undefined,
+      storage,
+    });
+
+    // Disable "runAttributionStrategy" function from instance
+    instance.runAttributionStrategy = () => Promise.resolve();
```
kevinpagtakhan, line 41: Lets avoid doing this anywhere please

packages/analytics-types/src/config.ts
```diff
@@ -77,16 +77,26 @@ export interface ReactNativeTrackingOptions extends TrackingOptions {
...
+  /** @deprecated New campaigns are now always tracked by default. Setting this to true also sets resetSessionOnNewCampaign to true. */
   trackNewCampaigns?: boolean;
+  /** @deprecated Use the new top level trackPageViews configuration instead. */
   trackPageViews?: boolean;
 }
 
+export interface PageTrackingBrowserOptions {
+  filter?: PageTrackingFilter;
```
kevinpagtakhan, line 95: Filter as a name is not very clear

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    // Web Attribution tracking
+    if (!attributionConfig.disabled) {
+      void campaignTracker.onPageChange(async ({ isNewCampaign, currentCampaign }) => {
+        if (isNewCampaign) {
+          if (attributionConfig.resetSessionOnNewCampaign) {
+            instance.setSessionId(Date.now());
+          }
+          await instance.track(campaignTracker.createCampaignEvent(currentCampaign));
```
kevinpagtakhan, line 50: Do we need to await?

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+  Storage,
+} from '@amplitude/analytics-types';
+import type { AmplitudeBrowser } from '@amplitude/analytics-browser';
+import { AdvancedCampaignTracker } from './advanced-campaign-tracker';
+
+export class MATrackingPlugin implements BeforePlugin {
+  name = 'MATracking';
+  type = PluginType.BEFORE as const;
+
+  constructor(private readonly instance: AmplitudeBrowser, private readonly options: AdditionalBrowserOptions = {}) {}
```
justin-fiedler, line 18: If this only works in browser we should add a postfix to the plugin name to specify `plugin-marketing-analytics-browser`

packages/analytics-types/src/campaign.ts
```diff
@@ -33,7 +33,14 @@ export interface CampaignTrackerOptions extends AttributionBrowserOptions {
 }
 
 export interface CampaignTracker extends CampaignTrackerOptions {
+  isNewCampaign(currentCampaign: Campaign, previousCampaign: Campaign): boolean;
```
justin-fiedler, line 36: For types, I think it would make sense to have each plugin export its own types. Also each top level client could reexport the types to prevent the addition package import by end users. If we follow that model we could export these options from the MA plugin, and re-export them from analytics-ma-browser as part of the Config.

packages/analytics-types/src/config.ts
justin-fiedler: I was thinking we add them to the config only in the `analytics-ma-browser` package. See other comment about types.

justin-fiedler (changes requested):

packages/plugin-ma-tracking/rollup.config.js
dnshi: We might still need it for snippet because we need another script for plugin.

dnshi: I will remove this file for now. Will add it back in separate PR.

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
dnshi: This value is created while calling `refreshCampaignState` function, but the linter complains about it. `!` is used to declare it is not undefined.

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
dnshi: When `getCurrentState` is called. It is to make sure `isNewCampaign` and `currentCampaign` are ready.

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: It would be great to support both classes and functions.

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: cookieStorage is `Storage<UserSession>`. In order to support `Storage<Campaign>`, need to set `as unknown` first.

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: Any suggestion of preventing calling `runAttributionStrategy` from core SDK?

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: It is optional. `void` also works.

packages/analytics-types/src/config.ts
kevinpagtakhan: i think we should add the plugin specific types to the plugins package and the wrapper package can import it from there

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
kevinpagtakhan: compiler or linter is correct when it called out that this property can be undefined. we should add a guard when accessing, instead of bypassing type checks

packages/plugin-ma-tracking/src/ma-tracking.ts
kevinpagtakhan: both are supported but i think functions are just more industry standard and better to show as examples to customers

packages/plugin-ma-tracking/src/ma-tracking.ts
kevinpagtakhan: again, this bypasses type checking. defeats the purpose of using typescript

packages/plugin-ma-tracking/src/ma-tracking.ts
kevinpagtakhan: shared here: https://amplitude.slack.com/archives/D035W3B4CJC/p1661982534275369

packages/analytics-browser/src/index.ts
kevinpagtakhan: is is possible to add these changes later when it's actually needed. it's better to keep related changes in the same commit/pr

packages/analytics-types/src/config.ts
kevinpagtakhan: also curious about the decision to make this `boolean | PageTrackingBrowserOptions` how does that work?

packages/analytics-types/src/config.ts
```diff
@@ -77,16 +77,26 @@ export interface ReactNativeTrackingOptions extends TrackingOptions {
...
+}
+
+export type PageTrackingFilter = 'onAttribution' | (() => boolean) | undefined;
```
kevinpagtakhan, line 98: i think it's better to do

```ts
export type PageTrackingFilter = 'onAttribution' | (() => boolean);
```

and leave it to the usage whether to allow undefined or not like

```ts
filter: PageTrackingFilter
```

or 

```ts
filter?: PageTrackingFilter
```

packages/plugin-ma-tracking/src/ma-tracking.ts
kevinpagtakhan: i think we can use `createFlexibleStorage` here instead of reusing cookie storage

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+        filter: 'onAttribution',
+      });
+
+    const instance = this.instance;
+    // Share cookie storage with user session storage
+    const storage = config.cookieStorage as unknown as Storage<Campaign>;
+
+    const campaignTracker = new AdvancedCampaignTracker(config.apiKey, {
+      ...attributionConfig,
+      trackPageViews: false,
```
kevinpagtakhan, line 34: looks like this is not used and can be removed?

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+      track: /* istanbul ignore next */ () => Promise.resolve(),
+      onNewCampaign: /* istanbul ignore next */ () => () => undefined,
```
kevinpagtakhan, line 36: hmm looks like these are only added for backwards compatibility? maybe we can just create a separate campaign tracker and not extend the old one

packages/plugin-ma-tracking/src/ma-tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+            instance.setSessionId(Date.now());
+          }
+          await instance.track(campaignTracker.createCampaignEvent(currentCampaign));
+        }
+      });
+    }
+
+    // Page View tracking
+    if (pageTrackingConfig) {
+      const filter = typeof pageTrackingConfig === 'boolean' || pageTrackingConfig.filter;
```
kevinpagtakhan, line 57: imo filter type being `boolean` and `{ filter: ... }` is a little hard to follow. do we need the boolean value or can we just have a default value for the object and its keys?

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
```diff
@@ -0,0 +1,43 @@
+import { Campaign, CampaignTrackerOptions } from '@amplitude/analytics-types';
+import { CampaignTracker } from '@amplitude/analytics-client-common';
+
+export class AdvancedCampaignTracker extends CampaignTracker {
```
kevinpagtakhan, line 4: this class seems to act as a bridge for `ma-tracking` and `CampaignTracker` and comes out a little odd as it tries to control two different versions of a feature. i'd recommend that we write out our own tracker for the plugin and let the old logic live on it's own. this way they're always separate and no changes impact both.

packages/plugin-ma-tracking/src/ma-tracking.ts
AlecBoutin: Given that page tracking is opt in and has no required options, it seems weird to pass `trackPageViews: {}` to enable it rather than supporting `true`

AlecBoutin: Since this tracking is now in a plugin, we could break the page view tracking options out from the on/off toggle pretty easily, ie. have `trackPageViews` and `pageViewTrackingOptions` as separate plugin configuration keys.

AlecBoutin:
Dean did you want to implement the additional page view tracking event properties in this PR? Alternately I'm happy to implement that improvement once this PR is merged https://amplitude.atlassian.net/wiki/spaces/GOV/pages/1874427995/Finalizing+Pagetracking+attribution+SDK+changes#Page-View-Tracking-Event-Properties

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: `createFlexibleStorage` is not a shared component. We update the plugin when it is moved into common package.

packages/analytics-types/src/config.ts
dnshi: `filter` is the name from the design. What would be suggested variable name?

packages/analytics-types/src/campaign.ts
dnshi: Removed the types

packages/analytics-types/src/config.ts
dnshi: Done

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: Removed

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: It is simpler to extend the existing one because both page tracking and web attribution plugins use the same functionality. 

Or make these two functions optional and add default values to these properties.

dnshi:
> Dean did you want to implement the additional page view tracking event properties in this PR? Alternately I'm happy to implement that improvement once this PR is merged https://amplitude.atlassian.net/wiki/spaces/GOV/pages/1874427995/Finalizing+Pagetracking+attribution+SDK+changes#Page-View-Tracking-Event-Properties

Maybe do it in a separate PR, after this is merged.

packages/analytics-browser/src/index.ts
dnshi: Removed any code related to `analytics-ma-browser` package.

packages/analytics-types/src/config.ts
dnshi: Removed it from typings. The boolean option will be handled in `analytics-ma-browser`. The options value is optional in the page view tracking plugin. Adding plugin means enabling the functionality.

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
dnshi: Added a default value

packages/plugin-ma-tracking/src/ma-tracking.ts
dnshi: `trackPageViews` is valid for `analytics-ma-browser` package. For the plugin, it does not take boolean value as an option.

packages/plugin-ma-tracking/src/advanced-campaign-tracker.ts
dnshi: In stead of extending the `CampaignTracker`, I initialize and use the `CampaignTracker` instance directly because it contains enough methods I need.

packages/analytics-types/src/config.ts
```diff
@@ -83,7 +84,9 @@ export interface AttributionBrowserOptions {
   disabled?: boolean;
   excludeReferrers?: string[];
   initialEmptyValue?: string;
+  /** @deprecated New campaigns are now always tracked by default. Setting this to true also sets resetSessionOnNewCampaign to true. */
   trackNewCampaigns?: boolean;
+  /** @deprecated Use the new top level trackPageViews configuration instead. */
   trackPageViews?: boolean;
```
kevinpagtakhan, line 90: these configs are technically not deprecated right? but are overwritten by plugin?

packages/plugin-page-view-tracking/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@amplitude/plugin-page-view-tracking",
```
kevinpagtakhan, line 2: let's get this name approved by dx please

packages/analytics-types/src/config.ts
AlecBoutin: Wouldn't deprecating them reduce confusion? since new implementers should use the plugin

packages/plugin-page-view-tracking/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { PageViewTrackingPlugin } from './page-view-tracking';
+export * from './typings/page-view-tracking';
```
kevinpagtakhan, line 2: Should we make type exposure similar to other packages?

packages/plugin-page-view-tracking/src/page-view-tracking.ts
```diff
@@ -0,0 +1,75 @@
...
+export class PageViewTrackingPlugin implements BeforePlugin {
+  name = 'page-view-tracking';
+  type = PluginType.BEFORE as const;
+
+  constructor(
+    private readonly instance: BrowserClient | IAmplitudeBrowser,
+    private readonly options: PageTrackingBrowserOptions = {},
+  ) {}
+
+  async setup(config: BrowserConfig): Promise<undefined> {
```
kevinpagtakhan, line 24: we can do implicit return if we do the following

```suggestion
  async setup(config: BrowserConfig): Promise<void> {
```

packages/plugin-page-view-tracking/package.json
justin-fiedler, line 2: Unless this is supported by all platforms I think we should include a postfix to be explicit `@amplitude/plugin-page-view-tracking-browser`

packages/plugin-page-view-tracking/src/page-view-tracking.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '@amplitude/analytics-types';
+import { PluginCampaignTracker } from './plugin-campaign-tracker';
+import { PageTrackingBrowserOptions } from './typings/page-view-tracking';
+
+export class PageViewTrackingPlugin implements BeforePlugin {
+  name = 'page-view-tracking';
+  type = PluginType.BEFORE as const;
+
+  constructor(
+    private readonly instance: BrowserClient | IAmplitudeBrowser,
```
kevinpagtakhan, line 20: Do we need `IAmplitudeBrowser`? I feel like customers do not have access to this one anyway
```suggestion
    private readonly instance: BrowserClient | IAmplitudeBrowser,
```

packages/plugin-page-view-tracking/src/page-view-tracking.ts
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    // Turn off sending page view event by "runAttributionStrategy" function
+    if (typeof attributionConfig === 'object') {
+      attributionConfig.trackPageViews = false;
+    }
+
+    const instance = this.instance;
+    // Share cookie storage with user session storage
+    const storage = config.cookieStorage as unknown as Storage<Campaign>;
```
kevinpagtakhan, line 38: I think we should just use call `createFlexibleStorage` here instead of reusing a storage for a different purpose. It should yield the same storage provider

packages/plugin-page-view-tracking/src/typings/page-view-tracking.ts
```diff
@@ -0,0 +1,5 @@
+export interface PageTrackingBrowserOptions {
+  filter?: PageTrackingFilter;
```
justin-fiedler, line 2: Depending on how we export this in `analytics-ma` we might want to give this a more descriptive name e.g. `pageTrackingFilter` so that it makes sense in the top-level config.

Do you plan to have the plugin config namespaced in the ma-client?
```
amplitude.init(apiKey, undefined, {
 pageTracking: { filter: 'onAttribution' }
}
```
or
```
amplitude.init(apiKey, undefined, {
  pageTrackingFilter: 'onAttribution'
}
```

I prefer the first one.

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,53 @@
...
+export class PluginCampaignTracker {
+  private _currentCampaign: Campaign = {
+    ...BASE_CAMPAIGN,
+  };
+  private _isNewCampaign = true;
+  ready: Promise<unknown>;
+  campaignTracker: CampaignTracker;
+
+  constructor(apiKey: string, storage: Storage<Campaign>, options: AttributionBrowserOptions) {
+    this.campaignTracker = new CampaignTracker(apiKey, {
```
kevinpagtakhan, line 15: i'm worried that with this shared with browser bundled and plugin bundled feature, it will be harder to maintain since you want to make sure they one does not affect the other. while many code is shared today, i think they have diverged already and should be maintained separately

packages/plugin-web-attribution/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { WebAttributionPlugin } from './web-attribution';
+export * from './typings/web-attribution';
```
kevinpagtakhan, line 2: same here for type exposure

packages/plugin-web-attribution/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@amplitude/plugin-web-attribution",
```
justin-fiedler, line 2: Even though the name implies this is Browser only I still think we should standardize Plugin naming to include the support platform (`-browser`) unless the plugin is platform agnostic.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -18,8 +17,10 @@ import { Context } from './plugins/context';
 import { useBrowserConfig, createTransport, createFlexibleStorage } from './config';
 import { parseOldCookies } from './cookie-migration';
 
+export type IAmplitudeBrowser = AmplitudeBrowser;
```
kevinpagtakhan, line 20: probably do not need to expose if `AmplitudeBrowser` is private?

packages/plugin-page-view-tracking/test/page-view-tracking.test.ts
```diff
@@ -0,0 +1,311 @@
...
+        const instance = createInstance();
+        const track = jest.spyOn(instance, 'track').mockReturnValueOnce({
+          promise: Promise.resolve({
+            code: 200,
+            message: '',
+            event: {
+              event_type: 'event_type',
+            },
+          }),
+        });
```
justin-fiedler, line 45: Maybe do this in beforeEach?

packages/analytics-types/src/config.ts
kevinpagtakhan: i guess i was thinking that for folks not interested in the plugin, these are not deprecated. plus the replacement `resetSessionOnNewCampaign` is not in core.

alternatively, i think we can just deprecate the entire attribution config and point them to the plugin

packages/plugin-page-view-tracking/src/index.ts
dnshi: We will need it when we are going to build `analytics-ma-browser`.  The type will be used for the init options.

packages/plugin-page-view-tracking/src/page-view-tracking.ts
justin-fiedler, line 20: Nice way to avoid the event bus for now. I think this approach works for this project. In the future though this would prevent us from adding the Plugin during init().

packages/plugin-page-view-tracking/src/page-view-tracking.ts
```diff
@@ -0,0 +1,75 @@
...
+    const attributionConfig = config.attribution;
+    const pageTrackingConfig = {
+      filter: attributionConfig?.trackPageViews ? 'onAttribution' : undefined,
+      ...this.options,
+    };
+
+    // Turn off sending page view event by "runAttributionStrategy" function
+    if (typeof attributionConfig === 'object') {
+      attributionConfig.trackPageViews = false;
+    }
```
kevinpagtakhan, line 34:
```suggestion
    if (config.attribution) {
      config.attribution.trackPageViews = false;
    }
```

justin-fiedler (changes requested):

packages/plugin-page-view-tracking/src/page-view-tracking.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor(
+    private readonly instance: BrowserClient | IAmplitudeBrowser,
+    private readonly options: PageTrackingBrowserOptions = {},
+  ) {}
+
+  async setup(config: BrowserConfig): Promise<undefined> {
+    const attributionConfig = config.attribution;
+    const pageTrackingConfig = {
+      filter: attributionConfig?.trackPageViews ? 'onAttribution' : undefined,
```
kevinpagtakhan, line 27: hmm i think we shouldnt use config.attribution to configure plugin. otherwise there are more combos. either purely core or purely plugin only

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Right. Because we need to add plugins in the `analytics-ma-browser` init function and pass browser options to plugins, then it would need `IAmplitudeBrowser`. 

```js
async init(apiKey, deviceId, options) {
  this.add(new AttributionPlugin(this, options?.attribution)); // the type of "this" is IAmplitudeBrowser
  ...
}
```

packages/plugin-page-view-tracking/src/page-view-tracking.ts
kevinpagtakhan: or even further

```suggestion
    if (config.attribution) {
      config.attribution.trackPageViews = false;
      config.loggerProvider.warn('put some warning message here')
    }
```

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Right. `createFlexibleStorage` is not exposed yet. Should we expose it?

packages/plugin-page-view-tracking/test/page-view-tracking.test.ts
kevinpagtakhan: perhaps better for making it DRY. i'd prefer that spies and mocks are in test itself since they should be custom to the test case.

packages/plugin-page-view-tracking/src/page-view-tracking.ts
kevinpagtakhan: yes please. we can expose anything from analytics-client-common since it's an internal package and customers shouldnt be using it directly

packages/plugin-page-view-tracking/src/index.ts
kevinpagtakhan: not saying we shouldn't expose. asking if we should make it similar to https://github.com/amplitude/Amplitude-TypeScript/blob/3909e7fcac0429eb047a9cf04ee96c36fc85e573/packages/analytics-browser/src/index.ts#L27

packages/plugin-page-view-tracking/src/page-view-tracking.ts
kevinpagtakhan: hmm can we commit this later when we do it? wanted to think about this since it's a little confusing for customers to see a foreign interface plus promise interface is different

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
dnshi: We only use a few functions from `CampaignTracker`. For any additional functionality, we can stay in the plugin. I am feeling `CampaignTracker` is more like a tool which provides some basic campaign functions. 

If MA team needs more features or function, only plugins need to be updated. We might not want to modify or change `CampaignTracker` with new features.

packages/plugin-web-attribution/src/index.ts
dnshi: The type will be used in the `analytics-ma-browser`.

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
kevinpagtakhan: i feel like campaign parser is the generic one, while campaign tracker is the one that's different. the config alone does not match

packages/analytics-types/src/config.ts
dnshi: Pointing them to the plugin is a good idea.

packages/plugin-page-view-tracking/src/index.ts
dnshi: Oh, I see what you means. Will change. Thanks.

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Sure, I will remove this part from this PR.

packages/analytics-browser/src/browser-client.ts
dnshi: I will remove it from this PR.

packages/plugin-page-view-tracking/src/page-view-tracking.ts
kevinpagtakhan: hmm sure but we do not repurpose storage instances in other places

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Sounds good.

packages/plugin-page-view-tracking/package.json
dnshi: Sure. Let's call `@amplitude/plugin-page-view-tracking-browser`

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...options,
+      trackPageViews: false,
+      track: /* istanbul ignore next */ () => Promise.resolve(),
+      onNewCampaign: /* istanbul ignore next */ () => () => undefined,
+      storage,
+    });
+    this.ready = new Promise((resolve) => resolve(this.refreshCampaignState()));
```
kevinpagtakhan, line 22: i think this is an example of why we should not write this as a class since no way to await on constructor

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,53 @@
+import { AttributionBrowserOptions, Campaign } from '@amplitude/analytics-types';
+import { CampaignTracker } from '@amplitude/analytics-client-common';
+import { BASE_CAMPAIGN } from './constant';
+import { Storage } from '@amplitude/analytics-types';
+
+export class PluginCampaignTracker {
```
kevinpagtakhan, line 6: also feels weird that we have campaign tracker in page view tracking since this plugin does not track campaigns. might just be a naming concern

packages/plugin-page-view-tracking/src/typings/page-view-tracking.ts
dnshi: It would be the first one. There are a few options:
```ts
amplitude.init(apiKey, undefined, {
 trackPageViews: { filter: 'onAttribution' }
}
```

```ts
amplitude.init(apiKey, undefined, {
 trackPageViews: { filter: () => true }
}
```

```ts
amplitude.init(apiKey, undefined, {
 trackPageViews: true
}
```

Currently we only have filter. It might have more options for `trackPageViews` in next a few iteration.

packages/plugin-web-attribution/src/typings/web-attribution.ts
```diff
@@ -0,0 +1,5 @@
+import { AttributionBrowserOptions } from '@amplitude/analytics-types';
+
+export interface AttributionPluginOptions extends AttributionBrowserOptions {
```
kevinpagtakhan, line 3: i prefer that we do not extend so we dont accidentally change in both places

packages/plugin-web-attribution/package.json
dnshi: Sure, I will add it to the package name.

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...options,
+      trackPageViews: false,
+      track: /* istanbul ignore next */ () => Promise.resolve(),
+      onNewCampaign: /* istanbul ignore next */ () => () => undefined,
+      storage,
+    });
+    this.ready = new Promise((resolve) => resolve(this.refreshCampaignState()));
+  }
+
+  async onPageChange(callback: (state: { isNewCampaign: boolean; currentCampaign: Campaign }) => Promise<unknown>) {
```
kevinpagtakhan, line 25: why is the return unknown?

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Agree. Initially I was thinking of adding instance as a parameter when we call `Plugin.setup` function, so that we did not need to pass the instance. But it would make inconsistency with other core SDKs.

packages/plugin-page-view-tracking/src/page-view-tracking.ts
dnshi: Ideally each plugin should have its own config. This project is a bit different that this plugin depends on attribution configuration.

packages/plugin-web-attribution/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+      track: /* istanbul ignore next */ () => Promise.resolve(),
+      onNewCampaign: /* istanbul ignore next */ () => () => undefined,
+      storage,
+    });
+    this.ready = new Promise((resolve) => resolve(this.refreshCampaignState()));
+  }
+
+  createCampaignEvent(currentCampaign: Campaign) {
+    return this.campaignTracker.createCampaignEvent(currentCampaign);
+  }
```
kevinpagtakhan, line 27: do we need this?

packages/plugin-page-view-tracking/src/page-view-tracking.ts
kevinpagtakhan: but wouldn't the customer pass a separate config in the plugin anyway?

packages/plugin-page-view-tracking/src/plugin-campaign-tracker.ts
dnshi: Okay, let me try to make it in a function.

packages/analytics-types/src/config.ts
AlecBoutin: Agree on deprecating entire config

packages/analytics-types/src/config.ts
```diff
@@ -32,6 +32,7 @@ export interface Config {
 
 export interface BrowserConfig extends Config {
   appVersion?: string;
+  attribution?: AttributionBrowserOptions;
```
dnshi, line 35: Add deprecated message here.

dnshi: Remove deprecation notes from AttributionBrowserOptions

packages/plugin-page-view-tracking-browser/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "description": "",
+  "author": "Amplitude Inc",
+  "homepage": "https://github.com/amplitude/Amplitude-TypeScript",
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "lib/esm/index.d.ts",
+  "publishConfig": {
+    "access": "public"
+  },
```
dnshi, line 13: Use `private: true` instead

packages/plugin-web-attribution-browser/src/typings/web-attribution.ts
```diff
@@ -0,0 +1,8 @@
+import { AttributionBrowserOptions } from '@amplitude/analytics-types';
+
+export interface AttributionPluginOptions extends AttributionBrowserOptions {
```
dnshi, line 3: Remove `AttributionBrowserOptions`

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -0,0 +1,78 @@
...
+  return pageViewEvent;
+};
+
+const isCampaignEvent = (event: Event) => {
+  if (event.event_type === '$identify' && event.user_properties) {
+    const properties = event.user_properties as IdentifyUserProperties;
+    const $set = properties[IdentifyOperation.SET] || {};
+    const $unset = properties[IdentifyOperation.UNSET] || {};
+    const userProperties = [...Object.keys($set), ...Object.keys($unset)];
+    return Object.keys(BASE_CAMPAIGN).some((r) => userProperties.includes(r));
```
dnshi, line 75: `Object.keys(BASE_CAMPAIGN).every((r) => userProperties.includes(r))`

packages/plugin-web-attribution-browser/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "description": "",
+  "author": "Amplitude Inc",
+  "homepage": "https://github.com/amplitude/Amplitude-TypeScript",
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "lib/esm/index.d.ts",
+  "publishConfig": {
+    "access": "public"
+  },
```
dnshi, line 13: Use `private: true` instead

packages/plugin-web-attribution-browser/src/plugin-campaign-tracker.ts
```diff
@@ -0,0 +1,42 @@
...
+  createCampaignEvent(currentCampaign: Campaign) {
+    return this.campaignTracker.createCampaignEvent(currentCampaign);
+  }
+
+  async onPageChange(callback: (state: { isNewCampaign: boolean; currentCampaign: Campaign }) => Promise<unknown>) {
+    await callback(await this.getCurrentState());
+
+    const currentState = await this.getCurrentState();
+    if (currentState.isNewCampaign) {
+      await this.campaignTracker.saveCampaignToStorage(currentState.currentCampaign);
```
dnshi, line 27: Use `void` instead of `await`

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -0,0 +1,52 @@
...
+  return {
+    name: 'web-attribution',
+    type: PluginType.BEFORE,
+
+    setup: async (config: BrowserConfig) => {
+      // Merge plugin options and attribution config from BrowserConfig
+      const attributionConfig = (config.attribution = Object.assign({}, config.attribution, options));
+
+      // Disable "runAttributionStrategy" function
+      attributionConfig.disabled = true;
```
dnshi, line 23: Use `config.attribution` directly

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -0,0 +1,52 @@
...
+import { AttributionPluginOptions } from './typings/web-attribution';
+
+export const webAttributionPlugin = (instance: BrowserClient, options: AttributionPluginOptions = {}): BeforePlugin => {
+  return {
+    name: 'web-attribution',
+    type: PluginType.BEFORE,
+
+    setup: async (config: BrowserConfig) => {
+      // Merge plugin options and attribution config from BrowserConfig
+      const attributionConfig = (config.attribution = Object.assign({}, config.attribution, options));
```
dnshi, line 20: `const attributionConfig = options`

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -0,0 +1,52 @@
...
+
+      // Disable "runAttributionStrategy" function
+      attributionConfig.disabled = true;
+
+      // New campaigns are now always tracked by default. Setting trackNewCampaigns to true also sets resetSessionOnNewCampaign to true.
+      if (attributionConfig.trackNewCampaigns && typeof attributionConfig.resetSessionOnNewCampaign === 'undefined') {
+        attributionConfig.resetSessionOnNewCampaign = true;
+      }
```
dnshi, line 28: Remove this block.

kevinpagtakhan (approved):

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -0,0 +1,78 @@
...
+        (typeof pageTrackingConfig.trackOn === 'function' && pageTrackingConfig.trackOn())
+      ) {
+        instance.track(createPageViewEvent());
+      }
+    },
+
+    execute: async (event: Event) => {
+      if (pageTrackingConfig.trackOn === 'attribution' && isCampaignEvent(event)) {
+        const pageViewEvent = createPageViewEvent();
+        event.event_type = pageViewEvent.event_type;
```
Mercy811, line 46: Hi @dnshi, any chance that you still remember why we override the event type from identify to page view? It feel makes more sense to keep the identify event and append attributions to the page view's event properties so that the identify event can actually set the user properties.

dnshi: Basically if the event is a campaign event, then the event type should match page view event type. If the event is not a campaign event or SDK does not track on attribution, then this condition is skipped.

Mercy811: > Basically if the event is a campaign event, then the event type should match page view event type.
In this case, there will be no identify event and the user properties will not update the user properties map in the backend and not take effect in the following events. Note that only `$identify` event can set user properties.

dnshi: I think there were some changes, which were different from the version I worked on. Feel free to update the code to match current scenarios. Thanks.

PR #216, opened 2022-09-14
feat: add campaign params to page view events

AlecBoutin:
### Summary

Adds standard campaign parameters to page view events, allowing additional flexibility for users making campaign reporting queries.

### Checklist


packages/analytics-client-common/src/attribution/campaign-parser.ts
```diff
@@ -26,6 +26,8 @@ import {
 export class CampaignParser implements ICampaignParser {
   utmCookieStorage = new UTMCookie();
 
+  constructor(protected queryStringParsingOnly: boolean = false) {}
```
AlecBoutin, line 29: Wasn't sure what to do here, since utm's stored in cookies is essentially behavior. Other possibilities I considered:
1) Add an optional options object to CampaignParser which includes queryStringParsingOnly, and could be expanded in the future.
2) Copy CampaignParser into the page tracking plugin and remove the cookie code.

Since CampaignParser isn't really a documented part of our SDK, I figured changes like this are less sensitive, but open to other solutions.

dnshi (approved):

packages/analytics-client-common/src/attribution/campaign-parser.ts
kevinpagtakhan: Should we just remove utmz?

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -58,9 +80,11 @@ const createPageViewEvent = (): Event => {
   const pageViewEvent: BaseEvent = {
     event_type: 'Page View',
     event_properties: {
-      page_title: /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
+      page_domain: location.hostname,
```
kevinpagtakhan, line 83: would just recommend adding the typeof check for location before accessing

kevinpagtakhan (approved):

PR #224, opened 2022-09-20
feat: add single page app page view tracking option

AlecBoutin:
### Checklist

### DIFF
--- a/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
+++ b/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
@@ -32,9 +32,42 @@ export const pageViewTrackingPlugin = (
     return campaignParams;
   };
 
-  return {
+  const shouldTrackOnPageLoad = () =>
+    typeof pageTrackingConfig.trackOn === 'undefined' ||
+    (typeof pageTrackingConfig.trackOn === 'function' && pageTrackingConfig.trackOn());
+
+  const createPageViewEvent = async (): Promise<Event> => {
+    const pageViewEvent: BaseEvent = {
+      event_type: 'Page View',
+      event_properties: {
+        ...(await getCampaignParamsForPageViewEvent()),
+        page_domain: /* istanbul ignore next */ (typeof location !== 'undefined' && location.hostname) || '',
+        page_location: /* istanbul ignore next */ (typeof location !== 'undefined' && location.href) || '',
+        page_path: /* istanbul ignore next */ (typeof location !== 'undefined' && location.pathname) || '',
+        page_title: /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
+        page_url: /* istanbul ignore next */ (typeof location !== 'undefined' && location.href.split('?')[0]) || '',
+      },
+    };
+    return pageViewEvent;
+  };
+
+  let previousURL: string | null = null;
+
+  const trackHistoryPageView = async (): Promise<void> => {
+    const newURL = location.href;
+
+    if (
+      shouldTrackHistoryPageView(pageTrackingConfig.trackHistoryChanges, newURL, previousURL || '') &&
+      shouldTrackOnPageLoad()
+    ) {
+      instance.track(await createPageViewEvent());
+    }
+    previousURL = newURL;
+  };
+
+  const plugin = {
     name: 'page-view-tracking',
-    type: PluginType.ENRICHMENT,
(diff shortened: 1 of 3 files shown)


packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -48,46 +81,48 @@ export const pageViewTrackingPlugin = (
...
+      if (pageTrackingConfig.trackHistoryChanges) {
+        window.addEventListener('popstate', () => {
+          void trackHistoryPageView();
+        });
+
+        // There is no global browser listener for changes to history, so we have
+        // to modify pushState directly.
+        // https://stackoverflow.com/a/64927639
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        window.history.pushState = new Proxy(window.history.pushState, {
```
dnshi, line 98: I am curious that if a user pushes a history state and sdk receives the state, do we send the page view event twice?

dnshi (approved):
I am good overall. The only concern is that when a history state is pushed, make sure not send the event twice.

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
AlecBoutin: Believe popstate happens when the user uses the forward or back buttons, pushState only happens from things like <Link>s

AlecBoutin: Ran some tests in a fresh create-react-app while working on this to make sure.

packages/plugin-page-view-tracking-browser/src/typings/page-view-tracking.ts
```diff
@@ -1,5 +1,8 @@
 export interface PageTrackingBrowserOptions {
   trackOn?: PageTrackingTrackOn;
+  trackHistoryChanges?: PageTrackingHistoryChanges;
 }
 
 export type PageTrackingTrackOn = 'attribution' | (() => boolean);
+
+export type PageTrackingHistoryChanges = 'all' | 'pathOnly';
```
kevinpagtakhan, line 8: im curious if we need to filter here. im thinking the popstate event and pushState method should be enough for us to tell if they navigate to a different page. i'd be curious if we can expand on PageTrackingTrackOn fn to work with this kind of filter

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -48,46 +81,48 @@ export const pageViewTrackingPlugin = (
...
-          ...(await getCampaignParamsForPageViewEvent()),
-          ...event.event_properties,
-        };
-        instance.track(event);
+      if (shouldTrackOnPageLoad()) {
+        instance.track(await createPageViewEvent());
+      }
+
+      /* istanbul ignore next */
+      if (pageTrackingConfig.trackHistoryChanges) {
```
kevinpagtakhan, line 89: curious if we truly need this, or if we can just treat SPA and non-SPA page transitions as the same

packages/plugin-page-view-tracking-browser/src/typings/page-view-tracking.ts
AlecBoutin: So my thinking was SPA's often use query params or hashes to save things that update frequently like user searches. Implementers should probably be using replaceState not pushState for things like that most of the time though, but seems like an easy mistake to make. Overall I added this to make he implementer feel a little safer they're not going to blow up their event volume with meaningless views. You're right this wouldn't be hard to implement themselves in trackOn. It's definitely a nice to have.

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
AlecBoutin: Like - why don't we track SPA page views by default?

kevinpagtakhan: Yeah

AlecBoutin: Other MA tools don't do it by default, so it's unexpected. Many customers considering switching to amplitude for marketing analytics want to see data parity with GA/Adobe before switching. Also replacing the global pushState is a bit of a hack, so I'm hesitant to do it by default.

PR #230, opened 2022-09-23
build: add integration testing for marketing analytics SDK

kevinpagtakhan:
### Summary

Adds integration testing (plugins + browser SDK). The tests verifies the following cases:

* Track attribution only
* Track page view on attribution
* Track page view on load
* Track attribution and page view on load separately

### Checklist


dnshi (approved):

packages/analytics-ma-browser/test/e2e.test.ts
```diff
@@ -0,0 +1,318 @@
...
+        expect(requestBody).toEqual({
+          api_key: 'API_KEY',
+          events: [
+            {
+              device_id: expectString,
+              event_id: 0,
+              event_properties: {
+                page_location: 'http://localhost/',
+                page_path: '/',
+                page_title: '',
```
AlecBoutin, line 130: would expect `page_domain` and `page_url` to also appear here
image

AlecBoutin (approved):
Looks good!

packages/analytics-ma-browser/test/e2e.test.ts
kevinpagtakhan: correct it should be there. let me check. might be because it has not been rebased yet

kevinpagtakhan: i updated the test and these field are being asserted including the additional click ids

Issue #238, opened 2022-09-29
`init()` function throws Unhandled Rejection TypeError when API KEY is missing


pitahorn:
Hi,

We encountered an issue while working with the package on a dev environment. We wanted Amplitude to track events in production only, so in development and staging environments the Amplitude API KEY would not be provided. We wrapped this logic in a `try/catch` sequence, but it throws an Unhandled Rejection TypeError nevertheless.

Function call on our code:
```js
/**
 * Initialize Amplitude / Google Analytics / Facebook Pixel / etc
 */
export function InitializeTrackers() {
  try {
    init(process.env.REACT_APP_AMPLITUDE_KEY);  // <--- Not provided on development or staging
  } catch (e) {
    console.log('Failed to load amplitude analytics', e);
  }
}
```

## Expected Behavior
The error should be handled downstream.

## Current Behavior
The error seems to be coming from this return in `src/cookie-name.ts:8`: 
```
return `${AMPLITUDE_PREFIX.toLowerCase()}_${apiKey.substring(0, 6)}`;
```
Since apiKey is not provided, the substring is problematic.
<img width="1042" alt="image" src="https://user-images.githubusercontent.com/26127375/193128417-b51ef81c-5387-4507-84cf-1f1b54461268.png">

## Possible Solution
- Having a fallback base value for apiKey so the error can be avoided.

## Steps to Reproduce
We are using a CRA app with TypeScript (^4.3.5) and React (^17.0.2),
1. Install the `@amplitude/analytics-browser` package on your project.
3. Import the `init()` method in your main App. 
4. Initialize with a potentially undefined value:
    ```js
    import { init } from '@amplitude/analytics-browser';
    // ... 
    const API_KEY = process.env.REACT_APP_AMPLITUDE_KEY;
    try {
        init(API_KEY);
    } catch(e) {
        console.log('Amplitude not available', e);
    }
    ```
6. Load your app and watch Unhandled Rejection TypeError mentioned above.

## Environment
- JS SDK Version: `node: >=16.0.0`
- Installation Method: yarn 
- `@amplitude/analytics-browser@^1.4.0`
- Browser and Version: Chrome 105.0.5195.125 (Official Build) (arm64)

Thanks! 👋🏼

liuyang1520:
Hi, @pitahorn ,

Thanks for reporting this! I think this issue can be solved with:
```javascript
export function InitializeTrackers() {
  try {
[...]

PR #240, opened 2022-09-30
fix: cover the case when apiKey is missing in the runtime

liuyang1520:
### Summary
Fix the edge case that `apiKey` might be null or undefined in the runtime.

Refer to issue: https://github.com/amplitude/Amplitude-TypeScript/issues/238

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -26,6 +26,7 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
     this.initializing = true;
 
     // Step 1: Read cookies stored by old SDK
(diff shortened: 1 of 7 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -26,6 +26,7 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
     this.initializing = true;
 
     // Step 1: Read cookies stored by old SDK
+    apiKey = apiKey ?? '';
```
kevinpagtakhan, line 29: alternatively, you can also modify the param list and provide a default when none is provided

```
async init(apiKey = '', ...
```

kevinpagtakhan (approved):

PR #243, opened 2022-10-05
fix: update userId in identify call if eventOptions contains userId

bohan-amplitude:
### Summary

update userId in identify call if eventOptions contains userId

### Checklist


kevinpagtakhan:
Is this how other SDKs work?

bohan-amplitude:
I think @yuhao900914 mentioned she had a PR for [kotlin](https://github.com/amplitude/Amplitude-Kotlin/pull/37/files). Other client SDK's I need to check if they are the same.

kevinpagtakhan:
Should we do something similar for device id and/or session id?

PR #245, opened 2022-10-05
build: fix lerna config to prevent release creation

kevinpagtakhan:
### Summary

fix lerna config to prevent release creation

### Checklist

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,12 +16,12 @@ jobs:
     name: Authorize
     runs-on: ubuntu-latest
     steps:
-    - name: ${{ github.actor }} permission check to do a release
-      uses: "lannonbr/repo-permission-check-action@2.0.2"
-      with:
-        permission: "write"
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -16,12 +16,12 @@ jobs:
...
-      with:
-        permission: "write"
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: ${{ github.actor }} permission check to do a release
+        uses: 'lannonbr/repo-permission-check-action@2.0.2'
+        with:
+          permission: 'write'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
yuhao900914, line 24: GITHUB_TOKEN is not used any more?

yuhao900914, line 24: Where we use the GITHUB_TOKEN?

kevinpagtakhan: `GITHUB_TOKEN` is required by `lannonbr/repo-permission-check-action` based on this doc: https://github.com/lannonbr/repo-permission-check-action#usage

PR #249, opened 2022-10-12
fix: run queued functions after attribution in browser-client.ts

### LINKED ISSUE #246: Web attributions are not set on events tracked before init
## Expected Behavior
When using the `@amplitude/analytics-browser` module, all events sent with `track` before calling the `init` method should have the web attribution properties.

## Current Behavior
Currently, whenever calls to `track` are made previous to the initialization are sent to Amplitude when `init` is called, but the call to `$identify` assigning the web attributions to the user properties is only made **after** those initial events are sent.

Here is an example of what's happening in the calls to the API:
image

We can see that the `$identify` event is only sent second, after a first custom event has been sent.

This results in unreconciled events and web attributions in the initial event:

image

While subsequent events tracked (post-`init`) do have these properties:

image

This means that it's impossible to actually track web attributions on initial events happening on our website.

Additionally, it's expected most of the times to track only on event such as a page view initially, before a user is navigating to another website. If no other event is sent to Amplitude in the meantime, it becomes impossible to use the UTM tracking and other web attributions.

## Possible Solution

I believe that the issue relies in the `init` method for the `AmplitudeBrowser` class.
At [this line](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L39), we are calling the `AmplitudeCore` `_init` method, that will [deliver all the previously queued events](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/core-client.ts#L43-L45).

The identify call is made only after all these events have been flushed, [at the end of the `init` function](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L82).

## Steps to Reproduce

1. Create a simple web page and `track` an event at page load and call `init` of the client
2. Access the web page with utm parameters
3. The initial event won't have web attributions attached

## Environment
- JS SDK Version: @amplitude/analytics-browser@1.5.3
- Installation Method: `import`
- Browser and Version: Chrome 106

justin-fiedler:
### Summary
Updated AmplitudeBrowser to run queued functions after other initialization completes.
 * There is now a separate pluginQueue for `add()` and `remove()` that runs before `this.q`
 * Added test for order of calls (identify with attribution, then track)
 * fixes #246 

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -37,7 +37,8 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
       lastEventTime: oldCookies.lastEventTime,
     });
 
-    const tempQ = this.getAndResetQueuedFunctions();
+    // delay running queued functions until after all initialization is complete
+    const delayedQ = this.getAndResetQueuedFunctions();
     await super._init(browserOptions);
 
     // Step 3: Manage session
@@ -85,7 +86,7 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
 
     // Step 7:
     // Run queued functions
-    this.q = [...tempQ, ...this.q];
+    this.q = [...delayedQ, ...this.q];
     await this.runQueuedFunctions();
   }
 
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -36,6 +36,8 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
       optOut: options?.optOut ?? oldCookies.optOut,
       lastEventTime: oldCookies.lastEventTime,
     });
+
+    const tempQ = this.getAndResetQueuedFunctions();
     await super._init(browserOptions);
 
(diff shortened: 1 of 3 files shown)


liuyang1520 (approved):
The idea looks good to me!

packages/analytics-browser/src/browser-client.ts
```diff
@@ -80,6 +83,11 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
 
     // Step 6: Track attributions
     await this.runAttributionStrategy(browserOptions.attribution, isNewSession);
+
+    // Step 7:
+    // Run queued functions
+    this.q = [...delayedQ, ...this.q];
+    await this.runQueuedFunctions();
```
kevinpagtakhan, line 90: do we need `delayedQ` separated out?

justin-fiedler: Pulling these items off the queue before init() allows the plugin queue to run in core _init() while waiting until other initialization is complete for AmplitudeBrowser.

kevinpagtakhan (approved):

PR #250, opened 2022-10-13
fix: globalThis polyfill

falconandy:
### Summary

Replace `globalThis` (unsupported in some legacy browsers) with `GlobalScope` polyfill (from Amplitude-JavaScript)

### Checklist

### DIFF
--- a/packages/analytics-browser/src/index.ts
+++ b/packages/analytics-browser/src/index.ts
@@ -23,5 +23,6 @@ export const {
   track,
 } = client;
 export { runQueuedFunctions } from './utils/snippet-helper';
+export { GlobalScope } from './utils/global-scope';
 export { Revenue, Identify } from '@amplitude/analytics-core';
 export * as Types from '@amplitude/analytics-types';
--- a/packages/analytics-browser/jest.config.js
+++ b/packages/analytics-browser/jest.config.js
@@ -6,5 +6,5 @@ module.exports = {
   displayName: package.name,
   rootDir: '.',
   testEnvironment: 'jsdom',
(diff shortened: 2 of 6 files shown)


kevinpagtakhan (approved):
in addition to `analytics-browser`, we should also add this to `marketing-analytics-browser`. we can merge this to unblock customer then do the changes to `marketing-analytics-browser` as a follow up.

packages/analytics-browser/src/index.ts
```diff
@@ -23,5 +23,6 @@ export const {
   track,
 } = client;
 export { runQueuedFunctions } from './utils/snippet-helper';
+export { GlobalScope } from './utils/global-scope';
```
kevinpagtakhan, line 26: probably don't need this. i see that we'll move `globalScope` to `analytics-client-common` once we do the follow up task

falconandy:
@kevinpagtakhan
marketing-analytics-browser was updated too

packages/analytics-browser/src/index.ts
falconandy: @kevinpagtakhan 
Initially I added it to `analytics-client-common` but I am not sure if it is required for `react-native` for example.
I moved it to `analytics-browser` because `marketing-analytics-browser` already uses code from `analytics-browser`

kevinpagtakhan:
> @kevinpagtakhan `marketing-analytics-browser` was updated too

Ah you're right! 👍 Sorry didn't see right away

justin-fiedler (approved):

PR #251, opened 2022-10-17
docs: adds plugin example for removing keys from event payload

kevinpagtakhan:
### Summary

* adds plugin example for removing keys from event payload
* adds an actual example for removing `event.time` from event payload

### Checklist

### DIFF
--- a/examples/plugins/remove-event-key/index.ts
+++ b/examples/plugins/remove-event-key/index.ts
@@ -34,9 +34,11 @@ export const removeEventKeyEnrichment = (keysToRemove: KeyOfEvent[] = []): Enric
  * This is an example plugin that enriches all events by removing `event.time` from all events.
  * `event.time` uses `Date.now()` which is controlled by the device where the browser runs on.
  * With `event.time` being `undefined`, the time of the event is determined when the event was sent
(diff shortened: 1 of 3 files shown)


yuhao900914 (approved):

examples/plugins/remove-event-key/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      return event;
+    },
+  };
+};
+
+/**
+ * This is an example plugin that enriches all events by removing `event.time` from all events.
+ * `event.time` uses `Date.now()` which is controlled by the device where the browser runs on.
+ * With `event.time` being `undefined`, the time of the event is determined when the event was sent
+ * by the browser, determined by the server clock, rather then when the event occurred which can be
```
liuyang1520, line 37: Thanks for adding this example! Maybe we can add some warnings, for example, this will potentially mark the events with a later timestamp in cases like internet interruption, server side rate limiting..

liuyang1520 (approved):

examples/plugins/remove-event-key/index.ts
kevinpagtakhan: This is a great call out! I will add this.

dnshi (approved):

PR #252, opened 2022-10-19
fix: add safe check for global scope before loading SDK

kevinpagtakhan:
### Summary

* Replaces unsafe access to global variables (`globalThis`, `window`, `self`, global`)
* Prevents SDK to load if `GlobalScope` is not defined

### Checklist


packages/marketing-analytics-browser/README.md
```diff
@@ -31,11 +31,6 @@ Alternatively, the package is also distributed through a CDN. Copy and paste the
 
 <!-- README_SNIPPET_BLOCK -->
 ```html
-<script type="text/javascript">
```
liuyang1520, line 34: Maybe remove the whole block, if we want to remove the snippet here.

packages/analytics-browser/src/snippet-index.ts
```diff
@@ -1,18 +1,28 @@
+import { getGlobalScope } from '@amplitude/analytics-client-common';
 import * as amplitude from './index';
 import { runQueuedFunctions } from './utils/snippet-helper';
-import { GlobalScope } from './utils/global-scope';
 
-GlobalScope.amplitude = Object.assign(GlobalScope.amplitude || {}, amplitude);
+// https://developer.mozilla.org/en-US/docs/Glossary/IIFE
```
liuyang1520, line 5: Curious to learn why we use IIFE here? Is it to isolate the scope?

liuyang1520 (approved):

packages/marketing-analytics-browser/README.md
kevinpagtakhan: ah, this is not intentional. will remove this diff

packages/analytics-browser/src/snippet-index.ts
kevinpagtakhan: the reason is very shallow 😅  only because i wanted to do 

```
if () {
  ...
  return;
}

...
```

than

```
if () {
  ...
} else {
  ...
}
```

we can only call `return` in function level, and not in top level

packages/marketing-analytics-browser/README.md
kevinpagtakhan: reverted the change

packages/analytics-browser/src/snippet-index.ts
liuyang1520: Got it, thanks for sharing!

liuyang1520 (approved):

PR #253, opened 2022-10-21
fix: invoke pre-init track fns after attribution

### LINKED ISSUE #246: Web attributions are not set on events tracked before init

kevinpagtakhan:
### Summary

invoke pre-init track fns after attribution

Fixed #246 

### Checklist

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -3,23 +3,7 @@ import { default as nock } from 'nock';
 import { success } from './responses';
 import 'isomorphic-fetch';
 import { path, SUCCESS_MESSAGE, url, uuidPattern } from './constants';
-import { PluginType, Response, Transport } from '@amplitude/analytics-types';
-import { BaseTransport } from '@amplitude/analytics-core';
-
-const mockWindowLocationFromURL = (url: URL) => {
-  window.location.href = url.toString();
-  window.location.search = url.search;
-  window.location.hostname = url.hostname;
-  window.location.pathname = url.pathname;
-};
-
-class DummyTransport extends BaseTransport implements Transport {
-  async send(): Promise<Response | null> {
-    return this.buildResponse({
-      code: 200,
-    });
-  }
-}
+import { PluginType } from '@amplitude/analytics-types';
 
 describe('integration', () => {
   const uuid: string = expect.stringMatching(uuidPattern) as string;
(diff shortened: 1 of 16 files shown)


packages/analytics-browser-test/test/index.test.ts
```diff
@@ -119,74 +103,122 @@ describe('integration', () => {
...
-        }),
-      );
-
-      expect(track).toHaveBeenCalledTimes(1);
-      expect(track).toHaveBeenCalledWith('Event Before Init');
-      expect(send.mock.invocationCallOrder[0]).toBe(1);
-      expect(track.mock.invocationCallOrder[0]).toBe(2);
+      await trackPromise;
+
+      expect(requestBody1).toEqual({
```
kevinpagtakhan, line 140: tests that attribution is sent before pre-init track calls

packages/analytics-browser/package.json
```diff
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@amplitude/analytics-client-common": "^0.3.1",
-    "@amplitude/analytics-core": "^0.9.6",
+    "@amplitude/analytics-core": "0.9.6",
```
kevinpagtakhan, line 47: hard pins on core version

packages/analytics-core/src/core-client.ts
```diff
@@ -39,31 +40,12 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
   async _init(config: T) {
     this.config = config;
     this.timeline.reset();
-    await this.runQueuedPluginFunctions();
-    await this.runQueuedFunctions();
+    await this.runQueuedFunctions('q');
```
kevinpagtakhan, line 43: this queue includes non event emitting fns, like add, remove, setDeviceId, etc

packages/analytics-core/src/core-client.ts
```diff
@@ -29,7 +30,7 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
   // @ts-ignore
   timeline: Timeline;
   protected q: CallableFunction[] = [];
-  protected pluginQueue: CallableFunction[] = [];
+  protected dispatchQ: CallableFunction[] = [];
```
kevinpagtakhan, line 33: this queue includes only event emitting fns like track, identify, etc

packages/analytics-browser/src/browser-client.ts
```diff
@@ -74,20 +72,11 @@ export class AmplitudeBrowser extends AmplitudeCore<BrowserConfig> {
...
-    // Step 6: Track attributions
+    // Step 5: Track attributions
     await this.runAttributionStrategy(browserOptions.attribution, isNewSession);
 
-    // Step 7:
-    // Run queued functions
-    this.q = [...delayedQ, ...this.q];
-    await this.runQueuedFunctions();
+    // Step 6: Run queued dispatch functions
+    await this.runQueuedFunctions('dispatchQ');
```
kevinpagtakhan, line 79: dispatchQ only invoked after runAttributionStrategy

kevinpagtakhan:
@amplitude/dx I added comments on the significant code changes to highlight them, would love feedback specifically on those areas

dnshi (approved):

packages/analytics-core/src/core-client.ts
```diff
@@ -98,37 +80,57 @@ export class AmplitudeCore<T extends Config> implements CoreClient<T> {
...
+      this.q.push(this.remove.bind(this, pluginName));
       return;
     }
     return this.timeline.deregister(pluginName);
   }
 
-  async dispatch(event: Event) {
+  dispatchWithCallback(event: Event, callback: (result: Result) => void): void {
+    if (!this.config) {
+      return callback(buildResult(event, 0, CLIENT_NOT_INITIALIZED));
```
liuyang1520, line 99: If `config` is not ready (not initialized), this will resolve the promise with `CLIENT_NOT_INITIALIZED`. Are we just dropping these events? I am not sure how this handles the case when `track` is called before `runAttributionStrategy`, that the tracked event will be sent after identify. Could you elaborate more? Thanks!

kevinpagtakhan: `dispatchWithCallback` is designed to only be called when in `this.delayedQ` as this ensures that `this.config` is defined. I only added this case just in case it's ever called (but probably never will) and it will return this error so dev would see.

kevinpagtakhan: I also created `dispatchWithCallback` to create a way where `dispatch` can always return `Promise<Result>`. The alternative was to add to fn call to queue but return `undefined`, which I personally didn't like.

https://github.com/amplitude/Amplitude-TypeScript/blob/DXOC-244/packages/analytics-core/src/core-client.ts#L106-L108

liuyang1520 (approved):

PR #254, opened 2022-10-25
feat: enhance logger with debug information

liuyang1520:
### Summary
This PR enhances the current debug logs:
- Add timestamp, function name, function args, function stack trace, internal states to all public function calls
- Skip overheads when not in debug mode

Note the current implementation:
- doesn't handle the initial debug logs before `init` with logLevel to `Debug`, as we don't have the logger and logLevel ready before `init`
- doesn't expose the global instance nor using `postMessage`, will do in a separate PR if needed

### Checklist


packages/analytics-types/src/logger.ts
```diff
@@ -4,11 +4,18 @@ export interface Logger {
...
   error(...args: any[]): void;
+  debug(...args: any[]): void;
 }
 
 export enum LogLevel {
   None = 0,
   Error = 1,
   Warn = 2,
   Verbose = 3,
+  Debug = 4,
```
kevinpagtakhan, line 15: thinking verbose and debug can be one. what do you think? browser and node seems to have varying opinions here

dnshi: I agree to merge debug and verbose. 
image

Just curious if the logLevel is set to None, can we still get logger information? It does not need to print on the console, but store somewhere or some extensions can retrieve the logger information.

liuyang1520: Good question, I added the debug level to match the `debug` level: "for debugging information. These messages are displayed only when user had opted-in to see debug logging for specific modules", https://webpack.js.org/api/logging/.

My plan is to have the following items in interface, with the actually logger implementation to decide.
- error
- warn
- log
- debug

Browser and Node has the `console.debug`, that maps the output to `Verbose` in browser, and `console.log` in Node.

For decision to whether merging debug into log, I am ok with both. Just some concern that `debug` outputs much more information than `log`, might be overwhelming information if user enables `log`.

@dnshi , for your question, no, the logged messages will not be stored/sent to anywhere, see https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/logger.ts#L34.

dnshi: @liuyang1520 I see. I am wondering if we have a new version of Chrome extension, how the extension gets the message if the logger level is set to None by the customer?

dnshi: Chatted with Marvin offline. The Chrome extension support will be in a separate topic.

packages/analytics-core/src/utils/debug.ts
```diff
@@ -0,0 +1,90 @@
...
+    if ((logLevel && logLevel < LogLevel.Debug) || !logLevel || !logger) {
+      return fn.apply(fnContext, args);
+    }
+    const logs: string[] = [];
+    logs.push(`start time: ${new Date().toISOString()}`);
+    logs.push(`function name "${fnName}" called with args: ${JSON.stringify(args, null, 2)}`);
+    if (getStates) {
+      logs.push(`states before execution ${JSON.stringify(getStates(), null, 2)}`);
+    }
+    logs.push(`function stacktrace: ${getStacktrace(1)}`);
```
dnshi, line 67: I am thinking of pushing objects instead. For example
```js
logs.push({
  startTime: new Date().toISOString(),
  fnName: fnName,
  fnArgs: JSON.stringify(args, null, 2),
  ...
})
```
Pushing string is not easy to change format.

liuyang1520: Thanks! Updated in [82139c2](https://github.com/amplitude/Amplitude-TypeScript/pull/254/commits/82139c2c12c8c673f7b55ef3eea652982bc5eb8e)!

yuhao900914 (approved):
LGTM;

liuyang1520:
@kevinpagtakhan , @dnshi , a quick follow up on this, any other comments on this PR? Thanks!

packages/analytics-core/src/utils/debug.ts
```diff
@@ -0,0 +1,92 @@
...
+      args,
+      stacktrace: getStacktrace(1),
+      time: {
+        start: new Date().toISOString(),
+      },
+      states: {},
+    };
+    if (getStates) {
+      debugContext.states!.before = getStates();
+    }
```
kevinpagtakhan, line 73: would prefer that we access keys safely instead of using non-null assertions

```suggestion
    if (getStates && debugContext.states) {
      debugContext.states.before = getStates();
    }
```

liuyang1520: Thanks! Updated in https://github.com/amplitude/Amplitude-TypeScript/pull/254/commits/d6a91b1bae2b46c7aafc6e8fb8681b4c6a98bbb2.

kevinpagtakhan (approved):
🚀

PR #260, opened 2022-10-27
feat: ignore subdomains when comparing newness of campaigns

AlecBoutin:
### Summary

For attribution plugin, don't reset session or campaign when subdomain changes.

### Checklist

### DIFF
--- a/packages/analytics-client-common/src/attribution/campaign-tracker.ts
+++ b/packages/analytics-client-common/src/attribution/campaign-tracker.ts
@@ -42,11 +42,7 @@ export class CampaignTracker implements ICampaignTracker {
     this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
   }
 
-  isNewCampaign(
-    current: Campaign,
-    previous: Campaign | undefined,
-    removeSubdomainsWhenComparingReferringDomains = false,
-  ) {
+  isNewCampaign(current: Campaign, previous: Campaign | undefined, ignoreSubdomainInReferrer = false) {
     const { referrer, referring_domain, ...currentCampaign } = current;
     const { referrer: _previous_referrer, referring_domain: prevReferringDomain, ...previousCampaign } = previous || {};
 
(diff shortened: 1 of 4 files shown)


packages/analytics-client-common/src/attribution/campaign-tracker.ts
```diff
@@ -42,17 +42,24 @@ export class CampaignTracker implements ICampaignTracker {
...
-  isNewCampaign(current: Campaign, previous: Campaign | undefined) {
-    const { referrer: _referrer, ...currentCampaign } = current;
-    const { referrer: _previous_referrer, ...previousCampaign } = previous || {};
+  isNewCampaign(
+    current: Campaign,
+    previous: Campaign | undefined,
+    removeSubdomainsWhenComparingReferringDomains = false,
+  ) {
+    const { referrer, referring_domain, ...currentCampaign } = current;
+    const { referrer: _previous_referrer, referring_domain: prevReferringDomain, ...previousCampaign } = previous || {};
```
kevinpagtakhan, line 51: Deconstructing `referring_domain: prevReferringDomain` will remove this key/value pair from `previousCampaign` and will not be part of the `hasNewCampaign` check. I think this will cause `hasNewCampaign` to always be true since the two objects do not contain the same key/value pair entries.

AlecBoutin: referring_domain is deconstructed from both current and previous campaign objects, so I wouldn't expect it to cause a problem. Am I misreading this?

kevinpagtakhan: Ah no i misread it, i didn't notice `referring_domain` deconstructed from current.

packages/analytics-client-common/src/attribution/campaign-tracker.ts
```diff
@@ -42,17 +42,24 @@ export class CampaignTracker implements ICampaignTracker {
     this.initialEmptyValue = options.initialEmptyValue ?? EMPTY_VALUE;
   }
 
-  isNewCampaign(current: Campaign, previous: Campaign | undefined) {
-    const { referrer: _referrer, ...currentCampaign } = current;
-    const { referrer: _previous_referrer, ...previousCampaign } = previous || {};
+  isNewCampaign(
+    current: Campaign,
+    previous: Campaign | undefined,
+    removeSubdomainsWhenComparingReferringDomains = false,
```
kevinpagtakhan, line 48:
```suggestion
    compareReferringSubdomains = false,
```

perhaps?

kevinpagtakhan (approved):

packages/analytics-client-common/src/attribution/campaign-tracker.ts
AlecBoutin: Changed to `ignoreSubdomainInReferrer`

PR #263, opened 2022-11-01
feat: auto-tracking session events

falconandy:
### Summary

1. Added new option (`options.trackingOptions.sessionEvents`) to auto-track session start/end events.
2. Added tests for session logic.

Tested only on Android emulator.

Existing differences between `Amplitude-Kotlin` and `React Native` implementations:
1. `React Native` has ability to set custom session id (client `options.sessionId`, `setSessionId`), `Amplitude-Kotlin` doesn't have.
2. `Amplitude-Kotlin` persists last event Id, `React Native` doesn't - always starts event Id from `0` on app restart.
3. Session timeout is 5 minutes in `Amplitude-Kotlin` and 30 minutes in `React Native`.

### Checklist

* Does your PR have a breaking change?:  No, but session logic could be affected.


packages/analytics-types/src/config.ts
```diff
@@ -76,6 +76,7 @@ export interface TrackingOptions {
 export interface ReactNativeTrackingOptions extends TrackingOptions {
   adid?: boolean;
   carrier?: boolean;
+  sessionEvents?: boolean;
```
kevinpagtakhan, line 79: I think this is called `trackingSessionEvents` in Kotlin, perhaps we can align?

falconandy: @kevinpagtakhan 
Do you mean `options.trackingSessionEvents` (as in Kotlin) vs `options.trackingOptions.trackingSessionEvents`?

kevinpagtakhan: this one: `options.trackingSessionEvents`

falconandy: Fixed to `options.trackingSessionEvents`. Currently in `ReactNativeConfig`, in the future can be moved to base `BrowserConfig` if session events auto-tracking will be implemented in browsers too

qingzhuozhen (approved):
Test on iOS and can see the session events. Will create tickets to handle the event id and default session.

Issue #274, opened 2022-11-11
Identify not working in "@amplitude/analytics-node"


hanna-becker:
I am trying to add user properties after initializing amplitude and before sending an event in a lambda function on AWS.

Without the identify part, the event gets sent and I can see it in the Amplitude console:
```
  await amplitude.init(AMPLITUDE_API_KEY, { logLevel: amplitude.Types.LogLevel.Debug }).promise;
  amplitude.track('some event', undefined, { user_id: userId });
  await amplitude.flush().promise;
```

Adding properties via identify like so, and the event does not make it to the console any longer:

```
  await amplitude.init(AMPLITUDE_API_KEY, { logLevel: amplitude.Types.LogLevel.Debug }).promise;
  
  // adding this part
  const identifyObj = new amplitude.Identify();
  identifyObj.set('email', email);
  identifyObj.set('env', process.env.ENV);
  amplitude.identify(identifyObj);

  amplitude.track('some event', undefined, { user_id: userId });
  await amplitude.flush().promise;
```
I also see this not very helpful log statement in the lambda logs: `ERROR Amplitude Logger [Error]: invalid`.

## Expected Behavior
The event should be visible in the Amplitude console, along with the user properties 'email' and 'env'.

## Current Behavior
The event does not get sent.

## Steps to Reproduce
See code above.

## Environment
- JS SDK Version: 0.6.0
- Installation Method: yarn
- Browser and Version: node 14

liuyang1520:
Hi @hanna-becker , thanks for supporting us!

Could you please try:
```js
amplitude.identify(identifyObj, { user_id: userId });
```

This works in my local. If this works for you too, then we will update the log and docs for this.

Thanks!

hanna-becker:
Yes, this works. Thank you!

PR #275, opened 2022-11-15
feat: add nest.js example app and improve response message with body content

liuyang1520:
### Summary
This PR includes the changes for:
- feat: add nest.js example app
- fix: improve response message with body content, related to: https://github.com/amplitude/Amplitude-TypeScript/issues/274

image

image

I found we don't have a Node example, so adding one here. I didn't choose Express as the official boilerplate/generator doesn't support TypeScript, and third-party ones introduces other dependencies like DB, Auth, see:
- https://github.com/w3tecch/express-typescript-boilerplate
- https://github.com/GeekyAnts/express-typescript
- https://github.com/ljlm0402/typescript-express-starter

While the Nest CLI is neat and clean, saving us time to tidying unnecessary code with the boilerplate. If needed, we can add Express example in the future for sure.

### Checklist


tsconfig.json
```diff
@@ -24,5 +24,8 @@
     "strictBindCallApply": true,
     "target": "es5",
     "types": ["jest", "node"],
+    "paths": {
+      "@amplitude/analytics-node": ["./packages/analytics-node/"]
```
liuyang1520, line 28: This is to allow the example to use the `@amplitude/analytics-node` without extra installation, also make life easier to debug local changes without extra `yarn link`.

kevinpagtakhan: was hoping the examples would be using published versions so they'd be more reliable. idk what do you think?

liuyang1520: Good call! I slightly prefer this way than the published versions, but I am good with both. The reasons I slightly prefer this way:
- easier for community developer to test their change locally, before opening PRs here
- react-native has a well-known issue when linking local dependency, this will make symlink not work. Seems the only easy way to test local change is to use this way, see the [legacy react-native](https://github.com/amplitude/Amplitude-ReactNative/blob/main/tsconfig.json#L5). I haven't test this on react-native example though, but I know `yarn/npm link` is not working for me (@yuhao900914 shared this tsconfig solution to me). Note, the published versions are working smoothly, but we will likely to have local uncommitted changes all the time for development purpose.
  - https://stackoverflow.com/questions/44061155/react-native-npm-link-local-dependency-unable-to-resolve-module
  - https://github.com/facebook/metro/issues/1

As you mention, the published versions are more stable, which will the examples more stable too. @kevinpagtakhan , let me know if you prefer the published versions, I can update to use it.

liuyang1520: Update to use the published versions to be aligned with other examples in this repo, we can probably revisit this for all the SDKs, for example Kotlin Android is using the local version: https://github.com/amplitude/Amplitude-Kotlin/blob/main/samples/kotlin-android-app/build.gradle#L45.

dnshi (approved):

kevinpagtakhan (approved):

PR #285, opened 2022-11-24
chore: add debug logs for session logic

falconandy:
### Summary

1. Removes some logic from `SessionManager.setSessionId`/`SessionManager.getSessionId`. If the logic is required in some cases, please let me know.
2. Adds some logging to try to debug session-logic issues. `Debug` level isn't used - too much output for other debug messages.

### Checklist


packages/analytics-client-common/src/session-manager.ts
```diff
@@ -4,7 +4,6 @@ import { getCookieName as getStorageKey } from './cookie-name';
 export class SessionManager implements ISessionManager {
   storageKey: string;
   cache: UserSession;
-  isSessionCacheValid = true;
```
qingzhuozhen, line 7: why remove this? cc @kevinpagtakhan do you know if this one is safe to remove?

falconandy: `isSessionCacheValid`  was used only in `getSessionId`/`setSessionId` logic and not used in others (`getDeviceId`/`setDeviceId`, `getUserId`/`setUserId`, `getLastEventTime`/`setLastEventTime`, etc)

qingzhuozhen: Maybe we can keep it if it doesn't hurt the debugging purpose, or you can move forward if you believe this doesn't break anything

qingzhuozhen (approved):

packages/analytics-client-common/src/session-manager.ts
falconandy: I believe it shouldn't break anytihng

PR #315, opened 2023-01-26
feat: allow opt out of deleting legacy sdk cookies

kevinpagtakhan:
### Summary

1. Adds `config.cookieUpgrade` to retain old SDK cookie and treat as source of truth.
3. This allows parallel instrumentation of legacy and new SDK
4. Best practice is to remove `config.cookieUpgrade` when no longer in parallel instrumentation
5. [BONUS] Integration tests for `@amplitude/analytics-browser` now runs faster and in parallel

#### Migration guide

**Step 0: Pre migration**
This assumes customer is instrumenting solely using `amplitude-js`

```ts
import amplitude from 'amplitude-js';
amplitude.getInstance().init(...);
```

**Step 1: Parallel instrumentation and partial migration**
This assumes customer is instrumenting using both `amplitude-js` and `@amplitude/analytics-browser`

```ts
import amplitude from 'amplitude-js';
import * as amplitude2 from '@amplitude/analytics-browser';

amplitude.getInstance().init(apiKey, userId, config, () => {
  // hold off init until legacy SDK cookies are persisted
  amplitude2.init(..., {
    cookieUpgrade: false, // reads and retains legacy SDK cookies
  });
});
```

**Step 2: Bye amplitude-js and full migration**
This assumes customer is instrumenting solely using `@amplitude/analytics-browser`

```ts

amplitude2.init(...);
// options.cookieUpgrade can be explicitly set to `true` or completely omitted
// this effectively reads and deletes legacy SDK cookies
// moving forward, new SDK solely maintain its own cookies
```

cc @jackmccloy

### Checklist

### DIFF
--- a/packages/analytics-browser-test/package.json
+++ b/packages/analytics-browser-test/package.json
@@ -10,7 +10,7 @@
     "url": "git+https://github.com/amplitude/Amplitude-TypeScript.git"
   },
   "scripts": {
-    "test": "jest -i"
+    "test": "jest"
   },
   "bugs": {
     "url": "https://github.com/amplitude/Amplitude-TypeScript/issues"
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -18,21 +18,20 @@ describe('integration', () => {
   };
 
   let apiKey = '';
+  let client = amplitude.createInstance();
 
   beforeEach(() => {
+    client = amplitude.createInstance();
     apiKey = UUID();
   });
 
   afterEach(() => {
     // clean up cookies
-    document.cookie = 'AMP_API_KEY=null; expires=-1';
     document.cookie = `AMP_${apiKey.substring(0, 6)}=null; expires=-1`;
+    document.cookie = `AMP_${apiKey.substring(0, 10)}=null; expires=-1`;
   });
 
-  // WARNING: This test has to run first
-  // This test is under the assumption that amplitude has not be initiated at all
-  // To achieve this condition, it must run before any other tests
-  describe('FIRST TEST: defer initialization', () => {
+  describe('defer initialization', () => {
     beforeAll(() => {
       Object.defineProperty(window, 'location', {
         value: {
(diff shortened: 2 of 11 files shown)


liuyang1520:
For `Step 1: Parallel instrumentation and partial migration`, think it might still cause minor issue because of the init order. TS SDK will probably not use the legacy cookie in the first time of the page visit (legacy cookie is not ready yet), this might affect the events sent without a page refresh. This might be an edge case though.

kevinpagtakhan:
@liuyang1520 what im thinking is that the cookies are already stored by step 0, so even if on step 1, amplitude-js has not finished initting, there will be some valid cookies to read. what do you think?

liuyang1520:
> @liuyang1520 what im thinking is that the cookies are already stored by step 0, so even if on step 1, amplitude-js has not finished initting, there will be some valid cookies to read. what do you think?

I think step 0 will make it work for the existing tracked user, the step 1 can read from them, but not for newly tracked user (first time visiting the web app) from step 1. Besides this edge case, the change looks good to me, I am not sure whether we have an easy solution for this though.

liuyang1520 (approved):

kevinpagtakhan:
> > @liuyang1520 what im thinking is that the cookies are already stored by step 0, so even if on step 1, amplitude-js has not finished initting, there will be some valid cookies to read. what do you think?
> 
> I think step 0 will make it work for the existing tracked user, the step 1 can read from them, but not for newly tracked user (first time visiting the web app) from step 1. Besides this edge case, the change looks good to me, I am not sure whether we have an easy solution for this though.

ahh I see what you mean. we can just update the code sample to await the legacy init

```ts
amplitude.getInstance().init(apiKey, userId, config, () => {
  amplitude2.init(..., {
    cookieUpgrade: false, // retains legacy SDK cookies
  });
});
```

do you think this would solve it?

liuyang1520:
> > > @liuyang1520 what im thinking is that the cookies are already stored by step 0, so even if on step 1, amplitude-js has not finished initting, there will be some valid cookies to read. what do you think?
> > 
> > 
> > I think step 0 will make it work for the existing tracked user, the step 1 can read from them, but not for newly tracked user (first time visiting the web app) from step 1. Besides this edge case, the change looks good to me, I am not sure whether we have an easy solution for this though.
> 
> ahh I see what you mean. we can just update the code sample to await the legacy init
> 
> ```ts
> amplitude.getInstance().init(apiKey, userId, config, () => {
>   amplitude2.init(..., {
>     cookieUpgrade: false, // retains legacy SDK cookies
>   });
> });
> ```
> 
> do you think this would solve it?

This looks good, think it should work, as long as the JS SDK sets the cookie before the TS SDK init.

jackmccloy:
@kevinpagtakhan catching up after being away for a long weekend. thanks so much for the fast turnaround on this ❤️

PR #322, opened 2023-02-07
feat: add file download tracker

kevinpagtakhan:
### Summary

This set of changes adds a internal plugin for tracking file downloads.

#### Example

Considering the following `<a>` tag:

```html
<a
  id="my-link-id"
  href="https://analytics.amplitude.com/files/my-file.pdf"
  target="_blank"
  rel="noopener noreferrer"
>
  Download
</a>
```

The plugin tracks  the following event:

```json
{
  "event_type": "file_download",
  "event_properties": {
    "file_extension": "pdf",
    "file_name": "/files/my-file.pdf",
    "link_id": "my-link-id",
    "link_text": "my-link-text",
    "link_url": "https://analytics.amplitude.com/files/my-file.pdf"
}
```

#### How to track
1. The download button must be an anchor tag
1. The download button must contain an href property
1. The href property must point to a file with the following file types
    - `pdf`, `xls`, `xlsx`, `doc`, `docx`, `txt`, `rtf`, `csv`, `exe`, `key`, `pps`, `ppt`, `pptx)`, `7z`, `pkg`, `rar`, `gz`, `zip`, `avi`, `mov`, `mp4`, `mpe`, `mpeg`, `wmv`, `mid`, `midi`, `mp3`, `wav`, `wma`
1. The download button must be clicked
1. All the rules above must be fulfilled to track a file download event

#### How not to track
1. The plugin does not support download button in an HTML form that lead to a file download
2. The plugin does not support download button in button element that lead to a file download
3. The plugin does not support download button `onclick` attributes that lead to a file download

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/file-download-tracking.ts
+++ b/packages/analytics-browser/src/plugins/file-download-tracking.ts
@@ -1,0 +1,77 @@
+import { BrowserClient, PluginType, Event, EnrichmentPlugin } from '@amplitude/analytics-types';
+import { BrowserConfig } from '../config';
+
+const FILE_DOWNLOAD_EVENT = 'file_download';
+
+export const fileDownloadTracking = (): EnrichmentPlugin => {
+  const name = 'fileDownloadTracking';
+  const type = PluginType.ENRICHMENT;
+  const setup = async (config: BrowserConfig, amplitude?: BrowserClient) => {
+    /* istanbul ignore if */
+    if (!amplitude) {
+      // TODO: Add required minimum version of @amplitude/analytics-browser
+      config.loggerProvider.warn(
+        'File download tracking requires a later version of @amplitude/analytics-browser. File download events are not tracked.',
+      );
+      return;
+    }
+
+    const addFileDownloadListener = (a: HTMLAnchorElement) => {
+      try {
+        const url = new URL(a.href);
+        const result = ext.exec(url.href);
+        const fileExtension = result?.[1];
+
+        if (fileExtension) {
+          a.addEventListener('click', () => {
+            if (fileExtension) {
+              amplitude.track(FILE_DOWNLOAD_EVENT, {
+                file_extension: fileExtension,
+                file_name: url.pathname,
+                link_id: a.id,
+                link_text: a.text,
+                link_url: a.href,
+              });
+            }
+          });
+        }
+      } catch {
+        config.loggerProvider.error(`Something went wrong. File download events are not tracked for a#{a.id}`);
+      }
+    };
+
+    const ext =
+      /\.(pdf|xlsx?|docx?|txt|rtf|csv|exe|key|pp(s|t|tx)|7z|pkg|rar|gz|zip|avi|mov|mp4|mpe?g|wmv|midi?|mp3|wav|wma)$/;
+
+    // Adds listener to existing anchor tags
+    const links = Array.from(document.getElementsByTagName('a'));
+    links.forEach(addFileDownloadListener);
+
+    // Adds listener to anchor tags added after initial load
+    /* istanbul ignore else */
+    if (typeof MutationObserver !== 'undefined') {
+      const observer = new MutationObserver((mutations) => {
+        mutations.forEach((mutation) => {
+          mutation.addedNodes.forEach((node) => {
+            if (node.nodeName == 'A') {
+              addFileDownloadListener(node as HTMLAnchorElement);
+            }
+          });
+        });
+      });
+
+      observer.observe(document.body, {
+        subtree: true,
+        childList: true,
+      });
+    }
+  };
+  const execute = async (event: Event) => event;
+
+  return {
+    name,
+    type,
+    setup,
+    execute,
+  };
+};
--- a/packages/analytics-browser/test/plugins/file-download-tracking.test.ts
+++ b/packages/analytics-browser/test/plugins/file-download-tracking.test.ts
@@ -1,0 +1,97 @@
+/* eslint-disable @typescript-eslint/unbound-method */
+
+import { createAmplitudeMock, createConfigurationMock } from '../helpers/mock';
+import { fileDownloadTracking } from '../../src/plugins/file-download-tracking';
+
+describe('fileDownloadTracking', () => {
+  let amplitude = createAmplitudeMock();
+
+  beforeEach(() => {
+    amplitude = createAmplitudeMock();
+
+    const link = document.createElement('a');
+    link.setAttribute('id', 'my-link-id');
+    link.setAttribute('class', 'my-link-class');
+    link.text = 'my-link-text';
+
+    document.body.appendChild(link);
+  });
+
+  afterEach(() => {
+    document.querySelector('a#my-link-id')?.remove();
+  });
+
+  test('should track file_download event', async () => {
+    // setup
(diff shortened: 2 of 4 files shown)


packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -0,0 +1,50 @@
...
+      // TODO: Add required minimum version of @amplitude/analytics-browser
+      config.loggerProvider.warn(
+        'File download tracking requires a later version of @amplitude/analytics-browser. File download events are not tracked.',
+      );
+      return;
+    }
+
+    const ext =
+      /\.(pdf|xlsx?|docx?|txt|rtf|csv|exe|key|pp(s|t|tx)|7z|pkg|rar|gz|zip|avi|mov|mp4|mpe?g|wmv|midi?|mp3|wav|wma)$/;
+    const links = Array.from(document.getElementsByTagName('a'));
```
dnshi, line 21: For SPA, anchor tag might not be displayed on the page initially. Maybe consider adding an event listener on the document / body and add conditional check to detect if the event target is anchor.

liuyang1520, line 21: I think this only supports pre-rendered static pages. Do we want to support dynamic pages like SPA? When the `setup` is invoked, if the `a` tag is not rendered on the page yet, or user interaction will create a `a` tag on the fly (like navbar, popup window), then this new `a` tag will be missed. If we want to support this, we can use [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe) to watch `document` sub tree addition and filter to `a` node. There might have other solutions too.

packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -0,0 +1,50 @@
...
+    links.forEach((a) => {
+      const url = new URL(a.href);
+      const result = ext.exec(url.href);
+      const fileExtension = result?.[1];
+
+      if (fileExtension) {
+        a.addEventListener('click', () => {
```
dnshi, line 28: Maybe consider adding `keypress` / `keydown` event if the download link is focused and users press `enter` key

packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -0,0 +1,50 @@
+import { BrowserClient, PluginType, Event, EnrichmentPlugin } from '@amplitude/analytics-types';
+import { BrowserConfig } from '../config';
+
+const FILE_DOWNLOAD_EVENT = 'file_download';
```
liuyang1520, line 4: Probably discussed somewhere else too. Do we want to have a special prefix/name for this event, avoiding naming conflicts with customer-defined event?

packages/analytics-browser/src/plugins/file-download-tracking.ts
liuyang1520, line 21: Do we want to add a special attribute or class to allow skipping from being tracked? E.g., `<a data-amplitude-ignore="true">` to allow opting out of tracking. Examples:
- https://support.pendo.io/hc/en-us/community/posts/6068892939163-Tech-Note-Excluding-specific-elements-from-tracking-in-Pendo
- https://posthog.com/docs/integrate/client/js#tuning-autocapture

packages/analytics-browser/src/plugins/file-download-tracking.ts
kevinpagtakhan: I tested this and it appears that `click` event is also fired hence the listener should be sufficient in this context.

packages/analytics-browser/src/plugins/file-download-tracking.ts
kevinpagtakhan: This is a great suggestion. It is likely that this out of scope for this project, but I can bring it up during the project sync to confirm.

packages/analytics-browser/src/plugins/file-download-tracking.ts
kevinpagtakhan: The event names are still TBD, and name will likely change! The placeholder name is inherited from a competitor. The plugin isn't auto added to the client yet so shouldn't be a concern yet.

packages/analytics-browser/src/plugins/file-download-tracking.ts
kevinpagtakhan: Good call out on dynamically added elements! Though adding an event listener to the document/body may be overkill and likely compromise performance. I will look at other options.

kevinpagtakhan: Good call out! I can confirm we want to handle this case as well. I'll look into how to best handle this scenario.

packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -0,0 +1,77 @@
...
+      /\.(pdf|xlsx?|docx?|txt|rtf|csv|exe|key|pp(s|t|tx)|7z|pkg|rar|gz|zip|avi|mov|mp4|mpe?g|wmv|midi?|mp3|wav|wma)$/;
+
+    // Adds listener to existing anchor tags
+    const links = Array.from(document.getElementsByTagName('a'));
+    links.forEach(addFileDownloadListener);
+
+    // Adds listener to anchor tags added after initial load
+    /* istanbul ignore else */
+    if (typeof MutationObserver !== 'undefined') {
+      const observer = new MutationObserver((mutations) => {
```
kevinpagtakhan, line 53: @liuyang1520 @dnshi covers dynamically added anchor tags

packages/analytics-browser/test/plugins/file-download-tracking.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    expect(amplitude.track).toHaveBeenNthCalledWith(1, 'file_download', {
+      file_extension: 'pdf',
+      file_name: '/files/my-file.pdf',
+      link_id: 'my-link-id',
+      link_text: 'my-link-text',
+      link_url: 'https://analytics.amplitude.com/files/my-file.pdf',
+    });
+  });
+
+  test('should track file_download event for a dynamically added achor tag', async () => {
```
kevinpagtakhan, line 45: @liuyang1520 @dnshi verifies tracking dynamically added anchor tags

dnshi (approved):

packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -0,0 +1,77 @@
+import { BrowserClient, PluginType, Event, EnrichmentPlugin } from '@amplitude/analytics-types';
+import { BrowserConfig } from '../config';
+
+const FILE_DOWNLOAD_EVENT = 'file_download';
+
+export const fileDownloadTracking = (): EnrichmentPlugin => {
+  const name = 'fileDownloadTracking';
+  const type = PluginType.ENRICHMENT;
```
qingzhuozhen, line 8: Do we need this to be an enrichment type, or before type is fine?

qingzhuozhen (approved):

packages/analytics-browser/src/plugins/file-download-tracking.ts
kevinpagtakhan: Logically, it could be either since these two type only differs when modifying the payload

PR #323, opened 2023-02-07
feat: add form interaction event tracker

kevinpagtakhan:
### Summary

This set of changes adds an internal plugin for tracking form interactions.

#### Example

Considering the following `<form>` tag:

```html
<form id="form-id" name="form-name" action="/hello">
  <input type="text" />
  <input type="submit" text="Submit" />
</form>
```

The plugin tracks  the following event:

```json
{
  "event_type": "form_start",
  "event_properties": {
  "form_id": "my-form-id",
   "form_name": "my-form-name",
    "form_destination": "http://localhost/hello",
}
```

```json
{
  "event_type": "form_submit",
  "event_properties": {
  "form_id": "my-form-id",
   "form_name": "my-form-name",
    "form_destination": "http://localhost/hello",
}
```

#### How to track
1. Form must have `<form>` tag
1. Form must contain `<input type="submit">` to track `form_submit` event
1. Form can contain `<input>` tags as children of `<form>` tag to track `form_start` independent of `form_submit`
1. When the first input field within the form was changed, a `form_start` event is tracked.
1. When the second input field within the form was changed, no succeeding `form_start` event is tracked.
1. The counter above resets on a new page load or on form submit.
1. When the submit button is clicked, a `form_submit` event is tracked.
1. When the submit button is clicked while no input fields have changed, `form_start` and `form_submit` events are tracked.

#### How not to track
1. The plugin is not compatible with forms not using `<form>` tags
3. The plugin is not compatible with forms not using `<input type="submit">`

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
+++ b/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
@@ -1,0 +1,87 @@
+import { BrowserClient, PluginType, Event, EnrichmentPlugin } from '@amplitude/analytics-types';
+import { BrowserConfig } from '../config';
+
+const FORM_START_EVENT = 'form_start';
+const FORM_SUBMIT_EVENT = 'form_submit';
+
+export const formInteractionTracking = (): EnrichmentPlugin => {
+  const name = 'formInteractionTracking';
+  const type = PluginType.ENRICHMENT;
+  const setup = async (config: BrowserConfig, amplitude?: BrowserClient) => {
+    /* istanbul ignore if */
+    if (!amplitude) {
+      // TODO: Add required minimum version of @amplitude/analytics-browser
+      config.loggerProvider.warn(
+        'Form interaction tracking requires a later version of @amplitude/analytics-browser. Form interaction events are not tracked.',
+      );
+      return;
+    }
+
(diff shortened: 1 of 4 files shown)


packages/analytics-browser/src/plugins/form-interaction-tracking.ts
```diff
@@ -0,0 +1,87 @@
...
+    const forms = Array.from(document.getElementsByTagName('form'));
+    forms.forEach(addFormInteractionListener);
+
+    // Adds listener to anchor tags added after initial load
+    /* istanbul ignore else */
+    if (typeof MutationObserver !== 'undefined') {
+      const observer = new MutationObserver((mutations) => {
+        mutations.forEach((mutation) => {
+          mutation.addedNodes.forEach((node) => {
+            if (node.nodeName == 'FORM') {
```
liuyang1520, line 66: Minor, `===` to avoid type conversion?

liuyang1520 (approved):

qingzhuozhen (approved):

packages/analytics-browser/src/plugins/form-interaction-tracking.ts
kevinpagtakhan: Good catch! It should be triple equal.

PR #330, opened 2023-02-16
fix: improper cookie usage

kevinpagtakhan:
### Summary

Actual behavior:
SDKs creates cookies despite being configured with `disableCookies: true`.

Expected behavior:
SDKS should not create cookies when configured with `disableCookies: true`.

### Checklist

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -1128,7 +1128,7 @@ describe('integration', () => {
                 insert_id: uuid,
                 ip: '$remote',
                 language: 'en-US',
-                library: 'amplitude-ts/1.8.0',
+                library,
                 os_name: 'WebKit',
                 os_version: '537.36',
                 partner_id: undefined,
@@ -1146,7 +1146,7 @@ describe('integration', () => {
                 insert_id: uuid,
                 ip: '$remote',
                 language: 'en-US',
-                library: 'amplitude-ts/1.8.0',
+                library,
                 os_name: 'WebKit',
                 os_version: '537.36',
                 partner_id: undefined,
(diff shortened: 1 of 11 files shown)


liuyang1520 (approved):

justin-fiedler (approved):

qingzhuozhen (approved):

lipper-li:
We encountered a problem when using amplitude, which generated a lot of cookies, causing the header to be too large when requesting and the page to fail to load. 
So I'm wondering if this PR will resolve this issue and if so, when will a new version be released?

packages/analytics-browser-test/test/index.test.ts
```diff
@@ -1675,12 +1675,10 @@ describe('integration', () => {
 
   describe('browser cookie existence', () => {
     test('should create cookies', async () => {
-      // intercept for attribution event
+      // intercept for attribution event and identify event
```
kevinpagtakhan, line 1678: default events work was able to combine these two events in a single payload

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -72,7 +72,7 @@ export const webAttributionPlugin: CreateWebAttributionPlugin = function (
 
       // Share cookie storage with user session storage
       const storage = config.cookieStorage as unknown as Storage<Campaign>;
-      const storageKey = getStorageKey(config.apiKey);
+      const storageKey = getStorageKey(config.apiKey, 'MKTG');
```
kevinpagtakhan, line 75: fix to prevent campaign cookies from overwriting user session cookies

yuhao900914 (approved):

PR #331, opened 2023-02-22
feat: retrofit web attribution and page view plugins to browser SDK

kevinpagtakhan:
### Summary

* Replaces browser SDK logic for attribution and page view tracking with first-party plugins
* Old logic for attribution and page view tracking is still supported
* New logic for page view tracking is enabled when using new config

To configure with old logic for page view tracking,

```ts
amplitude.init(API_KEY, USER_ID, {
  attribution: {
    trackPageViews: true,
  },
});
```

To configure with new logic for page view tracking,

```ts
amplitude.init(API_KEY, USER_ID, {
  defaultTracking: {
    pageViews: true,
  },
});
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -3,6 +3,7 @@ import {
   Event,
   BrowserOptions,
   BrowserConfig as IBrowserConfig,
+  DefaultTrackingOptions,
   Storage,
   TrackingOptions,
   TransportType,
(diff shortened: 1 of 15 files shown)


packages/analytics-browser/src/config.ts
```diff
@@ -48,6 +49,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
   cookieSecure: boolean;
   cookieUpgrade: boolean;
   cookieStorage: Storage<UserSession>;
+  defaultTracking?: DefaultTrackingOptions;
```
kevinpagtakhan, line 52: I will update this to the preferred name at a later PR cc @liuyang1520

liuyang1520 (approved):

PR #332, opened 2023-02-22
feat: add session start/end event tracker

kevinpagtakhan:
### Summary

Sends session start and session end events when a new session is assigned. Additionally, a new session is assigned when a new user is assigned

#### Testing

```ts
amplitude.init(API_KEY, USER_ID, {
  defaultTracking: {
    sessions: true,
  },
  sessionTimeout: 500,
});
// If previous session exists, sends `session_end` event
// Sends `session_start` event
amplitude.track('Event in first session');

setTimeout(() => {
  amplitude.track('Event in next session');
  // Sends `session_end` event for previous session
  // Sends `session_start` event for next session
}, 1000);

setTimeout(() => {
  amplitude.setUserId(USER_ID_2); // effectively creates a new session
  // Sends `session_end` event for previous user and session
  // Sends `session_start` event for next user and session
}, 2000);
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -3,6 +3,7 @@ import {
   getAnalyticsConnector,
   getPageViewTrackingConfig,
   IdentityEventSender,
+  isSessionTrackingEnabled,
 } from '@amplitude/analytics-client-common';
 import {
   BrowserClient,
@@ -19,11 +20,13 @@ import { useBrowserConfig, createTransport } from './config';
 import { parseOldCookies } from './cookie-migration';
 import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser';
 import { pageViewTrackingPlugin } from '@amplitude/plugin-page-view-tracking-browser';
+import { sessionHandlerPlugin, START_SESSION_EVENT, END_SESSION_EVENT } from './plugins/session-handler';
 
 export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   config: BrowserConfig;
+  previousSessionUserId: string | undefined;
 
   init(apiKey = '', userId?: string, options?: BrowserOptions) {
     return returnWrapper(this._init({ ...options, userId, apiKey }));
@@ -79,6 +82,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     // Do not track any events before this
     await this.add(new Destination()).promise;
     await this.add(new Context()).promise;
+    await this.add(sessionHandlerPlugin()).promise;
     await this.add(new IdentityEventSender()).promise;
 
     // Add web attribution plugin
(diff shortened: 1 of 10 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -145,8 +153,30 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
+        if (this.previousSessionUserId) {
+          eventOptions.user_id = this.previousSessionUserId;
+        }
+        this.track(END_SESSION_EVENT, undefined, eventOptions);
+        this.previousSessionUserId = undefined;
+      }
+
+      this.track(START_SESSION_EVENT, undefined, {
+        session_id: sessionId,
+        time: sessionId - 1,
```
liuyang1520, line 177: Is there any reason we `-1` here?

packages/analytics-browser/src/plugins/session-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+      browserConfig.lastEventTime = now;
+      return event;
+    }
+
+    const lastEventTime = browserConfig.lastEventTime || now;
+    const timeSinceLastEvent = now - lastEventTime;
+
+    if (timeSinceLastEvent > browserConfig.sessionTimeout) {
+      // assigns new session
+      amplitude.setSessionId(now);
```
qingzhuozhen, line 38: Curious about the order of the events, would this show before start session event in this case?

packages/analytics-browser/src/browser-client.ts
```diff
@@ -116,7 +120,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
       this.q.push(this.setUserId.bind(this, userId));
       return;
     }
-    this.config.userId = userId;
+    if (userId !== this.config.userId) {
+      this.previousSessionUserId = this.config.userId;
+      this.config.userId = userId;
+      this.setSessionId(Date.now());
```
qingzhuozhen, line 126: do you think we need an option for this?

packages/analytics-browser/src/plugins/session-handler.ts
kevinpagtakhan: The order of events in payload request are definitely not in the right order, but the session events themselves have the time property manipulated so analytics would see them in the intended order. I added some tests on that here: https://github.com/amplitude/Amplitude-TypeScript/blob/5a1f0e12fcd5d105d8764e7c2230cdc49ac1f6ef/packages/analytics-browser-test/test/index.test.ts#L926-L929

packages/analytics-browser/src/browser-client.ts
kevinpagtakhan: We can but i was thinking that having a different user id make them not share sessions anyway right?

packages/analytics-browser/src/browser-client.ts
kevinpagtakhan: I added this to guarantee that this is the first session of the event. The times must be manipulated so that analytics sees them in the right order

liuyang1520 (approved):

packages/analytics-browser/src/plugins/session-handler.ts
qingzhuozhen: I see, that sounds good to me.

packages/analytics-browser/src/browser-client.ts
qingzhuozhen: You mean they will appear on different users? Make sense.

qingzhuozhen (approved):

dnshi (approved):

PR #334, opened 2023-02-23
feat: add form interaction and file download plugins

kevinpagtakhan:
### Summary

Add form interaction and file download plugins
* Transform marketing analytics page view config to default event tracking config
* `form_start` -> `[Amplitude] Form Start`
* `form_submit` -> `[Amplitude] Form Submit`
* `file_download` -> `[Amplitude] File Download`

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -98,21 +98,22 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     }
 
     // Add web attribution plugin
-    const webAttribution = webAttributionPlugin({
-      disabled: this.config.attribution?.disabled,
-      excludeReferrers: this.config.attribution?.excludeReferrers,
-      initialEmptyValue: this.config.attribution?.initialEmptyValue,
-      resetSessionOnNewCampaign: this.config.attribution?.resetSessionOnNewCampaign,
-    });
+    if (!this.config.attribution?.disabled) {
+      const webAttribution = webAttributionPlugin({
+        excludeReferrers: this.config.attribution?.excludeReferrers,
+        initialEmptyValue: this.config.attribution?.initialEmptyValue,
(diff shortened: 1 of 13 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -81,6 +85,14 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     await this.add(new Context()).promise;
     await this.add(new IdentityEventSender()).promise;
 
+    if (isFileDownloadTrackingEnabled(this.config.autoTracking)) {
+      await this.add(fileDownloadTracking()).promise;
+    }
+
+    if (isFormInteractionTrackingEnabled(this.config.autoTracking)) {
```
liuyang1520, line 92: Minor, for page view, do we plan to make it similar like this to conditionally add the plugin or not?

kevinpagtakhan: For page views, I used `trackOn` config as toggle, although that means we will install it all the time. These two new plugins do not have options defined yet so the condition was added in browser-client level.

kevinpagtakhan: Any thoughts?

liuyang1520: I am good with the `trackOn`. If the `trackOn` config is used with some custom function, do we want to merge the logic there? I think currently it will be assigned with `attribution` value only. Example:
```
init(API_KEY, 'user@amplitude.com', {
  pageViewTracking: {
    trackOn: () => {
      return window.location.pathname === '/landing_page'
    },
  }
});
```

kevinpagtakhan: Yeah this makes sense. Although I was under the impression this is out of scope and we want to have a boolean config first. I can check with others.

kevinpagtakhan: @liuyang1520 the advanced controls are now added :)

liuyang1520 (approved):

PR #336, opened 2023-02-24
build: prevents script loader update for beta version

kevinpagtakhan:
### Summary

When publishing a new version, we update the script loader to include the latest version. The changes prevents this from happening when releasing a beta version. Production versions should update the script loader as usual.

### Checklist


liuyang1520 (approved):

yuhao900914:
Question: how do we use the beta version script then? Do we need to manually generate it?

kevinpagtakhan:
> Question: how do we use the beta version script then? Do we need to manually generate it?

We can generate it ourselves. I just didn't to make it public where people can easily access and install it. We should only let them install prod versions

PR #337, opened 2023-02-25
fix: inconsistent user and device id on session events

kevinpagtakhan:
### Summary

session end and session start shares may share user id or device id on `reset()` or `setUserId()`

#### Issue
```
Session Start
Event 1
Event 2
// User calls amplitude.reset() or amplitude.setUserId() 
Session End // has user id or device id of next session
Session Start
Event 3
Event 4
```

#### Solution
```
Session Start // store user id and event id
Event 1
Event 2
// User calls amplitude.reset() or amplitude.setUserId() 
Session End // use previously stored user id and event id
Session Start // store user id and event id
Event 3
Event 4
```

### Checklist

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -1213,16 +1213,22 @@ describe('integration', () => {
         setTimeout(() => {
           // Tranform events to be grouped by session and ordered by time
           // Helps assert that events will show up correctly on Amplitude UI
-          const compactEvents: { eventType: string; userId: string; sessionId: number; time: number }[] =
-            payload.events.map((e: BaseEvent) => ({
-              eventType: e.event_type,
-              userId: e.user_id,
-              sessionId: e.session_id,
-              time: e.time,
-            }));
+          const compactEvents: {
+            eventType: string;
+            userId: string;
+            sessionId: number;
+            time: number;
+            deviceId: string;
+          }[] = payload.events.map((e: BaseEvent) => ({
+            eventType: e.event_type,
+            userId: e.user_id,
+            sessionId: e.session_id,
+            deviceId: e.device_id,
+            time: e.time,
+          }));
           const compactEventsBySessionAndTime = Object.values(
             compactEvents.reduce<
-              Record<string, { eventType: string; userId: string; sessionId: number; time: number }[]>
+              Record<string, { eventType: string; userId: string; sessionId: number; time: number; deviceId: string }[]>
             >((acc, curr) => {
               if (!acc[curr.sessionId]) {
                 acc[curr.sessionId] = [];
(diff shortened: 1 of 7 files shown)


packages/analytics-browser-test/test/index.test.ts
```diff
@@ -1232,8 +1238,15 @@ describe('integration', () => {
...
-
+            .map((group) => group.map((c) => ({ eventType: c.eventType, userId: c.userId, deviceId: c.deviceId })));
+          const deviceIds = compactEventsBySessionAndTime.flat().map((c) => c.deviceId);
+          expect(deviceIds[0]).toEqual(deviceIds[1]);
+          expect(deviceIds[1]).toEqual(deviceIds[2]);
+          expect(deviceIds[2]).toEqual(deviceIds[3]);
+          expect(deviceIds[3]).toEqual(deviceIds[4]);
+          expect(deviceIds[4]).toEqual(deviceIds[5]);
+          expect(deviceIds[5]).toEqual(deviceIds[6]);
+          expect(deviceIds[6]).not.toEqual(deviceIds[7]);
```
kevinpagtakhan, line 1249: Asserts that only the last event (Which is a session start for a new user) has a different device id

packages/analytics-browser/src/browser-client.ts
```diff
@@ -43,17 +47,31 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
 
     // Step 2: Create browser config
+    const deviceId = options.deviceId ?? queryParams.deviceId ?? previousCookies?.deviceId ?? legacyCookies.deviceId;
+    const sessionId = options.sessionId ?? previousCookies?.sessionId ?? legacyCookies.sessionId;
+    const optOut = options.optOut ?? previousCookies?.optOut ?? legacyCookies.optOut;
+    const lastEventTime = previousCookies?.lastEventTime ?? legacyCookies.lastEventTime;
+    const userId = options.userId ?? previousCookies?.userId ?? legacyCookies.userId;
+
+    this.previousSessionDeviceId = previousCookies?.deviceId ?? legacyCookies.deviceId;
+    this.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
```
kevinpagtakhan, line 65: Pull out reading of cookies to client level to store `previousSessionDeviceId` and `previousSessionUserId`. This is used for sending the first session end with those IDs

packages/analytics-browser/src/browser-client.ts
```diff
@@ -181,17 +198,17 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
           session_id: previousSessionId,
           time: previousLastEventTime + 1,
         };
-        if (this.previousSessionUserId) {
-          eventOptions.user_id = this.previousSessionUserId;
-        }
+        eventOptions.device_id = this.previousSessionDeviceId;
+        eventOptions.user_id = this.previousSessionUserId;
```
kevinpagtakhan, line 202: Re-use recently used user id and device id for start event

liuyang1520 (approved):

qingzhuozhen (approved):
This seems a quick way to handle, ideally if we can get the order right, we could also fix this issue.

packages/analytics-browser/src/config.ts
```diff
@@ -168,24 +168,18 @@ export class BrowserConfig extends Config implements IBrowserConfig {
...
+  const deviceId = options?.deviceId ?? UUID();
+  const lastEventTime = options?.lastEventTime;
+  const optOut = options?.optOut;
+  const sessionId = options?.sessionId;
+  const userId = options?.userId;
 
   return new BrowserConfig(apiKey, {
     ...options,
-    cookieStorage,
+    cookieStorage: options?.cookieStorage,
```
yuhao900914, line 180: Why make cookieStorage and domain different from others?

yuhao900914 (approved):
LGTM

packages/analytics-browser/src/config.ts
kevinpagtakhan: Ah yeah good point. I can make it consistent

kevinpagtakhan: Fixed!

kevinpagtakhan:
> This seems a quick way to handle, ideally if we can get the order right, we could also fix this issue.

You're right @qingzhuozhen. I might need to get rid of the session handler plugin to guarantee order. I'll spend more time later to optimize if time permits.

PR #361, opened 2023-03-31
feat: update Plugin implementation to allow for dropping events

falconandy:
### Summary

Update Plugin implementation to allow for dropping events

### Checklist

### DIFF
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -74,7 +74,13 @@ export class Timeline {
     );
 
     for (const plugin of before) {
-      event = await plugin.execute({ ...event });
+      const e = await plugin.execute({ ...event });
+      if (e == null) {
+        resolve({ event, code: 0, message: '' });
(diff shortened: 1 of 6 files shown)


packages/analytics-core/src/timeline.ts
```diff
@@ -74,15 +74,27 @@ export class Timeline {
     );
 
     for (const plugin of before) {
-      event = await plugin.execute({ ...event });
+      const e = await plugin.execute({ ...event });
+      if (e == null) {
```
kevinpagtakhan, line 78: Would prefer threequals, which I should add to eslint rules to better enforce 😄 

```suggestion
      if (e === null) {
```

justin-fiedler (approved):
Other than the `===`, LGTM. Thanks @falconandy.

packages/analytics-core/src/timeline.ts
falconandy: Fixed

PR #374, opened 2023-04-24
feat: persist last event id - to be used on next initialization

kevinpagtakhan:
### Summary

* Persist last event it used in storage

### Testing

* Automated tests are passing
* Manually tested by initializing SDK on a website
  * decode cookies and verify event id is persisted
  * re-initialize SDK on a different page load and verify that event id is incremental across page loads

### Checklist


packages/analytics-types/src/config/react-native.ts
```diff
@@ -4,7 +4,6 @@ import { BrowserConfig, TrackingOptions } from './browser';
 export interface ReactNativeConfig extends BrowserConfig {
   trackingOptions: ReactNativeTrackingOptions;
   trackingSessionEvents?: boolean;
-  lastEventId?: number;
```
kevinpagtakhan, line 7: notes: removed from react native but moved to browser which is shared

liuyang1520 (approved):

qingzhuozhen (approved):

kevinpagtakhan:
> Thanks for the quick fix! Do we ever update `lastEventTime`? It feels like we should update those together in general.

@paladin8 Yes, we update it in the session handler plugin: https://github.com/amplitude/Amplitude-TypeScript/blob/0a4f8ede6e30ec3450ac0a468cf22b9266b0b23c/packages/analytics-browser/src/plugins/session-handler.ts#L42 because it is also used to determine if the most recently created session is valid/expired.

Not sure why build is failing for node 18 but will investigate further before merging.

PR #382, opened 2023-05-09
feat: use server side user agent parser

kevinpagtakhan:
### Summary

* Removes client-side user agent parsing
* Removes dependency on `@amplitude/ua-parser-js`
* Reduces minified bundle size `amplitude-min.js` file from ~80kb to ~65kb
* Contains breaking changes (PR opened to `browser-sdk-v2` branch)
* Updates @amplitude/analytics-browser version in package.json to `1.0.0-beta.0`

Note: Since our release automation does not support well maintaining multiple major versions, I plan to merge changes to `browser-sdk-v2` and do a "rebase and merge" (not "squash and merge") from `browser-sdk-v2` to `main` when all changes are ready

### Checklist


liuyang1520 (approved):

packages/analytics-types/src/config/browser.ts
```diff
@@ -33,14 +33,9 @@ export interface DefaultTrackingOptions {
 }
 
 export interface TrackingOptions {
-  deviceManufacturer?: boolean;
```
qingzhuozhen, line 36: Do we still want consumers to be able to set these values by themselves?

yuhao900914:
Question: by removing the ua parser, we are going to depend on the devices.json to get the device info?  Will each event still has deviceManufacturer, deviceModel etc? Will this break the existing analytics chart?

packages/analytics-types/src/config/browser.ts
kevinpagtakhan: Yes, they can set `event.device_manufacturer` through an enrichment plugin. This option was only used to enable client side user agent parsing which we no longer have.

kevinpagtakhan:
> Question: by removing the ua parser, will each event still have deviceManufacturer, deviceModel etc? Will this break the existing analytics chart?

@yuhao900914 without client side user agent parser, the SDK, be default, will no longer populate `event.device_manufacturer` and `event.device_model`. We are sending raw user agent values through `event.user_agent` which ingestion will parse and populate the appropriate fields, I believe here: https://github.com/amplitude/amplitude/blob/master/event/converters/http_event.py#L353-L367 However, our server side parser yields varying results than what we used to have, and charts will "break" hence these changes, and a few more, will be part of Browser SDK V2, a major version bump.

yuhao900914:
> > Question: by removing the ua parser, will each event still have deviceManufacturer, deviceModel etc? Will this break the existing analytics chart?
> 
> @yuhao900914 without client side user agent parser, the SDK, be default, will no longer populate `event.device_manufacturer` and `event.device_model`. We are sending raw user agent values through `event.user_agent` which ingestion will parse and populate the appropriate fields, I believe here: https://github.com/amplitude/amplitude/blob/master/event/converters/http_event.py#L353-L367 However, our server-side parser yields varying results than what we used to have, and charts will "break" hence these changes, and a few more, will be part of Browser SDK V2, a major version bump.

Thanks for the explanation. Hope we can also document in the doc center to notify the customer what will be broken after the version bumps up.

yuhao900914 (approved):

kevinpagtakhan:
> What about moving the `ua-parser` into a separate plugin that can optionally be included by users during migration? That would allow customers to migrate without breaking existing chart definitions if needed.

I like that, do you suggest that we publish new plugin or provide code samples that can be copy/pasted?

justin-fiedler:
> > What about moving the `ua-parser` into a separate plugin that can optionally be included by users during migration? That would allow customers to migrate without breaking existing chart definitions if needed.
> 
> I like that, do you suggest that we publish new plugin or provide code samples that can be copy/pasted?

I would lean towards publishing the plugin to make it easier for folks to migrate.

qingzhuozhen (approved):

PR #390, opened 2023-05-25
feat: simplify user identity storage options/configuration

kevinpagtakhan:
#### 1. Simplify "cookie storage" (better referred to as user identity storage) provider

Most customers would use default configuration with cookies as storage provider for user identity (user id, device id, session id, etc). However, cookie usage is well regulated and should have a simpler configuration interface. This includes opting out of cookies, opting into local storage, opting out of persistence.

Using `cookie` as storage provider for identity storage (note: no diff)

```diff
  amplitude.init(API_KEY);
```

Using `localStorage` as storage provider for identity storage

Pattern 1:
```diff
  amplitude.init(API_KEY, undefined, {
-   disableCookies: true,
+   identityStorage: 'localStorage',
  });
```

Pattern 2:
```diff
- import { LocalStorage } from '@amplitude/analytics-core';
  amplitude.init(API_KEY, undefined, {
-   cookieStorageProvider: new LocalStorage(),
+   identityStorage: 'localStorage',
  });
```

Opting out of cookie and local storage

```diff
- import { MemoryStorage } from '@amplitude/analytics-core';
  amplitude.init(API_KEY, undefined, {
-   cookieStorageProvider: new MemoryStorage(),
+   identityStorage: 'none',
  });
```

#### 2. Simplify "cookie storage" options

For customer using cookie as storage provider for user identity, this provides a simpler and more discoverable way of and managing cookies (expiration, domain, etc).

```diff
  amplitude.init(API_KEY, undefined, {
-   cookieExpiration: 365,
-   cookieSameSite: 'Lax',
-   cookieSecure: false,
-   cookieUpgrade: true,
-   domain: '',
+   cookieOptions: {
+     expiration: 365,
+     sameSite: 'Lax',
+     secure: false,
+     upgrade: true,
+     domain: '',
+   },
  });
```

### Checklist

* Does your PR have a breaking change?: Yes


packages/analytics-browser/src/config.ts
```diff
@@ -168,85 +173,93 @@ export class BrowserConfig extends Config implements IBrowserConfig {
...
-  if (options.disableCookies || !(await storage.isEnabled())) {
-    storage = new LocalStorage();
-    if (!(await storage.isEnabled())) {
-      storage = new MemoryStorage();
-    }
+export const createCookieStorage = <T>(
+  identityStorage: IdentityStorageType = DEFAULT_IDENTITY_STORAGE,
+  cookieOptions: CookieOptions = {},
+) => {
+  if (identityStorage === DEFAULT_IDENTITY_STORAGE) {
```
justin-fiedler, line 256: nit: I think it would be better to switch on the explicit value of identityStorage in case the default changes.

```suggestion
  switch (identityStorage) {
    case 'cookie':
      return new CookieStorage<T>(cookieOptions);
    case 'localStorage':
       ...
```

packages/analytics-browser/src/browser-client.ts
```diff
@@ -193,17 +160,17 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
         this.track(DEFAULT_SESSION_END_EVENT, undefined, eventOptions);
       }
 
       this.track(DEFAULT_SESSION_START_EVENT, undefined, {
         session_id: sessionId,
         time: sessionId - 1,
       });
-      this.previousSessionDeviceId = this.config.deviceId;
-      this.previousSessionUserId = this.config.userId;
+      this.config.lastSessionDeviceId = this.config.deviceId;
```
justin-fiedler, line 172: super nit: I think `previousSessionDeviceId` is a slightly better name as `last` can also imply "final". Just my 2c's.

kevinpagtakhan: Updated to `previousSessionDeviceId` and `previousSessionUserId`

packages/analytics-browser/src/config.ts
kevinpagtakhan: Updated to switch statement with 'cookie' being default value

packages/analytics-browser/src/config.ts
```diff
@@ -7,90 +7,95 @@ import {
...
+    deviceId?: string,
+    public flushIntervalMillis: number = 1000,
+    public flushMaxRetries: number = 5,
+    public flushQueueSize: number = 30,
+    public identityStorage: IdentityStorageType = DEFAULT_IDENTITY_STORAGE,
+    public ingestionMetadata?: IngestionMetadata,
+    lastEventId?: number,
+    lastEventTime?: number,
+    public previousSessionDeviceId?: string,
+    public previousSessionUserId?: string,
```
qingzhuozhen, line 57: So we have previousSessionUserId and previousSessionDeviceId in config now? and we also have userId I believe, what's the major difference between these two?

yuhao900914: I think this is used to assign the right sessionId and deviceId for the session end event. Check [here](https://github.com/amplitude/Amplitude-TypeScript/blob/browser-sdk-v2/packages/analytics-browser/src/browser-client.ts#L190-L206).
Also curious why we need to include those two in the config.

kevinpagtakhan: Hey, I moved it to config to consolidate it with the other metadata like `lastEventId`, `lastEventTime` which I believe serve similar purposes. What I made sure is that these are not included in public facing options on `init()`

Internal config definition: https://github.com/amplitude/Amplitude-TypeScript/blob/d1f528862211b96b13e4e3539338de76914699b6/packages/analytics-types/src/config/browser.ts#L23

 Browser options extends external config only: https://github.com/amplitude/Amplitude-TypeScript/blob/d1f528862211b96b13e4e3539338de76914699b6/packages/analytics-types/src/config/browser.ts#L63

yuhao900914: That makes sense to me. Thanks.

packages/analytics-browser/src/config.ts
```diff
@@ -7,90 +7,94 @@ import {
...
+    sessionId?: number,
+    public sessionTimeout: number = 30 * 60 * 1000,
+    public storageProvider: Storage<Event[]> = new LocalStorage(),
+    public trackingOptions: Required<TrackingOptions> = {
+      ipAddress: true,
+      language: true,
+      platform: true,
+    },
+    public transport: 'fetch' | 'xhr' | 'beacon' = 'fetch',
+    public transportProvider: Transport = new FetchTransport(),
```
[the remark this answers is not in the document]
kevinpagtakhan: that's a good point. i think they both have their own pros/cons and would be down to talk more about this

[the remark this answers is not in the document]
yuhao900914: Find the confluence page talked about the cons and pros.
https://amplitude.atlassian.net/wiki/spaces/GOV/pages/1492058287/Beacon+API+-+JS+SDK+Proposal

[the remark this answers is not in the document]
kevinpagtakhan: Took the discussion to Slack. Will be addressed in a different PR.

qingzhuozhen (approved):

PR #393, opened 2023-05-30
feat(plugins): session replay plugin

kwalker3690:
### Summary

This PR adds a new plugin to support session replay functionality. We will be adding additional configuration options in future revisions, this is the base functionality.

```typescript
import { SessionReplayPlugin } from '@amplitude/plugin-session-replay-browser';

amplitude.init(apiKey);
const sessionReplayTracking = new SessionReplayPlugin();
amplitude.add(sessionReplayTracking);
```

### Checklist


packages/plugin-session-replay/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@amplitude/plugin-session-replay",
+  "version": "1.0.0-beta.0",
```
kevinpagtakhan, line 3: Can we do 0.0.0?

```suggestion
  "version": "0.0.0",
```

packages/plugin-session-replay/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@amplitude/plugin-session-replay",
```
kevinpagtakhan, line 2: i believe our package naming convention is including the platform at the end cc @justin-fiedler 

```suggestion
  "name": "@amplitude/plugin-session-replay-browser",
```

packages/plugin-session-replay/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { SessionReplayPlugin as Plugin, SessionReplayPlugin } from './session-replay';
+export * as Types from './typings/session-replay';
```
kevinpagtakhan, line 2: do we need to expose this?

packages/plugin-session-replay/README.md
```diff
@@ -0,0 +1,46 @@
...
+# npm
+npm install @amplitude/plugin-session-replay
+
+# yarn
+yarn add @amplitude/plugin-session-replay
+```
+
+## Usage
+
+This plugin works on top of Amplitude Browser SDK and adds session replay features to built-in features. To use this plugin, you need to install `@amplitude/analytics-browser` version `v2.0.0` or later.
```
kevinpagtakhan, line 26: i think this plugin can also be used with Browser SDK 1.0

packages/plugin-session-replay/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "description": "",
+  "author": "Amplitude Inc",
+  "homepage": "https://github.com/amplitude/Amplitude-TypeScript",
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "lib/esm/index.d.ts",
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public"
```
kevinpagtakhan, line 13: maybe we can also set the dist tag to beta until we ga?

```suggestion
    "access": "public",
    "tag": "beta"
```

packages/plugin-session-replay/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "dependencies": {
+    "@amplitude/analytics-client-common": "^1.0.0-beta.0",
+    "@amplitude/analytics-types": "^1.0.0-beta.0",
+    "idb-keyval": "^6.2.1",
+    "rrweb": "^2.0.0-alpha.4",
+    "tslib": "^2.4.1"
+  },
+  "devDependencies": {
+    "@amplitude/analytics-browser": "^1.10.3",
```
kevinpagtakhan, line 46: if we can confirm the compatibility🤞 

```suggestion
    "@amplitude/analytics-browser": ">=1 <3",
```

packages/plugin-session-replay/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "publish": "node ../../scripts/publish/upload-to-s3.js",
+    "test": "jest",
+    "typecheck": "tsc -p ./tsconfig.json"
+  },
+  "bugs": {
+    "url": "https://github.com/amplitude/Amplitude-TypeScript/issues"
+  },
+  "dependencies": {
+    "@amplitude/analytics-client-common": "^1.0.0-beta.0",
+    "@amplitude/analytics-types": "^1.0.0-beta.0",
```
kevinpagtakhan, line 40:
```suggestion
    "@amplitude/analytics-client-common": ">=1 <3",
    "@amplitude/analytics-types": ">=1 <3",
```

packages/plugin-session-replay/src/typings/session-replay.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
+import { EnrichmentPlugin } from '@amplitude/analytics-types';
+
+export interface Options {}
+
+export interface CreateSessionReplayPlugin {
+  (options?: Options): EnrichmentPlugin;
+}
```
kevinpagtakhan, line 8: i think we don't need this?

packages/plugin-session-replay/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Change Log
```
kevinpagtakhan, line 1: we can also delete this file and let the publish github action create it from statch

packages/plugin-session-replay/rollup.config.js
```diff
@@ -0,0 +1,6 @@
+import { iife, umd } from '../../scripts/build/rollup.config';
+
+iife.input = umd.input;
+iife.output.name = 'sessionReplay';
+
+export default [umd, iife];
```
kevinpagtakhan, line 6: this uploads the bundle to s3 and makes them available through our CDN. can be used by customers installing the SDK via the snippet loader.

packages/plugin-session-replay/src/session-replay.ts
```diff
@@ -0,0 +1,285 @@
...
+    await this.emptyStoreAndReset();
+
+    this.recordEvents();
+  }
+
+  execute(event: Event) {
+    event.event_properties = {
+      ...event.event_properties,
+      session_replay_enabled: true,
+    };
```
kevinpagtakhan, line 44: curious if modifying the event props for $identify and $groupidentify events makes a difference

kevinpagtakhan:
@justin-fiedler we have this new plugin package getting published soon. what are your thoughts on which branch we should merge this to and publish from? i think it's compatible with v1 and v2 (but need to confirm)

justin-fiedler:
> we have this new plugin package getting published soon. what are your thoughts on which branch we should merge this to and publish from? i think it's compatible with v1 and v2 (but need to confirm)

Can we only create it in `main` so it can be maintained in a single location? Would be nice if new features are only in a single branch.

packages/plugin-session-replay/rollup.config.js
kwalker3690: Awesome!

packages/plugin-session-replay/src/session-replay.ts
kwalker3690: Makes a difference in what way?

kwalker3690: I'll add that to our test plan! I haven't seen any issues in testing it myself

packages/plugin-session-replay/README.md
```diff
@@ -0,0 +1,46 @@
+<p align="center">
+  <a href="https://amplitude.com" target="_blank" align="center">
+    <img src="https://static.amplitude.com/lightning/46c85bfd91905de8047f1ee65c7c93d6fa9ee6ea/static/media/amplitude-logo-with-text.4fb9e463.svg" width="280">
+  </a>
+  <br />
+</p>
+
+# @amplitude/plugin-session-replay
```
kevinpagtakhan, line 8: can we also update the title, installation instructions and the directory name?

packages/plugin-session-replay/README.md
```diff
@@ -0,0 +1,46 @@
...
+import * as amplitude from '@amplitude/analytics-browser';
+import { SessionReplayPlugin } from '@amplitude/plugin-session-replay-browser';
+```
+
+### 2. Instantiate session replay plugin and install plugin to Amplitude SDK
+
+The plugin must be registered with the amplitude instance via the following code:
+
+```typescript
+amplitude.init(API_KEY)
```
kevinpagtakhan, line 43:
```suggestion
amplitude.init(API_KEY);
```

packages/plugin-session-replay/README.md
```diff
@@ -0,0 +1,46 @@
...
+const sessionReplayTracking = new SessionReplayPlugin();
+amplitude.add(sessionReplayTracking)
```
kevinpagtakhan, line 45:
```suggestion
amplitude.add(sessionReplayTracking);
```

packages/plugin-session-replay/src/session-replay.ts
```diff
@@ -0,0 +1,286 @@
...
+import { shouldSplitEventsList } from './helpers';
+import { MAX_RETRIES_EXCEEDED_MESSAGE, STORAGE_FAILURE, SUCCESS_MESSAGE, UNEXPECTED_ERROR_MESSAGE } from './messages';
+import { Events, IDBStore, SessionReplayContext } from './typings/session-replay';
+
+const SESSION_REPLAY_SERVER_URL = 'https://api-secure.amplitude.com/sessions/track';
+const STORAGE_PREFIX = `${AMPLITUDE_PREFIX}_replay_unsent`;
+const PAYLOAD_ESTIMATED_SIZE_IN_BYTES_WITHOUT_EVENTS = 200; // derived by JSON stringifying an example payload without events
+const MAX_EVENT_LIST_SIZE_IN_BYTES = 20 * 1000000 - PAYLOAD_ESTIMATED_SIZE_IN_BYTES_WITHOUT_EVENTS;
+
+export class SessionReplayPlugin implements EnrichmentPlugin {
```
kevinpagtakhan, line 14: only pointing this out because it's public facing. our other public facing plugins are written as functions and are camel cased (vs pascal cased). given it's a class i think it's understandable have it pascal-cased but for consistency with other public facing plugins, should we write this as a function and be camel cased? @amplitude/dx any thoughts on this?

packages/plugin-session-replay/src/session-replay.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    this.config = config;
+    this.storageKey = `${STORAGE_PREFIX}_${this.config.apiKey.substring(0, 10)}`;
+    void this.emptyStoreAndReset();
+  }
+
+  execute(event: Event) {
+    event.event_properties = {
+      ...event.event_properties,
+      session_replay_enabled: true,
```
justin-fiedler, line 41: Are we sure we want to modify the `event properties` directly? This can cause collisions and override customer defined values. I think `session_replay_enabled` is fairly unique but still want to call this out.

If we are ok modifying the event properties directly, it might be good to create some sort of namespacing e.g. `amp_session_replay_enabled` so we can warn customers not to create properties that start with `amp_` in order to prevent data loss.

packages/plugin-session-replay/src/session-replay.ts
```diff
@@ -0,0 +1,286 @@
...
+      }
+      this.events = [];
+      this.currentSequenceId = 0;
+    }
+    return Promise.resolve(event);
+  }
+
+  async emptyStoreAndReset() {
+    const storedReplaySessions = await this.getAllSessionEventsFromStore();
+    console.log('in empty store after await');
```
justin-fiedler, line 62: 🧹

packages/plugin-session-replay/src/session-replay.ts
kwalker3690: Yep after a team discussion we decided to rename this property to `[Amplitude] Session Recorded` to keep it inline with the other default properties we already have

packages/plugin-session-replay/src/session-replay.ts
justin-fiedler: Can you please remove this or use the loggerProvider vs console.log.

packages/plugin-session-replay/src/session-replay.ts
kwalker3690: @kevinpagtakhan @justin-fiedler is this more about the name of the plugin or also about how it's written? Ie I can make it a function with a camel cased name by doing just
```
export const sessionReplayPlugin: SessionReplayPlugin = () => {
  return new SessionReplay();
};
```
Or are you asking the class also to be rewritten to match plugins like pageViewTracking? Asking before I go either direction because the changes have a large impact on my unit tests so hoping to avoid doing work that isn't what you're asking for

kwalker3690:
@justin-fiedler @kevinpagtakhan where did we land with which branch I should merge this to? `main` or `v1.x`?

packages/analytics-core/src/core-client.ts
kevinpagtakhan, line 1: do we need this or can we revert?

kwalker3690: Oh no, I'll remove, mistake addition!

kevinpagtakhan (approved):
🚀

qingzhuozhen (approved):

packages/plugin-session-replay-browser/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_EVENT_PROPERTY_PREFIX = '[Amplitude]';
+
+export const DEFAULT_SESSION_REPLAY_PROPERTY = `${DEFAULT_EVENT_PROPERTY_PREFIX} Session Recorded`;
+export const DEFAULT_SESSION_START_EVENT = 'session_start';
```
yuhao900914, line 4: Should we reuse [this](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/constants.ts#L9-L10) and DEFAULT_EVENT_PREFIX?

kevinpagtakhan: for SDKs, i'd lean on exposing less than more so there's less surface area for breaking changes. i think duplicating was a good call

PR #394, opened 2023-05-31
feat: track page dynamic page views in single page applications be de…

kevinpagtakhan:
#### 1. Enables SPA page view tracking by default

By default, page views are only tracked on initial load by default. To track dynamic page views (navigation through client-side routing), amplitude must be configured using `options.defaultTracking.pageViews.trackHistoryChanges`, and setting a value of either "all" or "pathOnly". Without a value, this feature is disabled.

With this change, initial and dynamic page loads are considered the same, and are enabled/disabled the same way. Although what remains to be configurable is how dynamic page views are defined to allow flexibility on client-side routing mechanism.

Enable SPA page view tracking (similar to how initial page view is enabled)

```diff
  amplitude.init(API_KEY, undefined, {
    defaultTracking: {
-     pageViews: {
-       trackHistoryChanges: 'all', // or 'pathOnly'
-     },
+     pageViews: true,
  },
});
```

Disables SPA page view tracking (similar to how dynamic page view is enabled)

```diff
  amplitude.init(API_KEY, undefined, {
    defaultTracking: {
-     pageViews: {
-       trackHistoryChanges: undefined, // or omit
-     },
+     pageViews: false,
  },
});
```

### Changes to @amplitude/plugin-page-view-tracking-browser

Same as above

### Checklist

### DIFF
--- a/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
+++ b/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
@@ -61,7 +61,7 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
       loggerProvider = config.loggerProvider;
       loggerProvider.log('Installing @amplitude/plugin-page-view-tracking-browser');
 
-      if (options.trackHistoryChanges && globalScope) {
+      if (globalScope) {
         /* istanbul ignore next */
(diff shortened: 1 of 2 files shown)


qingzhuozhen (approved):
The change lgtm, wonder how we configure the flexibility later on this.

liuyang1520 (approved):

PR #395, opened 2023-06-01
fix: dyanamic page view tracking to run after navigation

kevinpagtakhan:
### Summary

1. Ensures that page view event is tracked after dynamic page navigation
    * Prior to the fix, page view is optimistically tracked before dynamic page navigation. This causes an issue where `location` state references the source page, rather than the destination page.
3. Consolidate similar tests
4. Use actual `pushState` method to test dynamic page views
5. Remove internal methods designed for testing

### Checklist

### DIFF
--- a/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
+++ b/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
@@ -38,7 +38,7 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
   const shouldTrackOnPageLoad = () =>
     typeof options.trackOn === 'undefined' || (typeof options.trackOn === 'function' && options.trackOn());
 
-  let previousURL: string | null = null;
+  let previousURL: string | null = location.href;
 
   const trackHistoryPageView = async (): Promise<void> => {
     const newURL = location.href;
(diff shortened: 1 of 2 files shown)


packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -38,7 +38,7 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
   const shouldTrackOnPageLoad = () =>
     typeof options.trackOn === 'undefined' || (typeof options.trackOn === 'function' && options.trackOn());
 
-  let previousURL: string | null = null;
+  let previousURL: string | null = location.href;
```
kevinpagtakhan, line 41: establish baseline or the url on page load

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -74,9 +74,8 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
         // eslint-disable-next-line @typescript-eslint/unbound-method
         globalScope.history.pushState = new Proxy(globalScope.history.pushState, {
           apply: (target, thisArg, [state, unused, url]) => {
+            target.apply(thisArg, [state, unused, url]);
```
kevinpagtakhan, line 77: invoke `pushState` before `trackHistoryPageView`. this ensures that usages of `window.location` on page view tracking refers to the new page

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -102,11 +101,6 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
       return event;
     },
   };
-
-  // Required for unit tests
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-  (plugin as any).__trackHistoryPageView = trackHistoryPageView;
```
kevinpagtakhan, line 108: tests were improved to use `pushState` to remove dependency on this internal fn

liuyang1520 (approved):
LGTM, do we want to apply this fix to the V1 (current version) as well?

justin-fiedler (approved):

PR #399, opened 2023-06-05
chore: 1.x version

kevinpagtakhan:
### Summary

Cleans up all package versions to be on version 1.x
* New 1.x versions are on beta first
* However, release workflow should graduate these versions

### Checklist

* Does your PR have a breaking change?:  Np

### DIFF
--- a/packages/analytics-browser-test/package.json
+++ b/packages/analytics-browser-test/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amplitude/analytics-browser-test",
-  "version": "2.7.0",
+  "version": "1.4.0-beta.0",
   "private": true,
   "description": "",
   "author": "Amplitude Inc",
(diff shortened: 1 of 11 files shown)


packages/analytics-browser-test/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@amplitude/analytics-browser-test",
-  "version": "2.7.0",
+  "version": "1.4.0-beta.0",
```
kevinpagtakhan, line 3: This package started at 1.x and got a major version bump to 2.x when browser package got a version bump from 0.x to 1.x. Putting it on 1.4 since it is the first available one in 1.x.

This is a private package and is not published on npm.

qingzhuozhen (approved):

kelsonpw (approved):

justin-fiedler (approved):

PR #412, opened 2023-06-06
fix: uuid collision

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-76984

Replace `Math.random` with `crypto.getRandomValues` to avoid uuid collisions

### Checklist

### DIFF
--- a/packages/analytics-core/src/utils/uuid.ts
+++ b/packages/analytics-core/src/utils/uuid.ts
@@ -12,7 +12,7 @@ import * as crypto from 'crypto';
 const hex: Array<string> = Array.from({ length: 256 }, (_, index) => index.toString(16).padStart(2, '0'));
 
 export const UUID = (): string => {
-  const r: Uint8Array = crypto.getRandomValues(new Uint8Array(16));
+  const r: Buffer = crypto.randomBytes(16);
 
   r[6] = (r[6] & 0x0f) | 0x40;
   r[8] = (r[8] & 0x3f) | 0x80;
--- a/packages/analytics-core/src/utils/uuid.ts
+++ b/packages/analytics-core/src/utils/uuid.ts
@@ -1,32 +1,21 @@
 /**
- * Source: [jed's gist]{@link https://gist.github.com/982883}.
+ * Source: [jed's gist's comment]{@link https://gist.github.com/jed/982883?permalink_comment_id=3223002#gistcomment-3223002}.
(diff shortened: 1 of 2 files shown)


packages/analytics-core/src/utils/uuid.ts
```diff
@@ -1,32 +1,21 @@
...
-        String(-4e3) + // -4000 +
-        String(-8e3) + // -80000000 +
-        String(-1e11)
-      ) // -100000000000,
-        .replace(
-          // replacing
-          /[018]/g, // zeroes, ones, and eights with
-          UUID, // random hex digits
-        );
+import * as crypto from 'crypto';
```
kevinpagtakhan, line 9: is this package available on the browser?

kevinpagtakhan: should we use `window.crypto`?

kevinpagtakhan: also found out that there's `window.crypto.randomUUID()` that returns a random UUID

Mercy811: It's available on browser (https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues). 

This import is for nodejs environment as `yarn test` returns `Cannot find name 'crypto'` error without importing it. 

`yarn test` runs successfully locally.

kevinpagtakhan: i think this would throw an error on prod since crypto package does not exist. i think you mean `window.crypto`

kevinpagtakhan: one if the pr checks also caught the error

Mercy811: I looked at this post https://stackoverflow.com/questions/70033580/what-is-the-difference-between-import-crypto-from-crypto-vs-just-using-crypto:
- For browser environment, `crypto` is predefined globals. That's why it can be used directly in [JS SDK](https://github.com/amplitude/Amplitude-JavaScript/pull/594)
- For node environment, `crypto` is not predefined

kevinpagtakhan: synced offline and determined that the build throws an error 
image

Mercy811:
This seems to be not practically to simply replace `math.random` to `crypto.getRandomValues` as
- can only use window in browser based packages
- can only use built in modules in node based packages

PR #414, opened 2023-06-07
fix: expose server zone enum

kevinpagtakhan:
### Summary

Bring back ServerZone enum for ampli compatibility

### Checklist

### DIFF
--- a/packages/analytics-types/src/server-zone.ts
+++ b/packages/analytics-types/src/server-zone.ts
@@ -1,1 +1,6 @@
+export enum ServerZone {
+  US = 'US',
+  EU = 'EU',
+}
+
 export type ServerZoneType = 'US' | 'EU';
--- a/packages/analytics-types/src/server-zone.ts
+++ b/packages/analytics-types/src/server-zone.ts
@@ -1,3 +1,6 @@
+/**
+ * @deprecated use ServerZoneType instead
+ */
 export enum ServerZone {
   US = 'US',
   EU = 'EU',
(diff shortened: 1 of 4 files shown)


packages/analytics-types/src/server-zone.ts
```diff
@@ -1 +1,6 @@
+export enum ServerZone {
```
justin-fiedler, line 1:
```suggestion
/**
 * @deprecated use ServerZoneType instead
 */
export enum ServerZone {
```

justin-fiedler (approved):
Looks good. Thanks @kevinpagtakhan. I think we should mark the enum deprecated to steer people to the union type.

falconandy (commented):
@justin-fiedler @kevinpagtakhan 

> for ampli compatibility

1. What compatibility case should be supported?
2. Is it the only incompatibility?
3. Will new codegen `v3` be added to support new `2.x` SDK version? If yes, will codegen use new `ServerZoneType` type (actually constants `US`/`EU`)?

kevinpagtakhan:
> @justin-fiedler @kevinpagtakhan
> 
> > for ampli compatibility
> 
> 1. What compatibility case should be supported?
> 2. Is it the only incompatibility?
> 3. Will new codegen `v3` be added to support new `2.x` SDK version? If yes, will codegen use new `ServerZoneType` type (actually constants `US`/`EU`)?

Hi @falconandy! I removed customer facing enums which includes `ServerZone` for Browser SDK 2.0. I then noticed that ServerZone is used by ampli v2. @justin-fiedler and I agreed that we should make Browser SDK 2.0 compatible with ampli v2, so it does not force us to work on an ampli v2. The work here is to bring back `ServerZone` enum but only for ampli v2 purposes.

PR #415, opened 2023-06-08
feat: log response body from API to logger

justin-fiedler:
### Summary
* [AMP-77375](https://amplitude.atlassian.net/browse/AMP-77375) FIX: Correctly [resolve `send()` promise on error](https://github.com/amplitude/Amplitude-TypeScript/pull/415/files#r1223588296)
* [AMP-77261](https://amplitude.atlassian.net/browse/AMP-77261) log intermediate response body to logger on retries for Node SDK

### Checklist

[AMP-77375]: https://amplitude.atlassian.net/browse/AMP-77375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AMP-77261]: https://amplitude.atlassian.net/browse/AMP-77261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-node/src/transports/http.ts
+++ b/packages/analytics-node/src/transports/http.ts
@@ -27,7 +27,7 @@ export class Http extends BaseTransport implements Transport {
       port: url.port,
       protocol: url.protocol,
     };
-    return new Promise((resolve) => {
+    return new Promise((resolve, reject) => {
       const req = protocol.request(options, (res) => {
         res.setEncoding('utf8');
         let responsePayload = '';
@@ -43,13 +43,13 @@ export class Http extends BaseTransport implements Transport {
               const result = this.buildResponse(parsedResponsePayload);
               resolve(result);
               return;
-            } catch {
-              resolve(null);
+            } catch (e) {
+              reject(e);
             }
           }
         });
       });
-      req.on('error', this.buildResponse.bind(this));
+      req.on('error', reject);
       req.end(requestPayload);
     });
   }
--- a/packages/analytics-node/src/transports/http.ts
+++ b/packages/analytics-node/src/transports/http.ts
@@ -27,7 +27,7 @@ export class Http extends BaseTransport implements Transport {
       port: url.port,
       protocol: url.protocol,
     };
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
       const req = protocol.request(options, (res) => {
         res.setEncoding('utf8');
         let responsePayload = '';
@@ -43,13 +43,13 @@ export class Http extends BaseTransport implements Transport {
               const result = this.buildResponse(parsedResponsePayload);
               resolve(result);
               return;
-            } catch (e) {
-              reject(e);
+            } catch {
+              resolve(null);
             }
           }
         });
       });
-      req.on('error', reject);
+      req.on('error', () => resolve(null));
       req.end(requestPayload);
     });
   }
--- a/packages/analytics-node/test/transport/http.test.ts
+++ b/packages/analytics-node/test/transport/http.test.ts
@@ -121,7 +121,7 @@ describe('http transport', () => {
     expect(request).toHaveBeenCalledTimes(1);
   });
 
-  test('should handle unexpected error', async () => {
+  test('should re-throw unexpected error', async () => {
     const provider = new Http();
     const url = 'http://localhost:3000';
     const payload = {
@@ -151,8 +151,7 @@ describe('http transport', () => {
       } as any;
     });
 
-    const response = await provider.send(url, payload);
-    expect(response).toBe(null);
+    await expect(provider.send(url, payload)).rejects.toThrow('Unexpected token < in JSON at position 0');
     expect(request).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/analytics-node/test/transport/http.test.ts
+++ b/packages/analytics-node/test/transport/http.test.ts
@@ -121,7 +121,7 @@ describe('http transport', () => {
     expect(request).toHaveBeenCalledTimes(1);
   });
 
-  test('should re-throw unexpected error', async () => {
+  test('should handle unexpected error', async () => {
     const provider = new Http();
     const url = 'http://localhost:3000';
     const payload = {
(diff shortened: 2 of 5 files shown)


falconandy (changes requested):
1. Test (status 429) is fixed
2. Added logging for network errors (errors are rejected now `req.on('error', reject);` instead of strange-looking logic `req.on('error', this.buildResponse.bind(this));`)

There is existing logging per event (https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/core-client.ts#L124):
```
result.code === 200
  ? this.config.loggerProvider.log(result.message)
  : this.config.loggerProvider.error(result.message);
```
Now logging is duplicated in some cases. For example, for 7 events (3+3+1):
```
Amplitude Logger [Log]: {"code":200,"eventsIngested":0,"payloadSizeBytes":0,"serverUploadTime":0}
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: {"code":200,"eventsIngested":0,"payloadSizeBytes":0,"serverUploadTime":0}
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: {"code":200,"eventsIngested":0,"payloadSizeBytes":0,"serverUploadTime":0}
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: Event tracked successfully
Amplitude Logger [Log]: Event tracked successfully
```

```
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
Amplitude Logger [Error]: socket hang up
```

Additional notes:
1. maybe events should be re-sent on some/all (network) errors - https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/plugins/destination.ts#L157
2. some strange-lloking flag `useRetry` prevents event re-sending (it is set to `false` on explicit `flush()` call) - I do not understand why the flag is required

packages/analytics-node/test/transport/http.test.ts
```diff
@@ -121,7 +121,7 @@ describe('http transport', () => {
     expect(request).toHaveBeenCalledTimes(1);
   });
 
-  test('should handle unexpected error', async () => {
+  test('should re-throw unexpected error', async () => {
```
justin-fiedler: @falconandy in the latest SDKs we don't throw any errors. We never want the SDK to cause a users app to crash, no matter the state of the SDK.

justin-fiedler:
> maybe events should be re-sent on some/all (network) errors - https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/plugins/destination.ts#L157
Thanks, maybe we can fix in another PR. I want to release this to help us debug the 429s in more detail.

> some strange-looking flag useRetry prevents event re-sending (it is set to false on explicit flush() call) - I do not understand why the flag is required

Thanks for calling this out. I think the intent is that in Node `flush()`  is usually called at shutdown when the user wants to kill the application. In this cause we "fire and forget" all events. In general the events will be retried based on the set configuration. And flush() should be reserved for end of app. We should add this to docs.

packages/analytics-node/src/transports/http.ts
```diff
@@ -49,7 +49,7 @@ export class Http extends BaseTransport implements Transport {
           }
         });
       });
-      req.on('error', this.buildResponse.bind(this));
+      req.on('error', () => resolve(null));
```
justin-fiedler, line 52: Correctly resolve promise on `error`

akramarev (approved):

packages/analytics-types/src/response.ts
```diff
@@ -65,7 +65,7 @@ export interface TimeoutResponse {
   statusCode: number;
 }
 
-export interface OtherReponse {
+export interface OtherResponse {
```
kevinpagtakhan, line 68: 😅

kevinpagtakhan:
> 2\. it is set to `false` on explicit `flush()` call

@falconandy We assumed that flush would be commonly invoked before a node app exits. Retry was disabled to allow node apps to exist immediately after one last attempt.

kevinpagtakhan (approved):

falconandy:
> I think the intent is that in Node flush() is usually called at shutdown when the user wants to kill the application. In this case we "fire and forget" all events. In general the events will be retried based on the provided configuration. And flush() should be reserved for end of app. We should add this to docs.

Shouldn't it be a method like `shutdown()`?

packages/analytics-node/src/transports/http.ts
falconandy: @justin-fiedler 
1. An original network error is thrown away here - it could be logged for debugging.
2. Other transports do rejects (for example, https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/transports/send-beacon.ts#L28) - errors are caught/logged by callers.

packages/analytics-node/test/transport/http.test.ts
falconandy: @justin-fiedler 
I believe, transport objects are used by SDK code and errors are caught/processed in callers on SDK level. For example, https://github.com/amplitude/Amplitude-TypeScript/blob/0c03a57239e1048b1730fa58464b125ba9b8b6ba/packages/analytics-core/src/plugins/destination.ts#L168-L172

justin-fiedler:
> Shouldn't it be a method like `shutdown()`?

We had some strong push back on adding an extra `shutdown()` method as it is seen as a "special case" and breaks from the shared interface of the Next-Gen SDKs.

Maybe it would be better to update `amplitude.flush()` to accept a optional `useRetry = true` in all SDKs. I think it would be better to default to retries to prevent data loss. And in the case that a customer wants a "hard exit" (less likely) they can pass in `amplitude.flush(false)`

PR #416, opened 2023-06-08
feat: simplify init interface

kevinpagtakhan:
### Summary

Backward compatible simplification on init function signature. Extends current interface to allow omitting `userId` as top level parameter.

```ts
init(apiKey: string, options?: BrowserOptions): AmplitudeReturn<void>;
init(apiKey: string, userId?: string, options?: BrowserOptions): AmplitudeReturn<void>;
```

#### Setting user id through init()

Method 1: Omit user id
```ts
amplitude.init(apiKey);
```

```ts
amplitude.init(apiKey, {
  defaultTracking: false,
});
```

Method 2: Setting via top level parameter
```ts
amplitude.init(apiKey, userId);
```

Method 3: Setting via options
```ts
amplitude.init(apiKey, {
  userId
});
```

Method 4: Setting to undefined
```ts
amplitude.init(apiKey, undefined);
```

```ts
amplitude.init(apiKey, {
  userId: undefined,
});
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -45,6 +45,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
       userId = userIdOrOptions;
       options = maybeOptions;
     } else {
+      userId = userIdOrOptions?.userId;
       options = userIdOrOptions;
     }
     return returnWrapper(this._init({ ...options, userId, apiKey }));
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -41,10 +41,10 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     let userId: string | undefined;
     let options: BrowserOptions | undefined;
 
-    if (arguments.length === 3) {
+    if (arguments.length > 2) {
       userId = userIdOrOptions as string | undefined;
       options = maybeOptions;
-    } else if (arguments.length === 2) {
+    } else {
       if (typeof userIdOrOptions === 'string') {
         userId = userIdOrOptions;
         options = undefined;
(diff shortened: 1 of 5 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -37,7 +37,22 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
   previousSessionDeviceId: string | undefined;
   previousSessionUserId: string | undefined;
 
-  init(apiKey = '', userId?: string, options?: BrowserOptions) {
+  init(apiKey = '', userIdOrOptions?: string | BrowserOptions, maybeOptions?: BrowserOptions) {
```
liuyang1520, line 40: Change looks good to me. Just to confirm, should we apply the similar change to [react-native](https://github.com/amplitude/Amplitude-TypeScript/blob/dcf21ca0a2603c042076b6a5d02ea3f184c1ed8c/packages/analytics-react-native/src/react-native-client.ts#L45-L47) one as well for consistency?

liuyang1520 (approved):

qingzhuozhen (approved):

packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -51,6 +51,48 @@ describe('browser-client', () => {
...
+
+    test('should set user to options.userId', async () => {
+      client.setOptOut(true);
+      await client.init(apiKey, {
+        userId,
+      }).promise;
+      expect(client.getUserId()).toBe(userId);
+    });
+
+    test('should set user id using top level parameter as priority', async () => {
```
justin-fiedler, line 88: I agree with this but it sort of goes against our standard of `options` overriding previous values (e.g. `track(event, options)`.

That said I think it is probably what most people would expect. So ✅

justin-fiedler (approved):

packages/analytics-browser/test/browser-client.test.ts
kevinpagtakhan: yeah, same thought! this is also status quo for 1.0, so just carried over here

packages/analytics-browser/src/browser-client.ts
liuyang1520: No need to be in single PR for sure, thank you!

PR #425, opened 2023-06-09
feat: added extendSession() method to Browser Client

justin-fiedler:
### Summary

** Add extendSession() method to Browser SDK to allow customers to extend the current session without sending events / using quota.

```typescript
amplitude.extendSession();
```

### Checklist

### DIFF
--- a/packages/analytics-browser/test/browser-client.test.ts
+++ b/packages/analytics-browser/test/browser-client.test.ts
@@ -570,7 +570,7 @@ describe('browser-client', () => {
       const eventTime1 = client.config.lastEventTime ?? -1;
       expect(eventTime1 > 0).toBeTruthy();
 
-      // wait for session to almost expire
+      // wait for session to almost expire, then extend it
       await new Promise<void>((resolve) =>
         setTimeout(() => {
           client.extendSession();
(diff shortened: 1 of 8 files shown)


packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -551,6 +551,92 @@ describe('browser-client', () => {
...
+          await client.track('test 3').promise;
+          resolve();
+        }, 21),
+      );
+      // assert session id is changed
+      expect(client.config.sessionId).not.toBe(firstSessionId);
+      expect(client.config.sessionId ?? -1 > firstSessionId).toBeTruthy();
+    });
+
+    test('should extendSession using proxy', async () => {
```
Mercy811, line 617: How's the testing code reflecting that a proxy is used?

Mercy811 (approved):

yuhao900914 (approved):

packages/analytics-browser/test/browser-client.test.ts
justin-fiedler: It fakes it by not calling init() under after calling other methods, causing them to be queued and run upon init(), similar to the proxy.

kevinpagtakhan:
we should also update the snippet loader to include the new public facing method

PR #428, opened 2023-06-13
feat: add option for instance name

kevinpagtakhan:
### Summary

Add options for instance name to be used for Analytics connector

```ts
amplitude.init(API_KEY, {
  instanceName: 'env',
});
```

### Checklist


justin-fiedler (commented):
Changes look good for AnalyticsConnector.

Wondering if we should also use `instanceName` for Storage creation like in the Mobile SDKs  i.e. all instances with the same name also share the same storage.

kevinpagtakhan:
> all instances with the same name also share the same storage.

Are they not using api key?

qingzhuozhen (approved):

justin-fiedler (approved):

tmchow:
@kevinpagtakhan what exactly is this for? are there docs explaining this attribute?

PR #435, opened 2023-06-15
feat(storage): add support for sessionStorage as identityStorage

nutgaard:
Proposed fix for #434

### Summary

Adds support for sessionStorage as identityStorage

```ts
amplitude.init(<YOUR_API_KEY>, {
  identityStorage: 'sessionStorage',
});
```

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:  No

PR #437, opened 2023-06-15
feat: adds session storage option for identity storage

### LINKED ISSUE #434: Introduce sessionStorage for identityStorage option
Hi! Is it possible to copy this feature to this sdk? https://github.com/amplitude/amplitude-javascript/issues/317

## Summary

Se issue: https://github.com/amplitude/amplitude-javascript/issues/317

## Motivations

Because of privacy we want to limit our users to sessions. While localStorage is an improvement over cookies, we would love to use sessionStorage for our use.

kevinpagtakhan:
### Summary

Resolves #434

Add a new identity storage option: `sessionStorage`.

```ts
amplitude.init(<YOUR_API_KEY>, {
  identityStorage: 'sessionStorage',
});
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/storage/session-storage.ts
+++ b/packages/analytics-browser/src/storage/session-storage.ts
@@ -1,0 +1,67 @@
+import { getGlobalScope } from '@amplitude/analytics-client-common';
+import { Storage } from '@amplitude/analytics-types';
+
+export class SessionStorage<T> implements Storage<T> {
+  async isEnabled(): Promise<boolean> {
+    /* istanbul ignore if */
+    if (!getGlobalScope()) {
+      return false;
+    }
+
+    const random = String(Date.now());
+    const testStorage = new SessionStorage<string>();
+    const testKey = 'AMP_TEST';
+    try {
+      await testStorage.set(testKey, random);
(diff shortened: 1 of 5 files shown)


nutgaard:
This seems quite similar to #435 😿

packages/analytics-browser/src/storage/session-storage.ts
```diff
@@ -0,0 +1,67 @@
...
+  async get(key: string): Promise<T | undefined> {
+    try {
+      const value = await this.getRaw(key);
+      if (!value) {
+        return undefined;
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return JSON.parse(value);
+    } catch {
+      /* istanbul ignore next */
```
kelvin-lu, line 35: i would expect a warning or error in the console in these cases for better debugging - in case the error is that we do not have access to session storage

kelvin-lu (approved):
lgtm - expected that this looks almost identical to the local storage variant

kevinpagtakhan:
> This seems quite similar to #435 😿

@nutgaard apologies, i did not notice #435.

@kelvin-lu i will close this in favor of #435.

PR #447, opened 2023-06-20
fix: updated session logic and tests

falconandy:
### Summary

Slightly updates session logic and added new tests (copied from Amplitude-Kotlin).
Additional notes - in the Jira task.

### Checklist

### DIFF
--- a/packages/analytics-react-native/src/react-native-client.ts
+++ b/packages/analytics-react-native/src/react-native-client.ts
@@ -230,12 +230,17 @@ export class AmplitudeReactNative extends AmplitudeCore {
         event = { ...event, time: eventTime };
       }
 
-      if (event.event_type != START_SESSION_EVENT && event.event_type != END_SESSION_EVENT) {
-        if (this.appState !== 'active') {
-          this.startNewSessionIfNeeded(eventTime);
+      const isSessionEvent = event.event_type === START_SESSION_EVENT || event.event_type === END_SESSION_EVENT;
+      const isCustomEventSessionId =
+        !isSessionEvent && event.session_id != undefined && event.session_id !== this.getSessionId();
+      if (!isCustomEventSessionId) {
+        if (!isSessionEvent) {
+          if (this.appState !== 'active') {
+            this.startNewSessionIfNeeded(eventTime);
+          }
         }
+        this.config.lastEventTime = eventTime;
       }
-      this.config.lastEventTime = eventTime;
 
       if (event.session_id == undefined) {
         event.session_id = this.getSessionId();
--- a/packages/analytics-react-native/src/react-native-client.ts
+++ b/packages/analytics-react-native/src/react-native-client.ts
@@ -84,11 +84,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
 
     // Step 4: Manage session
     this.appState = AppState.currentState;
-    const currentTimestamp = this.currentTimeMillis();
-    const isNewSession =
-      this.appState === 'active'
-        ? this.enterForeground(currentTimestamp)
-        : this.shouldStartNewSession(currentTimestamp);
+    const isNewSession = this.startNewSessionIfNeeded(this.currentTimeMillis());
     this.config.loggerProvider?.log(
       `Init: startNewSessionIfNeeded = ${isNewSession ? 'yes' : 'no'}, sessionId = ${
         this.getSessionId() ?? 'undefined'
@@ -174,7 +170,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
   }
 
   getSessionId() {
-    return this.explicitSessionId ?? this.config?.sessionId;
+    return this.config?.sessionId;
   }
 
   setSessionId(sessionId: number) {
--- a/packages/analytics-react-native/src/react-native-client.ts
+++ b/packages/analytics-react-native/src/react-native-client.ts
@@ -84,7 +84,11 @@ export class AmplitudeReactNative extends AmplitudeCore {
 
     // Step 4: Manage session
     this.appState = AppState.currentState;
-    const isNewSession = this.startNewSessionIfNeeded();
+    const currentTimestamp = this.currentTimeMillis();
+    const isNewSession =
+      this.appState === 'active'
+        ? this.enterForeground(currentTimestamp)
+        : this.shouldStartNewSession(currentTimestamp);
     this.config.loggerProvider?.log(
       `Init: startNewSessionIfNeeded = ${isNewSession ? 'yes' : 'no'}, sessionId = ${
         this.getSessionId() ?? 'undefined'
@@ -170,7 +174,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
   }
 
   getSessionId() {
-    return this.config?.sessionId;
+    return this.explicitSessionId ?? this.config?.sessionId;
   }
 
   setSessionId(sessionId: number) {
@@ -242,7 +246,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
       }
 
       if (event.event_id === undefined) {
-        const eventId = (this.config.lastEventId ?? -1) + 1;
+        const eventId = (this.config.lastEventId ?? 0) + 1;
         event = { ...event, event_id: eventId };
         this.config.lastEventId = eventId;
       }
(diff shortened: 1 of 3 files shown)


packages/analytics-react-native/src/react-native-client.ts
```diff
@@ -170,7 +174,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
   }
 
   getSessionId() {
-    return this.config?.sessionId;
+    return this.explicitSessionId ?? this.config?.sessionId;
```
justin-fiedler, line 177: I think we should revert this so it matches [browser-client](https://github.com/amplitude/Amplitude-TypeScript/blob/0f5fccc83d3f7f0a80adc4a0807fbd7e71c72e4a/packages/analytics-react-native/src/react-native-client.ts#L173).

I would expect this to set the `sessionId` but also adhere to other session rules and expire with sessionTimeout.

packages/analytics-react-native/test/react-native-client.test.ts
```diff
@@ -572,26 +572,26 @@ describe('react-native-client', () => {
...
+        events.forEach((event, i) => expect(event.event_id).toEqual(i + 1));
 
         expect(events[0].event_type).toEqual('session_end');
         expect(events[0].session_id).toEqual(500);
         expect(events[0].time).toEqual(851);
 
         expect(events[1].event_type).toEqual('session_start');
-        expect(events[1].session_id).toEqual(950);
-        expect(events[1].time).toEqual(950);
+        expect(events[1].session_id).toEqual(1000);
```
justin-fiedler, line 582: Is this because the app starts in the background? I would expect the session to start on init.

justin-fiedler (commented):
Thanks @falconandy. Appreciate the updates. See questions.

packages/analytics-react-native/src/react-native-client.ts
```diff
@@ -242,7 +246,7 @@ export class AmplitudeReactNative extends AmplitudeCore {
       }
 
       if (event.event_id === undefined) {
-        const eventId = (this.config.lastEventId ?? -1) + 1;
+        const eventId = (this.config.lastEventId ?? 0) + 1;
```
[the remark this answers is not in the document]
falconandy: Amplitude-Kotlin/Amplitude-Swift start with 1
legacy Amplitude-Andorid/Amplitude-iOS use sqlite `AUTOINCREMENT` column - it starts with 1, I believe

packages/analytics-react-native/src/react-native-client.ts
falconandy: ok, it depends on comment below

packages/analytics-react-native/test/react-native-client.test.ts
falconandy: Yes, in this test the app starts in the background.
Case A: the app starts at time 950, does something (not tracked to Amplitude), closes
Case B: the app starts at time 950, track an event  at time 1000, closes.

Option 1:
Case A: session_start(950), session_end(950) - here we have empty session
Case B: session_start(950), event(1000), session_end(1000)

Option 2:
Case A: -
Case B: session_start(950), event(1000), session_end(1000)

Option 3:
Case A: -
Case B: session_start(1000), event(1000), session_end(1000)

Now I do Option 3. Do you expect Option 1?

packages/analytics-react-native/src/react-native-client.ts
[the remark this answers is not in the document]
justin-fiedler: Thanks. That makes sense. Maybe we should update web to match at some point.

packages/analytics-react-native/test/react-native-client.test.ts
justin-fiedler: Thanks, in this case I would expect `Option 1`. This will become more important to track "empty" sessions since we are being to add session replay functionality that will depend on session start/end regardless of other Amplitude track()ing.

packages/analytics-react-native/src/react-native-client.ts
falconandy: Reverted

packages/analytics-react-native/test/react-native-client.test.ts
falconandy: Reverted to `Option 1`

packages/analytics-react-native/test/react-native-sessions.test.ts
```diff
@@ -0,0 +1,523 @@
...
+    expect(events[1].session_id).toEqual(950);
+    expect(events[1].time).toEqual(1000);
+
+    expect(events[2].event_type).toEqual('event-2');
+    expect(events[2].session_id).toEqual(3000);
+    expect(events[2].time).toEqual(1050);
+
+    expect(events[3].event_type).toEqual('event-3');
+    expect(events[3].session_id).toEqual(950);
+    expect(events[3].time).toEqual(1100);
```
justin-fiedler, line 374: Not 100% what should happen here. Some options

1. 1 session like you have now
2. "1.5" sessions - see code below.
3. 2.5 sessions - same as 2 but with session_start/end events around event-2.

I think 3 would be too hard to implement, but what are your thoughts on 2? Should an out of session event extend the current session?

```
'session_start', 950, 950
'event-1', 950, 1000
'session_end', 950, 1001

'event-2', 3000, 1050

'session_start', 1100, 1099
'event-3', 1100, 1100
```

packages/analytics-react-native/test/react-native-sessions.test.ts
```diff
@@ -0,0 +1,523 @@
...
+  test('close foreground and background events should not start new session', async () => {
+    const send = jest.fn().mockReturnValue(sendResponse);
+    const cookieStorage = new core.MemoryStorage<UserSession>();
+
+    const client = new AmplitudeReactNativeTest(950);
+    await client.init(API_KEY, undefined, clientOptions(send, cookieStorage, true)).promise;
+
+    client.setForeground(1000);
+    client.track(createEvent(1500, 'event-1'));
+    client.setForeground(2000);
```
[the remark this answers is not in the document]
falconandy: Fixed.

packages/analytics-react-native/test/react-native-sessions.test.ts
```diff
@@ -0,0 +1,523 @@
...
+    expect(events[0].session_id).toEqual(950);
+    expect(events[0].time).toEqual(950);
+
+    expect(events[1].event_type).toEqual('event-1');
+    expect(events[1].session_id).toEqual(950);
+    expect(events[1].time).toEqual(1500);
+
+    expect(events[2].event_type).toEqual('session_end');
+    expect(events[2].session_id).toEqual(950);
+    expect(events[2].time).toEqual(1501);
```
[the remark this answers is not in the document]
falconandy: Thank you - found a bug, fixed.

packages/analytics-react-native/test/react-native-sessions.test.ts
```diff
@@ -0,0 +1,523 @@
...
+
+    expect(events[3].event_type).toEqual('session_start');
+    expect(events[3].session_id).toEqual(3000);
+    expect(events[3].time).toEqual(3000);
```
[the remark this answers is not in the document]
falconandy: I can't find `eventTime - 1` code for `session_start` event in browser. Only `eventTime + 1` for `session_end` event.
https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L190-L196

packages/analytics-react-native/test/react-native-sessions.test.ts
falconandy: @justin-fiedler 
Maybe the test is not realistic. Based on documentation (https://www.docs.developers.amplitude.com/data/sdks/browser-2/#custom-session-id) ` When setting a custom session ID, make sure the value is in milliseconds since epoch` - I violated it in the test (seesion _id 3000 is greater than timestamp 1050).
I would stay with 1 for now - I think, the behavior should be the same in all client SDKs. WDYT?

packages/analytics-react-native/test/react-native-sessions.test.ts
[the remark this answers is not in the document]
justin-fiedler: Thanks for checking. I was mistaken.

packages/analytics-react-native/test/react-native-sessions.test.ts
justin-fiedler: I discussed with the team and we feel that option #2 is the right approach, i.e. out-of-session should not extend the current session.

In general, if the `event.sessionId` doesn't equal `config.sessionId` then it shouldn't update `lastEventTime`

falconandy: @justin-fiedler Please, take a look at updated code and tests

justin-fiedler (approved):
Thanks @falconandy. LGTM.

PR #454, opened 2023-06-22
fix: Fix React Native 0.72 deprecation warning

johnf:
### Summary

React Native 0.72 has update gradle things and has the following warning

```
> Task :amplitude_analytics-react-native:processDebugManifest
package="com.amplitude.reactnative" found in source AndroidManifest.xml: /home/johnf/work/gladly/mobile/node_modules/@amplitude/analytics-react-native/android/src/main/AndroidManifest.xml.
Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated.
Please instead set the namespace (or testNamespace) in the module's build.gradle file, as described here: https://developer.android.com/studio/build/configure-app-module#set-namespace
This migration can be done automatically using the AGP Upgrade Assistant, please refer to https://developer.android.com/studio/build/agp-upgrade-assistant for more information.
```
This PR updates to the newer format.

### Checklist

* Does your PR have a breaking change?: I don't think so, from the docs it sounds like this has been in the works for a while but maybe it would break for someone using a very old gradle?

### DIFF
--- a/packages/analytics-react-native/android/build.gradle
+++ b/packages/analytics-react-native/android/build.gradle
@@ -19,6 +19,8 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
+    namespace "com.amplitude.reactnative"
+
     compileSdkVersion safeExtGet('compileSdkVersion', 29)
     buildToolsVersion safeExtGet('buildToolsVersion', '29.0.2')
     defaultConfig {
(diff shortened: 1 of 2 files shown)


packages/analytics-react-native/android/src/main/AndroidManifest.xml
```diff
@@ -1,4 +1,3 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.amplitude.reactnative">
```
alexhernandez, line 2: You can leave this alone for backwards compatibility I believe

packages/analytics-react-native/android/build.gradle
```diff
@@ -19,6 +19,8 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
+    namespace "com.amplitude.reactnative"
```
alexhernandez, line 22: Wrap this line for backwards compatibility:
```def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
  namespace "com.amplitude.reactnative"
}

PR #456, opened 2023-06-22
fix: add app_set_id, idfa and idfv support

falconandy:
### Summary

Add support for:
Android (app_set_id)
iOS (idfa, idfv)

1. For some reason idfa support in amplitude-ios/amplitude-swift is "external":  callback in amplitude-ios (https://www.docs.developers.amplitude.com/data/sdks/ios/#advertiser-id), example plugin in amplitude-swift (https://github.com/amplitude/Amplitude-Swift/blob/main/Examples/AmplitudeSwiftUIExample/AmplitudeSwiftUIExample/ExamplePlugins/IDFACollectionPlugin.swift). Here, in React Native, I added new tracking options `idfa`/`idfv` (both `true` by default) - I don't know how to implement something "external" similar to the callback/plugin.
2. To be able to get `idfa` : Set up a [NSUserTrackingUsageDescription](https://developer.apple.com/documentation/bundleresources/information_property_list/nsusertrackingusagedescription?language=objc) to display a system-permission alert request for your app installed on end-user devices.
3. Tested in emulators only.

### Checklist


packages/analytics-react-native/ios/AppleContextProvider.swift
```diff
@@ -1,3 +1,5 @@
+import AdSupport
+import AppTrackingTransparency
```
justin-fiedler, line 2: Thanks @falconandy.  I see why Swift has this externalized. To avoid the extra dependencies and warnings, I think we should remove the IDFA access from the core SDK and instead provide an example ReactNative plugin (TS) for [accessing IDFA](https://www.npmjs.com/package/@sparkfabrik/react-native-idfa-aaid) and setting it on the event payload.

justin-fiedler (changes requested):

packages/analytics-react-native/ios/AppleContextProvider.swift
falconandy: > remove the IDFA access from the core SDK and instead provide an example ReactNative plugin

Ready

examples/plugins/react-native-idfa-plugin/ios/IdfaPlugin.swift
```diff
@@ -0,0 +1,42 @@
+import AdSupport
+import AppTrackingTransparency
+import Foundation
+
+@objc(IdfaPlugin)
+class IdfaPlugin: NSObject {
```
justin-fiedler, line 6: Thanks for implementing this @falconandy.

I think I would prefer to use a 3rd party RN NPM module for the IDFA access.

That way don't need to maintain/test this functionality ourselves and it will be easier for users to add our example to their project (just copy 1 file and add the npm dependency). wdyt?

cc @yuhao900914 @liuyang1520

Mercy811:
Are we going to add `idfa` in this PR. I saw it in the description but not in the code.

examples/plugins/react-native-idfa-plugin/ios/IdfaPlugin.swift
falconandy: @justin-fiedler Yes, we can adopt the example plugin to use a 3rd party package.
Pros: easier to setup for users.
Cons: new dependency.

Also we could provide both options (with/without 3rd party package).

packages/analytics-react-native/ios/AppleContextProvider.swift
yuhao900914: More info here. I think including idfa in the app will cause the app to get rejected from the app store. That's why we decided to  be customer driven. 
https://github.com/amplitude/Amplitude-iOS/blob/94b4dcfda9ed9edd6463d9a657bf30c9cf075f09/CHANGELOG.md?plain=1#L309

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeModule.kt
```diff
@@ -20,9 +20,10 @@ class AmplitudeReactNativeModule(private val reactContext: ReactApplicationConte
     }
 
     @ReactMethod
-    private fun getApplicationContext(shouldTrackAdid: Boolean, promise: Promise) {
+    private fun getApplicationContext(options: ReadableMap, promise: Promise) {
+        val trackAdid = if (options.hasKey("adid")) options.getBoolean("adid") else false
```
yuhao900914, line 24: Should adid be included in the TrackingOptions? Like how other mobile sdk did? 
https://github.com/amplitude/Amplitude-Kotlin/blob/main/android/src/main/java/com/amplitude/android/plugins/AndroidContextPlugin.kt#L120

falconandy: @yuhao900914
https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-types/src/config/react-native.ts#L28 ?

falconandy:
@Mercy811 
> Are we going to add idfa in this PR. I saw it in the description but not in the code.

https://github.com/amplitude/Amplitude-TypeScript/blob/AMP-60685-appsetid-idfa-idfv-support/examples/plugins/react-native-idfa-plugin/idfaPlugin.ts

examples/plugins/react-native-idfa-plugin/ios/IdfaPlugin.swift
falconandy: Ready

justin-fiedler (approved):

PR #460, opened 2023-06-29
fix: allow plugins to teardown to remove listeners

### LINKED ISSUE #457: Undefined document in file-download-tracking
## Expected Behavior

No error if `document` is undefined.

## Current Behavior

In a next.js app with SSR enabled, and amplitude v2.1, I'm seeing the error `document is not defined` at this line 

https://github.com/amplitude/Amplitude-TypeScript/blob/8f4ea010279fb21190a2c0595d4ae8a7d9e987ce/packages/analytics-browser/src/plugins/file-download-tracking.ts#L49

This seems similar to https://github.com/amplitude/Amplitude-TypeScript/issues/444. I have not seen this with prior versions of the SDK.

## Possible Solution

Check if `document` is defined; fall back to an empty array.

## Steps to Reproduce

I believe this would have the same reproduction as https://github.com/amplitude/Amplitude-TypeScript/issues/444. Run a next.js app with SSR, and include amplitude with file download tracking enabled.

## Environment

- JS SDK Version: `"@amplitude/analytics-browser": "2.1.0"`
- Installation Method: yarn
- Browser and Version:

kevinpagtakhan:
### Summary

* Adds teardown method to plugins interface
* Removes event listeners on plugin teardown
* Fixes #457

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
+++ b/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
@@ -27,7 +27,8 @@ export const formInteractionTracking = (): EnrichmentPlugin => {
   };
   const removeClickListeners = () => {
     eventListeners.forEach(({ element, type, handler }) => {
-      element.removeEventListener(type, handler);
+      /* istanbul ignore next */
+      element?.removeEventListener(type, handler);
     });
     eventListeners = [];
   };
--- a/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
+++ b/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
@@ -8,7 +8,30 @@ import {
 } from '../constants';
 import { BrowserConfig } from '../config';
 
+interface EventListener {
+  element: Element;
+  type: 'change' | 'submit';
+  handler: () => void;
+}
+
 export const formInteractionTracking = (): EnrichmentPlugin => {
+  let observer: MutationObserver | undefined;
+  let eventListeners: EventListener[] = [];
+  const addEventListener = (element: Element, type: 'change' | 'submit', handler: () => void) => {
+    element.addEventListener(type, handler);
+    eventListeners.push({
+      element,
+      type,
+      handler,
+    });
+  };
+  const removeClickListeners = () => {
+    eventListeners.forEach(({ element, type, handler }) => {
+      element.removeEventListener(type, handler);
+    });
+    eventListeners = [];
+  };
+
   const name = '@amplitude/plugin-form-interaction-tracking-browser';
   const type = 'enrichment';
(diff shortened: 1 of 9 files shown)


justin-fiedler (commented):
LGTM

packages/analytics-browser/src/plugins/form-interaction-tracking.ts
```diff
@@ -8,7 +8,30 @@ import {
...
+    element.addEventListener(type, handler);
+    eventListeners.push({
+      element,
+      type,
+      handler,
+    });
+  };
+  const removeClickListeners = () => {
+    eventListeners.forEach(({ element, type, handler }) => {
+      element.removeEventListener(type, handler);
```
liuyang1520, line 30: Maybe `element?.removeEventListener`, not sure whether if `element` is removed from the DOM, this reference will be null or not.

liuyang1520 (approved):

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
```diff
@@ -290,6 +290,24 @@ describe('pageViewTrackingPlugin', () => {
     });
   });
 
+  describe('teardown', () => {
```
justin-fiedler, line 293: In a recent session fix for mobile, sometimes `onExitForeground` would not get called (crash, etc). Do you see any potential for a similar issue here? 

Is there any harm in calling `teardown()` in `setup()` to be safe?

packages/analytics-browser/src/plugins/form-interaction-tracking.ts
kevinpagtakhan: fixed!

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
justin-fiedler: Discussed in Slack.

image

justin-fiedler (approved):
Please also add to docs 🙏

packages/analytics-browser/src/plugins/form-interaction-tracking.ts
```diff
@@ -8,7 +8,31 @@ import {
...
+interface EventListener {
+  element: Element;
+  type: 'change' | 'submit';
+  handler: () => void;
+}
+
 export const formInteractionTracking = (): EnrichmentPlugin => {
+  let observer: MutationObserver | undefined;
+  let eventListeners: EventListener[] = [];
+  const addEventListener = (element: Element, type: 'change' | 'submit', handler: () => void) => {
```
kelsonpw, line 20: nit: I find it confusing to see a function named `addEventListener`, my brain immediately jumps to "is this calling `window.addEventListener` implicitly? Same with remove. Nbd

PR #471, opened 2023-07-06
fix(plugins): update timing of sending batches of session replay events

kwalker3690:
### Summary
This PR updates the timing of when batches of session replay events as the following:
- As soon as the first two events in a session sequence are recorded they are sent to the API
- Then batches are sent every 1000, 2000, 3000....up to 10000 max milliseconds.
- There is also a size limit kept on the event batch of 20MB, if the next event would cause the batch of events to go over 20MB the list will be sent.

This PR also adds a default of masking all inputs to the recorder, this will be updated to be configurable in future revisions.

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -35,7 +35,7 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
   stopRecordingEvents: ReturnType<typeof record> | null = null;
   maxPersistedEventsSize = MAX_EVENT_LIST_SIZE_IN_BYTES;
   interval = MIN_INTERVAL;
-  timeSinceLastSend: number | null = null;
+  timeAtLastSend: number | null = null;
 
   async setup(config: BrowserConfig) {
     config.loggerProvider.log('Installing @amplitude/plugin-session-replay.');
@@ -122,13 +122,19 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     });
   }
 
+  /**
+   * Determines whether to send the events list to the backend and start a new
+   * empty events list, based on the size of the list as well as the last time sent
+   * @param nextEventString
+   * @returns boolean
+   */
   shouldSplitEventsList = (nextEventString: string): boolean => {
     const sizeOfNextEvent = new Blob([nextEventString]).size;
     const sizeOfEventsList = new Blob(this.events).size;
     if (sizeOfEventsList + sizeOfNextEvent >= this.maxPersistedEventsSize) {
       return true;
     }
-    if (this.timeSinceLastSend !== null && Date.now() - this.timeSinceLastSend > this.interval) {
+    if (this.timeAtLastSend !== null && Date.now() - this.timeAtLastSend > this.interval) {
       this.interval = Math.min(MAX_INTERVAL, this.interval + MIN_INTERVAL);
       return true;
     }
@@ -315,7 +321,7 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     if (err) {
       this.config.loggerProvider.error(err);
     } else if (success) {
-      this.timeSinceLastSend = Date.now();
+      this.timeAtLastSend = Date.now();
       this.config.loggerProvider.log(success);
     }
   }
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -3,7 +3,6 @@ import { BrowserConfig, Event, Status } from '@amplitude/analytics-types';
 import * as IDBKeyVal from 'idb-keyval';
 import { pack, record } from 'rrweb';
 import { DEFAULT_SESSION_END_EVENT, DEFAULT_SESSION_REPLAY_PROPERTY, DEFAULT_SESSION_START_EVENT } from './constants';
-import { shouldSplitEventsList } from './helpers';
 import { MAX_RETRIES_EXCEEDED_MESSAGE, STORAGE_FAILURE, SUCCESS_MESSAGE, UNEXPECTED_ERROR_MESSAGE } from './messages';
 import {
   Events,
@@ -17,6 +16,8 @@ const SESSION_REPLAY_SERVER_URL = 'https://api-secure.amplitude.com/sessions/tra
 const STORAGE_PREFIX = `${AMPLITUDE_PREFIX}_replay_unsent`;
 const PAYLOAD_ESTIMATED_SIZE_IN_BYTES_WITHOUT_EVENTS = 200; // derived by JSON stringifying an example payload without events
 const MAX_EVENT_LIST_SIZE_IN_BYTES = 20 * 1000000 - PAYLOAD_ESTIMATED_SIZE_IN_BYTES_WITHOUT_EVENTS;
+const MIN_INTERVAL = 1 * 1000; // 1 second
+const MAX_INTERVAL = 10 * 1000; // 10 seconds
 
 class SessionReplay implements SessionReplayEnrichmentPlugin {
   name = '@amplitude/plugin-session-replay-browser';
@@ -33,6 +34,8 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
   queue: SessionReplayContext[] = [];
   stopRecordingEvents: ReturnType<typeof record> | null = null;
   maxPersistedEventsSize = MAX_EVENT_LIST_SIZE_IN_BYTES;
+  interval = MIN_INTERVAL;
+  timeSinceLastSend: number | null = null;
 
   async setup(config: BrowserConfig) {
     config.loggerProvider.log('Installing @amplitude/plugin-session-replay.');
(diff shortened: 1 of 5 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -103,9 +118,23 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
         void this.storeEventsForSession(this.events, this.currentSequenceId);
       },
       packFn: pack,
+      maskAllInputs: true,
```
kickerbear, line 121: what does maskAllInputs do?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -89,7 +92,19 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     this.stopRecordingEvents = record({
       emit: (event) => {
         const eventString = JSON.stringify(event);
-        const shouldSplit = shouldSplitEventsList(this.events, eventString, this.maxPersistedEventsSize);
+        // Send the first two recorded events immediately
```
kickerbear, line 95: I don't quite follow how this is executing for the first 2 events instead of just the first event. What's the case?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -89,7 +92,19 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
+        if (this.events.length === 1 && this.currentSequenceId === 0) {
+          this.sendEventsList({
+            events: this.events.concat(eventString),
+            sequenceId: this.currentSequenceId,
```
kickerbear, line 99: Sorry that I'm just hooking into this PR for the config - if you don't mind, what is the sequenceId used for? Is it the batch?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -89,7 +92,19 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
+            sessionId: this.config.sessionId as number,
```
kickerbear, line 100: why is this typed as non-number initially?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -103,9 +118,23 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
         void this.storeEventsForSession(this.events, this.currentSequenceId);
       },
       packFn: pack,
+      maskAllInputs: true,
     });
   }
 
+  shouldSplitEventsList = (nextEventString: string): boolean => {
```
kickerbear, line 125: Can we add a comment to what the purpose of this function is?

kickerbear: Reading this - it seems to be controlling the batching logic?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -131,7 +160,7 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     tryable.forEach((context) => {
       this.queue = this.queue.concat(context);
       if (context.timeout === 0) {
-        this.schedule(this.config.flushIntervalMillis);
+        this.schedule(0);
```
kickerbear, line 163: What does this schedule function do?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -286,6 +315,7 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     if (err) {
       this.config.loggerProvider.error(err);
     } else if (success) {
+      this.timeSinceLastSend = Date.now();
```
kickerbear, line 318: time since implies it's the elapsed MS or S since last send - but this looks like it's MS since epoch instead... If i've understood this right, can we rename this?

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -247,27 +246,82 @@ describe('SessionReplayPlugin', () => {
...
   describe('recordEvents', () => {
+    test('should send the first two emitted events immediately', () => {
+      const sessionReplay = sessionReplayPlugin();
+      const send = jest.spyOn(sessionReplay, 'send').mockReturnValueOnce(Promise.resolve());
+      sessionReplay.config = mockConfig;
+      sessionReplay.recordEvents();
+      expect(sessionReplay.events).toEqual([]);
+      const recordArg = record.mock.calls[0][0];
+      recordArg?.emit && recordArg?.emit(mockEvent);
+      recordArg?.emit && recordArg?.emit(mockEvent);
```
kickerbear, line 257: Woah - i'm unfamiliar with this library, can you help me understand what lines 255-257 are doing?

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -247,27 +246,82 @@ describe('SessionReplayPlugin', () => {
   });
 
   describe('recordEvents', () => {
+    test('should send the first two emitted events immediately', () => {
+      const sessionReplay = sessionReplayPlugin();
+      const send = jest.spyOn(sessionReplay, 'send').mockReturnValueOnce(Promise.resolve());
+      sessionReplay.config = mockConfig;
+      sessionReplay.recordEvents();
+      expect(sessionReplay.events).toEqual([]);
```
kickerbear, line 254: Can we add a comment for what each expect is doing? This seems to be saying "right after setup, there's no session data to send, so we shouldn't have sent any session events" - is that right? How is it different than the next expect on line 258?

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -247,27 +246,82 @@ describe('SessionReplayPlugin', () => {
...
+      const recordArg = record.mock.calls[0][0];
+      // Emit first two events, which get sent immediately
+      recordArg?.emit && recordArg?.emit(mockEvent);
+      recordArg?.emit && recordArg?.emit(mockEvent);
+      expect(sendEventsList).toHaveBeenCalledTimes(1);
+      sendEventsList.mockClear();
+      // Emit third event, which is not sent immediately
+      recordArg?.emit && recordArg?.emit(mockEvent);
+      expect(sendEventsList).toHaveBeenCalledTimes(0);
+      // Emit fourth event and advance timers to interval
```
kickerbear, line 307: This test is clearer to me

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -289,7 +343,7 @@ describe('SessionReplayPlugin', () => {
       expect(update.mock.calls[0][1]({})).toEqual({
         123: {
           events: [mockEventString],
-          sequenceId: 1,
+          sequenceId: 2,
```
kickerbear, line 346: Before, this test wasn't splitting into 2 distinct batches?

kickerbear (approved):
LGTM - I added a few comments to help me understand how this code works and the changes we're making - if you have time, would love to learn but they're non-blocking

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: At this point in the code, we haven't added the second emitted event to `this.events` yet, so there is the first emitted event in `this.events`, then the second `event`, and we send the two of those together. We end up never adding the second emitted event to `this.events` at all

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: It's just typed as `number | undefined`, but at this point in the code we know that it's been defined, this allows us to not have to add unnecessary logic to ensure that it is defined (as well as unnecessary tests, the SDK requires 100% test coverage)

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: pretty much what it says, see the guide: https://github.com/rrweb-io/rrweb/blob/master/guide.md, we'll be adding configuration to this down the line to allow customers to unmask some inputs, but legal asked that this be in place before we start dogfooding

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: Yeah thats right

packages/plugin-session-replay-browser/test/session-replay.test.ts
kwalker3690: Yeah I can add some more comments

packages/plugin-session-replay-browser/test/session-replay.test.ts
kwalker3690: I needed a way to test the internal functionality of the callback that we pass to rrweb here: https://github.com/amplitude/Amplitude-TypeScript/blob/721a3997673d700b6bda9302b076be0f3fab7c09/packages/plugin-session-replay-browser/src/session-replay.ts#L93-L93
So all this is doing is grabbing the callback from the argument passed to the rrweb record (which is spied on as `record` here) and manually calling it with a mock event, like rrweb itself will do when performing recordings

packages/plugin-session-replay-browser/test/session-replay.test.ts
kwalker3690: It was split into two batches before, this actually perserves the specific behavior of this test, notice I added a manual override to the currentSequenceId on line 324, that was to evade this if case https://github.com/amplitude/Amplitude-TypeScript/blob/721a3997673d700b6bda9302b076be0f3fab7c09/packages/plugin-session-replay-browser/src/session-replay.ts#L96-L96, which is unnecessary to test for the set of behavior I'm trying to validate in this test specifically

packages/plugin-session-replay-browser/src/session-replay.ts
kickerbear: Gotcha - do we even end up with a batch Id?

kwalker3690: Nope, there's no batch id, just the sequence id

PR #476, opened 2023-07-10
feat(plugins): Adding configuration options to session replay plugin

kwalker3690:
### Summary

This PR contains three changes:
1. Add default tracking of sessions upon setup of session replay plugin. Its required for the session replay plugin to work to have sessions tracked, this allows us to enforce that requirement
2. Add a configuration option for the sampling rate. This allows users to specify what percent of the sessions they would like to have recorded
3. Abide by the global configuration for opting out of tracking in the session recording - ie don't record if opt out is true

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/README.md
+++ b/packages/plugin-session-replay-browser/README.md
@@ -51,7 +51,7 @@ const sessionReplayTracking = sessionReplayPlugin({
 
 |Name|Type|Default|Description|
 |-|-|-|-|
-|`sampleRate`|`number`|`undefined`|Use this option to control how many sessions will be selected for recording. <br></br>The number should be a decimal between 0 and 1, ie `0.4`, representing the fraction of sessions you would like to have randomly selected for recording. Over a large number of sessions, `0.4` would select `40%` of those sessions.|
+|`sampleRate`|`number`|`undefined`|Use this option to control how many sessions will be selected for recording. A selected session will be recorded, while sessions that are not selected will not be recorded.  <br></br>The number should be a decimal between 0 and 1, ie `0.4`, representing the fraction of sessions you would like to have randomly selected for recording. Over a large number of sessions, `0.4` would select `40%` of those sessions.|
 
 ### 3. Install plugin to Amplitude SDK
 
--- a/packages/plugin-session-replay-browser/README.md
+++ b/packages/plugin-session-replay-browser/README.md
@@ -35,12 +35,26 @@ import * as amplitude from '@amplitude/analytics-browser';
 import { sessionReplayPlugin } from '@amplitude/plugin-session-replay-browser';
 ```
 
-### 2. Instantiate session replay plugin and install plugin to Amplitude SDK
+### 2. Instantiate Session Replay plugin
 
-The plugin must be registered with the amplitude instance via the following code:
+The plugin must be registered with the amplitude instance via the following code. The plugin accepts an optional parameter which is an `Object` to configure the plugin based on your use case.
 
 ```typescript
 amplitude.init(API_KEY);
-const sessionReplayTracking = sessionReplayPlugin();
+const sessionReplayTracking = sessionReplayPlugin({
+  sampleRate: undefined
+});
+```
+
+
+#### Options
+
+|Name|Type|Default|Description|
+|-|-|-|-|
+|`sampleRate`|`number`|`undefined`|Use this option to control how many sessions will be selected for recording. <br></br>The number should be a decimal between 0 and 1, ie `0.4`, representing the fraction of sessions you would like to have randomly selected for recording. Over a large number of sessions, `0.4` would select `40%` of those sessions.|
+
+### 3. Install plugin to Amplitude SDK
+
+```typescript
 amplitude.add(sessionReplayTracking);
 ```
--- a/packages/plugin-session-replay-browser/README.md
+++ b/packages/plugin-session-replay-browser/README.md
@@ -25,6 +25,8 @@ yarn add @amplitude/plugin-session-replay-browser
 
 This plugin works on top of Amplitude Browser SDK and adds session replay features to built-in features. To use this plugin, you need to install `@amplitude/analytics-browser` version `v1.0.0` or later.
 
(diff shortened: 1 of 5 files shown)


packages/plugin-session-replay-browser/README.md
```diff
@@ -35,12 +35,26 @@ import * as amplitude from '@amplitude/analytics-browser';
...
+  sampleRate: undefined
+});
+```
+
+
+#### Options
+
+|Name|Type|Default|Description|
+|-|-|-|-|
+|`sampleRate`|`number`|`undefined`|Use this option to control how many sessions will be selected for recording. <br></br>The number should be a decimal between 0 and 1, ie `0.4`, representing the fraction of sessions you would like to have randomly selected for recording. Over a large number of sessions, `0.4` would select `40%` of those sessions.|
```
kickerbear, line 54: To make this clearer - I would change the first sentence to:
`Use this option to control how many sessions will be selected for recording. A selected session will be recorded, while sessions that are not selected will not be recorded`

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -38,12 +44,36 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
+  }
 
   async setup(config: BrowserConfig) {
     config.loggerProvider.log('Installing @amplitude/plugin-session-replay.');
 
     this.config = config;
+    this.config.sessionId = config.sessionId;
     this.storageKey = `${STORAGE_PREFIX}_${this.config.apiKey.substring(0, 10)}`;
+
+    if (typeof config.defaultTracking === 'boolean') {
```
kickerbear, line 61: Why do we have this handling of type which implies it's handling a type for backwards compatibility

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -108,6 +142,20 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     }
   }
 
+  setShouldRecord(sessionStore?: IDBStoreSession) {
+    if (sessionStore?.shouldRecord === false) {
+      this.shouldRecord = false;
+    } else if (this.options && this.options.sampleRate) {
+      this.shouldRecord = Math.random() < this.options.sampleRate;
```
kickerbear, line 149: Will this result essentially in the _next_ session being selected or unselected for recording?

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -338,10 +386,26 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     }
   }
 
+  async storeShouldRecordForSession(sessionId: number, shouldRecord: boolean) {
```
kickerbear, line 389: Can you share a little context on why we need to store the should record for session? Is it for multiple tabs or for the nexts session load?

kickerbear (approved):
LGTM - a couple of clarifying questions non-blocking

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: defaultTracking can be of either type boolean or an object: https://www.docs.developers.amplitude.com/data/sdks/browser-2/#configuration

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: No, it'll capture that the current session should not be recorded, like this:
1. On plugin initialization, if user has a matching session in the indexdb store, we'll check there for if shouldRecord should be set to false
2. If they don't have a matching session in the store (even if they are in the middle of a session, ie if a customer has just set up the plugin on their site), we'll calculate it based on the sample rate
3. Otherwise we'll check if optOut is configured

We'll do all of the same things when session_start fires, telling us that a new session has begun and we should recalculate whether to record or not.

kwalker3690: I wonder if config.optOut should override any other values actually, I think it should, so I'll move it to be the first check in this if statement

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: Yeah its for consistency across multiple tabs or a page refresh. We don't want to recalculate whether a session should record when a page refreshes, otherwise we might decide to record part of a session only, that would result in some really confusing recordings when part of a session is deemed included in a sample and another part is not. This allows for consistent behavior for an entire session, no matter how many page loads there are

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -38,12 +44,36 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
+          pageViews: false,
+          formInteractions: false,
+          fileDownloads: false,
+          sessions: true,
+        };
+      }
+    } else {
+      config.defaultTracking = {
+        ...config.defaultTracking,
+        sessions: true,
```
jackson-amplitude, line 73: This forces sessions to be tracked when using session replay plugin. I think it's the right behavior, but i wonder if we need to document this. Why do customers want this off?

kwalker3690: Yeah this was a suggestion from Kevin, as the plugin stands now I do think we need to require this as a part of using the plugin. I'll add a note to the README

kwalker3690: I'm not sure why people would want this off, in browser SDK 2.0 they changed it to be on by default, so I wonder if it's an edge case for customers

packages/plugin-session-replay-browser/src/session-replay.ts
jackson-amplitude: it's starting to feel like we have more states on `this` that we need to manage/cleanup when a new session starts.

jackson-amplitude (approved):
LGTM! minor comments. High level thought is it feels like we are starting to have more states on `this`, might be worth it to grouping these states depending on how they need to managed.

PR #488, opened 2023-07-13
fix(plugins): change timing of record upon initialization for session replay plugin

kwalker3690:
### Summary

This PR modifies the timing of initializing the recording of events for the session replay plugin, so that events are always recorded upon plugin initialization.

### Checklist


jackson-amplitude (approved):
LGTM. looks like it wasn't recording before because we were expecting some previously stored events.

PR #497, opened 2023-07-19
feat: ua-parser-plugin

yuhao900914:
### Summary

Moving the ua-parser into the plugin. This makes the customer able to choose if they want to use the server-side device mapping or the original one(@amplitude/ua-parser-js).
 
Reference: https://github.com/amplitude/Amplitude-TypeScript/pull/382#issuecomment-1542623700

This is mainly for avoiding the breaking change for [GTM Browser 2.0 project](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2126053457/GTM+upgrade+SDK+to+Browser+2.0). 

### Checklist


packages/plugin-ua-parser-browser/README.md
```diff
@@ -0,0 +1,6 @@
+<p align="center">
+  <a href="https://amplitude.com" target="_blank" align="center">
+    <img src="https://static.amplitude.com/lightning/46c85bfd91905de8047f1ee65c7c93d6fa9ee6ea/static/media/amplitude-logo-with-text.4fb9e463.svg" width="280">
+  </a>
+  <br />
```
justin-fiedler, line 5: Can you please add a short description of the plugin here and link to the docs calling out the difference in 2.0?

justin-fiedler: Hmmm, seems the UA parser change is not currently mentioned in the 2.0 migration docs. https://www.docs.developers.amplitude.com/data/sdks/browser-2/migration/

packages/plugin-ua-parser-browser/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "test": "jest",
+    "typecheck": "tsc -p ./tsconfig.json"
+  },
+  "bugs": {
+    "url": "https://github.com/amplitude/Amplitude-TypeScript/issues"
+  },
+  "dependencies": {
+    "@amplitude/analytics-client-common": "^2.0.4",
+    "@amplitude/analytics-types": "^2.1.1",
+    "tslib": "^2.4.1"
```
justin-fiedler, line 41: Shouldn't `@amplitude/ua-parser-js` be added here?

packages/plugin-ua-parser-browser/test/ua-parser.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    flushIntervalMillis: 0,
+    flushMaxRetries: 0,
+    flushQueueSize: 0,
+    logLevel: LogLevel.None,
+    loggerProvider: new Logger(), //        log: jest.fn()
+    optOut: false,
+    serverUrl: undefined,
+    transportProvider: new FetchTransport(),
+    useBatch: false,
+
```
justin-fiedler, line 21: extra space

packages/plugin-ua-parser-browser/CHANGELOG.md
kevinpagtakhan, line 1: we can delete this and let the publish workflow create this instead

packages/plugin-ua-parser-browser/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@amplitude/plugin-ua-parser-browser",
+  "version": "1.0.0-beta.0",
```
kevinpagtakhan, line 3: 0.0.0

packages/plugin-ua-parser-browser/rollup.config.js
```diff
@@ -0,0 +1,6 @@
+import { iife, umd } from '../../scripts/build/rollup.config';
+
+iife.input = umd.input;
+iife.output.name = 'uaParser';
```
kevinpagtakhan, line 4: might need to say amplitude at least

kevinpagtakhan: and maybe plugin

packages/plugin-ua-parser-browser/src/typings/ua-parser-plugin.ts
```diff
@@ -0,0 +1,13 @@
...
+import { EnrichmentPlugin } from '@amplitude/analytics-types';
+
+export interface Options {
+  osName?: boolean;
+  osVersion?: boolean;
+  deviceManufacturer?: boolean;
+  deviceModel?: boolean;
+}
+
+export interface CreateUaParserPlugin {
```
kevinpagtakhan, line 11:
```suggestion
export interface EnrichUserAgentPlugin {
```

kevinpagtakhan:
```suggestion
export interface UserAgentEnrichmentPlugin {
```

packages/plugin-ua-parser-browser/rollup.config.js
kevinpagtakhan: `amplitudeUserAgentEnrichmentPlugin`?

packages/plugin-ua-parser-browser/src/ua-parser-plugin.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CreateUaParserPlugin, Options } from './typings/ua-parser-plugin';
+import UAParser from '@amplitude/ua-parser-js';
+
+export const uaParserPlugin: CreateUaParserPlugin = function (
+  options: Options = {
+    osName: true,
+    osVersion: true,
+    deviceManufacturer: true,
+    deviceModel: true,
+  },
```
kevinpagtakhan, line 11:
```suggestion
  options: Options = {},
```

the default values for each key should be taken care of by the next line

packages/plugin-ua-parser-browser/src/ua-parser-plugin.ts
```diff
@@ -0,0 +1,50 @@
...
+) {
+  const { osName = true, osVersion = true, deviceManufacturer = true, deviceModel = true } = options;
+
+  let uaResult: UAParser.IResult;
+
+  const plugin: EnrichmentPlugin = {
+    name: '@amplitude/plugin-ua-parser',
```
kevinpagtakhan, line 18:
```suggestion
    name: '@amplitude/plugin-user-agent-enrichment-browser',
```

packages/plugin-ua-parser-browser/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@amplitude/plugin-ua-parser-browser",
```
kevinpagtakhan, line 2:
```suggestion
  "name": "@amplitude/plugin-user-agent-enrichment-browser",
```

i suggest that we change the name to something referring to an user agent enrichment than user agent parser. the parsing is internal implementation and and the value this plugin provide is the enrichment portion and not the parser portion. i'd also avoid using abbreviations like ua since it makes our plugin less discoverable

kevinpagtakhan:
@justin-fiedler which branch should we merge this pr to?

kevinpagtakhan:
@yuhao900914 how do we plan consume this package in the gtm template?

justin-fiedler:
> which branch should we merge this pr to?

@kevinpagtakhan Hmm, I feel it will only be used by 2.0 so `main` could make sense. wdyt?

yuhao900914:
@kevinpagtakhan  I plan to merge it to 2.0. So in the GTM template, if we wrap on Browser 2.0. We can provide options for the customer to use server device mapping or client ua-parser, maybe default to client ua-parser. 
How to consume this plugin: If the customers choose to use the client ua-parser-js. They can enable it through the GTM template on the init tag. Then we will add this plugin before the init call.

packages/plugin-ua-parser-browser/CHANGELOG.md
yuhao900914: Thanks for pointing this out. Don't know the whole CHANGELOG file is init by the publish workflow. I will delete it.

kevinpagtakhan:
> > which branch should we merge this pr to?
> 
> @kevinpagtakhan Hmm, I feel it will only be used by 2.0 so `main` could make sense. wdyt?

i feel that it become subjective where the plugin should be merged to. the relationship of a package with others evolve overtime and so its relationship with other. for example we could extend this plugin to take a custom user agent parser which could be useful for 1.0. i still feel that the major version determines where it gets merged and make the criteria clear and less dependent on the interpretation of the author

kevinpagtakhan:
> @kevinpagtakhan I plan to merge it to 2.0. So in the GTM template, if we wrap on Browser 2.0. We can provide options for the customer to use server device mapping or client ua-parser, maybe default to client ua-parser. How to consume this plugin: If the customers choose to use the client ua-parser-js. They can enable it through the GTM template on the init tag. Then we will add this plugin before the init call.

i meant to ask how are we going to load this plugin to the template?

packages/plugin-ua-parser-browser/src/typings/ua-parser-plugin.ts
yuhao900914: use `CreateEnrichUserAgentPlugin` to align the names with other plugins.

packages/plugin-ua-parser-browser/src/ua-parser-plugin.ts
```diff
@@ -0,0 +1,50 @@
+import { BrowserConfig, EnrichmentPlugin, Event } from '@amplitude/analytics-types';
+import { CreateUaParserPlugin, Options } from './typings/ua-parser-plugin';
+import UAParser from '@amplitude/ua-parser-js';
+
+export const uaParserPlugin: CreateUaParserPlugin = function (
```
kevinpagtakhan, line 5:
```suggestion
export const userAgentEnrichmentPlugin: CreateUaParserPlugin = function (
```

packages/plugin-ua-parser-browser/src/typings/ua-parser-plugin.ts
kevinpagtakhan: `CreateUserAgentEnrichmentPlugin`?

yuhao900914:
> i meant to ask how are we going to load this plugin to the template?

I feel this plugin can be installed in the wrapper, and expose to the template. Then we can add this plugin in if it's needed.
Wdyt?

yuhao900914:
> i feel that it become subjective where the plugin should be merged to. the relationship of a package with others evolve overtime and so its relationship with other. for example we could extend this plugin to take a custom user agent parser which could be useful for 1.0. i still feel that the major version determines where it gets merged and make the criteria clear and less dependent on the interpretation of the author

- What's the plan for the 1.0. Are we going to deprecate 1.0 finally?
- This plugin is the patch for 2.0 if the customer still wants to use ua-parser-js. The current logic is identical with ua-parser-js in 1.x.  It's not allowed to use other custom user agent parser. Also, it's not auto-installed in any of the SDKs. I don't think it will increase the dependency on Browser 2.0.

packages/plugin-user-agent-enrichment-browser/test/user-agent.test.ts
```diff
@@ -0,0 +1,120 @@
...
+describe('uaParserPlugin', () => {
+  let event: BaseEvent;
+
+  const mockConfig: BrowserConfig = {
+    apiKey: UUID(),
+    flushIntervalMillis: 0,
+    flushMaxRetries: 0,
+    flushQueueSize: 0,
+    logLevel: LogLevel.None,
+    loggerProvider: new Logger(), //        log: jest.fn()
```
Mercy811, line 16: is this comment intentional?

packages/plugin-page-view-tracking-browser/src/typings/ua-parser.d.ts
Mercy811, line 1: Curious why do we declare it in page view packing plugin?

packages/plugin-user-agent-enrichment-browser/src/typings/user-agent-enrichment-plugin.ts
```diff
@@ -0,0 +1,13 @@
...
+import { EnrichmentPlugin } from '@amplitude/analytics-types';
+
+export interface Options {
+  osName?: boolean;
+  osVersion?: boolean;
+  deviceManufacturer?: boolean;
+  deviceModel?: boolean;
+}
+
+export interface CreateUserAgentEnrichmentPlugin {
```
Mercy811, line 11: nits: can we use `UserAgentEnrichmentPlugin` to align with the naming convention of this repo?

Mercy811 (approved):
Thank you Alyssa for building it 🙏 See nits

packages/plugin-page-view-tracking-browser/src/typings/ua-parser.d.ts
yuhao900914: nits. Removed.

packages/plugin-user-agent-enrichment-browser/test/user-agent.test.ts
yuhao900914: Removed.

packages/plugin-user-agent-enrichment-browser/src/typings/user-agent-enrichment-plugin.ts
yuhao900914: The interface is starting with `Create` in other packages.
Like `CreatePageViewTrackingPlugin`, `CreateWebAttributionPlugin`, etc.

PR #501, opened 2023-07-20
fix(plugins): Solve timing and duplicated batch issues with session replay plugin

kwalker3690:
### Summary
This PR reworks the initialization logic as well as the IndexDB storage format to better support scenarios where users move between multiple tabs, and to provide resiliency against duplication of batch sequences. The core changes are:
- the initialize function, which in the changes in this PR will be run on both full page load as well as tab refocus
- the format of the IndexDB data, which gives us better insight into the status of the current sequence for a session.

See the following diagram for an explanation of the flow of data of the plugin after these changes:
Session Replay Flow V1

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -1,7 +1,8 @@
+import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { AMPLITUDE_PREFIX, BaseTransport } from '@amplitude/analytics-core';
 import { BrowserConfig, Event, Status } from '@amplitude/analytics-types';
 import * as IDBKeyVal from 'idb-keyval';
-import { pack, record } from 'rrweb';
+import { record } from 'rrweb';
 import { DEFAULT_SESSION_END_EVENT, DEFAULT_SESSION_REPLAY_PROPERTY, DEFAULT_SESSION_START_EVENT } from './constants';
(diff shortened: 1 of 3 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -61,48 +74,65 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
-      void this.storeEventsForSession([], this.currentSequenceId);
     }
-    this.recordEvents();
   }
 
   recordEvents() {
     this.stopRecordingEvents = record({
       emit: (event) => {
         const eventString = JSON.stringify(event);
-        // Send the first two recorded events immediately
```
kwalker3690, line 95: I decided to get rid of this immediate send since we are batching events on the backend every 90 seconds, there is no point to immediately sending the first two events

jackson-amplitude: I think the other case for this was to avoid having dangling references in the regular events. I.e. the user opens browser, triggers some amplitude events then leaves, that would lead to missing recording error.

jackson-amplitude (approved):

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: Hmm that's true, I have the minimum interval set to 1 second right now, I could just reduce it even further? This specific logic was causing some weird edge cases so instead of fixing those I thought I'd just simplify the logic.

Issue #492, opened 2023-07-19
Disabled ipAddress tracking option doesn't reflect on IP address of a user on amplitude.com


alok1111:
## Expected Behavior
IP address of a user doesn't appear in `User Look-Up > some_user > Event Stream > some_event > Other > IP Address`

## Current Behavior
IP address appears

## Possible Solution
-

## Steps to Reproduce
1. Initialize SDK with `ipAddress` tracking option disabled
2. Execute `track` method add see ip address on the amplitude.com

Sample code:
```
        init(apiKey, userId, {
            logLevel: Types.LogLevel.None,
            defaultTracking: {
                attribution: false,
                pageViews: false,
                sessions: true,
                formInteractions: false,
                fileDownloads: false
            },
            trackingOptions: {
                ipAddress: false,
                language: true,
                platform: true
            }
        })

        track(actionName, {
            elementName
        })
```

## Environment
- JS SDK Version: 2.1.2
- Installation Method: Yarn
- Browser and Version: Firefox 115

Mercy811:
Hi @alok1111, have you ever sent IP before disabling it? If so, Amplitude's backend uses history IP data. Could you log the payload of events sent to see if IP and location are include (or share your project link so I can check raw requests)? If not, it means the SDK works as expected. You can:

- simply test with another user id
- create another project and send events to the new project
- clear previous properties by 
```
amplitude.identify(
  new amplitude.Identify().identify.clearAll()
)
```

CoryMortimer:
@Mercy811 I believe I am running into this issue as well. I was looking through the source and I found [this section of code](https://github.com/amplitude/Amplitude-TypeScript/blob/c6355eff3f8e366bee43b3af5f0d2c57d90941d5/packages/analytics-browser/src/config.ts#L215-L220) that may be the culprit. It appears that any option passed to `options.trackingOptions` will be forced to `true`.

I do not know the code base that well, so in case I'm wrong, I ran your testing steps as well.

I am using ampli instead of directly interfacing with the amplitude SDK. Here are my steps:
```typescript
ampli.load({
  client: {
    configuration: {
      identityStorage: 'none',
      serverUrl: '/custom-endpoint',
      trackingOptions: {
        ipAddress: false
      }
    }
  }
});
```
I then triggered an event.
At `/custom-endpoint`, I added a log statement to see what the body contained. Here's a small snippet:
```typescript
{
  "api_key": "...",
  "events": [
    {
      ...
[...]

PR #505, opened 2023-07-26
fix: get trackingOptions from config

Mercy811:
### Summary

Fix https://github.com/amplitude/Amplitude-TypeScript/issues/492 where ipAddress is set to be false but is still tracked in the event payload.

The original code uses a spread operator (...) to spread out the properties of the options.trackingOptions object into the new trackingOptions object being created. It creates a shallow copy of the options.trackingOptions object. And then ipAddress, language, and platform properties are added to the trackingOptions object setting true which overrides the value in options.trackingOptions

To fix it, check if each property is undefined. If so, default it to be true. Otherwise use the property value of trackingOptions.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -213,10 +213,9 @@ export const useBrowserConfig = async (
   amplitudeInstance.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
 
   const trackingOptions = {
-    ...options.trackingOptions,
-    ipAddress: true,
-    language: true,
-    platform: true,
+    ipAddress: options.trackingOptions?.ipAddress === undefined ? true : options.trackingOptions.ipAddress,
+    language: options.trackingOptions?.language === undefined ? true : options.trackingOptions.language,
+    platform: options.trackingOptions?.platform === undefined ? true : options.trackingOptions.platform,
   };
 
   return new BrowserConfig(
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -213,9 +213,10 @@ export const useBrowserConfig = async (
   amplitudeInstance.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
 
   const trackingOptions = {
-    ipAddress: options.trackingOptions?.ipAddress === undefined ? true : options.trackingOptions.ipAddress,
-    language: options.trackingOptions?.language === undefined ? true : options.trackingOptions.language,
-    platform: options.trackingOptions?.platform === undefined ? true : options.trackingOptions.platform,
+    ipAddress: true,
+    language: true,
+    platform: true,
+    ...options.trackingOptions,
   };
 
   return new BrowserConfig(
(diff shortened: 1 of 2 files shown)


Mercy811:
Don't merge it. Adding a test.

packages/analytics-browser/src/config.ts
```diff
@@ -213,10 +213,9 @@ export const useBrowserConfig = async (
   amplitudeInstance.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
 
   const trackingOptions = {
-    ...options.trackingOptions,
-    ipAddress: true,
-    language: true,
-    platform: true,
+    ipAddress: options.trackingOptions?.ipAddress === undefined ? true : options.trackingOptions.ipAddress,
```
justin-fiedler, line 216: nit: I think `??` could be useful here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
```suggestion
    ipAddress: options.trackingOptions?.ipAddress ?? true,
```

packages/analytics-browser/src/config.ts
```diff
@@ -213,10 +213,9 @@ export const useBrowserConfig = async (
...
+    language: options.trackingOptions?.language === undefined ? true : options.trackingOptions.language,
+    platform: options.trackingOptions?.platform === undefined ? true : options.trackingOptions.platform,
```
kevinpagtakhan, line 218: would this be enough?

```suggestion
    ipAddress: true,
    language: true,
    platform: true,
    ...options.trackingOptions,
```

packages/analytics-browser/src/config.ts
Mercy811: Take the solution Kevin provided to reverse the order.

packages/analytics-browser/test/config.test.ts
```diff
@@ -241,6 +241,23 @@ describe('config', () => {
...
+            language: false,
+            platform: false,
+          },
+        },
+        new AmplitudeBrowser(),
+      );
+      expect(config.trackingOptions.ipAddress).toEqual(false);
+      expect(config.trackingOptions.language).toEqual(false);
+      expect(config.trackingOptions.platform).toEqual(false);
+    });
```
kevinpagtakhan, line 260: we can also test the possible values

```suggestion
    test.each([
      [true, true]
      [undefined, true]
      [false, false]
    ])('should use trackingOptions', async (input, expected) => {
      const config = await Config.useBrowserConfig(
        apiKey,
        {
          trackingOptions: {
            ipAddress: input,
            language: input,
            platform: input,
          },
        },
        new AmplitudeBrowser(),
      );
      expect(config.trackingOptions.ipAddress).toEqual(expected);
      expect(config.trackingOptions.language).toEqual(expected);
      expect(config.trackingOptions.platform).toEqual(expected);
    });
```

kevinpagtakhan (approved):

Mercy811:
Fixed https://github.com/amplitude/Amplitude-TypeScript/pull/508/

PR #506, opened 2023-07-26
fix(plugins): stop recording when document.hasFocus is false

kwalker3690:
### Summary

This PR adds logic to stop recording for session replay when document.hasFocus is false. I have noticed that there is some flakiness and timing issues in the window.blur event so the stop recording functionality does not always occur at the right time. This should add extra protection against recording when the tab is not active.

I also updated the max persisted event size to be 10 MB due to changes in the backend.

### Checklist


jackson-amplitude (approved):
Nice! Having that in the send events path feels like a more robust validation!

Issue #179, opened 2022-08-10
[bug] `Version` and `location_lat` and `location_lng` not being tracked


BrodaNoel:
After migrating to this new library, I'm seeing events without `Version`, which was automatically tracked before (`expo-analytics-amplitude` who was a wrapper of `Amplitude-Android` and `Amplitude-iOS`).

(I'm also seeing `null` the `location_lat` and `location_lng`, which was automatically tracked before as well)

As you can see, there is a `None` now:

<img width="1161" alt="Screen Shot 2022-08-10 at 15 09 05" src="https://user-images.githubusercontent.com/1837969/183985606-7be3bef3-814e-489e-9953-3945d347f6ef.png">

<img width="1132" alt="Screen Shot 2022-08-10 at 15 09 23" src="https://user-images.githubusercontent.com/1837969/183985651-5af68fdc-02c3-40f9-9135-c7417fa34c78.png">

Raw:
```
{
  "$insert_id": "f82b7e89-3e36-4c0b-b508-00c21d09ee74",
  "$insert_key": "010cf28863bc2a1de01d0666c5da3fe85f#208",
  "$schema": 13,
  "_time": 1659654506324,
  "adid": null,
  "amplitude_attribution_ids": null,
  "amplitude_event_type": null,
  "amplitude_id": 441735958251,
  "app": 227935,
  "city": null,
  "client_event_time": "2022-08-04 23:08:26.324000",
  "client_upload_time": "2022-08-04 23:08:29.271000",
  "country": "United States",
  "data": {
    "group_first_event": {
    },
    "group_ids": {
    }
  },
  "data_type": "event",
  "device_brand": null,
  "device_carrier": "Cell-c",
  "device_family": "Google Nexus Phone",
  "device_id": "ae40a07e-8a16-4a0d-96c4-beb80f4960a2",
  "device_manufacturer": "LGE",
  "device_model": "Nexus 5X",
  "device_type": "Google Nexus 5X",
  "display_name": "Go to Explore",
  "dma": null,
  "event_id": 0,
  "event_properties": {
  },
  "event_time": "2022-08-04 23:08:26.324000",
  "event_type": "PAGE_EXPLORE",
  "global_user_properties": {
  },
  "group_properties": {
  },
  "groups": {
  },
  "idfa": null,
  "ip_address": "xxx",
  "is_attribution_event": false,
  "language": "English",
  "library": "amplitude-react-native-ts/0.2.1",
  "location_lat": null,
[...]

PR #504, opened 2023-07-25
fix: fix app_version tracking

falconandy:
### Summary

Fixed `app_version` tracking.

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:  No

PR #507, opened 2023-07-26
fix: react-native fix context plugin to include app version on mobile

bgiori:
### Summary

Fix the context plugin to include app version from native mobile.

related to #179 

### Checklist

* Does your PR have a breaking change?:   NO

### DIFF
--- a/packages/analytics-react-native/src/plugins/context.ts
+++ b/packages/analytics-react-native/src/plugins/context.ts
@@ -81,7 +81,7 @@ export class Context implements BeforePlugin {
       device_id: this.config.deviceId,
       session_id: this.config.sessionId,
       time,
-      ...(this.config.appVersion && { app_version: appVersion }),
+      ...(appVersion && { app_version: appVersion }),
       ...(this.config.trackingOptions.platform && { platform: platform }),
       ...(this.config.trackingOptions.osName && { os_name: osName }),
       ...(this.config.trackingOptions.osVersion && { os_version: osVersion }),
--- a/packages/analytics-react-native/test/plugins/context.test.ts
+++ b/packages/analytics-react-native/test/plugins/context.test.ts
@@ -146,13 +146,6 @@ describe('context', () => {
       const firstContextEvent = await context.execute(event);
 
       expect(firstContextEvent.app_version).toEqual(isWeb() ? undefined : '1.0.0');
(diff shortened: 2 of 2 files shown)


packages/analytics-react-native/src/plugins/context.ts
```diff
@@ -81,7 +81,7 @@ export class Context implements BeforePlugin {
       device_id: this.config.deviceId,
       session_id: this.config.sessionId,
       time,
-      ...(this.config.appVersion && { app_version: appVersion }),
+      ...(appVersion && { app_version: appVersion }),
```
justin-fiedler, line 84: Thanks @bgiori, this change makes sense.

Noticed during review, do you know why the Native Module overrides the `config` values. I would expect any explicit config values to take precedence.

justin-fiedler (approved):

justin-fiedler:
Duplicate [PR](https://github.com/amplitude/Amplitude-TypeScript/pull/504)

packages/analytics-react-native/src/plugins/context.ts
bgiori: Hmm, not sure, though I'd be the one to ask since I wrote it. 

I think that may have been an oversight on my part.

kevinpagtakhan (approved):
🔥

packages/analytics-react-native/src/plugins/context.ts
bgiori: I agree that explicit config should take precedence over system values.

justin-fiedler: 👍 I'll make a new ticket for that fix

bgiori: FWIW analytics-browser does the same, we use uaresult values not the config values

yuhao900914 (approved):

packages/analytics-react-native/src/plugins/context.ts
justin-fiedler: https://amplitude.atlassian.net/browse/AMP-81082

akramarev:
seems related https://amplitude.atlassian.net/browse/AMP-81081, the ticket also asks about Carrier though

PR #508, opened 2023-07-26
fix: should use trackingOptions

Mercy811:
### Summary

Fix a previous commit https://github.com/amplitude/Amplitude-TypeScript/pull/505 
- Add missing commas
- Putting `...options.trackingOptions` at the end will override previous value. Test fails when it is undefined.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -213,10 +213,9 @@ export const useBrowserConfig = async (
   amplitudeInstance.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
 
   const trackingOptions = {
-    ...options.trackingOptions,
-    ipAddress: true,
-    language: true,
-    platform: true,
+    ipAddress: options.trackingOptions?.ipAddress ?? true,
+    language: options.trackingOptions?.ipAddress ?? true,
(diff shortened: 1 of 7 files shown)


kevinpagtakhan (approved):

packages/analytics-browser/src/config.ts
```diff
@@ -213,10 +213,9 @@ export const useBrowserConfig = async (
   amplitudeInstance.previousSessionUserId = previousCookies?.userId ?? legacyCookies.userId;
 
   const trackingOptions = {
-    ipAddress: true,
-    language: true,
-    platform: true,
-    ...options.trackingOptions,
+    ipAddress: options.trackingOptions?.ipAddress ?? true,
+    language: options.trackingOptions?.ipAddress ?? true,
+    platform: options.trackingOptions?.ipAddress ?? true,
```
justin-fiedler, line 218: These don't match

justin-fiedler (changes requested):

packages/analytics-browser/src/config.ts
Mercy811: Just saw it after creating this PR. Fixed in the later commit.

justin-fiedler (approved):

PR #509, opened 2023-07-27
feat(plugins): add masking controls to session replay and reorganize constants and helper fns

kwalker3690:
### Summary

This PR adds support for masking non-input elements, as well as unmasking some input elements. See the README update for further details.

I also split out the constants from the top of the session-replay.ts file into their own file, and added a helper file for smaller functions (like the maskInputFn added here)

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/package.json
+++ b/packages/plugin-session-replay-browser/package.json
@@ -39,7 +39,7 @@
   "dependencies": {
     "@amplitude/analytics-types": ">=1 <3",
     "idb-keyval": "^6.2.1",
-    "rrweb": "^2.0.0-alpha.4",
+    "rrweb": "^2.0.0-alpha.9",
     "tslib": "^2.4.1"
   },
   "devDependencies": {
--- a/packages/plugin-session-replay-browser/src/constants.ts
+++ b/packages/plugin-session-replay-browser/src/constants.ts
@@ -1,5 +1,22 @@
+import { AMPLITUDE_PREFIX } from '@amplitude/analytics-core';
+import { IDBStoreSession } from './typings/session-replay';
+
 export const DEFAULT_EVENT_PROPERTY_PREFIX = '[Amplitude]';
 
 export const DEFAULT_SESSION_REPLAY_PROPERTY = `${DEFAULT_EVENT_PROPERTY_PREFIX} Session Recorded`;
 export const DEFAULT_SESSION_START_EVENT = 'session_start';
 export const DEFAULT_SESSION_END_EVENT = 'session_end';
+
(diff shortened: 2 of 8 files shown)


packages/plugin-session-replay-browser/package.json
```diff
@@ -39,7 +39,7 @@
   "dependencies": {
     "@amplitude/analytics-types": ">=1 <3",
     "idb-keyval": "^6.2.1",
-    "rrweb": "^2.0.0-alpha.4",
+    "rrweb": "^2.0.0-alpha.9",
```
jackson-amplitude, line 42: Alpha worries me a little lol.  Should we be using the stable 1.x version instead? what does datadog use?

packages/plugin-session-replay-browser/src/constants.ts
```diff
@@ -1,5 +1,22 @@
...
+
 export const DEFAULT_EVENT_PROPERTY_PREFIX = '[Amplitude]';
 
 export const DEFAULT_SESSION_REPLAY_PROPERTY = `${DEFAULT_EVENT_PROPERTY_PREFIX} Session Recorded`;
 export const DEFAULT_SESSION_START_EVENT = 'session_start';
 export const DEFAULT_SESSION_END_EVENT = 'session_end';
+
+export const MASK_TEXT_CLASS = 'amp-mask';
+export const UNMASK_TEXT_CLASS = 'amp-unmask';
+export const SESSION_REPLAY_SERVER_URL = 'https://api-secure.amplitude.com/sessions/track';
```
jackson-amplitude, line 12: I think EU has a different url, this should also be config based.

jackson-amplitude (approved):
LGTM! The comments are not directly related to this pr, can be discussed/addressed in another PR.

packages/plugin-session-replay-browser/package.json
kwalker3690: Yeah I was wondering about that too... I'll see what the others are using. All of the docs are referencing the most recent alpha version (I only updated to this because the maskInputFn signature did not match the docs, so it broke and made me realize we weren't using the latest package), so I wonder how much we'd need to change to go to 1.x

packages/plugin-session-replay-browser/src/constants.ts
kwalker3690: It looks like we use a zone option in the config: 
image
and we toggle the server url (if its not overridden completely) based on that: https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/config.ts#L86-L86
I'll add a ticket to use this top level config as well.

PR #512, opened 2023-07-27
fix(plugins): prevent any activity from occurring in session replay plugin if document does not have focus

kwalker3690:
### Summary

This fixes some confusing behavior observed when users open a new tab but do not have it in focus. We only want to record the tab that is in focus at any time.

### Checklist


jackson-amplitude (approved):
LGTM. There are some log statements we may want to clean up, though I am ok with it be there now for debugging purposes.

kwalker3690:
Oh yeah ill remove those before merging

PR #520, opened 2023-08-03
fix(plugins): use hash to calculate is session is included in sample

kwalker3690:
### Summary

This PR reduces our dependence on IndexedDB (and the timing issues related to using it) by using a hash function to deterministically calculate whether a session id is included in the sample or not. @jackson-amplitude and I ran a series of automated tests using 10,000 generated session ids to ensure that this calculation would allow for accurate sampling.

### Checklist


jackson-amplitude (approved):
LGTM!

PR #523, opened 2023-08-07
feat(plugins): add teardown functionality to session replay plugin

kwalker3690:
### Summary
This PR adds a teardown method to the session replay plugin, which is called when either `amplitude.remove()` or `amplitude.init()` is called for a subsequent time. This method removes event listeners from the window, stops recording, and flushes the event queue, sending all remaining events.

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -76,39 +76,46 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
       };
     }
 
-    const GlobalScope = getGlobalScope();
-    if (GlobalScope && GlobalScope.window) {
-      GlobalScope.window.addEventListener('blur', () => {
-        this.stopRecordingEvents && this.stopRecordingEvents();
-        this.stopRecordingEvents = null;
-      });
-      GlobalScope.window.addEventListener('focus', () => {
-        void this.initialize();
-      });
+    const globalScope = getGlobalScope();
+    if (globalScope) {
+      globalScope.addEventListener('blur', this.blurListener);
+      globalScope.addEventListener('focus', this.focusListener);
     }
 
-    if (GlobalScope && GlobalScope.document && GlobalScope.document.hasFocus()) {
+    if (globalScope && globalScope.document && globalScope.document.hasFocus()) {
       await this.initialize(true);
     }
   }
 
+  blurListener = () => {
+    this.stopRecordingAndSendEvents();
+  };
+  focusListener = () => {
+    void this.initialize();
+  };
+
+  stopRecordingAndSendEvents(sessionId?: number) {
+    this.stopRecordingEvents && this.stopRecordingEvents();
+    this.stopRecordingEvents = null;
+    const sessionIdToSend = sessionId || this.config.sessionId;
+    if (this.events.length && sessionIdToSend) {
+      this.sendEventsList({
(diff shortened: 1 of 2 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -76,39 +76,46 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
-    const GlobalScope = getGlobalScope();
-    if (GlobalScope && GlobalScope.window) {
-      GlobalScope.window.addEventListener('blur', () => {
-        this.stopRecordingEvents && this.stopRecordingEvents();
-        this.stopRecordingEvents = null;
-      });
-      GlobalScope.window.addEventListener('focus', () => {
-        void this.initialize();
-      });
+    const globalScope = getGlobalScope();
```
kwalker3690, line 79: This change was based on looking at the code for page view tracking plugin, to make the reference on globalScope look the same

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -76,39 +76,46 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
   }
 
+  blurListener = () => {
+    this.stopRecordingAndSendEvents();
+  };
+  focusListener = () => {
+    void this.initialize();
+  };
+
+  stopRecordingAndSendEvents(sessionId?: number) {
```
kwalker3690, line 97: Consolidating logic here, and actually discovered a bug doing so - any time we call `stopRecordingEvents`, we have to set it to null afterwards to signal that recording can begin again

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -195,9 +202,9 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
     }
     this.stopRecordingEvents = record({
       emit: (event) => {
-        const GlobalScope = getGlobalScope();
-        if (GlobalScope && GlobalScope.document && !GlobalScope.document.hasFocus()) {
-          this.stopRecordingEvents && this.stopRecordingEvents();
```
kwalker3690, line 200: This is where we weren't setting stopRecordingEvents to null

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -76,39 +76,46 @@ class SessionReplay implements SessionReplayEnrichmentPlugin {
...
+    this.stopRecordingAndSendEvents();
+  };
+  focusListener = () => {
+    void this.initialize();
+  };
+
+  stopRecordingAndSendEvents(sessionId?: number) {
+    this.stopRecordingEvents && this.stopRecordingEvents();
+    this.stopRecordingEvents = null;
+    const sessionIdToSend = sessionId || this.config.sessionId;
```
jackson-amplitude, line 100: when would this be different?

jackson-amplitude (approved):
LGTM! Love the organized teardown logic.

kevinpagtakhan (approved):
lgtm!

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: It's a safety check for the end session event, I want to make sure we send events with the session id associated with the end session event, not the next session id that gets created when a start session event fires

PR #503, opened 2023-07-21
feat: user agent plugin

yuhao900914:
### Summary

This pr is the same as [this PR](https://github.com/amplitude/Amplitude-TypeScript/pull/497), but merge to v1.x branch since we made the decision to merge to which branch is based on the version in the package.json in[ this discussion](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2148827164/Process+to+maintain+the+typescript+repo.). 

Where to use this plugin:
This plugin will be used to avoid breaking changes for GTM Browser 2.0 template. 
The plugin will be added to [ the GTM SDK Wrapper ](https://github.com/amplitude/amplitude-js-gtm/blob/v3.x/src/amplitude-wrapper.js#L146-L154) by default.
Still investigating on if this can be configured based on customer input, if not, or if there is no quick solution at this time, we are going to force everyone to use this plugin for the user agent and investigate further.

### Checklist

* [ ] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:

PR #524, opened 2023-08-08
feat: page view v1 enrichment plugin

yuhao900914:
### Summary

Based on[ the comment here](https://github.com/amplitude/amplitude-js-gtm/pull/23#discussion_r1286310001). I think it makes sense to have the plugin released in the typescript SDK.  

This is a pr to create a page view v1 enrichment plugin, which will be used in the gtm sdk wrapper. to make the page view tracking consistent like how it looks like before.

### Checklist


packages/plugin-page-view-v1-enrichment/src/page-view-v1-enrichment.ts
```diff
@@ -0,0 +1,49 @@
...
+            page_domain: page_domain ?? '',
+            page_location: page_location ?? '',
+            page_path: page_path ?? '',
+            page_title: page_title ?? '',
+            page_url: page_url ?? '',
+            /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+          },
+        };
+
+        delete event.event_properties?.['[Amplitude] Page Domain'];
```
yuhao900914, line 38: The current test coverage didn't reach 100% because of those lines. 

Screen Shot 2023-08-08 at 12 39 34 PM

However, I tried to add the cases, and none of it will increase the test coverage. 
- event_properties is missing in the event
- event_properties is undefined in the event
- has event_properties and there is no key for what we are deleting here. 
- expect the delete call won't throw any error. 

Do you have any idea on this?

yuhao900914: cc: @Mercy811 @justin-fiedler

justin-fiedler: Hmm, I have had similar issues before. In my case I added a test like `event_properties is undefined in the event` to get the error to go away.

In this case maybe it is to try to delete a prop off a potentially undefined field? Maybe this could work instead?
```
 if('event_properties' in event) {
  delete event.event_properties.['...']
}
```

yuhao900914: That does work. I spent too much time on this.  I feel that's the jest test issue.
Very much appreciate your recommendation. It saved me.

packages/plugin-page-view-v1-enrichment/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@amplitude/plugin-page-view-v1-enrichment-browser",
+  "version": "0.0.0",
```
justin-fiedler, line 3: Are you going to bump this version later?

justin-fiedler (approved):

packages/plugin-page-view-v1-enrichment/package.json
yuhao900914: Yes, it will be bumped up by the release.

https://github.com/amplitude/Amplitude-TypeScript/pull/503/files#diff-235400fe4397dc5a6273af81b761054e1729653c744418fb046a2e6e5bfcd6de

PR #527, opened 2023-08-09
chore: add publish script to plugins to upload the script into cdn

yuhao900914:
### Checklist

### DIFF
--- a/packages/plugin-page-view-v1-enrichment/package.json
+++ b/packages/plugin-page-view-v1-enrichment/package.json
@@ -29,6 +29,7 @@
     "lint": "yarn lint:eslint & yarn lint:prettier",
     "lint:eslint": "eslint '{src,test}/**/*.ts'",
     "lint:prettier": "prettier --check \"{src,test}/**/*.ts\"",
+    "publish": "node ../../scripts/publish/upload-to-s3.js",
     "test": "jest",
     "typecheck": "tsc -p ./tsconfig.json"
   },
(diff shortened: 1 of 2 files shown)


packages/plugin-page-view-v1-enrichment/package.json
```diff
@@ -29,6 +29,7 @@
     "lint": "yarn lint:eslint & yarn lint:prettier",
     "lint:eslint": "eslint '{src,test}/**/*.ts'",
     "lint:prettier": "prettier --check \"{src,test}/**/*.ts\"",
+    "publish": "node ../../scripts/publish/upload-to-s3.js",
```
justin-fiedler, line 32: I think adding this in each package will cause the `node ../../scripts/publish/upload-to-s3.js` to run 3 times (browser, page-view-plugin, ua-plugin). Will that be handled correctly? If not maybe we should update upload-to-s3.js to accept args for what package to publish and update each package.json to only publish itself.

yuhao900914: @justin-fiedler 
I don't think it works that way. 
In the upload-to-s3 file [here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/scripts/publish/upload-to-s3.js#L7C7-L7C15), it only read the files in the current directory. 

Currently, packages that have been uploaded into the s3 cdn include the publish in the script in each package's package.json. 

Some examples:
https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/plugin-session-replay-browser/package.json#L32
https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/plugin-page-view-tracking-browser/package.json#L32

kevinpagtakhan: yup, thats ok. the script runs with a working directory of the package's root, so the upload job is scoped down

justin-fiedler (approved):

PR #528, opened 2023-08-09
feat: add savedMaxCount

Mercy811:
### Summary

To fix this [P1 issue](https://amplitude.atlassian.net/browse/AMP-82144) where app is crashing because localStorage exceeds its capacity:
- expose `config.savedMaxCount` and defaults it to 1000 like [JS SDK ](https://github.com/amplitude/Amplitude-JavaScript/blob/main/src/amplitude-client.js#L1512-L1525)
- drop events when exceeding `saveMaxCount` in destination plugin

Add this config to core package as planning to limit storage usage on all platforms:
- Browser local storage has capacity of 5MB
- [React Native async storage](https://react-native-async-storage.github.io/async-storage/docs/limits/#:~:text=Total%20storage%20size%20is%20capped,new%20size%20using%20feature%20flag.) has capacity of 6MB
- Node.js SDK defaults to memory storage which has much higher capacity. "by default v8 has a memory limit of 512mb on 32-bit systems, and 1gb on 64-bit systems" But it's still not a good practice to allow unlimited storage usage. 

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -67,7 +67,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     public serverZone: ServerZoneType = DEFAULT_SERVER_ZONE,
     sessionId?: number,
     public sessionTimeout: number = 30 * 60 * 1000,
-    public storageProvider: Storage<Event[]> = new LocalStorage(),
+    public storageProvider: Storage<Event[]> = new LocalStorage({ loggerProvider }),
     public trackingOptions: Required<TrackingOptions> = {
       ipAddress: true,
       language: true,
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -3,6 +3,8 @@ import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
   constructor(private storage?: Storage) {}
 
+  savedMaxCount?: number;
+
   async isEnabled(): Promise<boolean> {
     /* istanbul ignore if */
     if (!this.storage) {
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -1,4 +1,3 @@
-// import { Logger } from '@amplitude/analytics-types';
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -1,9 +1,12 @@
+import { Logger } from '@amplitude/analytics-types';
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
-  constructor(private storage?: Storage) {}
+  logger?: Logger;
 
-  savedMaxCount?: number;
+  constructor(private storage?: Storage, logger?: Logger) {
+    this.logger = logger;
+  }
 
   async isEnabled(): Promise<boolean> {
     /* istanbul ignore if */
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -1,11 +1,11 @@
-import { Logger } from '@amplitude/analytics-types';
+// import { Logger } from '@amplitude/analytics-types';
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
-  logger?: Logger;
+  // logger?: Logger;
 
-  constructor(private storage?: Storage, logger?: Logger) {
-    this.logger = logger;
+  constructor(private storage?: Storage) {
+    // this.logger = logger;
   }
 
   async isEnabled(): Promise<boolean> {
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -2,11 +2,7 @@
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
-  // logger?: Logger;
-
-  constructor(private storage?: Storage) {
-    // this.logger = logger;
-  }
+  constructor(private storage?: Storage) {}
 
   async isEnabled(): Promise<boolean> {
     /* istanbul ignore if */
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -2,7 +2,8 @@ import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
 
 export class LocalStorage<T> extends BrowserStorage<T> {
-  constructor() {
+  constructor(savedMaxCount = 1000) {
     super(getGlobalScope()?.localStorage);
+    this.savedMaxCount = savedMaxCount;
   }
 }
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -1,9 +1,21 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
+import { Logger } from '@amplitude/analytics-types';
 
 export class LocalStorage<T> extends BrowserStorage<T> {
-  constructor(savedMaxCount = 1000) {
-    super(getGlobalScope()?.localStorage);
-    this.savedMaxCount = savedMaxCount;
+  logger?: Logger;
+
+  constructor(logger?: Logger) {
+    super(getGlobalScope()?.localStorage, logger);
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
+      const droppedEvents = value.splice(0, value.length - 1000);
+      const errorMessage = `Dropped ${droppedEvents.length} events because the queue length exceeded 1000.`;
+      this.logger ? this.logger.error(errorMessage) : console.error(errorMessage);
+    }
+
+    await super.set(key, value);
   }
 }
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -17,8 +17,8 @@ export class LocalStorage<T> extends BrowserStorage<T> {
   async set(key: string, value: T): Promise<void> {
     if (Array.isArray(value) && value.length > MAX_ARRAY_LENGTH) {
       const droppedEventsCount = value.length - MAX_ARRAY_LENGTH;
-      await super.set(key, value.slice(droppedEventsCount) as T);
-      this.loggerProvider?.error(`Dropped ${droppedEventsCount} events because the queue length exceeded 1000.`);
+      await super.set(key, value.slice(0, MAX_ARRAY_LENGTH) as T);
+      this.loggerProvider?.error(`Failed to save ${droppedEventsCount} events because the queue length exceeded 1000.`);
     } else {
       await super.set(key, value);
     }
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -2,17 +2,22 @@ import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
 import { Logger } from '@amplitude/analytics-types';
 
+interface LocalStorageOptions {
+  logger?: Logger;
+}
 export class LocalStorage<T> extends BrowserStorage<T> {
   logger?: Logger;
 
-  constructor(logger?: Logger) {
-    super(getGlobalScope()?.localStorage, logger);
+  constructor(config?: LocalStorageOptions) {
+    super(getGlobalScope()?.localStorage);
+    this.logger = config ? config.logger : undefined;
   }
 
   async set(key: string, value: T): Promise<void> {
     if (Array.isArray(value) && value.length > 1000) {
-      const droppedEvents = value.splice(0, value.length - 1000);
-      const errorMessage = `Dropped ${droppedEvents.length} events because the queue length exceeded 1000.`;
+      const droppedEventsCount = value.length - 1000;
+      value.slice(droppedEventsCount);
+      const errorMessage = `Dropped ${droppedEventsCount} events because the queue length exceeded 1000.`;
       this.logger ? this.logger.error(errorMessage) : console.error(errorMessage);
     }
 
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -11,7 +11,7 @@ export class LocalStorage<T> extends BrowserStorage<T> {
 
   constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
-    this.loggerProvider = config ? config.loggerProvider : undefined;
+    this.loggerProvider = config?.loggerProvider;
   }
 
   async set(key: string, value: T): Promise<void> {
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -1,7 +1,8 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
 import { Logger } from '@amplitude/analytics-types';
-import { MAX_ARRAY_LENGTH } from '../constants';
+
+const MAX_ARRAY_LENGTH = 1000;
 
 interface LocalStorageOptions {
   loggerProvider?: Logger;
@@ -18,7 +19,9 @@ export class LocalStorage<T> extends BrowserStorage<T> {
     if (Array.isArray(value) && value.length > MAX_ARRAY_LENGTH) {
       const droppedEventsCount = value.length - MAX_ARRAY_LENGTH;
       await super.set(key, value.slice(0, MAX_ARRAY_LENGTH) as T);
-      this.loggerProvider?.error(`Failed to save ${droppedEventsCount} events because the queue length exceeded ${MAX_ARRAY_LENGTH}.`);
+      this.loggerProvider?.error(
+        `Failed to save ${droppedEventsCount} events because the queue length exceeded ${MAX_ARRAY_LENGTH}.`,
+      );
     } else {
       await super.set(key, value);
     }
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -1,26 +1,26 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
 import { Logger } from '@amplitude/analytics-types';
+import { MAX_ARRAY_LENGTH } from '../constants';
 
 interface LocalStorageOptions {
-  logger?: Logger;
+  loggerProvider?: Logger;
 }
 export class LocalStorage<T> extends BrowserStorage<T> {
-  logger?: Logger;
+  loggerProvider?: Logger;
 
   constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
-    this.logger = config ? config.logger : undefined;
+    this.loggerProvider = config ? config.loggerProvider : undefined;
   }
 
   async set(key: string, value: T): Promise<void> {
-    if (Array.isArray(value) && value.length > 1000) {
-      const droppedEventsCount = value.length - 1000;
-      value.slice(droppedEventsCount);
-      const errorMessage = `Dropped ${droppedEventsCount} events because the queue length exceeded 1000.`;
-      this.logger ? this.logger.error(errorMessage) : console.error(errorMessage);
+    if (Array.isArray(value) && value.length > MAX_ARRAY_LENGTH) {
+      const droppedEventsCount = value.length - MAX_ARRAY_LENGTH;
+      await super.set(key, value.slice(droppedEventsCount) as T);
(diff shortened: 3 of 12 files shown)


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,11 +297,24 @@ export class Destination implements DestinationPlugin {
...
+    if (storedEvents && storedEvents.length + events.length > this.config.savedMaxCount) {
+      const removeCount = storedEvents.length + events.length - this.config.savedMaxCount;
+      deletedEvents = storedEvents.splice(0, removeCount);
+      this.config.loggerProvider.error(
+        `Exceeded maximum saved events count. ${deletedEvents.length} events are dropped.`,
+      );
+    }
+
+    const combinedEvents = [...events, ...storedEvents];
+    await this.config.storageProvider.set(this.storageKey, combinedEvents);
```
kevinpagtakhan, line 318: all the events we plan to persist isare in `events` array already and no need to read from storage. can we just truncate the `events` array to enforce the limit?

```suggestion
void this.config.storageProvider.set(this.storageKey, events.slice(0, this.savedMaxCount);
```

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -59,7 +59,7 @@ export class Destination implements DestinationPlugin {
 
     this.storageKey = `${STORAGE_PREFIX}_${this.config.apiKey.substring(0, 10)}`;
     const unsent = await this.config.storageProvider?.get(this.storageKey);
-    this.saveEvents(); // sets storage to '[]'
+    await this.saveEvents(); // sets storage to '[]'
```
kevinpagtakhan, line 62: all these `await`s have performance impact. please use cautiously

packages/analytics-browser/src/config.ts
```diff
@@ -244,6 +245,7 @@ export const useBrowserConfig = async (
     sessionId,
     options.sessionTimeout,
     options.storageProvider,
+    options.savedMaxCount,
```
kevinpagtakhan, line 248: do we need to update the params in `BrowserConfig`?

packages/analytics-browser/src/config.ts
```diff
@@ -68,6 +68,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     sessionId?: number,
     public sessionTimeout: number = 30 * 60 * 1000,
     public storageProvider: Storage<Event[]> = new LocalStorage(),
+    public savedMaxCount: number = 1000,
```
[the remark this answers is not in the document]
kevinpagtakhan: i think it makes it harder to set a value unless we have a default number that's reasonable and if they want to modify then that's when they'd need to find out how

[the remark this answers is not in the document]
justin-fiedler: I feel we use a default for local storage (e.g. 1000)
```
init()
```

and configure via storageProvider vs extending the client config. e.g.
```
// user config
init({storageProvider: new LocalStorage({ maxEventCount: 100000 })
```

Other storages might have different limitations. wdyt?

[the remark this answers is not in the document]
kevinpagtakhan: local storage can also be used for user session and marketing attribution, which stores plain objects, instead of arrays. so I'm concerned that maxEventCount is specific to a single use case of persisting unsent events

packages/analytics-core/src/config.ts
```diff
@@ -30,6 +30,7 @@ export const getDefaultConfig = () => ({
   serverUrl: AMPLITUDE_SERVER_URL,
   serverZone: 'US' as ServerZoneType,
   useBatch: false,
+  savedMaxCount: 1000,
```
justin-fiedler, line 33: If we keep this at the top-level maybe change it to `storageSize` to align with `storageProvider` and `flushQueueSize`?

packages/analytics-core/src/plugins/destination.ts
Mercy811: Thank you Kevin. It's really helpful to know that.

packages/analytics-core/src/plugins/destination.ts
Mercy811: As all events should be in queue, I deleted `get()` so all await should not be needed.

packages/analytics-browser/src/config.ts
Mercy811: Put `savedMaxCount` in `Storage`. Please see the latest commits.

Mercy811:
Thanks for your previous comments. I did some changes accordingly:
- `saveMaxCount` is put to `Storage` as it is more like a config of `Storage` rather than Amplitude class
- make `saveMaxCount` as optional in `Storage` so that it won't affect existing users with custom storages 
- default `saveMaxCount` to 1000 in the constructor of `LocalStorage` because so far we want a quick fix for local storage. We can add it to other storages similarly if its necessary. 1000 is the default value of legacy JS SDK. And according to the storage size estimation, it should take about 1/5 of browser local storage. 
- leave the drop event logic in destination plugin for easier logging

Please let me know if there is a better solution or any concern. @kevinpagtakhan @justin-fiedler

packages/analytics-core/src/config.ts
Mercy811: Not a naming expert. I used the same name from legacy JS SDK. `storageSize` sounds good to me!

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -302,6 +302,14 @@ export class Destination implements DestinationPlugin {
...
     }
     const events = Array.from(this.queue.map((context) => context.event));
+
+    if (this.config.storageProvider.savedMaxCount && events.length > this.config.storageProvider.savedMaxCount) {
+      const deletedEvents = events.splice(0, events.length - this.config.storageProvider.savedMaxCount);
+      this.config.loggerProvider.error(
+        `Exceeded maximum saved events count. ${deletedEvents.length} events are dropped.`,
+      );
+    }
+
```
kevinpagtakhan, line 312: imo the storage provider should enforce the limits and not its consumers if the config is set in storage provider

packages/analytics-browser/src/storage/browser-storage.ts
```diff
@@ -3,6 +3,8 @@ import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
   constructor(private storage?: Storage) {}
 
+  savedMaxCount?: number;
```
kevinpagtakhan, line 6: how is this useful for user session and marketing attribution?

kevinpagtakhan (changes requested):
i see the potential value in having max count in storage interface, but it makes us responsible to enforce the limits within the implementation; otherwise consumers can/will bypass the limits and defeat its purpose. additionally, if we are to enforce the limtis within the implementation, then it should apply to any data persisted by these storage providers like user session and marketing attribution.

Mercy811:
Thanks for the comment @kevinpagtakhan. Do you have a recommended way to achieve the final goal (limit storage so that it will not crash the local storage and the app using the SDK)? 

What I can think of based on the current implementation and comments is to move limit enforcement logic out of destination plugin to `LocalStorage.set()` and let `set()` returns dropped events if there is any so that we can keep logging in destination plugin.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Should we check in the storage, if the current events length is > savedMaxCount,  while trying to save the upcoming event, instead of doing some modification in the destination plugins?

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,21 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
+import { Logger } from '@amplitude/analytics-types';
 
 export class LocalStorage<T> extends BrowserStorage<T> {
-  constructor() {
-    super(getGlobalScope()?.localStorage);
+  logger?: Logger;
+
+  constructor(logger?: Logger) {
```
kevinpagtakhan, line 8: i think we want the param to be an object with `loggerProvider` as a key. that way we can extend this to be of type `BrowserOptions` (not to be mistaken with `BrowserConfig`) if needed in the future

packages/analytics-browser/src/storage/browser-storage.ts
```diff
@@ -1,7 +1,12 @@
+import { Logger } from '@amplitude/analytics-types';
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
-  constructor(private storage?: Storage) {}
+  logger?: Logger;
+
+  constructor(private storage?: Storage, logger?: Logger) {
```
kevinpagtakhan, line 7: do we need this?

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,21 @@
...
+    super(getGlobalScope()?.localStorage, logger);
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
+      const droppedEvents = value.splice(0, value.length - 1000);
```
kevinpagtakhan, line 14: `splice` mutates the array and could have side effects when the same reference is used elsewhere. would suggest to use [pure functions](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-a-pure-function-d1c076bec976#:~:text=pure%20function%E2%80%9D%20mean%3F-,A%20pure%20function%20is%20a%20function,-which%3A)

this one is good article: https://medium.com/@fknussel/arrays-objects-and-mutations-6b23348b54aa

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: I make an interface called `localStorageOptions` with logger as an property and use this interface as the type of the param of local storage constructor.

packages/analytics-browser/src/storage/browser-storage.ts
Mercy811: I was thinking about extend this to session storage as well. But it's necessary for this fix. Deleted it.

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: Thanks Kevin for letting me know the trap. Use `slice()` instead

packages/analytics-browser/src/storage/browser-storage.ts
```diff
@@ -1,7 +1,12 @@
+// import { Logger } from '@amplitude/analytics-types';
 import { Storage as AmplitudeStorage } from '@amplitude/analytics-types';
 
 export class BrowserStorage<T> implements AmplitudeStorage<T> {
-  constructor(private storage?: Storage) {}
+  // logger?: Logger;
```
justin-fiedler, line 5: remove?

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
     super(getGlobalScope()?.localStorage);
+    this.logger = config ? config.logger : undefined;
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
+      const droppedEventsCount = value.length - 1000;
+      value.slice(droppedEventsCount);
+      const errorMessage = `Dropped ${droppedEventsCount} events because the queue length exceeded 1000.`;
+      this.logger ? this.logger.error(errorMessage) : console.error(errorMessage);
```
justin-fiedler, line 21: We should avoid console.log() to honor logLevel.

```suggestion
      this.logger?.error(errorMessage);
```

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
-  constructor() {
+  logger?: Logger;
+
+  constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
+    this.logger = config ? config.logger : undefined;
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
```
justin-fiedler, line 17: Can you please extract this to a variable to make it easy to update if needed.

```
const MAX_ARRAY_LENGTH = 1000;
```

justin-fiedler (changes requested):

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
+import { Logger } from '@amplitude/analytics-types';
 
+interface LocalStorageOptions {
+  logger?: Logger;
+}
 export class LocalStorage<T> extends BrowserStorage<T> {
-  constructor() {
+  logger?: Logger;
```
kevinpagtakhan, line 9: i think we want the keys to be exactly what config uses?
```suggestion
  loggerProvider?: Logger;
```

packages/analytics-types/src/storage.ts
```diff
@@ -1,10 +1,13 @@
+import { Logger } from './logger';
+
 export interface Storage<T> {
   isEnabled(): Promise<boolean>;
   get(key: string): Promise<T | undefined>;
   getRaw(key: string): Promise<string | undefined>;
   set(key: string, value: T): Promise<void>;
   remove(key: string): Promise<void>;
   reset(): Promise<void>;
+  logger?: Logger;
```
kevinpagtakhan, line 10: are we able to avoid this? would prefer that we do not touch public interfaces

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
+
+  constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
+    this.logger = config ? config.logger : undefined;
```
kevinpagtakhan, line 13:
```suggestion
    this.loggerProvider = config?.loggerProvider;
```

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
+      const droppedEventsCount = value.length - 1000;
+      value.slice(droppedEventsCount);
+      const errorMessage = `Dropped ${droppedEventsCount} events because the queue length exceeded 1000.`;
```
yuhao900914, line 20: Please use the constant instead.

packages/analytics-browser/test/storage/local-storage.test.ts
```diff
@@ -12,4 +13,23 @@ describe('local-storage', () => {
...
+
+  test('should drop events when set more than 1000 events and use custom logger', async () => {
+    const logger = new Logger();
+    const localStorage = new LocalStorage({ logger });
+    const errorMock = jest.spyOn(logger, 'error');
+
+    await localStorage.set('storage-key', new Array(1001).fill(1));
+    expect(errorMock).toHaveBeenCalledTimes(1);
+    expect(errorMock).toHaveBeenCalledWith('Dropped 1 events because the queue length exceeded 1000.');
+  });
```
kevinpagtakhan, line 34:
```suggestion
  test('should drop events when set more than 1000 events and use custom logger', async () => {
    const logger = new Logger();
    const localStorage = new LocalStorage<number[]>({ loggerProvider });
    const errorMock = jest.spyOn(logger, 'error');

    await localStorage.set('storage-key', new Array(1001).fill(1));
    const value = await localStorage.get('storage-key');
    expect(value.length).toBe(1000);
    expect(errorMock).toHaveBeenCalledTimes(1);
    expect(errorMock).toHaveBeenCalledWith('Dropped 1 events because the queue length exceeded 1000.');
  });
```

can we also assert on the actual persisted values?

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
+
+  constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
+    this.logger = config ? config.logger : undefined;
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > 1000) {
+      const droppedEventsCount = value.length - 1000;
+      value.slice(droppedEventsCount);
```
kevinpagtakhan, line 19: i think we're still missing slicing the array and persisting the result of slice

packages/analytics-browser/src/storage/browser-storage.ts
Mercy811: Done.

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: So it will drop without any logs.

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: Extracted it to `constants.ts`

packages/analytics-types/src/storage.ts
Mercy811: Removed it

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: Extracted it to `constants.ts`

packages/analytics-browser/test/storage/local-storage.test.ts
Mercy811: added it

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: Yes I just failed the test and added it back

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
+
+  constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
+    this.loggerProvider = config?.loggerProvider;
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > MAX_ARRAY_LENGTH) {
+      const droppedEventsCount = value.length - MAX_ARRAY_LENGTH;
+      await super.set(key, value.slice(droppedEventsCount) as T);
```
kevinpagtakhan, line 20: Should it be

```suggestion
      await super.set(key, value.slice(0, droppedEventsCount) as T);
```

Mercy811: Should we drop the most early received events like what we did in [JS SDK](https://github.com/amplitude/Amplitude-JavaScript/blob/4cbe557a81ca981d03e140bebed6134c49595a5e/src/amplitude-client.js#L1512-L1525)?

kevinpagtakhan: sorry i mean to  say

```suggestion
      await super.set(key, value.slice(0, 1000) as T);
```

this should keep the top 1000 of the list

packages/analytics-browser/test/storage/local-storage.test.ts
```diff
@@ -12,4 +13,26 @@ describe('local-storage', () => {
...
     expect(await localStorage.isEnabled()).toBe(false);
   });
+
+  test('should drop events when set more than 1000 events without logging', async () => {
+    const localStorage = new LocalStorage<number[]>();
+
+    await localStorage.set('storage-key', new Array<number>(1001).fill(1));
+    const value = await localStorage.get('storage-key');
+
+    expect(value!.length).toBe(1000);
```
kevinpagtakhan, line 23: would prefer to use optional chaining (`?`) vs non-null assertion (`!`) as it acknowledges that the object can be `undefined` as opposed to overruling type rules

```suggestion
    expect(value?.length).toBe(1000);
```

would probably yield to more meaning full error messages too I think

packages/analytics-browser/src/storage/local-storage.ts
Mercy811:
```
await super.set(key, value.slice(droppedEventsCount) as T);
```
Feel like we should keep this and drop oldest events. For example:
[1, 2, 3, 4, 5, 6], with storage size 5, we drop the first element and save [2, 3, 4, 5, 6]

However, this drops the youngest events and keep [1,2,3,4,5] 
```
await super.set(key, value.slice(0, STORAGE_SIZE) as T);
```

packages/analytics-core/src/plugins/destination.ts
Mercy811: Move the logic to local storage rather than destination plugin

packages/analytics-browser/src/storage/local-storage.ts
Mercy811: Reasons to use `slice(0, MAX_ARRAY_LENGTH)`
1. the event queue doesn't guarantee order of events because of retry logic. retried events get pushed at the back of the queue
2. not persisting in storage doesn't drop events, the events are still in memory
3. the first x events is because they are the ones about to be sent out hence have priority

Change the log info too

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
...
+  constructor(config?: LocalStorageOptions) {
     super(getGlobalScope()?.localStorage);
+    this.loggerProvider = config?.loggerProvider;
+  }
+
+  async set(key: string, value: T): Promise<void> {
+    if (Array.isArray(value) && value.length > MAX_ARRAY_LENGTH) {
+      const droppedEventsCount = value.length - MAX_ARRAY_LENGTH;
+      await super.set(key, value.slice(0, MAX_ARRAY_LENGTH) as T);
+      this.loggerProvider?.error(`Failed to save ${droppedEventsCount} events because the queue length exceeded 1000.`);
```
justin-fiedler, line 21:
```suggestion
      this.loggerProvider?.error(`Failed to save ${droppedEventsCount} events because the queue length exceeded ${MAX_ARRAY_LENGTH}.`);
```

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -1,8 +1,26 @@
 import { getGlobalScope } from '@amplitude/analytics-client-common';
 import { BrowserStorage } from './browser-storage';
+import { Logger } from '@amplitude/analytics-types';
+import { MAX_ARRAY_LENGTH } from '../constants';
```
justin-fiedler, line 4: Since it's not configurable, I would probably declare `MAX_ARRAY_LENGTH`in `local-storage.ts`.

Mercy811: I was thinking about it for a sec. Move it to `local-storage.ts` without export.

yuhao900914 (approved):

justin-fiedler (approved):

kevinpagtakhan (approved):

PR #533, opened 2023-08-11
feat: update Android files to support recent React Native versions

falconandy:
### Summary

Updated Android files to match files produced by [create-react-native-library tool](https://www.npmjs.com/package/create-react-native-library). It removes a few deprecation warnings.
It looks like iOS part doesn't contain significant changes.

Tested RN versions: 0.72.3, 0.71.12, 0.70.13

### Checklist

### DIFF
--- a/packages/analytics-react-native/android/build.gradle
+++ b/packages/analytics-react-native/android/build.gradle
@@ -1,59 +1,108 @@
 buildscript {
-    ext.kotlinVersion = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : "1.5.30"
-    repositories {
-        google()
-        jcenter()
-    }
+  // Buildscript is evaluated before everything else so we can't use getExtOrDefault
+  def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["AmplitudeReactNative_kotlinVersion"]
 
-    dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.3'
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-    }
+  repositories {
+    google()
+    mavenCentral()
+  }
+
+  dependencies {
+    classpath "com.android.tools.build:gradle:7.2.1"
+    // noinspection DifferentKotlinGradleVersion
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+  }
 }
 
-apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+def isNewArchitectureEnabled() {
+  return rootProject.hasProperty("newArchEnabled") && rootProject.getProperty("newArchEnabled") == "true"
+}
 
-def safeExtGet(prop, fallback) {
-    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+apply plugin: "com.android.library"
+apply plugin: "kotlin-android"
+
+
+def appProject = rootProject.allprojects.find { it.plugins.hasPlugin('com.android.application') }
+
+if (isNewArchitectureEnabled()) {
+  apply plugin: "com.facebook.react"
+}
+
+def getExtOrDefault(name) {
+  return rootProject.ext.has(name) ? rootProject.ext.get(name) : project.properties["AmplitudeReactNative_" + name]
+}
+
+def getExtOrIntegerDefault(name) {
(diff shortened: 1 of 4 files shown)


packages/analytics-react-native/android/build.gradle
```diff
@@ -1,59 +1,108 @@
...
+def appProject = rootProject.allprojects.find { it.plugins.hasPlugin('com.android.application') }
+
+if (isNewArchitectureEnabled()) {
+  apply plugin: "com.facebook.react"
 }
 
-apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+def getExtOrDefault(name) {
+  return rootProject.ext.has(name) ? rootProject.ext.get(name) : project.properties["AwesomeLibrary_" + name]
```
justin-fiedler, line 32: AwesomeLibrary? Also line 36

packages/analytics-react-native/android/src/main/AndroidManifestNew.xml
```diff
@@ -0,0 +1,3 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
justin-fiedler, line 2: Can we remove this permission? Concerned it will block app approvals.

Maintenance - /android/src/main/AndroidManifest.xml

justin-fiedler (changes requested):

packages/analytics-react-native/android/build.gradle
```diff
@@ -1,59 +1,108 @@
...
+
+def isNewArchitectureEnabled() {
+  return rootProject.hasProperty("newArchEnabled") && rootProject.getProperty("newArchEnabled") == "true"
+}
+
+apply plugin: "com.android.library"
+apply plugin: "kotlin-android"
+
+
+def appProject = rootProject.allprojects.find { it.plugins.hasPlugin('com.android.application') }
```
yuhao900914, line 25: Is this being used anywhere?

packages/analytics-react-native/android/build.gradle
falconandy: Fixed

packages/analytics-react-native/android/src/main/AndroidManifestNew.xml
falconandy: The file is a copy of existing https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-react-native/android/src/main/AndroidManifest.xml
Remove in both files?

falconandy: Amplitude-ReactNative didn't have this permission.

packages/analytics-react-native/android/build.gradle
falconandy: The `build.gradle` content is generated by [create-react-native-library tool](https://www.npmjs.com/package/create-react-native-library). I would not modify it in any way to simplify possible future updates.

packages/analytics-react-native/android/src/main/AndroidManifestNew.xml
yuhao900914: I think we determined to use an example plugins to fetch the location? @justin-fiedler

justin-fiedler: Thanks I updated this ticket with proposed changes. PTAL - https://amplitude.atlassian.net/browse/AMP-62860?focusedCommentId=204509

falconandy: Without the permission there is a lint error in a sample application - let's try to fix the issue in the AMP-62860 ticket:
```
Error: Missing permissions required by LocationManager.getLastKnownLocation: android.permission.ACCESS_COARSE_LOCATION or android.permission.ACCESS_FINE_LOCATION [MissingPermission]
          location = locationManager.getLastKnownLocation(provider!!)
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "MissingPermission":
   This check scans through your code and libraries and looks at the APIs
   being used, and checks this against the set of permissions required to
   access those APIs. If the code using those APIs is called at runtime, then
   the program will crash.

   Furthermore, for permissions that are revocable (with targetSdkVersion 23),
   client code must also be prepared to handle the calls throwing an exception
   if the user rejects the request for permission at runtime.
```

justin-fiedler:
Hi @falconandy I merged the change to remove location access. Can you please rebase. Thanks

falconandy:
@justin-fiedler 
Rebased.

justin-fiedler (approved):

Issue #551, opened 2023-08-21
Cookie options attribute from `init` call mostly not respected


michal-filip:
While attempting to modify cookie options, I noticed that despite setting the corresponding configuration options, the cookies are still generated with same original attributes.

## Expected Behavior
I expected the `sameSite: 'None'` and `secure: true` attributes to be applied to newly created cookies

## Current Behavior
Original attributes `sameSite: 'Lax'` and `secure: false` are still set.

## Possible Solution
It looks like the object spreading on the following line might be done in an incorrect location, keeping most properties fixed and not overridable:
https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/config.ts#L189

## Steps to Reproduce
1. add the following code to be executed in the browser on page load
```
amplitude.init(<apikey>, {
        // ...
        cookieOptions: {
          sameSite: 'None',
          secure: true,
        }
})
```
2. make sure no `AMP_` cookies are set in the browser
3. load the application
4. check the attributes of the auto-created cookies - observe that sameSite is 'Lax' and secure is false

## Environment
- JS SDK Version: 2.2.0
- Installation Method: yarn
- Browser and Version: Chrome 114.0.5735.106

alex-geopro:
Is this still an issue?  I'm initializing the browser sdk like this:

```
amplitude.init('f01c77366af58e34a071913e9f75ab7f', { 
        logLevel: 4,
        cookieOptions: {
          sameSite: 'None',
          secure: true,
        }
      }
```

And the cookie still has sameSite: 'Lax' (after deleting cookies and trying again)

PR #552, opened 2023-08-22
fix(browser): allow cookie options to be customized #548

michal-filip:
### Summary

Fixes https://github.com/amplitude/Amplitude-TypeScript/issues/551

### Checklist

* Does your PR have a breaking change?: Not really

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -186,13 +186,13 @@ export const useBrowserConfig = async (
   // Step 1: Create identity storage instance
   const identityStorage = options.identityStorage || DEFAULT_IDENTITY_STORAGE;
   const cookieOptions = {
-    ...options.cookieOptions,
     domain:
       identityStorage !== DEFAULT_IDENTITY_STORAGE ? '' : options.cookieOptions?.domain ?? (await getTopLevelDomain()),
     expiration: 365,
     sameSite: 'Lax' as const,
     secure: false,
     upgrade: true,
+    ...options.cookieOptions,
   };
   const cookieStorage = createCookieStorage<UserSession>(options.identityStorage, cookieOptions);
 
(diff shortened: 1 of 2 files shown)


yuhao900914 (approved):

kevinpagtakhan (approved):

packages/analytics-browser/src/config.ts
```diff
@@ -186,13 +186,13 @@ export const useBrowserConfig = async (
   // Step 1: Create identity storage instance
   const identityStorage = options.identityStorage || DEFAULT_IDENTITY_STORAGE;
   const cookieOptions = {
-    ...options.cookieOptions,
     domain:
       identityStorage !== DEFAULT_IDENTITY_STORAGE ? '' : options.cookieOptions?.domain ?? (await getTopLevelDomain()),
```
yuhao900914, line 190:
```
      identityStorage !== DEFAULT_IDENTITY_STORAGE ? '' : options.cookieOptions?.domain ?? (await getTopLevelDomain()),
```

This change will also overwrite the domain.
I think we should move the domain after the  `...options.cookieOptions`, it will also capture the domain in the  options.cookieOptions?.domain at the end.

kevinpagtakhan: i think we want to allow `domain` to be modified right?

yuhao900914: yes, 
but [here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/config.ts#L44), we set the domain = '' by default.
The domain value can never be getTopLevelDomain().

kevinpagtakhan: but that only applies when no `cookieOptions` object is passed to `BrowserConfig` which in our case we always pass an object right?

kevinpagtakhan: i added tests to make sure that the values are set correctly: https://github.com/amplitude/Amplitude-TypeScript/pull/564

yuhao900914: I got it now after thinking more. 
if there is no cookiesOption.domain, the domain value will be set through     
``` identityStorage !== DEFAULT_IDENTITY_STORAGE ? '' : options.cookieOptions?.domain ?? (await getTopLevelDomain())```
I think it works fine.

PR #554, opened 2023-08-22
fix: fix web attribution behavior for no referrer in the same session

yuhao900914:
### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -10,6 +10,7 @@ import {
   isFormInteractionTrackingEnabled,
   setConnectorDeviceId,
   setConnectorUserId,
+  isNewSession,
 } from '@amplitude/analytics-client-common';
 import {
   BrowserClient,
@@ -250,14 +251,13 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
   async process(event: Event) {
     const currentTime = Date.now();
-    const lastEventTime = this.config.lastEventTime || Date.now();
-    const timeSinceLastEvent = currentTime - lastEventTime;
+    const _isNewSession = isNewSession(this.config.sessionTimeout, this.config.lastEventTime);
 
     if (
       event.event_type !== DEFAULT_SESSION_START_EVENT &&
       event.event_type !== DEFAULT_SESSION_END_EVENT &&
       (!event.session_id || event.session_id === this.getSessionId()) &&
-      timeSinceLastEvent > this.config.sessionTimeout
+      _isNewSession
     ) {
       this.setSessionId(currentTime);
     }
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -251,13 +251,13 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
   async process(event: Event) {
     const currentTime = Date.now();
-    const _isNewSession = isNewSession(this.config.sessionTimeout, this.config.lastEventTime);
+    const isEventInNewSession = isNewSession(this.config.sessionTimeout, this.config.lastEventTime);
 
     if (
       event.event_type !== DEFAULT_SESSION_START_EVENT &&
       event.event_type !== DEFAULT_SESSION_END_EVENT &&
       (!event.session_id || event.session_id === this.getSessionId()) &&
-      _isNewSession
+      isEventInNewSession
     ) {
       this.setSessionId(currentTime);
     }
--- a/packages/analytics-client-common/src/session.ts
+++ b/packages/analytics-client-common/src/session.ts
@@ -1,7 +1,6 @@
-export const isNewSession = (sessionTimeout: number, lastEventTime?: number): boolean => {
-  const _currentTime = Date.now();
-  const _lastEventTime = lastEventTime || Date.now();
-  const timeSinceLastEvent = _currentTime - _lastEventTime;
+export const isNewSession = (sessionTimeout: number, lastEventTime: number = Date.now()): boolean => {
+  const currentTime = Date.now();
+  const timeSinceLastEvent = currentTime - lastEventTime;
 
   return timeSinceLastEvent > sessionTimeout;
 };
--- a/packages/analytics-client-common/src/session.ts
+++ b/packages/analytics-client-common/src/session.ts
@@ -1,0 +1,7 @@
+export const isNewSession = (sessionTimeout: number, lastEventTime?: number): boolean => {
+  const _currentTime = Date.now();
+  const _lastEventTime = lastEventTime || Date.now();
+  const timeSinceLastEvent = _currentTime - _lastEventTime;
+
+  return timeSinceLastEvent > sessionTimeout;
+};
--- a/packages/analytics-client-common/test/session.test.ts
+++ b/packages/analytics-client-common/test/session.test.ts
@@ -3,7 +3,7 @@ import { isNewSession } from '../src/session';
 describe('session', () => {
   const sessionTimeout: number = 30 * 60 * 1000;
 
-  test('should be a new session for the first event', () => {
+  test('should be in a same session for undefined lastEventTime', () => {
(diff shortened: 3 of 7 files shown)


packages/plugin-web-attribution-browser/src/helpers.ts
```diff
@@ -17,13 +17,25 @@ const domainWithoutSubdomain = (domain: string) => {
...
+  options: Options,
+  isNewSession = false,
+) => {
   const { referrer, referring_domain, ...currentCampaign } = current;
   const { referrer: _previous_referrer, referring_domain: prevReferringDomain, ...previousCampaign } = previous || {};
 
   if (isExcludedReferrer(options.excludeReferrers, current.referring_domain)) {
     return false;
   }
+
```
yuhao900914, line 32: [Here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-web-attribution-browser/src/web-attribution.ts), we will add the domain into the exclude referrers, so direct traffic will return to false, and won't be treat as new campaign.

packages/plugin-web-attribution-browser/src/helpers.ts
```diff
@@ -17,13 +17,25 @@ const domainWithoutSubdomain = (domain: string) => {
...
+  //In the same session, no referrer should not override or unset any persisting query params
+  const isNoReferrer = !referring_domain;
+  if (isNewSession && isNoReferrer) {
```
kevinpagtakhan, line 35:
```suggestion
  if (isNewSession && !referring_domain) {
```

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -26,7 +26,13 @@ export const webAttributionPlugin: CreateWebAttributionPlugin = function (option
         storage.get(storageKey),
       ]);
 
-      if (isNewCampaign(currentCampaign, previousCampaign, pluginConfig)) {
+      const currentTime = Date.now();
+      const lastEventTime = config.lastEventTime || Date.now();
+      const timeSinceLastEvent = currentTime - lastEventTime;
+
+      const isNewSession = timeSinceLastEvent > config.sessionTimeout;
```
kevinpagtakhan, line 33: let's create a shared function that checks for new session between browser SDK and this plugin to make sure they always yield the same results

kevinpagtakhan (changes requested):

packages/plugin-web-attribution-browser/src/web-attribution.ts
yuhao900914: where is the good place to declare the shared function? In analytics-client-common?

kevinpagtakhan: yup i think so

yuhao900914: moved this in analytics-client-common.

packages/analytics-client-common/test/session.test.ts
```diff
@@ -0,0 +1,25 @@
+import { isNewSession } from '../src/session';
+
+describe('session', () => {
+  const sessionTimeout: number = 30 * 60 * 1000;
+
+  test('should be a new session for the first event', () => {
+    const _isNewSession = isNewSession(sessionTimeout, undefined);
+
+    expect(_isNewSession).toBe(false);
+  });
```
Mercy811, line 10: Kind of confusing here that if the first event should init a new session, then we should expect `isNewSession()` to be true instead of false.

packages/analytics-client-common/src/session.ts
```diff
@@ -0,0 +1,7 @@
+export const isNewSession = (sessionTimeout: number, lastEventTime?: number): boolean => {
+  const _currentTime = Date.now();
+  const _lastEventTime = lastEventTime || Date.now();
+  const timeSinceLastEvent = _currentTime - _lastEventTime;
+
+  return timeSinceLastEvent > sessionTimeout;
```
Mercy811, line 6: if it is the first event (timeSinceLastEvent is 0). should we return true?

justin-fiedler (approved):

packages/analytics-client-common/test/session.test.ts
yuhao900914: In the init method [here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L75), we will set the sessionId at the beginning, the first event will compare with the init session.

packages/analytics-client-common/src/session.ts
kevinpagtakhan: while the first event of the session is being instrumented, it would probably be during a state where timeSinceLastEvent > sessionTimeout. so if timeSinceLastEvent === 0 or more generally timeSinceLastEvent < sessionTimeout, a valid session existed by the time this logic executed

packages/analytics-client-common/src/session.ts
```diff
@@ -0,0 +1,7 @@
+export const isNewSession = (sessionTimeout: number, lastEventTime?: number): boolean => {
+  const _currentTime = Date.now();
+  const _lastEventTime = lastEventTime || Date.now();
+  const timeSinceLastEvent = _currentTime - _lastEventTime;
```
kevinpagtakhan, line 4: the duplicated names make it harder to read so suggesting this

```suggestion
export const isNewSession = (sessionTimeout: number, lastEventTime: number = Date.now()): boolean => {
  const currentTime = Date.now();
  const timeSinceLastEvent = currentTime - lastEventTime;
```

packages/plugin-web-attribution-browser/src/web-attribution.ts
```diff
@@ -26,7 +27,9 @@ export const webAttributionPlugin: CreateWebAttributionPlugin = function (option
         storage.get(storageKey),
       ]);
 
-      if (isNewCampaign(currentCampaign, previousCampaign, pluginConfig)) {
+      const _isNewSession = isNewSession(config.sessionTimeout, config.lastEventTime);
```
kevinpagtakhan, line 30: nit: `_isNewSession` bothers me a little since it looks identical to the function at first glance. additionally, there are existing conventions that use the practice of pretending variable names with underscore.

1. Intentionally unused variable ([ref](https://www.sonarsource.com/blog/common-typescript-issues-no-3-unused-local-variables-and-functions/#:~:text=Naming%20a%20variable%20that%20is%20unused%20with%20an%20underscore%2C%20or%20prefixing%20a%20variable%20name%20with%20an%20underscore%2C%20are%20conventions%20to%20show%20other%20developers%20that%20we%20had%20to%20name%20the%20variable%2C%20but%20we%20don%27t%20intend%20to%20use%20it.))
2. Private variables ([ref](https://dev.to/somedood/emulating-private-variables-in-javascript-with-closures-and-factory-functions-2314#:~:text=It%20is%20simply%20the%20convention%20of%20prepending%20an%20underscore%20(_)%20to%20a%20variable%20name.%20This%20is%20done%20to%20indicate%20that%20a%20variable%20is%20private%20and%20should%20not%20be%20toyed%20with.))

yuhao900914: Any recommendation to the variable name?

yuhao900914: change it to `isEventInNewSession`.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -250,14 +251,13 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
   async process(event: Event) {
     const currentTime = Date.now();
-    const lastEventTime = this.config.lastEventTime || Date.now();
-    const timeSinceLastEvent = currentTime - lastEventTime;
+    const _isNewSession = isNewSession(this.config.sessionTimeout, this.config.lastEventTime);
```
kevinpagtakhan, line 254: would be great to fix the _ here also and in the tests

kevinpagtakhan (approved):

PR #555, opened 2023-08-22
feat(session replay): introduce standalone version of session replay package

kwalker3690:
### Summary

This PR introduces a standalone version of the session replay SDK, that is also used by the existing plugin. The interface and functionality of the plugin itself has not changed at all, the functionality has been mostly moved to the standalone SDK. This will allow us to support customers who want to use session replay who aren't instrumented through the browser SDK.

To make reviewing this PR a bit easier, I've made the following gists to help compare functional logic movement, view the diffs of the most recent versions to see the file comparisons:
1. The existing plugin-session-replay-browser/src/session-replay.ts file vs the updated plugin-session-replay-browser/src/session-replay.ts ([gist](https://gist.github.com/kwalker3690/92cd79f06b9288b57ba25ae07e4f0117/revisions))
2. The existing plugin-session-replay-browser/src/session-replay.ts file vs the new session-replay-browser/src/session-replay.ts (ie, the old version of the plugin vs the new standalone sdk) ([gist](https://gist.github.com/kwalker3690/2b90ee274573afc2846c6747764939d2/revisions))
3. The existing plugin-session-replay-browser/test/session-replay.test.ts file vs the updated plugin-session-replay-browser/test/session-replay.test.ts ([gist](https://gist.github.com/kwalker3690/41ee2d0de510c3c3ffe6a4df8ed972fe/revisions))
3. The existing plugin-session-replay-browser/test/session-replay.test.ts file vs the new session-replay-browser/test/session-replay.test.ts ([gist](https://gist.github.com/kwalker3690/f06d3f449206482e5345348a165bb3df/revisions))

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:  no

PR #559, opened 2023-08-23
feat(session replay): introduce standalone version of session replay

kwalker3690:
### Summary

This PR introduces a standalone version of the session replay SDK, that will also used by the existing plugin. There should be no functional changes in separating this to it's own package. This will allow us to support customers who want to use session replay who aren't instrumented through the browser SDK.

Please take a look at this closed PR which includes how the standalone SDK will be used in the plugin: https://github.com/amplitude/Amplitude-TypeScript/pull/555. I will make a revision to incorporate the changes into the plugin once this PR has merged.

### Checklist


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -0,0 +1,530 @@
...
+      return;
+    }
+    this.stopRecordingAndSendEvents(this.config.sessionId);
+    this.config.sessionId = sessionId;
+    this.events = [];
+    this.currentSequenceId = 0;
+    this.recordEvents();
+  }
+
+  getSessionRecordingProperties() {
```
jackson-amplitude, line 90: this interface implies that these properties are static. we probably want to be clear when the user should refetch these properties. I.e., a new session might be created and not sampled, opt-out might be triggered in the middle of a recording.

jackson-amplitude: oh actually, it looks like opting out in the middle of a session isn't supported, which I think is the right behavior.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -0,0 +1,530 @@
...
+  stopRecordingEvents: ReturnType<typeof record> | null = null;
+  maxPersistedEventsSize = MAX_EVENT_LIST_SIZE_IN_BYTES;
+  interval = MIN_INTERVAL;
+  timeAtLastSend: number | null = null;
+
+  constructor() {
+    this.loggerProvider = new Logger();
+  }
+
+  async init(apiKey: string, options: SessionReplayOptions) {
```
jackson-amplitude, line 59: do we need to handle the case when init is called multiple times? That was a way to disable/reenable with browser sdk.

packages/session-replay-browser/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Options
+
+|Name|Type|Required|Default|Description|
+|-|-|-|-|-|
+|`deviceId`|`string`|Yes|`undefined`|Sets an identifier for the device running your application.|
+|`sessionId`|`number`|Yes|`undefined`|Sets an identifier for the users current session. The value must be in milliseconds since epoch (Unix Timestamp).|
+|`sampleRate`|`number`|No|`undefined`|Use this option to control how many sessions will be selected for recording. A selected session will be recorded, while sessions that are not selected will not be recorded.  <br></br>The number should be a decimal between 0 and 1, ie `0.4`, representing the fraction of sessions you would like to have randomly selected for recording. Over a large number of sessions, `0.4` would select `40%` of those sessions.|
+|`optOut`|`boolean`|No|`false`|Sets permission to record sessions. Setting a value of true prevents Amplitude from recording sessions.|
+|`flushMaxRetries`|`number`|No|`5`|Sets the maximum number of retries for failed upload attempts. This is only applicable to retryable errors.|
```
jackson-amplitude, line 80: didn't notice that this was exposed before, but I think it's actually good for performance tuning so retries don't "hog" all the network connections.

packages/session-replay-browser/src/session-replay.ts
jackson-amplitude: oh just saw, there is teardown. should we enforce the need to teardown before re-initializing?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -0,0 +1,530 @@
...
+      globalScope.addEventListener('blur', this.blurListener);
+      globalScope.addEventListener('focus', this.focusListener);
+    }
+
+    if (globalScope && globalScope.document && globalScope.document.hasFocus()) {
+      await this.initialize(true);
+    }
+  }
+
+  setSessionId(sessionId: number) {
```
jackson-amplitude, line 78: hmm we have setSessionId on both config object the SessionReplay object. I wonder if we should make a distinction of these properties, i.e. things in `config` are static and would require teardown/re-init if update is needed. Then the dynamic properties can be exposed only on the SessionReplay object.

jackson-amplitude (approved):
Looks good! Mostly focused on the interface changes. Left comments/questions/thoughts mostly to help me understand some implications. Nothing blocking for the initial test version.

justin-fiedler (commented):
@kwalker3690 I think it would be good to see how this standalone version integrates with the plugin before releasing the standalone package, but ok releasing it if it makes the PRs cleaner.

kwalker3690:
@justin-fiedler yes! please see https://github.com/amplitude/Amplitude-TypeScript/pull/555, also linked in the description

packages/session-replay-browser/src/session-replay.ts
kwalker3690: @jackson-amplitude take a look at the README and let me know if the way I've described it there provides that clarity! Or if you have a suggestion for a better name for this method very happy to hear it also

packages/session-replay-browser/src/session-replay.ts
kwalker3690: No it's actually not necessary, I tested calling init multiple times without teardown, and there are no negative repercussions. Teardown does 2 things, 1) stops recording events and 2) removes event listeners. 

If teardown is not called before init is called a subsequent time, recording has not been stopped, which means that sessionReplay.stopRecordingEvents is defined. We return early [here](https://github.com/amplitude/Amplitude-TypeScript/blob/78ad90a58053d4971bf54231bcf96b42ff90b6b1/packages/session-replay-browser/src/session-replay.ts#L143-L143) and prevent a new `record` call from occurring is sessionReplay.stopRecordingEvents is already defined.

Additionally, there are no additional event listeners added because addEventListener is called with the same reference as it has already been called with, which does not result in a new listener (https://stackoverflow.com/a/26146974).

So teardown is really just a utility that can be called if the customer wants to stop recording altogether for a while, it doesn't need to come before init.

packages/session-replay-browser/src/session-replay.ts
kwalker3690: When you call `setSessionId` it changes the session id on `this.config`, there aren't two representations of `sessionId`. See https://github.com/amplitude/Amplitude-TypeScript/blob/78ad90a58053d4971bf54231bcf96b42ff90b6b1/packages/session-replay-browser/src/session-replay.ts#L84-L84

There are three properties on config that can be updated outside of init, `deviceId`, `optOut`, and `sessionId`. We currently provide only an interface on the standalone SDK for changing `sessionId` mid-recording, but `deviceId` and `optOut` can be changed when used as a plugin, if `browserSDK.setDeviceID` or `browserSDK.setOptOut` are called.   I'm definitely open to adding an interface to the standalone sdk to support changing these properties directly as well, we can make that change in a later update I think if we decide to, since it'd additive.

The mutability of some properties on config and immutability of others matches the behavior of the Browser SDK so I'd like to overall keep that consistent!

packages/session-replay-browser/README.md
```diff
@@ -0,0 +1,95 @@
...
+### 4. Update session id
+Any time that the session id for the user changes, the session replay SDK must be notified of that change. Update the session id via the following method:
+```typescript
+sessionReplay.setSessionId(UNIX_TIMESTAMP)
+```
+
+### 5. Teardown (optional)
+If at any point you would like to discontinue recording, for example in a part of your application where you would not like sessions to be recorded, you can use the following method to stop recording and remove recording event listeners.
+```typescript
+sessionReplay.teardown()
```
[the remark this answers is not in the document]
kwalker3690: I'll rename to shutdown, that's what I see for python

packages/session-replay-browser/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import sessionReplay from './session-replay-factory';
+export const { init, setSessionId, getSessionRecordingProperties, teardown } = sessionReplay;
+export { AmplitudeSessionReplay } from './typings/session-replay';
```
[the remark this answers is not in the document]
kwalker3690: hmm no I'll remove

packages/session-replay-browser/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import sessionReplay from './session-replay-factory';
+export const { init, setSessionId, getSessionRecordingProperties, teardown } = sessionReplay;
```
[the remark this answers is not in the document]
kwalker3690: We haven't discussed whether we'd support more than one instance of session replay (or really what that means for something like session replay), so I figured I'd default to exposing a single instance, and we can add on the exposure of multiple instances when we've vetted how that would work

packages/session-replay-browser/README.md
```diff
@@ -0,0 +1,95 @@
...
+This SDK provides access to the Amplitude Session Replay product.
+
+This plugin requires that default tracking for sessions is enabled. If default tracking for sessions is not enabled in the config, the plugin will automatically enable it.
+
+### 1. Import Amplitude packages
+
+* `@amplitude/session-replay-browser`
+
+```typescript
+import { sessionReplay } from '@amplitude/session-replay-browser';
```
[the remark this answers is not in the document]
kwalker3690: Nope! good catch thanks

packages/session-replay-browser/src/session-replay.ts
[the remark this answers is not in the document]
kwalker3690: Yeah I think that's valid, the main difference in this file is the change from listening for the session start/session end events in [execute](https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/plugin-session-replay-browser/src/session-replay.ts#L115) to 'reset' the recording, now that happens in the `setSessionId` function

kevinpagtakhan (approved):

PR #564, opened 2023-08-24
test: add tests to verify setting cookie domain option works correctly

kevinpagtakhan:
### Summary

Add a test to verify that
1. default cookie domain is set correctly
2. custom cookie domain is set correctly

### Checklist

### DIFF
--- a/packages/analytics-browser/test/config.test.ts
+++ b/packages/analytics-browser/test/config.test.ts
@@ -84,6 +84,7 @@ describe('config', () => {
 
   describe('useBrowserConfig', () => {
     test('should create default config', async () => {
+      const getTopLevelDomain = jest.spyOn(Config, 'getTopLevelDomain').mockResolvedValueOnce('.amplitude.com');
       const logger = new core.Logger();
       logger.enable(LogLevel.Warn);
       const config = await Config.useBrowserConfig(apiKey, undefined, new AmplitudeBrowser());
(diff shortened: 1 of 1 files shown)


packages/analytics-browser/test/config.test.ts
```diff
@@ -159,6 +161,7 @@ describe('config', () => {
           },
           sessionTimeout: 1,
           cookieOptions: {
+            domain: '.amplitude.com',
```
yuhao900914, line 164: Should we give the domain and getTopLevelDomain a different value?

kevinpagtakhan: what do you mean? what gap does it address?

yuhao900914: if getTopLevelDomain and cookieOption have the same value, we are not sure the expected value comes from where.

kevinpagtakhan: the mock result for `getTopLevelDomain` is only valid once and only for the earlier test block. the mock is no longer in effect on this test block. without a valid mock the `getTopLevelDomain` returns an empty string on a test environment.

yuhao900914 (approved):

packages/analytics-browser/test/config.test.ts
kevinpagtakhan: in addition, mocks are also restored after each test block to ensure the mock behavior does not leak onto the next test block as configured here: https://github.com/amplitude/Amplitude-TypeScript/blob/c5c3d62cf505e3df949a4225e3fa3ae2b56d5a0a/jest.config.js#L10

PR #567, opened 2023-08-25
refactor(session replay plugin): use session replay sdk to power plugin

kwalker3690:
### Summary

This PR updates the session replay plugin to use the standalone session replay SDK added in https://github.com/amplitude/Amplitude-TypeScript/pull/559

### Checklist


packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -1,62 +1,36 @@
-/* eslint-disable jest/expect-expect */
-import * as AnalyticsClientCommon from '@amplitude/analytics-client-common';
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
kevinpagtakhan, line 1: what are the unused vars?

packages/plugin-session-replay-browser/package.json
```diff
@@ -41,6 +41,7 @@
     "@amplitude/analytics-client-common": ">=1 <3",
     "@amplitude/analytics-core": ">=1 <3",
     "@amplitude/analytics-types": ">=1 <3",
+    "@amplitude/session-replay-browser": "0.1.0",
```
kevinpagtakhan, line 44: are the dependencies above still accurate?

kevinpagtakhan (approved):

jackson-amplitude (approved):
LGTM! so much red!

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -1,62 +1,35 @@
-/* eslint-disable jest/expect-expect */
-import * as AnalyticsClientCommon from '@amplitude/analytics-client-common';
 import { CookieStorage, FetchTransport } from '@amplitude/analytics-client-common';
```
justin-fiedler, line 1: Will importing `@amplitude/analytics-client-common` create circular dependencies? Should we mock these instead?

justin-fiedler (approved):

packages/plugin-session-replay-browser/test/session-replay.test.ts
kwalker3690: The session replay plugin isn't imported anywhere, definitely not to analytics-client-common, so I don't think this should be an issue! But actually I'll remove the use of this anyways, these values aren't used by the plugin anyways

PR #568, opened 2023-08-25
feat: add diagnostic

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-81179

Send diagnostic events when the SDK drops events

This feature is enabled by default and can be disabled by setting `config.diagnosticProvider` to `undefined`

### Checklist


packages/analytics-browser/src/browser-client.ts
```diff
@@ -86,7 +87,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     // Step 4: Install plugins
     // Do not track any events before this
-    await this.add(new Destination()).promise;
+    await this.add(new Destination(this.config)).promise;
```
kevinpagtakhan, line 90: all plugins receive a config object through setup method, probably not needed

packages/analytics-browser/src/plugins/diagnostic.ts
```diff
@@ -0,0 +1,8 @@
+import { Diagnostic as CoreDiagnostic } from '@amplitude/analytics-core';
+
+export class Diagnostic extends CoreDiagnostic {
+  async flush(): Promise<void> {
+    await fetch(this.serverUrl, this.requestPayloadBuilder(this.queue));
```
kevinpagtakhan, line 5: the pattern is we pass a `transportProvider` instead of creating subclass

packages/analytics-browser/src/plugins/destination.ts
```diff
@@ -0,0 +1,17 @@
+import { Destination as CoreDestination, buildResult } from '@amplitude/analytics-core';
+import { BrowserConfig, DestinationContext as Context } from '@amplitude/analytics-types';
+
+export class Destination extends CoreDestination {
```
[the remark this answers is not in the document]
kevinpagtakhan: why do we need a subclass?

packages/analytics-core/src/plugins/diagnostic.ts
```diff
@@ -0,0 +1,66 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: '*/*',
+      },
+      events: events,
+      method: 'POST',
+    };
+  }
+
+  execute(_context: Event): Promise<Result> {
```
kevinpagtakhan, line 56: should this be a plugin of setup and execute aren't used?

packages/analytics-types/src/diagnostic-event.ts
```diff
@@ -0,0 +1,10 @@
+export interface DiagnosticEvent {
+  time: number;
+  event_properties: {
+    response_error_code: number;
```
yuhao900914, line 4: Should error code and others be enum?

packages/analytics-core/src/plugins/diagnostic.ts
yuhao900914: +1

justin-fiedler: Agreed, the only part of the Plugin lifecycle that we need to hook into is `flush()` but that can be called from the client instead.

packages/analytics-browser/src/plugins/destination.ts
```diff
@@ -0,0 +1,17 @@
...
+  config: BrowserConfig;
+
+  constructor(config: BrowserConfig) {
+    super();
+    this.config = config;
+  }
+
+  async fulfillRequest(list: Context[], code: number, message: string) {
+    await this.config.diagnosticProvider?.track(list.length, code, message);
+    this.saveEvents();
```
justin-fiedler, line 14: This will save events to the `config.storageProvider` which could cause them to be picked up by the AmplitudeDestination plugin. I think we will need a separate storageProvider or maybe just a new storageKey.

packages/analytics-core/src/plugins/diagnostic.ts
```diff
@@ -0,0 +1,66 @@
+import { DiagnosticEvent, DiagnosticPlugin, Event, Result } from '@amplitude/analytics-types';
+
+export class Diagnostic implements DiagnosticPlugin {
+  name = '@amplitude/plugin-diagnostic';
+  type = 'destination' as const;
+  public serverUrl = new URL('http://localhost:8000');
```
justin-fiedler, line 6: For configuration options see the Options pattern in other Plugins - https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-web-attribution-browser/src/typings/web-attribution.ts

packages/analytics-types/src/diagnostic-event.ts
```diff
@@ -0,0 +1,10 @@
+export interface DiagnosticEvent {
```
justin-fiedler, line 1: We might want to make this internal to the DiagnosticProvider. Since it won't be immediately used by customers, it might be confusing compared to "real" events.

justin-fiedler (changes requested):
Thanks @Mercy811. See comments. Since we only need to hook into "flush" configuring this as a Plugin is probably not needed, instead we can mostly treat it like the other providers.

We can probably use the config.transportProvider to send the diagnotics but should be careful with the storage (config.storageProvider) so these events aren't mixed with "real" events.

packages/analytics-browser/src/browser-client.ts
Mercy811: [The setup method](https://github.com/amplitude/Amplitude-TypeScript/blob/ad928eff9556a3456667593e9acae9d47c5bf1b9/packages/analytics-core/src/timeline.ts#L25-L30) in `CoreClient.register()` accepts an `CoreConfig` type without a `DiagnosticProvider`. 

I only add a `DiagnosticProvider` to `BrowserConfig`.

As this is internal, I don't see any harm to pass a config here. Otherwise, have to create a `BrowserTimeline` to overrider the register method in Core.

packages/analytics-browser/src/config.ts
```diff
@@ -76,6 +78,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     public transport: 'fetch' | 'xhr' | 'beacon' = 'fetch',
     public useBatch: boolean = false,
     userId?: string,
+    public diagnosticProvider: DiagnosticPlugin | undefined = new Diagnostic(),
```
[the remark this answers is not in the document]
Mercy811: An error will return "Parameter cannot have question mark and initializer." if adding a question mark

What I want to achieve here is that 
- diagnosticProvider property can either hold a value of type DiagnosticPlugin or it can be undefined
   - users can opt out by passing undefined
   - the default value of this property is a Diagnostic instance

packages/analytics-browser/src/plugins/destination.ts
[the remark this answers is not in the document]
Mercy811: This project only includes the Browser SDK so I want to minimize the impact on other SDKs. That's why I only add the `diagnosticProvider` property to `BrowserConfig` interface rather than `Config` interface in core package. wdyt?

packages/analytics-browser/src/browser-client.ts
kevinpagtakhan: Pretty sure that the browser specific config would be present in the config object passed to setup. We might just need to adjust the type definition of the plugin

packages/analytics-browser/src/config.ts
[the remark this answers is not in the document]
Mercy811: [`new BrowserConfig()`](https://github.com/amplitude/Amplitude-TypeScript/blob/ebd668595b1f5adf1365a9d8106325c159d2119d/packages/analytics-browser/src/config.ts#L223-L224) passes an optional diagnosticProvider. So here must accept undefined as well. Then two options left:
1. `public diagnosticProvider?: DiagnosticPlugin`
2. `public diagnosticProvider: DiagnosticPlugin | undefined`
But I want to assign a default value to this property, so use method 2. 

However, users can only opt out by passing a Diagnostic class with empty track()

packages/analytics-browser/src/plugins/destination.ts
Mercy811: I feel it will not cause unwanted events to be saved. I only add one line to [this function](https://github.com/amplitude/Amplitude-TypeScript/blob/e5c1ebafc300e517e809f42b015a26adebe7d79b/packages/analytics-core/src/plugins/destination.ts#L289-L292). Originally these events should be saved.

As moving diagnosticProvider to core package, I add  `diagnosticProvider.track()` to core destination. Please see the latest change.

packages/analytics-types/src/diagnostic-event.ts
Mercy811: Put it `packages/analytics-core/src/diagnostic.ts` without export.

packages/analytics-core/src/plugins/diagnostic.ts
Mercy811: Thanks for sharing the patter. But as we not don't treat Diagnostic as a plugin, I think it would be better to have `constructor(serverUrl)` instead. Please see the latest change in `packages/analytics-core/src/diagnostic.ts`.

packages/analytics-browser/test/plugins/diagnostic.test.ts
```diff
@@ -0,0 +1,15 @@
+import { Diagnostic } from '../../src/plugins/diagnostic';
+
+describe('Diagnostic', () => {
```
justin-fiedler, line 3: We should also add tests for the client. e.g.
1. Should track diagnostics on 429 error
2. Should track diagnostics on 500 error
3. ... others
4. Should not track diagnostics on 200

packages/analytics-core/src/diagnostic.ts
```diff
@@ -0,0 +1,82 @@
...
+export class Diagnostic implements IDiagnostic {
+  serverUrl: string;
+  queue: DiagnosticEvent[] = [];
+
+  private scheduled: ReturnType<typeof setTimeout> | null = null;
+  // deault delay is 1 minute
+  // make it private to prevent users from changing it to smaller value
+  private delay = 60000;
+
+  constructor(serverUrl?: string) {
```
justin-fiedler, line 24: To make the config easier to maintain (expand) can you please create an DiagnosticsOptions object for config vs constructor args.

packages/analytics-browser/src/plugins/diagnostic.ts
Mercy811: `transportProvider` has `send(serverUrl: string, payload: Payload)` where Payload is designed for normal events and is not compatible with diagnostic events.

packages/analytics-types/src/diagnostic-event.ts
Mercy811: Keep status code as number but make messages as constants.

packages/analytics-browser/src/plugins/diagnostic.ts
Mercy811: After team discussion with Ondrej and Justin, we decided that in this version, we just use fetch method instead of re-using transportProvider

packages/analytics-core/src/diagnostic.ts
```diff
@@ -0,0 +1,82 @@
+import { Diagnostic as IDiagnostic, Event, Result } from '@amplitude/analytics-types';
+import { DIAGNOSTIC_ENDPOINT } from './constants';
+
+interface DiagnosticEvent {
```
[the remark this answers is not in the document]
Mercy811: put it under diagnostics folder

packages/analytics-core/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,74 @@
+import { Diagnostic as IDiagnostic, DiagnosticOptions } from '@amplitude/analytics-types';
+import { DIAGNOSTIC_ENDPOINT } from '../constants';
+import { DiagnosticEvent } from './typings';
+import { DIAGNOSTIC_METADATA_TYPE } from './constants';
+
+export class Diagnostic implements IDiagnostic {
```
justin-fiedler, line 6:
```suggestion
export class BaseDiagnostic implements Diagnostic {
```

packages/analytics-browser/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,8 @@
+import { Diagnostic as CoreDiagnostic } from '@amplitude/analytics-core';
+
+export class Diagnostic extends CoreDiagnostic {
```
justin-fiedler, line 3:
```suggestion
export class BrowserDiagnostic extends BaseDiagnostic {
```

packages/analytics-browser/test/diagnostic.test.ts
```diff
@@ -0,0 +1,15 @@
...
+describe('Diagnostic', () => {
+  test('should fetch', async () => {
+    const diagnostic = new Diagnostic();
+    const fetchMock = jest.fn().mockResolvedValueOnce({} as Response);
+    global.fetch = fetchMock;
+
+    diagnostic.track(5, 200, 'Test message');
+    await diagnostic.flush();
+
+    expect(fetchMock).toHaveBeenCalledTimes(1);
```
justin-fiedler, line 12: Can you also please verify the track payload here?

packages/analytics-browser/test/helpers/mock.ts
```diff
@@ -54,6 +54,7 @@ export const createConfigurationMock = (options?: Partial<BrowserConfig>) => {
       send: jest.fn(),
     },
     useBatch: false,
+    diagnosticProvider: new Diagnostic(),
```
justin-fiedler, line 57: Should this be a mock object like the other providers?

packages/analytics-core/src/config.ts
```diff
@@ -30,6 +33,7 @@ export const getDefaultConfig = () => ({
   serverUrl: AMPLITUDE_SERVER_URL,
   serverZone: 'US' as ServerZoneType,
   useBatch: false,
+  diagnosticProvider: new Diagnostic(),
```
justin-fiedler, line 36: Since this is being added in core, it would be good to have some functional implementation for the `BaseDiagnostic`. Would it make sense to log these to the logger provider in the default (non-browser) case? Or will that duplicate messages?

packages/analytics-core/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,74 @@
...
+      this.scheduled = setTimeout(() => {
+        void this.flush();
+      }, this.delay);
+    }
+  }
+
+  async flush(): Promise<void> {
+    // send http request based on environment
+    // implemented in its child class
+
```
justin-fiedler, line 42: Maybe add a default implemetion here? Log the diagnostics to loggerProvider? See other comment above.

packages/analytics-core/src/diagnostics/typings.ts
```diff
@@ -0,0 +1,14 @@
...
+  metadata_type: string;
+  library: string;
+  accounting_time_min: number;
+  response_code: number;
+  trigger: string;
+  action: string;
+  event_count: number;
+}
+
+export interface DiagnosticEvent {
```
justin-fiedler, line 11: I'm concerned with the naming `DiagnosticEvent` since this is not a real `BaseEvent`.

Can we use a different name e.g. `DiagnosticPayload` or `DiagnosticRequest`?

packages/analytics-browser/src/config.ts
```diff
@@ -75,6 +78,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     },
     public transport: 'fetch' | 'xhr' | 'beacon' = 'fetch',
     public useBatch: boolean = false,
+    public diagnosticProvider: IDiagnostic | DiagnosticOptions = new Diagnostic(),
```
justin-fiedler, line 81: This make the usage elsewhere a little crufty with type casts. Can you change this to only accept a provider (not options) similar to transportProvider and storage. 
e.g.
https://github.com/amplitude/Amplitude-TypeScript/pull/568/files#diff-379abfc2a0d72883f73638e5ee6b8a9e0fb89793f5cebc6751a0716731bfa1d7R99

packages/analytics-core/src/diagnostics/diagnostic.ts
Mercy811: leave it as how it is now but we will see how we can improve it as Node and RN have diagnosticProvdier

packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -261,6 +261,163 @@ describe('browser-client', () => {
...
+      test('should diagnostic track when 429 rate limit when flush', async () => {
+        const transportProvider = {
+          send: jest.fn().mockImplementationOnce(() => {
+            return Promise.resolve({
+              status: Status.RateLimit,
+              statusCode: 429,
+              body: {
+                exceededDailyQuotaUsers: { user_0: 1 },
+                exceededDailyQuotaDevices: {},
+                throttledEvents: [],
```
justin-fiedler, line 348: Just to confirm, does the diagnostic endpoint return the same response payloads as HTTPv2 API?

packages/analytics-core/test/helpers/default.ts
```diff
@@ -1,21 +1,25 @@
...
-    set: async () => undefined,
-    remove: async () => undefined,
-    reset: async () => undefined,
-    getRaw: async () => undefined,
-  },
-  ...getDefaultConfig(),
-});
+export const useDefaultConfig = (): Config => {
+  const defaultConfig = getDefaultConfig();
+  defaultConfig.diagnosticProvider.apiKey = API_KEY;
```
justin-fiedler, line 6: Maybe this should be mocked like the other providers below?

packages/analytics-browser/test/browser-client.test.ts
Mercy811: Is there any concern with it? Afaik, the success response is similar but not the same. 

diagnostic endpoint: 
```
{
    "code": 200,
    "events_ingested": 1
}
```

HTTT V2 API:
```
{
  "code": 200,
  "events_ingested": 50,
  "payload_size_bytes": 50,
  "server_upload_time": 1396381378123
}
```

packages/analytics-core/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,74 @@
+import { Diagnostic as IDiagnostic, DiagnosticOptions } from '@amplitude/analytics-types';
```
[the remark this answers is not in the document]
Mercy811: Renamed as suggested

packages/analytics-browser/test/plugins/diagnostic.test.ts
Mercy811: Tests added in `analytics-browser/test/browser-client.test.ts`

packages/analytics-browser/src/config.ts
Mercy811: verify the payload 

```
{
      method: 'POST',
      headers: {
      },
      body: JSON.stringify({
        api_key: '',
        omni_metrics: [
          {
          metadata_type: DIAGNOSTIC_METADATA_TYPE,
          library: 'amplitude-ts',
          accounting_time_min: 12345,
          response_code: 400,
          trigger: 'test message 0',
          action: 'drop events',
          event_count: 5,
        }
      ],
    })
}
```

packages/analytics-browser/test/diagnostic.test.ts
Mercy811: Verify the payload to be something like 

```
{
  method: 'POST',
  headers: {
  },
  body: JSON.stringify({
    api_key: '',
    omni_metrics: [{
      metadata_type: DIAGNOSTIC_METADATA_TYPE,
      library: 'amplitude-ts',
      accounting_time_min: 12345,
      response_code: 400,
      trigger: 'test message 0',
      action: 'drop events',
      event_count: 5,
    }],
  }),
}
```

packages/analytics-core/src/diagnostics/typings.ts
Mercy811: Don't call it `event` to prevent customer confusion. Ping dx-pod [here](https://amplitude.slack.com/archives/C02E1H87E2C/p1697055738127019). Will use `DiagnosticRequest`.

packages/analytics-core/src/config.ts
Mercy811: After discussion, we decided not to have functional implementation for `BaseDiagnostic` as it is shared by Node and RN but is disabled and haven't planned to implement them yet. Log to logger provider will double logging.

packages/analytics-core/src/diagnostics/typings.ts
Mercy811: Also renamed `diagnosticEventBuilder` to `diagnosticRequestBuilder`

packages/analytics-browser/src/config.ts
Mercy811: Removed DiagnosticOptions in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/3ad286e0a5203570c62c5cd6099e64f3f2a905cf

packages/analytics-core/test/helpers/default.ts
Mercy811: True. Changed in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/860bde26d816642ae744ad8d4d14c214f805d6ae

packages/analytics-browser/test/helpers/mock.ts
Mercy811: Changed in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/860bde26d816642ae744ad8d4d14c214f805d6ae

packages/analytics-browser/test/browser-client.test.ts
justin-fiedler: Sorry, reviewing to quick. This makes sense.

packages/analytics-core/src/index.ts
```diff
@@ -2,6 +2,16 @@ export { AmplitudeCore } from './core-client';
...
 export { Destination } from './plugins/destination';
+export { Diagnostic } from './diagnostics/diagnostic';
+export {
+  EXCEEDED_MAX_RETRY_DIAGNOSTIC_MESSAGE,
+  MISSING_API_KEY_DIAGNOSTIC_MESSAGE,
+  UNEXPECTED_DIAGNOSTIC_MESSAGE,
+  INVALID_OR_MISSING_FIELDS_DIAGNOSTIC_MESSAGE,
+  EVENT_ERROR_DIAGNOSTIC_MESSAGE,
+  PAYLOAD_TOO_LARGE_DIAGNOSTIC_MESSAGE,
+  EXCEEDED_DAILY_QUOTA_DIAGNOSTIC_MESSAGE,
```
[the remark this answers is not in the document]
Mercy811: Done in commit https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/266c52ccfc8d0bcf9355c67395ef25ad3090971a

packages/analytics-core/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,71 @@
+import { Diagnostic, DiagnosticOptions } from '@amplitude/analytics-types';
+import { DIAGNOSTIC_ENDPOINT } from '../diagnostics/constants';
+import { DiagnosticOmniMetrics } from './typings';
+import { DIAGNOSTIC_METADATA_TYPE } from './constants';
+
+export class BaseDiagnostic implements Diagnostic {
+  isDisabled = false;
```
justin-fiedler, line 7: Since this is going to be added automatically in Node and ReactNative can we have the BaseDiagnostic be disabled by default, and then have BrowserDiagnostic enabled by default?

packages/analytics-types/src/diagnostic.ts
```diff
@@ -0,0 +1,9 @@
+export interface DiagnosticOptions {
+  isDisabled?: boolean;
+  serverUrl?: string;
+  apiKey?: string;
+}
+
+export interface Diagnostic extends DiagnosticOptions {
```
justin-fiedler, line 7: I don't think we should have `Diagnostics` extend `DiagnosticsOptions`. Instead I think we can enforce the `DiagnosticsOptions` to be provided in the constructor of `BaseDiagnostic`. This will let each `Diagnostic` implementation to decide it's own necessary options.

packages/analytics-core/src/config.ts
```diff
@@ -30,6 +32,7 @@ export const getDefaultConfig = () => ({
   serverUrl: AMPLITUDE_SERVER_URL,
   serverZone: 'US' as ServerZoneType,
   useBatch: false,
+  diagnosticProvider: new BaseDiagnostic(),
```
justin-fiedler, line 35: Can we disable this for non-Browser usage?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -205,9 +202,10 @@ export class Destination implements DestinationPlugin {
     this.fulfillRequest(list, res.statusCode, SUCCESS_MESSAGE);
   }
 
-  handleInvalidResponse(res: InvalidResponse, list: Context[]) {
+  handleInvalidResponse(res: InvalidResponse, list: Context[], useRetry: boolean) {
     if (res.body.missingField || res.body.error.startsWith(INVALID_API_KEY)) {
-      this.fulfillRequest(list, res.statusCode, res.body.error);
+      this.fulfillRequest(list, res.statusCode, `${res.status}: ${getResponseBodyString(res)}`);
```
justin-fiedler, line 207: Does this handle all of the previous cases?
https://github.com/amplitude/Amplitude-TypeScript/pull/568/files#diff-379abfc2a0d72883f73638e5ee6b8a9e0fb89793f5cebc6751a0716731bfa1d7L158-L165

The original code had a special handling for `res` without `res.body` which I think would not be fulfilled with this new logic.

packages/analytics-core/src/config.ts
```diff
@@ -75,6 +79,14 @@ export class Config implements IConfig {
     this.storageProvider = options.storageProvider;
     this.transportProvider = options.transportProvider;
     this.useBatch = options.useBatch ?? defaultConfig.useBatch;
+
+    if (options.diagnosticProvider == undefined) {
+      this.diagnosticProvider = defaultConfig.diagnosticProvider;
+    } else {
+      this.diagnosticProvider = options.diagnosticProvider;
+    }
+    this.diagnosticProvider.apiKey = this.apiKey;
```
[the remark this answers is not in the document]
justin-fiedler: Or maybe

```
if (!this.diagnosticProvider.apiKey) {
  this.diagnosticProvider.apiKey = this.apiKey;
}
```

packages/analytics-core/src/diagnostics/diagnostic.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private scheduled: ReturnType<typeof setTimeout> | null = null;
+  // deault delay is 1 minute
+  // make it private to prevent users from changing it to smaller value
+  private delay = 60000;
+
+  constructor(options?: DiagnosticOptions) {
+    this.isDisabled = options && options.isDisabled ? options.isDisabled : false;
+    this.serverUrl = options && options.serverUrl ? options.serverUrl : DIAGNOSTIC_ENDPOINT;
+    this.apiKey = options && options.apiKey ? options.apiKey : '';
```
Mercy811, line 20: abstract these properties to be inside an option object

packages/analytics-core/src/config.ts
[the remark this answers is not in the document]
Mercy811: I like the latter one! Fixed in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/899bd6d37c7774a024339e5dcc76fb7f06fb1cce

packages/analytics-core/test/diagnostic.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks(); // Restore mocked functions after each test
+  });
+
+  describe('constructor', () => {
+    test('should set default values if not provided', () => {
+      expect(diagnostic.serverUrl).toBe(DIAGNOSTIC_ENDPOINT);
+      expect(diagnostic.isDisabled).toBe(false);
```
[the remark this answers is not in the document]
Mercy811: fixed in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/899bd6d37c7774a024339e5dcc76fb7f06fb1cce

packages/analytics-types/src/diagnostic.ts
Mercy811: Remove `extends DiagnosticOptions` here. Add api key property to Diagnostic interface as it's used in core config.

Done in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/fb672f43a99a8c7188e6c2979ed1db48bb6e6242

packages/analytics-core/src/diagnostics/diagnostic.ts
Mercy811: Disable BaseDiagnostic in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/80a811483b2725fc81124966fb664c6640f5c37c

packages/analytics-core/src/config.ts
Mercy811: Change the default value of `isDisable` to true in BaseDiagnostic in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/80a811483b2725fc81124966fb664c6640f5c37c. So leave here as it is.

packages/analytics-core/src/diagnostics/diagnostic.ts
Mercy811: save it for now as Diagnostic not extend from options

packages/analytics-core/src/plugins/destination.ts
Mercy811: `filfillRequest()` triggers callback which can be customized by customers to monitor the SDK errors. So have to make sure we don't miss any.

Since we don't have tests for `filfillRequest()`, its hard to manually check them. 

Recover `filfillRequest()` in send when `!useTry` in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/afe973135b0d8a2607d49d5e5a6b3b28cecf28e7. Should not missing any `fulfilllRequest()` not.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -269,6 +297,11 @@ export class Destination implements DestinationPlugin {
       return true;
     });
 
+    const dropEvents = list.filter((element) => !retry.includes(element));
```
justin-fiedler, line 300: Efficiency - the same list is already filtered above which could be heavy for large lists. Also `list.includes()` is linear. `Set.has()` is O(1) and can be better for frequent lookups.

It would probably be more efficient to do it all at once. e.g.
```
const retryList = [];
const dropList = [];
list.forEach(element => {
   if(shouldRetry(element)) {
     ...
     retryList.push();
   } else {
      ...
      dropList.push();
   }
})
```

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -161,53 +164,60 @@ export class Destination implements DestinationPlugin {
         } else {
           this.fulfillRequest(list, res.statusCode, res.status);
         }
-        return;
       }
-      this.handleResponse(res, list);
+      this.handleResponse(res, list, useRetry);
```
justin-fiedler, line 168: Distributing the `useRetry` across the multiple handlers makes the logic a little hard to follow.

From what I can tell `useRetry` is needed to determine if diagnostics should be tracked in each `!useRetry` case. However the diagnostics in these cases all seem based on the full list (not retries or drops) and could be done at the top level instead with a separate method. That way we can revert the other response handlers to only handle `useRetry`.

Update lines 166 - 168
```
        processResponseDiagnostics(res, list);
        return;
      }
      this.handleResponse(res, list);
```

New method
```
function processResponseDiagnostics(res, list) {
  const { status } = res;
  switch (status) {
    case Status.Invalid: {
        if (!res.body.error.startsWith(INVALID_API_KEY)) {
          this.config.diagnosticProvider.track(list.length, 400, DIAGNOSTIC_MESSAGES.INVALID_OR_MISSING_FIELDS);
        } else {
            this.config.diagnosticProvider.track(list.length, 400, DIAGNOSTIC_MESSAGES.EVENT_ERROR);
      }
      break;
    }
    case Status.PayloadTooLarge: {
      this.config.diagnosticProvider.track(list.length, 413, DIAGNOSTIC_MESSAGES.PAYLOAD_TOO_LARGE);
      break;
    }
    case Status.RateLimit: { 
      ...
    }
  }
}
```

justin-fiedler (changes requested):

packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -261,6 +261,190 @@ describe('browser-client', () => {
...
+          send: jest.fn().mockImplementationOnce(() => {
+            return Promise.resolve(res);
+          }),
+        };
+
+        await client.init(apiKey, {
+          defaultTracking: false,
+        }).promise;
+        const diagnosticTrack = jest.spyOn(client.config.diagnosticProvider, 'track');
+        client.config.transportProvider = transportProvider;
```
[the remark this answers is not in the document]
Mercy811: `transportProvider` is omitted. https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-types/src/config/browser.ts#L58-L61

packages/analytics-core/src/plugins/destination.ts
Mercy811: How about keeping the original code and only add `dropList.push()` 

```
const dropList = []

    const retry = list.filter((context, index) => {
      if ( shouldDrop ) {
        this.fulfillRequest([context], res.statusCode, res.body.error);
        dropList.push(context);
        return;
      }
      ...
      return true;
    });
```
It also only iterates once.

packages/analytics-core/src/plugins/destination.ts
Mercy811: Done in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/f52c10d544db3037af15b595763e438dfe835813

packages/analytics-core/src/plugins/destination.ts
justin-fiedler: That sounds good.

Mercy811: Done in https://github.com/amplitude/Amplitude-TypeScript/pull/568/commits/025b4b079a1a2f452918647f99897abff920f476

justin-fiedler (approved):

PR #529, opened 2023-08-10
feat: add bookmarklet

liuyang1520:
### Summary
- feat: add bookmarklet

To use it:
1. In browser bookmark bar, "Add Page"
1. Put any name
1. Copy the code below, update the `YOUR_API_KEY` to your API key
1. Put the code in the URL input box
1. Save
1. Open any web site
1. Click on the bookmark
1. Now Amplitude SDK with more trackings is enabled, click/change any item to track

Limitations:
- Ads blocker might block the script loading
- CSP might block the script loading, depending on the target website security configuration

The bookmarklet will:
- Load the SDK from CDN
- Initialize the SDK with your API key (and user id if updated)
- Initialize the inline [auto-tracking plugin](https://github.com/amplitude/Amplitude-TypeScript/compare/add-bookmarklet?expand=1#diff-6c206852d91cf818395e72ab4930bbffffbb605dc369e24d783b005432c9e0c2R58-R213)
  - tracking `click` events
  - tracking `change` events

Steps to generate the bookmarket:
- `cd packages/analytics-browser`, then `node ../../scripts/version/create-bookmarklet-snippet.js` (this will update the file in `packages/analytics-browser/generated/amplitude-bookmarklet-snippet.js`)
- Go back to the root of the repo, then `yarn build`
- `vim packages/analytics-browser/lib/scripts/amplitude-bookmarklet.html` to update the `YOUR_API_KEY`, `YOUR_USER_ID`, `YOUR_SERVER_ZONE`
- Create a new bookmark and copy the content to it

Final bookmarklet looks like (updated with remote loading auto-tracking plugin):
```js
[...]

PR #570, opened 2023-08-29
feat: add auto-tracking plugin

liuyang1520:
### Summary
Extract the auto-tracking logic from https://github.com/amplitude/Amplitude-TypeScript/pull/529 to a new package.

- feat: add auto-tracking plugin

#### Instantiate the plugin

The plugin accepts 1 optional parameter, which is an `Object` to configure the allowed tracking options.

```typescript
const plugin = autoTrackingPlugin({
  cssSelectorAllowlist: ['.amp-auto-tracking', '[amp-auto-tracking]'],
  tagAllowlist: ['button', 'a'],
});
```

Examples:
- The above `cssSelectorAllowlist` will only allow tracking elements like:
    - `<button amp-auto-tracking>Click</button>`
    - `<a class="amp-auto-tracking">Link</a>`
- The above `tagAllowlist` will only allow `button` and `a` tags to be tracked.

Note `ingestionMetadata` is for internal use only, you don't need to provide it.

#### Options

|Name|Type|Default|Description|
|-|-|-|-|
|`cssSelectorAllowlist`|`string[]`|`undefined`| When provided, only allow elements matching any selector to be tracked. |
|`tagAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label']`| Only allow elements with tag in this list to be tracked. |

### Checklist

### DIFF
--- a/packages/plugin-auto-tracking-browser/README.md
+++ b/packages/plugin-auto-tracking-browser/README.md
@@ -1,0 +1,76 @@
+<p align="center">
+  <a href="https://amplitude.com" target="_blank" align="center">
+    <img src="https://static.amplitude.com/lightning/46c85bfd91905de8047f1ee65c7c93d6fa9ee6ea/static/media/amplitude-logo-with-text.4fb9e463.svg" width="280">
+  </a>
+  <br />
+</p>
+
+# @amplitude/plugin-auto-tracking-browser
+
+Official Browser SDK plugin for auto-tracking.
+
+## Installation
+
+This package is published on NPM registry and is available to be installed using npm and yarn.
+
+```sh
+# npm
+npm install @amplitude/plugin-auto-tracking-browser
+
+# yarn
+yarn add @amplitude/plugin-auto-tracking-browser
+```
+
+## Usage
+
+This plugin works on top of the Amplitude Browser SDK, generating auto-tracked events and sending to Amplitude.
+
+To use this plugin, you need to install `@amplitude/analytics-browser` version `v1.9.1` or later.
+
+### 1. Import Amplitude packages
+
+* `@amplitude/analytics-browser`
+* `@amplitude/plugin-auto-tracking-browser`
+
+```typescript
+import * as amplitude from '@amplitude/analytics-browser';
+import { autoTrackingPlugin } from '@amplitude/plugin-auto-tracking-browser';
+```
+
+### 2. Instantiate the plugin
+
+The plugin accepts 1 optional parameter, which is an `Object` to configure the allowed tracking options.
+
+```typescript
+const plugin = autoTrackingPlugin({
+  cssSelectorAllowlist: ['.amp-auto-tracking', '[amp-auto-tracking]'],
+  tagAllowlist: ['button', 'a'],
+});
+```
+
+Examples:
+- The above `cssSelectorAllowlist` will only allow tracking elements like:
+    - `<button amp-auto-tracking>Click</button>`
+    - `<a class="amp-auto-tracking">Link</a>`
+- The above `tagAllowlist` will only allow `button` and `a` tags to be tracked.
+
+Note `ingestionMetadata` is for internal use only, you don't need to provide it.
+
+#### Options
+
+|Name|Type|Default|Description|
+|-|-|-|-|
+|`cssSelectorAllowlist`|`string[]`|`undefined`| When provided, only allow elements matching any selector to be tracked. |
+|`tagAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label']`| Only allow elements with tag in this list to be tracked. |
+
+### 3. Install plugin to Amplitude SDK
+
+```typescript
+amplitude.add(plugin);
+```
+
+### 4. Initialize Amplitude SDK
+
+```typescript
+amplitude.init('API_KEY');
+```
--- a/packages/plugin-auto-tracking-browser/package.json
+++ b/packages/plugin-auto-tracking-browser/package.json
@@ -1,0 +1,57 @@
+{
+  "name": "@amplitude/plugin-auto-tracking-browser",
+  "version": "0.0.1",
+  "description": "",
+  "author": "Amplitude Inc",
+  "homepage": "https://github.com/amplitude/Amplitude-TypeScript",
+  "license": "MIT",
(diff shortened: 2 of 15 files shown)


packages/plugin-ga-events-forwarder-browser/test/ga-events-forwarder.test.ts
```diff
@@ -114,7 +114,7 @@ describe('gaEventsForwarderPlugin', () => {
   });
 
   describe('execute', () => {
-    test('should return the return the same event type', async () => {
+    test('should return the same event type', async () => {
```
kevinpagtakhan, line 117: ❤️

packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+  let text = '';
+  if (isNonSensitiveElement(element) && element.childNodes && element.childNodes.length) {
+    element.childNodes.forEach((child) => {
+      let childText = '';
+      if (isTextNode(child)) {
+        if (child.textContent) {
+          childText = child.textContent;
+        }
+      } else {
+        childText = getText(child as Element);
```
liuyang1520, line 41: @justin-fiedler , this should have the inner text not concatenated issue fixed.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -0,0 +1,191 @@
...
+interface EventListener {
+  element: Element;
+  type: ActionType;
+  handler: () => void;
+}
+
+interface Options {
+  cssSelectorAllowlist?: string[];
+  tagAllowlist?: string[];
+  ingestionMetadata?: IngestionMetadata; // Should avoid overriding this if unplanned to do so, this is not available in the charts.
```
liuyang1520, line 20: Add the ability to distinguish the ingestion source, this can be configured by the wrapped source like bookmarklet. This is only for internal usage.

packages/plugin-auto-tracking-browser/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@amplitude/plugin-auto-tracking-browser",
+  "version": "0.0.1",
```
yuhao900914, line 3: 0.0.0?

liuyang1520: Thanks for reviewing! Updated!

packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -0,0 +1,53 @@
+const SENTITIVE_TAGS = ['input', 'select', 'textarea'];
+
+export const isNonSensitiveString = (text: string | null) => {
+  if (text == null) {
+    return false;
+  }
```
yuhao900914, line 6: Will email also be the sensitive info for some customers?

yuhao900914 (approved):

packages/plugin-auto-tracking-browser/src/helpers.ts
liuyang1520: I think email is probably fine, not sensitive but PII, as the normal instrumentation would require user id/email to send events.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
justin-fiedler, line 20: @liuyang1520 instead of including `ingestionMetadata` on the plugin `Options`, can it instead be retrieved in `setup()` via the client config?

liuyang1520: Thanks @justin-fiedler ! Good call, I can remove this override and use the existing client config one.

packages/plugin-auto-tracking-browser/README.md
```diff
@@ -0,0 +1,76 @@
+<p align="center">
+  <a href="https://amplitude.com" target="_blank" align="center">
+    <img src="https://static.amplitude.com/lightning/46c85bfd91905de8047f1ee65c7c93d6fa9ee6ea/static/media/amplitude-logo-with-text.4fb9e463.svg" width="280">
+  </a>
+  <br />
+</p>
+
+# @amplitude/plugin-auto-tracking-browser
```
kevinpagtakhan, line 8: can we get this name approved by growth and dx teams (or other stakeholders)?

packages/plugin-auto-tracking-browser/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+### 2. Instantiate the plugin
+
+The plugin accepts 1 optional parameter, which is an `Object` to configure the allowed tracking options.
+
+```typescript
+const plugin = autoTrackingPlugin({
+  cssSelectorAllowlist: ['.amp-auto-tracking', '[amp-auto-tracking]'],
+  tagAllowlist: ['button', 'a'],
```
kevinpagtakhan, line 47: do these need to be separate or can we unify?

packages/plugin-auto-tracking-browser/README.md
liuyang1520: Good call! This plugin is for internal use at present, will definitely need more discussion/approval about the naming and interface if we want to make it ready to public users.

kevinpagtakhan: are we publishing to npm? it'll still be public if so. and we can't rename packages in npm

packages/plugin-auto-tracking-browser/README.md
liuyang1520: I thought about this. The function is a little bit overlapping but not all are easy to configure. For `tagAllowlist`, if it is not provided, we have a list of default tags that we want to track, while `cssSelectorAllowlist` is for more advanced usage, if doesn't have a default list. If we just rely on `cssSelectorAllowlist` with a default list like `['button', 'a']`, then problem is, when we want to add a allowed class, we have to do `['button.amp-auto-tracking', 'a.amp-auto-tracking']`, need to explicitly copy all tags here (this is what I think not easy to configure). Let me know if you have any good thoughts on this, this interface is not ready to public users at the moment though.

PR #572, opened 2023-08-30
fix: ignore non-string form.name values for form interaction tracking

kevinpagtakhan:
### Summary

ignore non-string `form.name` values for form interaction tracking

We've seen in a customer's production site where `form.name` can return a value of type Element. This yields an error when this value is included in the event payload as it eventually gets stringified and the element has circular references. The following error is thrown.

```
logger.ts:38 Amplitude Logger [Error]: Converting circular structure to JSON
    --> starting at object with constructor 'HTMLInputElement'
    |     property 'xxx' -> object with constructor 'x'
    --- property 'stateNode' closes the circle
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
+++ b/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
@@ -17,6 +17,7 @@ interface EventListener {
 export const formInteractionTracking = (): EnrichmentPlugin => {
   let observer: MutationObserver | undefined;
   let eventListeners: EventListener[] = [];
+
   const addEventListener = (element: Element, type: 'change' | 'submit', handler: () => void) => {
     element.addEventListener(type, handler);
     eventListeners.push({
@@ -25,6 +26,7 @@ export const formInteractionTracking = (): EnrichmentPlugin => {
       handler,
     });
   };
+
   const removeClickListeners = () => {
     eventListeners.forEach(({ element, type, handler }) => {
       /* istanbul ignore next */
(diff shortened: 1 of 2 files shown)


packages/analytics-browser/src/plugins/form-interaction-tracking.ts
```diff
@@ -122,3 +124,15 @@ export const formInteractionTracking = (): EnrichmentPlugin => {
     teardown,
   };
 };
+
+export const stringOrUndefined = (name: string) => {
```
kelsonpw, line 128: wondering if we can set this up to do proper type narrowing?

```ts
export const stringOrUndefined = <T>(
  name: T,
): T extends string ? string : undefined => {
  if (typeof name !== 'string') {
    return undefined as T extends string ? string : undefined;
  }

  return name as T extends string ? string : undefined;
};
```

kelsonpw: i.e. 
```ts

const str = stringOrUndefined('hello');
// ?^ string
const str2 = stringOrUndefined(undefined);
// ?^ undefined
```

yuhao900914 (approved):

packages/analytics-browser/src/plugins/form-interaction-tracking.ts
kevinpagtakhan: ohh i like this!

kevinpagtakhan: fixed!

PR #577, opened 2023-09-11
feat: add global user properties plugin 

kelvin-lu:
### Summary

This PR adds a simple enrichment plugin that transforms the user properties field into `global_user_properties` - a beta feature used internally first 

### Checklist


packages/plugin-global-user-properties/README.md
```diff
@@ -0,0 +1,96 @@
+<p align="center">
+  <a href="https://amplitude.com" target="_blank" align="center">
+    <img src="https://static.amplitude.com/lightning/46c85bfd91905de8047f1ee65c7c93d6fa9ee6ea/static/media/amplitude-logo-with-text.4fb9e463.svg" width="280">
+  </a>
+  <br />
+</p>
+
+# @amplitude/plugin-web-attribution-browser
```
kelvin-lu, line 8: todo: need to update this

packages/plugin-global-user-properties/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@amplitude/plugin-global-user-properties",
+  "version": "2.1.0",
+  "description": "",
```
kelsonpw, line 4: Any value in adding desc here?

packages/plugin-global-user-properties/README.md
```diff
@@ -0,0 +1,66 @@
...
+  initialEmptyValue: undefined,
+  resetSessionOnNewCampaign: undefined,
+});
+```
+
+#### Options
+
+|Name|Type|Default|Description|
+|-|-|-|-|
+|`shouldKeepOriginalUserProperties`|`boolean`| `false` | Use this option if you want the user properties to be sent along with the global user properties. This would do nothing for analyses but as of the beta version, global user properties do not appear in Data Governance.|
```
kelsonpw, line 53: I personally find this a bit confusing, especially the part about "This would do nothing for analyses but as of the beta version, global user properties do not appear in Data Governance." Maybe some clarity or further elaboration around relevance would be helpful for users.

packages/plugin-global-user-properties/src/global-user-properties.ts
```diff
@@ -0,0 +1,25 @@
...
+import { isAmplitudeIdentifyEvent, isNotSpecialAmplitudeEvent } from './helpers';
+
+export const globalUserPropertiesPlugin: GlobalUserPropertiesPlugin = function (options: Options = {}) {
+  const plugin: EnrichmentPlugin = {
+    name: '@amplitude/plugin-global-user-properties',
+    type: 'enrichment',
+
+    /* Note: The promise is because of the interface, not because this has any asynchronous behavior */
+    execute: async (event: Event): Promise<Event> => {
+      if (isNotSpecialAmplitudeEvent(event) || isAmplitudeIdentifyEvent(event)) {
```
kelsonpw, line 12: should we add a comment on why we don't attach globals to special events or is that well known logic already?

kelsonpw (approved):
It seems like there is a new CHANGELOG.md file, but the diff indicates it's empty. 

Just a few comments and questions to address, otherwise LGTM.

packages/plugin-global-user-properties/src/global-user-properties.ts
kelvin-lu: I can add a comment - revenue charts do not have user properties!

packages/plugin-global-user-properties/package.json
kelvin-lu: oh good catch, I copy-pasta'd this from another plugin and should add this

packages/plugin-global-user-properties/README.md
```diff
@@ -0,0 +1,66 @@
...
+import * as Amplitude from '@amplitude/analytics-browser'
+import { globalUserPropertiesPlugin } from '@amplitude/plugin-global-user-properties';
+```
+
+### 2. Instantiate page view plugin
+
+The plugin accepts an optional parameter of type `Object` to configure the plugin based on your use case.
+
+```typescript
+const globalUserPropertiesPlugin = globalUserPropertiesPlugin({
```
liuyang1520, line 42: Maybe typo? These options are not mentioned in below.

liuyang1520 (approved):

packages/plugin-global-user-properties/src/global-user-properties.ts
kelsonpw: oh interesting! TIL!

packages/plugin-global-user-properties/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@amplitude/plugin-global-user-properties",
+  "version": "2.1.0",
```
qingzhuozhen, line 3: version starts from 2?

packages/analytics-types/src/event.ts
```diff
@@ -107,11 +113,8 @@ export interface TrackEvent extends BaseEvent {
 
 export interface IdentifyEvent extends BaseEvent {
   event_type: SpecialEventType.IDENTIFY;
-  user_properties:
-    | IdentifyUserProperties
-    | {
-        [key in Exclude<string, IdentifyOperation>]: any;
-      };
+  global_user_properties?: UserProperties;
```
justin-fiedler, line 116: Are we 100% final on this property being part of the event definition? Naming, etc? Changing it after it's in production will be hard.

If this feature is in beta only I would prefer to leaving the base types unchanged.

TS has "[declaration merging](https://github.com/amplitude/Amplitude-Kotlin/pull/144#discussion_r1323554003)" for interfaces. Maybe you could define another `interface IdenitfyEvent { global_user_properties?: UserProperties; } ` in the plugin package for now?
Or type cast the event to `any` in the plugin for assignment `(event as any).global_user_properties = user_properties`

kelvin-lu: oh, I like this idea! I'll do that, especially since this would limit this definition to this plugin, and I'm not sure if we're driving this feature outside of our org.

kevinpagtakhan:
@justin-fiedler should we point this pr to `v1.x`?

packages/plugin-global-user-properties/README.md
```diff
@@ -0,0 +1,64 @@
...
+# npm
+npm install @amplitude/plugin-global-user-properties
+
+# yarn
+yarn add @amplitude/plugin-global-user-properties
+```
+
+## Usage
+
+This plugin works on top of Amplitude Browser SDK and adds web attribution tracking features to built-in features. To use this plugin, you need to install `@amplitude/plugin-global-user-properties `v0.0.0` or later.
```
kevinpagtakhan, line 26: This was added in other plugins to say that the Browser SDK should be from a certain version. Is that what we meant by this?

kevinpagtakhan: Looking at the plugin, it doesn't seem that it requires some version of the Browser SDK.

yuhao900914: adds web attribution tracking features to built-in features?

packages/plugin-global-user-properties/README.md
```diff
@@ -0,0 +1,64 @@
...
+const globalUserPropertiesPlugin = globalUserPropertiesPlugin({
+  shouldKeepOriginalUserProperties: true,
+});
+```
+
+#### Options
+
+|Name|Type|Default|Description|
+|-|-|-|-|
+|`shouldKeepOriginalUserProperties`|`boolean`| `false` | Use this option if you want the user properties to be sent along with the global user properties. Since global user properties do not appear in Governance yet, this would |
```
yuhao900914, line 51: The sentence is not completed.

packages/plugin-global-user-properties/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public",
+    "tag": "latest"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/amplitude/Amplitude-TypeScript.git"
+  },
+  "scripts": {
```
yuhao900914, line 20: Do we want to upload this to cdn? if yes, add `"publish": "node ../../scripts/publish/upload-to-s3.js"` will upload the script into S3.

packages/plugin-global-user-properties/src/global-user-properties.ts
yuhao900914: for other events which are not belong to amplitude special events, if there is no user_properties, do we want to set it to null?

yuhao900914: Please correct me if I'm wrong. 
Based on the definition of the `SpecialEventType`, it already includes the identify event. 

```
export enum SpecialEventType {
  IDENTIFY = '$identify',
  GROUP_IDENTIFY = '$groupidentify',
  REVENUE = 'revenue_amount',
}
```
`isNotSpecialAmplitudeEvent` means no event_type is '$identify', isAmplitudeIdentifyEvent means the event_type = '$identify'.
Does that conflict with each other?

packages/plugin-global-user-properties/README.md
kelvin-lu: I think I will just delete this - this was copy-pasted from the previous plugin since I wasn't sure which version started plugin support.

kelvin-lu:
> for other events which are not belong to amplitude special events, if there is no user_properties, do we want to set it to null?

I think returning the event unmodified would work in these cases

> isNotSpecialAmplitudeEvent means no event_type is '$identify', isAmplitudeIdentifyEvent means the event_type = '$identify'.

The type narrowing is not what i'm expecting it to do, but ideally I'm only modifying events of type `IdentifyEvent` and `BaseEvent` - really, I'm trying to avoid mutating group identify and revenue events, and other special event types that may come along the way

kelvin-lu:
Thanks everyone for the comments (and sorry for the amount of copy-pasta I left on the ground). I've updated the readme and the tests to include a `modules.d.ts` that fixes the types. 

Unfortunately, ts-jest is having some issues picking up this file, so the tests use a couple of any's. I'll make this a follow up ticket if this isn't a blocker to this pr.

packages/plugin-global-user-properties/package.json
kelvin-lu: we do not need to publish to cdn, since this will be used on sites with npm, thanks for asking about this though!

kelvin-lu:
@justin-fiedler @yuhao900914 I've limited my changes to the plugin package - I'll be merging this if there are no concerns

PR #584, opened 2023-09-15
fix: fix attribution to capture UTMs even if there is no referral info

yuhao900914:
### Summary

Fix the issue at [Jira](https://amplitude.atlassian.net/browse/AMP-84339?filter=18277), based on the comment 

> Direct = No external referral, no UTMs, no click-ids, and no other customer-identified marketing campaign url params.
> 
> Therefore we should always capture UTMs even if there is no referral info.  Stripping referral info is common with redirects.

### Checklist

### DIFF
--- a/packages/plugin-web-attribution-browser/src/helpers.ts
+++ b/packages/plugin-web-attribution-browser/src/helpers.ts
@@ -17,13 +17,15 @@ const domainWithoutSubdomain = (domain: string) => {
   return parts.slice(parts.length - 2, parts.length).join('.');
 };
 
+//Direct traffic mean no external referral, no UTMs, no click-ids, and no other customer identified marketing campaign url params.
 const isDirectTraffic = (current: Campaign) => {
   const { referrer, referring_domain, ...currentCampaign } = current;
-  Object.keys(currentCampaign).forEach(
-    (key) => (currentCampaign[key] === undefined || currentCampaign[key] === '') && delete currentCampaign[key],
+  const currentCampaignWithValue = Object.keys(currentCampaign).filter(
+    (key) => currentCampaign[key] !== undefined && currentCampaign[key] !== '',
   );
+  const noCampaign = Object.keys(currentCampaignWithValue).length === 0;
 
-  return !referrer && Object.keys(currentCampaign).length === 0;
+  return !referrer && noCampaign;
 };
 
 export const isNewCampaign = (
@@ -40,7 +42,6 @@ export const isNewCampaign = (
   }
 
   //In the same session, direct traffic should not override or unset any persisting query params
-  //Direct traffic mean no external referral, no UTMs, no click-ids, and no other customer identified marketing campaign url params.
   if (!isNewSession && isDirectTraffic(current) && previous) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


packages/plugin-web-attribution-browser/src/helpers.ts
```diff
@@ -30,8 +39,9 @@ export const isNewCampaign = (
     return false;
   }
 
-  //In the same session, no referrer should not override or unset any persisting query params
-  if (!isNewSession && !referrer && previous) {
+  //In the same session, direct traffic should not override or unset any persisting query params
+  //Direct traffic mean no external referral, no UTMs, no click-ids, and no other customer identified marketing campaign url params.
```
justin-fiedler, line 43: nit: this might be better above the function def at line 19

justin-fiedler (approved):

packages/plugin-web-attribution-browser/src/helpers.ts
```diff
@@ -17,6 +17,15 @@ const domainWithoutSubdomain = (domain: string) => {
   return parts.slice(parts.length - 2, parts.length).join('.');
 };
 
+const isDirectTraffic = (current: Campaign) => {
+  const { referrer, referring_domain, ...currentCampaign } = current;
+  Object.keys(currentCampaign).forEach(
+    (key) => (currentCampaign[key] === undefined || currentCampaign[key] === '') && delete currentCampaign[key],
+  );
```
kevinpagtakhan, line 24: do we really need to mutate this? i prefer that we do not mutate. if we're using this for the purposes of a conditional, we can just use `filter` or `every`

yuhao900914: update to `  const currentCampaignWithValue = Object.keys(currentCampaign).filter(key => (currentCampaign[key] === undefined || currentCampaign[key] === ''))`

packages/plugin-web-attribution-browser/test/helpers.test.ts
```diff
@@ -125,6 +125,20 @@ describe('isNewCampaign', () => {
 
     expect(isNewCampaign(currentCampaign, previousCampaign, {}, false)).toBe(false);
   });
+
+  test('should return true for no referrer with any new campaign in the same session', () => {
+    const previousCampaign = {
+      ...BASE_CAMPAIGN,
+      utm_campaign: 'utm_campaign',
+      referring_domain: 'a.b.c.d',
```
kevinpagtakhan, line 133: we can also remove this since in practice `referring_domain` exists when a `referrer` exists

kevinpagtakhan: ohh i misread, this is for the previous campaign. please ignore

packages/plugin-web-attribution-browser/src/helpers.ts
```diff
@@ -17,6 +17,17 @@ const domainWithoutSubdomain = (domain: string) => {
...
 
+//Direct traffic mean no external referral, no UTMs, no click-ids, and no other customer identified marketing campaign url params.
+const isDirectTraffic = (current: Campaign) => {
+  const { referrer, referring_domain, ...currentCampaign } = current;
+  const currentCampaignWithValue = Object.keys(currentCampaign).filter(
+    (key) => currentCampaign[key] !== undefined && currentCampaign[key] !== '',
+  );
+  const noCampaign = Object.keys(currentCampaignWithValue).length === 0;
+
+  return !referrer && noCampaign;
```
kevinpagtakhan, line 28: would it be as good as saying

```suggestion
  return Object.values(current).every(value => !value)
```

as in return true if every value in `current` is falsey otherwise return false?

yuhao900914: That's much clearer. I previously though the value could be boolean. Like this.

PR #585, opened 2023-09-18
fix(session replay): Update endpoint for session replay and only retry for 5xx responses

kwalker3690:
### Summary
We've updated the endpoint used to track sessions, so this PR updates the URL for those endpoints.

This PR also updates the logic to only retry requests if the response status is 500 or greater

### Checklist


jackson-amplitude (approved):
lgtm!

PR #586, opened 2023-09-20
fix(session replay): opt out should take effect immediately

jxiwang:
### Summary

Fix issue with session replay opt out not applying immediately. This change will immediately stop recordings and opt the user out of session replay. 

Also fixed issue where identityStoreOptOut would default to this.config.optOut when identityStoreOptOut was false. This misses the case where optOut should be false and not defaulted to this.config.optOut which could be the previous optOut state.


kwalker3690 (changes requested):
Awesome this looks great! Do you mind adding some tests to cover both of these changes? I know the test coverage says 100% (which is interesting in and of itself...) but clearly we didn't have the test coverage for this logic already. Great catch on the identityStoreOptOut bug too!

kwalker3690 (approved):

PR #587, opened 2023-09-21
fix(session replay) set max retries to cap at 1 and update to new api endpoint

kwalker3690:
### Summary

- Set max retries for request failures to 1 (flushMaxRetries size of 2)
- Max payload size of 1 MB
- Update to use new API endpoint

### Checklist


jackson-amplitude (approved):
Looks great! Thanks for the quick fix!

PR #588, opened 2023-09-21
fix(session replay plugin): update plugin to use the sessionReplayProperties fn

kwalker3690:
### Summary

We want to update all places we mention 'record' to mention 'replay' instead, so we will eventually deprecate `getSessionRecordingProperties`

### Checklist


jackson-amplitude (approved):
LGTM!

PR #589, opened 2023-09-22
feat: update auto tracking plugin

liuyang1520:
### Summary
feat: update auto tracking plugin
- Remove `tagAllowList` option to avoid confusion (if we want this, we can add in the future)
- Add `pageUrlAllowlist` option
- Add `shouldTrackEventResolver` option
- Add `dataAttributePrefix` option
- Add "arial label", "attributes", "selector" to the event property
- Add "Element Parent Label" to the event property
- Fork [`finder` lib](https://github.com/antonmedv/finder/blob/master/finder.ts), reasons
  - Package install works, but it fails when building with the code, as it contains TS source file, seems to be an [infamous bug](https://github.com/rollup/plugins/issues/287) of rollup, and the code has TS errors
  - Fork to local to make it more adjustable and testable
- Remove property keys when the value is empty

### Checklist


packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -51,3 +51,45 @@ export const getText = (element: Element): string => {
   }
   return text;
 };
+
+export const isPageUrlAllowed = (url: string, pageUrlAllowlist: string[] | undefined) => {
+  if (!pageUrlAllowlist || !pageUrlAllowlist.length) {
+    return true;
+  }
+  return pageUrlAllowlist.some((allowedUrl) => {
+    const allowedUrlRegex = new RegExp(allowedUrl);
```
kelsonpw, line 60: might be overkill, but a library like this one could provide really nice DX for defining path list

https://github.com/pillarjs/path-to-regexp

liuyang1520:
@justin-fiedler , re: meta tag info. I found the meta is at the page/site level. Do you think it is a little noisy to add for every event? If we want to add, maybe with a JSON event property like attributes?

packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -51,3 +51,45 @@ export const getText = (element: Element): string => {
   }
   return text;
 };
+
+export const isPageUrlAllowed = (url: string, pageUrlAllowlist: string[] | undefined) => {
```
kelsonpw, line 55: Is this matched against full url or path names? As written I would expect to have to define entire URLs and not just a list of paths

packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -51,3 +51,45 @@ export const getText = (element: Element): string => {
...
+  if (!pageUrlAllowlist || !pageUrlAllowlist.length) {
```
yuhao900914, line 56: No allowlist means, we white list any url?

packages/plugin-auto-tracking-browser/src/libs/finder.ts
```diff
@@ -0,0 +1,322 @@
+/* istanbul ignore file */
```
yuhao900914, line 1: Can we provide some explanation for this helper function?

justin-fiedler:
> @justin-fiedler , re: meta tag info. I found the meta is at the page/site level. Do you think it is a little noisy to add for every event? If we want to add, maybe with a JSON event property like attributes?

Thanks, you can omit meta tags info for now. I will add it if needed for semantic autotrack.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -52,6 +64,15 @@ export const autoTrackingPlugin = (options: Options = {}): BrowserEnrichmentPlug
     if (!element) {
       return false;
     }
+
+    if (shouldTrackEventCallback) {
+      return shouldTrackEventCallback(actionType, element);
```
justin-fiedler, line 69: I think "callback" in the name is a little confusing to me. wdyt about `shouldTrackEventResolver` or just `shouldTrackEvent`?

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -15,12 +18,21 @@ interface EventListener {
 }
 
 interface Options {
+  tagAllowlist?: AllowedTag[];
   cssSelectorAllowlist?: string[];
```
justin-fiedler, line 22: Can you please also add descriptions of these settings here so they appear in intellisense?

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -62,9 +83,11 @@ export const autoTrackingPlugin = (options: Options = {}): BrowserEnrichmentPlug
...
-    const tag = element.tagName.toLowerCase();
+    /* istanbul ignore next */
+    const tag = element?.tagName?.toLowerCase?.();
     /* istanbul ignore if */
-    if (!tagAllowlist.includes(tag)) {
+    if (!DEFAULT_TAG_ALLOWLIST.includes(tag) || !tagAllowlist.includes(tag)) {
+      // Tag needs to be in the default allowlist and the user provided allowlist.
       return false;
     }
     if (cssSelectorAllowlist) {
```
justin-fiedler, line 93: As discussed in the last meeting, can we update this to make it easier to allow a single new element without needing to change `tagAllowList`? 

E.g. `cssSelectorAllowlist = ['amp-allow']` would allow tracking this div and button without other configuration changes.

```
<body>
  <button>Click Me</button>
  <div amp-allow>Click me too<./div>
</body>
```

packages/plugin-auto-tracking-browser/src/helpers.ts
liuyang1520: Thanks for reviewing! Yes, if it is not provided, then we don't filter on this.

packages/plugin-auto-tracking-browser/src/libs/finder.ts
liuyang1520: Yes, there is a detailed doc here: https://github.com/antonmedv/finder. In brief, we use this helper function to get a unique CSS selector of the element (in current page).

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
liuyang1520: Thanks for reviewing! I like the name `shouldTrackEventResolver`, `shouldTrackEvent` sounds like a boolean value, I will update to `shouldTrackEventResolver`.

packages/plugin-auto-tracking-browser/src/helpers.ts
liuyang1520: Thanks for reviewing! The lib looks nice (thanks for sharing!), I don't think we need it at the moment, might be a little bit overkill.

packages/plugin-auto-tracking-browser/src/helpers.ts
liuyang1520: This is a regex match. As I think in most of the cases it would be a regex-like entry instead of a specific url, but it can be achievable as we support regex here. Example of allow list:
- `["https://amplitude.com/blog"]`, then `https://amplitude.com/blog/abc` should work here.
- `["http.?://*.amplitude"]`, then subdomains should work here.
- `["https://amplitude.com/blog$"]`, then should only match `https://amplitude.com/blog`.
I will add more tests in code and examples in the doc. (cc @justin-fiedler if you have any thoughts on this).

liuyang1520: @justin-fiedler , in this case, I can update the type to be a list of string or regex, so if it is a regex, it will work like above (user needs to provide a regex instead of string), if it is a string, then we do the exact match. How about this approach?

justin-fiedler: Nice, I like that 👍

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
liuyang1520: Thank you @justin-fiedler for discussing this offline. Updates:
- Remove `tagAllowList` for now, if needed, we can add in the future, addition is easier than deprecation.
- Will track the 6 tags & `cssSelectorAllowlist`.

packages/plugin-auto-tracking-browser/src/helpers.ts
```diff
@@ -51,3 +51,61 @@ export const getText = (element: Element): string => {
...
+    return filteredProperties;
+  }, {});
+};
+
+export const getNearestLabel = (element: Element): string => {
+  const parent = element.parentElement;
+  if (!parent) {
+    return '';
+  }
+  const labelElement = parent.querySelector(':scope>span,h1,h2,h3,h4,h5,h6');
```
liuyang1520, line 104: @justin-fiedler , I updated:
- `&>` to `:scope>` as `&` is not valid
- redact the string with `isNonSensitiveString`

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -62,11 +105,16 @@ export const autoTrackingPlugin = (options: Options = {}): BrowserEnrichmentPlug
...
       }
     }
-    const tag = element.tagName.toLowerCase();
+    /* istanbul ignore next */
+    const tag = element?.tagName?.toLowerCase?.();
+
+    // We only track these limited tags, as general text tag is not interactive and might contain sensitive information.
     /* istanbul ignore if */
-    if (!tagAllowlist.includes(tag)) {
+    if (!DEFAULT_TAG_ALLOWLIST.includes(tag)) {
```
justin-fiedler, line 113: Won't this prevent users from selecting other elements? Should we remove this and only use cssSelectorAllowList?

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -140,15 +196,15 @@ export const autoTrackingPlugin = (options: Options = {}): BrowserEnrichmentPlug
         });
       }
     };
-    const allElements = Array.from(document.body.querySelectorAll(tagAllowlist.join(',')));
+    const allElements = Array.from(document.body.querySelectorAll(DEFAULT_TAG_ALLOWLIST.join(',')));
```
justin-fiedler, line 199: cssSelectorAllowList instead?

packages/plugin-auto-tracking-browser/src/libs/finder.ts
```diff
@@ -0,0 +1,322 @@
...
+
+// License: MIT
+// Author: Anton Medvedev <anton@medv.io>
+// Source: https://github.com/antonmedv/finder
+
+type Knot = {
```
justin-fiedler, line 7: nit: would it make sense to fork this repo and import it? That way we can pull in changes from the main repo in the future for bug fixes etc.

liuyang1520: Good question, I should have mentioned more in the PR desc. Reasons why I copy to folk here:
- I was trying to use that package from npm directly, but the package install works. It fails when building with the code, as it contains TS source file, seems to be an https://github.com/rollup/plugins/issues/287 of rollup, and the code also has TS errors unless manually fixing minor types to pass in our build. Not sure whether there is a good solution on this, but I cannot figure it out after investigating for a day.
- If we folk to our repo, as we don't have a private registry, we probably need to publish to npm to installation either, which will likely introduce the similar issue above and we need to manage a new package.
- The original repo is not updated very often, and we will probably add our own logic when tuning the selector results.

I think we can definitely folk to a new repo (so we can sync from main repo easier) and manage it in the future if needed. It should be easily replaceable here.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -15,12 +17,44 @@ interface EventListener {
...
+   * Prefix for data attributes to allow auto collecting.
+   * Default is 'data-amp-auto-track-'.
+   */
+  dataAttributePrefix?: string;
 }
 
 export const autoTrackingPlugin = (options: Options = {}): BrowserEnrichmentPlugin => {
-  const { tagAllowlist = DEFAULT_TAG_ALLOWLIST, cssSelectorAllowlist } = options;
+  const {
+    cssSelectorAllowlist,
```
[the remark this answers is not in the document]
liuyang1520: Sorry, correct me if I miss anything from our last discussion.
-  `DEFAULT_TAG_ALLOWLIST` is the internal superset of all tags, I think we only track these 6 tags initially. The only way to overwrite is by providing the `shouldTrackEventResolver` function. That's why I have added a hard-coded check with `DEFAULT_TAG_ALLOWLIST.includes(tag)`.
- `cssSelectorAllowlist` is filtering based on the 6 tags.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
```diff
@@ -1,12 +1,14 @@
...
 import * as constants from './constants';
-import { getText } from './helpers';
+import { getText, isPageUrlAllowed, getAttributesWithPrefix, removeEmptyProperties, getNearestLabel } from './helpers';
+import { finder } from './libs/finder';
 
 type BrowserEnrichmentPlugin = EnrichmentPlugin<BrowserClient, BrowserConfig>;
 type ActionType = 'click' | 'change';
 
 const DEFAULT_TAG_ALLOWLIST = ['a', 'button', 'input', 'select', 'textarea', 'label'];
+const DEFAULT_DATA_ATTRIBUTE_PREFIX = 'data-amp-auto-track-';
```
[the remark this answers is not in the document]
liuyang1520: I can export `DEFAULT_DATA_ATTRIBUTE_PREFIX`, do we want to export `DEFAULT_TAG_ALLOWLIST` as it is only internally used?

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
liuyang1520: Explained in https://github.com/amplitude/Amplitude-TypeScript/pull/589#discussion_r1349205879.

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
liuyang1520: Selector is too dynamic, e.g., `[":scope>span", "button.abc"]`, I think a simple `join` combining might cause unexpected results. We can technically add another loop, iterating through all selectors then check, but there might have duplicates depending on the quality of selectors. I filter with the default tags at first then filter with any custom configurations later on. Do you have any concern on this?

packages/plugin-auto-tracking-browser/src/auto-tracking-plugin.ts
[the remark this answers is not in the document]
justin-fiedler: Although I understand we are being defensive or customers changing the defaults but I still think this is a little strange that the `cssSelectorAllowList` is dependent on `DEFAULT_TAG_ALLOWLIST`.

I would prefer a single configuration value that defines what elements are included. e.g. only `cssSelectorAllowList` vs `DEFAULT_TAG_LIST` + `cssSelectorAllowList` + `shouldTrackEventResolver()`

That said I think this clearly defined in the config comments and we will be dogfooding this before publishing publicly allowing us to change it later if it turns out to be clunky in practice.

justin-fiedler (approved):

packages/plugin-auto-tracking-browser/src/libs/finder.ts
lewgordon-amplitude: BTW I think I fixed this issue upstream (https://github.com/antonmedv/finder/pull/83). I'm hoping once the new build is published it will fix the compilation issue.

PR #591, opened 2023-09-22
fix: proxy non-promised methods in browser-snippet

justin-fiedler:
### Summary

[AMP-85112](https://amplitude.atlassian.net/browse/AMP-85112)

Non-promised functions are not being properly queued in the Browser snippet. There is [existing logic in browser-client.ts](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L131), but I don't think that works in the proxy case.

```
<script>
// load snippet function
!function(...)

amplitude.init()
<script>
<script>
   assertFalse(amplitude.invoked)
   amplitude.setUserId("user-1")

   setTimeout(() => {
      assertTrue(amplitude.invoked)
      const userId = amplitude.getUserId()
      // userId is undefined
   }, 3000)
<script>
```

I added new logic in the snippet to proxy non-promise methods.

[AMP-85112]: https://amplitude.atlassian.net/browse/AMP-85112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevinpagtakhan:
> There is [existing logic in browser-client.ts](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L131), but I don't think that works in the proxy case.

This logic is added for the situation where the SDK is already loaded, but operations are called before init.

packages/analytics-browser/generated/amplitude-snippet.js
```diff
@@ -35,6 +41,7 @@
         if (isPromise) return {
           promise: new Promise(getPromiseResult(instance, fn, Array.prototype.slice.call(arguments)))
         };
+        proxyInstance(instance, fn, Array.prototype.slice.call(arguments));
```
kevinpagtakhan, line 44: should it just be `proxy(instance, fn)`?

kevinpagtakhan: i was trying to look back but couldn't identify what caused the regression (or did it never work?)

justin-fiedler: Thanks, I don't think it ever worked.

I tried using proxy(instance, fn) but it fails to properly handle args.

scripts/templates/browser-snippet.template.js
```diff
@@ -97,11 +97,18 @@ const snippet = (name, integrity, version, globalVar) => `
         });
       };
     }
+    function proxyInstance(instance, fn, args) {
+      instance._q.push({
+        name: fn,
+        args: Array.prototype.slice.call(args, 0),
+      });
```
yuhao900914, line 104: Will no `resolve` be passed in the push function cause any issues?

yuhao900914 (approved):

packages/analytics-browser/generated/amplitude-snippet.js
```diff
@@ -89,4 +89,4 @@
     };
     window.amplitude = amplitude;
   }
-}(window, document);
\ No newline at end of file
+}(window, document);
```
kevinpagtakhan, line 92: ok to keep but want to note that diffs on this file will be overwritten by the publish action

kevinpagtakhan (commented):
🚀

kevinpagtakhan (approved):

PR #598, opened 2023-10-06
feat(session replay): add metadata to headers

jxiwang:
### Summary

Adding sample rate, full url, and version number to header. When adding metadata to the body, the request throws 400 error. Also added version file update script to match sdk-bot. 

* Does your PR have a breaking change?:  Possible issues with SSR frameworks such as Next.js when accessing the window, but there is a check that should account for that in helper file under getCurrentUrl.


packages/session-replay-browser/src/constants.ts
```diff
@@ -6,6 +6,7 @@ export const DEFAULT_EVENT_PROPERTY_PREFIX = '[Amplitude]';
 export const DEFAULT_SESSION_REPLAY_PROPERTY = `${DEFAULT_EVENT_PROPERTY_PREFIX} Session Recorded`;
 export const DEFAULT_SESSION_START_EVENT = 'session_start';
 export const DEFAULT_SESSION_END_EVENT = 'session_end';
+export const DEFAULT_SAMPLE_RATE = 1;
```
kwalker3690, line 9: Default sample rate should be 0

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -400,6 +406,9 @@ export class SessionReplay implements AmplitudeSessionReplay {
           'Content-Type': 'application/json',
           Accept: '*/*',
           Authorization: `Bearer ${apiKey}`,
+          'X-Client-Version': version,
```
kwalker3690, line 409: Check with backend that they can parse these

packages/session-replay-browser/src/verstion.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.2.5';
```
kwalker3690, line 1: Nit: file is misspelled

packages/session-replay-browser/test/session-replay.test.ts
```diff
@@ -1007,6 +1016,15 @@ describe('SessionReplayPlugin', () => {
     });
   });
 
+  describe('getSampleRate', () => {
+    test('should return undefined if no config set', () => {
```
kwalker3690, line 1020: should return default if no config set

packages/session-replay-browser/src/helpers.ts
```diff
@@ -25,3 +25,8 @@ export const isSessionInSample = function (sessionId: number, sampleRate: number
   const mod = absHashMultiply % 100;
   return mod / 100 < sampleRate;
 };
+
+export const getCurrentUrl = () => {
+  // eslint-disable-next-line no-restricted-globals
+  return typeof window !== 'undefined' ? window.location.href : '';
```
kwalker3690, line 31: Can you use `getGlobalScope` here instead? You won't have to disable the lint rule then too and its a bit more standard in the codebase

jxiwang: Yeah I can. Did not realize there was a function for this already.

kwalker3690 (approved):

PR #601, opened 2023-10-18
feat: add client upload time

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-87178

### Checklist

### DIFF
--- a/packages/analytics-types/src/payload.ts
+++ b/packages/analytics-types/src/payload.ts
@@ -8,4 +8,5 @@ export interface Payload {
   api_key: string;
   events: readonly Event[];
   options?: PayloadOptions;
+  client_upload_time?: string;
 }
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -25,6 +25,9 @@ describe('integration', () => {
   let apiKey = '';
   let client = amplitude.createInstance();
 
+  const event_upload_time = '2023-01-01T12:00:00:000Z';
+  Date.prototype.toISOString = jest.fn(() => event_upload_time);
+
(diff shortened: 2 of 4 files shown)


packages/analytics-types/src/payload.ts
```diff
@@ -8,4 +8,5 @@ export interface Payload {
   api_key: string;
   events: readonly Event[];
   options?: PayloadOptions;
+  client_upload_time?: string;
```
kevinpagtakhan, line 11: Is this different from `time`?

Mercy811: yes, they are different. `time` in SDK is [`client_event_time` in background](https://github.com/amplitude/amplitude/blob/90b19342d7d991bb4854102ad85fda22b299a3b3/event/converters/http_event.py#L254-L258).

liuyang1520: Same question, seems similar as `time` field except that this is ISO format.

Mercy811: Please see more background [here](https://amplitude.slack.com/archives/C02E1H87E2C/p1697495952472679)

liuyang1520 (approved):
Discussed with @Mercy811 offline, `client_upload_time` can be used by the server to estimate the clock drift of that device (server received time - upload time), then adjust the client event time in the server side.

packages/analytics-types/src/payload.ts
Mercy811: Closed it as getting approved from Kevin from mid week oncall sync

ondrej, line 11: Are there cases where we wouldn't set `client_upload_time`?

PR #605, opened 2023-10-25
fix: log error trace in catch blocks

kevinpagtakhan:
### Summary

Logs caught errors before stringifying
* Allows complete error trace to be logged using the logger provider

### Checklist

### DIFF
--- a/packages/analytics-core/src/core-client.ts
+++ b/packages/analytics-core/src/core-client.ts
@@ -127,8 +127,8 @@ export class AmplitudeCore implements CoreClient {
 
       return result;
     } catch (e) {
+      this.config.loggerProvider.error(e);
       const message = String(e);
-      this.config.loggerProvider.error(message);
       const result = buildResult(event, 0, message);
 
       return result;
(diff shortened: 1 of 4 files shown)


Mercy811 (approved):

packages/analytics-core/src/core-client.ts
```diff
@@ -127,8 +127,8 @@ export class AmplitudeCore implements CoreClient {
 
       return result;
     } catch (e) {
+      this.config.loggerProvider.error(e);
```
justin-fiedler, line 130: I see that the Logger interface accepts `...args: any[]` but wonder if this will make implementing custom loggers harder for customers vs simple strings (each logger will need to implement the string conversion itself).

Since this doesn't change the public interface I think this is ok, just something to think about.

justin-fiedler (approved):

packages/analytics-core/src/core-client.ts
kevinpagtakhan: I think because it's already of `any` type and we never committed to a string parameter, the custom implementation should handle conversion to the preferred type. So i think we should be ok.

PR #606, opened 2023-10-26
feat: make auto-tracking solely use css selector allowlist

liuyang1520:
### Summary
feat: make auto-tracking solely use css selector allowlist

This change make `cssSelectorAllowlist` the single source of truth for the selector configuration, which allows us to track clicking of elements not in the default tag list: `['a', 'button', 'input', 'select', 'textarea', 'label']`. Reason is some button might be built with `div`, not with the `button`.

Add `event.stopPropagation();` specifically, as nested element clicking may cause unexpected events flood.

### Checklist

### DIFF
--- a/packages/plugin-default-event-tracking-advanced-browser/README.md
+++ b/packages/plugin-default-event-tracking-advanced-browser/README.md
@@ -65,7 +65,7 @@ Examples:
 
 |Name|Type|Default|Description|
 |-|-|-|-|
-|`cssSelectorAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label']`| When provided, only allow elements matching any selector to be tracked. |
+|`cssSelectorAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label', '.amp-default-track']`| When provided, only allow elements matching any selector to be tracked. |
 |`pageUrlAllowlist`|`(string\|RegExp)[]`|`undefined`| When provided, only allow elements matching URLs to be tracked. |
 |`shouldTrackEventResolver`|`(actionType: ActionType, element: Element) => boolean`|`undefined`| When provided, overwrite the default filter behavior. |
 |`dataAttributePrefix`|`string`|`'data-amp-track-'`| Allow data attributes to be collected in event property. |
--- a/packages/plugin-default-event-tracking-advanced-browser/README.md
+++ b/packages/plugin-default-event-tracking-advanced-browser/README.md
@@ -65,7 +65,7 @@ Examples:
 
 |Name|Type|Default|Description|
 |-|-|-|-|
-|`cssSelectorAllowlist`|`string[]`|`undefined`| When provided, only allow elements matching any selector to be tracked. |
+|`cssSelectorAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label']`| When provided, only allow elements matching any selector to be tracked. |
 |`pageUrlAllowlist`|`(string\|RegExp)[]`|`undefined`| When provided, only allow elements matching URLs to be tracked. |
 |`shouldTrackEventResolver`|`(actionType: ActionType, element: Element) => boolean`|`undefined`| When provided, overwrite the default filter behavior. |
 |`dataAttributePrefix`|`string`|`'data-amp-track-'`| Allow data attributes to be collected in event property. |
--- a/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
+++ b/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
@@ -14,7 +14,15 @@ import { finder } from './libs/finder';
 type BrowserEnrichmentPlugin = EnrichmentPlugin<BrowserClient, BrowserConfig>;
 type ActionType = 'click' | 'change';
 
-export const DEFAULT_CSS_SELECTOR_ALLOWLIST = ['a', 'button', 'input', 'select', 'textarea', 'label'];
+export const DEFAULT_CSS_SELECTOR_ALLOWLIST = [
+  'a',
+  'button',
+  'input',
+  'select',
+  'textarea',
+  'label',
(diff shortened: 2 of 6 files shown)


packages/plugin-default-event-tracking-advanced-browser/README.md
```diff
@@ -65,7 +65,7 @@ Examples:
 
 |Name|Type|Default|Description|
 |-|-|-|-|
-|`cssSelectorAllowlist`|`string[]`|`undefined`| When provided, only allow elements matching any selector to be tracked. |
+|`cssSelectorAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label']`| When provided, only allow elements matching any selector to be tracked. |
```
kevinpagtakhan, line 68: since we are supporting tags, should this be better named as querySelectorAllowlist or something?

packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
```diff
@@ -184,27 +178,29 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
     }
     const addListener = (el: Element) => {
       if (shouldTrackEvent('click', el)) {
-        addEventListener(el, 'click', () => {
+        addEventListener(el, 'click', (event: Event) => {
           /* istanbul ignore next */
           amplitude?.track(constants.AMPLITUDE_ELEMENT_CLICKED_EVENT, getEventProperties('click', el));
+          event.stopPropagation();
```
kevinpagtakhan, line 184: could this impact events listener added by the customer for some other purpose?

liuyang1520: Thanks for reviewing! Good question, it shouldn't, `stopPropagation` only prevents this event handler bubbling, there is another method [`stopImmediatePropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation) preventing other event listeners:
> If several listeners are attached to the same element for the same event type, they are called in the order in which they were added. If stopImmediatePropagation() is invoked during one such call, no remaining listeners will be called, either on that element or any other element.

packages/plugin-default-event-tracking-advanced-browser/README.md
liuyang1520: I prefer css selector more as query selector is a method name for querying elements with CSS selectors, e.g., https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#selectors:
> A string containing one or more selectors to match. This string must be a valid CSS selector string; if it isn't, a SyntaxError exception is thrown.

It mentions the [CSS selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors) concept, which includes a list of all [selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors#selectors), like id, class, type (tag).

kevinpagtakhan: ahh i see i see

packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
```diff
@@ -108,19 +115,6 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
...
-      return false;
-    }
-
-    /* istanbul ignore if */
-    if (cssSelectorAllowlist) {
-      const hasMatchAnyAllowedSelector = cssSelectorAllowlist.some((selector) => element.matches(selector));
-      if (!hasMatchAnyAllowedSelector) {
-        return false;
-      }
-    }
```
justin-fiedler, line 123: Curious, why was the cssSelectorAllowlist match check removed?

packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
```diff
@@ -1,19 +1,26 @@
...
+  querySelectUniqueElements,
+} from './helpers';
 import { finder } from './libs/finder';
 
 type BrowserEnrichmentPlugin = EnrichmentPlugin<BrowserClient, BrowserConfig>;
 type ActionType = 'click' | 'change';
 
-const DEFAULT_TAG_ALLOWLIST = ['a', 'button', 'input', 'select', 'textarea', 'label'];
-const DEFAULT_DATA_ATTRIBUTE_PREFIX = 'data-amp-track-';
+export const DEFAULT_CSS_SELECTOR_ALLOWLIST = ['a', 'button', 'input', 'select', 'textarea', 'label'];
```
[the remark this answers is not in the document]
liuyang1520: Thanks for reviewing! Good idea, will add it.

packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
liuyang1520: The elements are already filtered out based on the `cssSelectorAllowlist`(https://github.com/amplitude/Amplitude-TypeScript/pull/606/files#diff-f0d884147fd15c876ca81736143fe16eff39e7d5547db18718c304519a7a9eabR195-R203). We need this before, because we filter the element by tags then by this filter.

justin-fiedler: Thank you!

packages/plugin-default-event-tracking-advanced-browser/README.md
justin-fiedler, line 68:
```suggestion
|`cssSelectorAllowlist`|`string[]`|`['a', 'button', 'input', 'select', 'textarea', 'label', '.amp-default-track']`| When provided, only allow elements matching any selector to be tracked. |
```

justin-fiedler (approved):
Thanks @liuyang1520! Looks good. See fix on README.

PR #611, opened 2023-10-31
feat: add session replay ID generic identifier to init/turnover

jxiwang:
Adding in generic session replay ID to initialization and session turnover:

### Browser SDK
```
// Your existing initialization logic with Browser SDK
amplitude.init(API_KEY);

// Create and Install Session Replay Plugin
const sessionReplayTracking = sessionReplayPlugin();
amplitude.add(sessionReplayTracking);

amplitude.getSessionReplay
```

### Standalone SDK
```
// Can include generic sessionReplayId 
await sessionReplay.init(AMPLITUDE_API_KEY, {
  deviceId: <string>,
  sessionId: <number>,
  // sessionReplayId: <string>,
  optOut: <boolean>,
  sampleRate: <number>
}).promise;

// This would generate from sessionId/deviceId
sessionReplay.setSessionId({sessionId:123});

// This would use passed in generic
//sessionReplay.setSessionId({sessionReplayId:"123/456"});

// This will include [NEW_SESSION_REPLAY_PROPERTY]: sessionReplayId / "device_id/session_id" / null
const sessionReplayProperties = sessionReplay.getSessionReplayProperties();
```

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -51,7 +51,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
 
   async execute(event: Event) {
     if (event.event_type === DEFAULT_SESSION_START_EVENT && event.session_id) {
-      sessionReplay.setSessionId(event.session_id);
+      sessionReplay.setSessionId({ sessionId: event.session_id });
     }
 
     const sessionRecordingProperties = sessionReplay.getSessionReplayProperties();
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -51,7 +51,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
 
   async execute(event: Event) {
     if (event.event_type === DEFAULT_SESSION_START_EVENT && event.session_id) {
-      sessionReplay.setSessionId({ sessionId: event.session_id });
+      sessionReplay.setSessionId(event.session_id);
     }
 
     const sessionRecordingProperties = sessionReplay.getSessionReplayProperties();
@@ -65,6 +65,10 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
 
   async teardown(): Promise<void> {
     sessionReplay.shutdown();
+  }
+
+  getSessionReplayProperties() {
+    return sessionReplay.getSessionReplayProperties();
   }
 }
 
--- a/packages/session-replay-browser/src/config.ts
+++ b/packages/session-replay-browser/src/config.ts
@@ -3,6 +3,7 @@ import { Config, Logger } from '@amplitude/analytics-core';
 import { LogLevel } from '@amplitude/analytics-types';
 import { SessionReplayConfig as ISessionReplayConfig, SessionReplayOptions } from './typings/session-replay';
 import { DEFAULT_SAMPLE_RATE } from './constants';
+import { generateSessionReplayId } from './helpers';
 
 export const getDefaultConfig = () => ({
   flushMaxRetries: 2,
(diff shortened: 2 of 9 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -82,13 +89,24 @@ export class SessionReplay implements AmplitudeSessionReplay {
     }
   }
 
-  setSessionId(sessionId: number) {
+  setSessionId({ sessionId, sessionReplayId }: { sessionId?: number; sessionReplayId?: string }) {
```
jackson-amplitude, line 92: would this be a breaking change for customers who are on segment for example.

packages/session-replay-browser/src/helpers.ts
```diff
@@ -31,3 +31,23 @@ export const getCurrentUrl = () => {
...
+};
+
+export const parseSessionReplayId = (
+  sessionReplayId: string | undefined,
+): { deviceId?: string; sessionId?: string } => {
+  if (!sessionReplayId) {
+    return {};
+  }
+
+  const parts = sessionReplayId.split('/');
```
jackson-amplitude, line 46: this is a bit brittle, can break if customer passes in the wrong sessionReplayId that's not made of device/session id. I wonder if we should drop the option to directly pass in a SessionReplayId before we can better support it.

packages/session-replay-browser/src/typings/session-replay.ts
```diff
@@ -39,14 +39,15 @@ export interface SessionReplayConfig extends Config {
...
 
 export type SessionReplayOptions = Omit<Partial<SessionReplayConfig>, 'apiKey'>;
 
 export interface AmplitudeSessionReplay {
   init: (apiKey: string, options: SessionReplayOptions) => AmplitudeReturn<void>;
-  setSessionId: (sessionId: number) => void;
-  getSessionRecordingProperties: () => { [key: string]: boolean };
-  getSessionReplayProperties: () => { [key: string]: boolean };
+  setSessionId: ({ sessionId, sessionReplayId }: { sessionId?: number; sessionReplayId?: string }) => void;
+  getSessionRecordingProperties: () => { [key: string]: boolean | string | null };
```
jackson-amplitude, line 50: the legal requirement is to remove any reference to "getSessionRecordingProperties", we should do that as part of this work. I think we've had this deprecated for a while now.

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -51,7 +51,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
 
   async execute(event: Event) {
     if (event.event_type === DEFAULT_SESSION_START_EVENT && event.session_id) {
-      sessionReplay.setSessionId(event.session_id);
+      sessionReplay.setSessionId({ sessionId: event.session_id });
```
jackson-amplitude, line 54: should we just keep this as a number instead of object? Doesn't feel like a necessary change right now.

packages/session-replay-browser/src/config.ts
```diff
@@ -32,7 +34,10 @@ export class SessionReplayConfig extends Config implements ISessionReplayConfig
     this.apiKey = apiKey;
     this.sampleRate = options.sampleRate || DEFAULT_SAMPLE_RATE;
 
-    this.deviceId = options.deviceId;
-    this.sessionId = options.sessionId;
+    if (options.sessionId && options.deviceId) {
```
jackson-amplitude, line 37: this seems like we would fail silently. should log an error if these are no available?

jackson-amplitude (approved):
LGTM!

packages/session-replay-browser/src/config.ts
jxiwang: Added a check!

packages/plugin-session-replay-browser/src/session-replay.ts
jxiwang: Should be number now, was an older commit.

PR #613, opened 2023-11-02
fix: skip parent element event when it is not target

liuyang1520:
### Summary
- fix: skip parent element event when it is not target
- feat: limit to only the innermost allowed element

In the following case, the plugin will fire 3 events, as the clicking event is propagated to the parent (and parent's parent):
```html
<div data-testid="details">
  <div data-testid="container">
    <button data-testid="button">Click Me</button>
  </div>
</div>

// configure
cssSelectorAllowlist: ["[data-testid]"]
```

To avoid this, we can check the `event?.target != event?.currentTarget`. However, this will potential introduce another issue:
```html
<button>
  <span>Click Me</span>
</button>

// configure
cssSelectorAllowlist: ["button"]
```
When clicking on `span`, it won't track as `span` is not in the allow list, and due to the change we check the target, it will not track the event for `button` as well.

To fix the above issue, we introduce another check `getClosestElement(event?.target, cssSelectorAllowlist) != event?.currentTarget`, so it will track the event for the innermost allowed element, e.g.:
```html
<div id="container2">
  <div id="container1">
    <div id="inner">
      click me
    </div>
  </div>
</div>

cssSelectorAllowlist: ['div']
expect: only track inner, as we should only track the innermost allowed element

<div id="container2">
  <div id="container1">
    <span id="inner">
      click me
    </span>
  </div>
</div>

cssSelectorAllowlist: ['div']
expect: only track container1, as we should only track the innermost allowed element
```

There is a case like below:
```html
<button id="container" data-track>
  <div id="inner" data-track>
    click me
  </div>
</button>

cssSelectorAllowlist: ['[data-track]']
expect: only track div click, as div is innermost element that matches allowlist
```
Note: we do not track the button click here, this is a rare case that the inner div is also allowed.

### Checklist

### DIFF
--- a/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
+++ b/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
@@ -200,13 +200,23 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
     }
     const addListener = (el: Element) => {
       if (shouldTrackEvent('click', el)) {
-        addEventListener(el, 'click', () => {
+        addEventListener(el, 'click', (event: Event) => {
+          // Limit to only the element that was clicked, not the propagated event.
+          /* istanbul ignore next */
+          if (event?.target != event?.currentTarget) {
+            return;
+          }
           /* istanbul ignore next */
           amplitude?.track(constants.AMPLITUDE_ELEMENT_CLICKED_EVENT, getEventProperties('click', el));
         });
       }
       if (shouldTrackEvent('change', el)) {
-        addEventListener(el, 'change', () => {
+        addEventListener(el, 'change', (event: Event) => {
+          // Limit to only the element that was clicked, not the propagated event.
+          /* istanbul ignore next */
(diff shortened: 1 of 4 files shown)


packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
```diff
@@ -200,13 +200,23 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
     }
     const addListener = (el: Element) => {
       if (shouldTrackEvent('click', el)) {
-        addEventListener(el, 'click', () => {
+        addEventListener(el, 'click', (event: Event) => {
+          // Limit to only the element that was clicked, not the propagated event.
+          /* istanbul ignore next */
+          if (event?.target != event?.currentTarget) {
```
justin-fiedler, line 206: What about checking that the child nodes don't contain a match the CSS selector list?

```
 if (event?.currentTarget?.querySelector(cssSelectorsList)) {
```

justin-fiedler (approved):
Thanks @liuyang1520! Looks good.

PR #621, opened 2023-11-15
ci: prerelease feature branch

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-88911
Add a Github actions workflow to prerelease from a feature branch
- Can run a dry run without pushing to remote repo 
- Must input the branch name. It is then used when checkout and used as the preid

Example:
If current version is `@amplitude/analytics-browser: 2.2.1`, the feature branch is `feat-sdk-diagnostics`, it will prerelease to ` 2.3.0-featsdkdiagnostics.0`

### Checklist

* [ ] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:

PR #622, opened 2023-11-16
ci: update dryRun type to string

Mercy811:
### Summary

A fix PR to https://github.com/amplitude/Amplitude-TypeScript/pull/621

image
Cannot go into If condition because GitHub Actions treats all input types as strings, even if they are defined as boolean

### Checklist


justin-fiedler (commented):
I feel the check box is a little unclear if your doing a dryRun vs release.

To make it explicit how about making it a dropdown like the release type in the `Publish vX` actions? e.g. `dry-run` and `prerelease`
image

justin-fiedler (approved):

PR #625, opened 2023-11-16
chore(session replay): manually bump version to 0.8.0

jxiwang:
### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/package.json
+++ b/packages/plugin-session-replay-browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amplitude/plugin-session-replay-browser",
-  "version": "0.7.1",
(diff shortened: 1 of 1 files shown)


packages/plugin-session-replay-browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@amplitude/plugin-session-replay-browser",
-  "version": "0.7.1",
+  "version": "0.8.0",
```
justin-fiedler, line 3: I _think_ that this will work, but not 100% how it will play with lerna, I have been relying on the CI for versioning for too long. Do you know @kevinpagtakhan?

Another option would be to create a `feat: update to new version` PR and update the README or another non-functional code in the package.

PR #633, opened 2023-12-05
feat: convert session replay plugin to be a destination plugin with flush support

justin-fiedler:
### Summary
Alternative to #628 

* [AMP-89820](https://amplitude.atlassian.net/browse/AMP-89820) - update Browser SR plugin to allow explicit flushing
* [fix: binding of sessionReplay shutdown method](https://github.com/amplitude/Amplitude-TypeScript/pull/633/commits/65812f376aa5c0129eeb9759a0dd3297166271d5)

For Mobile session replay I am using the Browser SR plugin to convert the mobile layouts to `rrweb` events. This works well but I need a way to explicitly flush the SR events, which is not currently possible as the Browser SR plugin doesn't expose a flush method.

This PR converts the session replay plugin to a DestinationPlugin to add `flush()`. Since we also need to do enrichment the new DestinationPlugin adds a EnrichmentPlugin in `setup()`. This allows the change to be (mostly) backwards compatible.

"Breaking" Changes
1. SessionReplayPlugin now extends DestinationPlugin (vs EnrichmentPlugin). This should be transparent to users as long as they are using `sessionReply.plugin()` and `createSessionReplayPlugin()` and not manually extending the class.
2. SessionReplayPlugin now requires access to the Amplitude BrowserClient which is passed into `plugin.setup(config, client)`. The `client` parameter is only supported in `v1.9.1+`. A [majority of customers are using version](https://app.amplitude.com/analytics/amplitude/chart/new/iuc36ny3) `v1.9.1+`.

Customer comm updates
1. We should call out the SessionReplay will now require `v1.9.1+`

### Checklist

* Does your PR have a breaking change?:  Kind of, but should be transparent to end users

[AMP-89820]: https://amplitude.atlassian.net/browse/AMP-89820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -45,7 +45,12 @@ export class SessionReplayPlugin implements DestinationPlugin {
     this.options = { ...options };
   }
 
-  async setup(config: BrowserConfig, client: BrowserClient) {
+  async setup(config: BrowserConfig, client?: BrowserClient) {
+    if (!client) {
+      config.loggerProvider.error('SessionReplayPlugin requires v1.9.1+ of the Amplitude SDK.');
+      return;
+    }
+
     config.loggerProvider.log('Installing @amplitude/plugin-session-replay.');
 
     this.config = config;
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -1,10 +1,40 @@
-import { BrowserConfig, EnrichmentPlugin, Event } from '@amplitude/analytics-types';
+import {
+  BrowserClient,
+  BrowserConfig,
+  DestinationPlugin,
+  EnrichmentPlugin,
+  Event,
+  Result,
+} from '@amplitude/analytics-types';
 import * as sessionReplay from '@amplitude/session-replay-browser';
 import { DEFAULT_SESSION_START_EVENT } from './constants';
 import { SessionReplayOptions } from './typings/session-replay';
-export class SessionReplayPlugin implements EnrichmentPlugin {
-  name = '@amplitude/plugin-session-replay-browser';
+
+class SessionReplayEnrichmentPlugin implements EnrichmentPlugin {
+  name = '@amplitude/plugin-session-replay-browser-enrichment';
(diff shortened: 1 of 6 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -15,7 +45,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
     this.options = { ...options };
   }
 
-  async setup(config: BrowserConfig) {
+  async setup(config: BrowserConfig, client: BrowserClient) {
```
kevinpagtakhan, line 48:
```suggestion
  async setup(config: BrowserConfig, client?: BrowserClient) {
  
  if (!client) {
    config.loggerProvider.error('...')
    return;
  }
```

i think to be safe, we can make client optional and handle it accordingly to prevent runtime errors for versions <= 1.9.0.

justin-fiedler: Fixed.

jxiwang (approved):
LGTM

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -15,7 +45,12 @@ export class SessionReplayPlugin implements EnrichmentPlugin {
     this.options = { ...options };
   }
 
-  async setup(config: BrowserConfig) {
+  async setup(config: BrowserConfig, client?: BrowserClient) {
+    if (!client) {
+      config.loggerProvider.error('SessionReplayPlugin requires v1.9.1+ of the Amplitude SDK.');
+      return;
+    }
```
jxiwang, line 52: In our docs/guides we already recommend v1.9.1+

kevinpagtakhan: good to know we already are! there might be some spots that need updating like: https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/plugin-session-replay-browser/README.md?plain=1#L26

justin-fiedler: Updated README

PR #635, opened 2023-12-16
feat: adds support for setting offline mode

kevinpagtakhan:
This PR introduces online/offline mode support for the Amplitude SDK. Online mode indicates that the platform has an active connection to the network, while offline mode indicates that the platform does not have an active connection to the network. During online mode, the SDK works as usual and sends events at the configured time interval, while during offline mode, the SDK does not send events at all, but stores unsent events to the configured storage. Events persisted in the configured storage are available to be sent at a later time. The SDK by default is set to online mode, unless explicitly set through initialization options or the `setOffline()` API. Note that the online/offline mode support does not include automatic toggles between modes. The mode must explicitly be set by the instrumenter. 

#### To set Amplitude SDK to offline mode

* via initialization options
```ts
amplitude.init(apiKey, {
  offline: true,
});
```

* via `setOffline()` API
```ts
amplitude.setOffline(true)
```

#### To set Amplitude SDK to online mode

* via initialization options (unnecessary because offline mode is false by default)
```ts
amplitude.init(apiKey, {
  offline: false,
});
```

* via `setOffline()` API
```ts
amplitude.setOffline(false)
```

### Checklist


qingzhuozhen (approved):

justin-fiedler (approved):
Thanks @kevinpagtakhan !

packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -937,6 +937,28 @@ describe('browser-client', () => {
...
+    });
+
+    test('should set offline to false', async () => {
+      await client.init(apiKey, {
+        defaultTracking: false,
+        offline: true,
+      }).promise;
+
+      client.setOffline(false);
+      expect(client.config.offline).toBe(false);
```
justin-fiedler, line 959: Maybe check that flush is called here?

PR #644, opened 2024-01-05
feat: add offline mode

Mercy811:
### Summary
Add support for offline. When online, it behaves exactly the same as previous versions. When offline, tracking an event will not schedule a flush not will only be stored in storage. When back to online, the SDK will flush all stored events.

Offline feature is support by default

#### To opt of the default offline logic provided by network checker plugin
```
amplitude.init('API_KEY', {
  offline: OfflineDisabled
});
```

### Checklist


packages/analytics-browser/src/browser-client.ts
```diff
@@ -130,6 +133,22 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     connector.eventBridge.setEventReceiver((event) => {
       void this.track(event.eventType, event.eventProperties);
     });
+
+    // Step 8: Add network listeners for offline mode
```
justin-fiedler, line 137: I would prefer if this is done in the Plugin instead. Plugins should  have access to `client.flush()` via `plugin.setup(config, client)`

justin-fiedler (changes requested):

packages/analytics-browser/src/browser-client.ts
Mercy811: I moved all the offline logic to the plugin at https://github.com/amplitude/Amplitude-TypeScript/pull/644/commits/8d14bfc77bb7a7c98fd82309b9cd9a4e990f80bd#diff-0c1d0fcf8156764bdf906a8f23ee6243eff9afeda0799e3551ad83237562e4f0

Mercy811:
It requires `setOffline()` to enable custom offline logic by first opt out the default plugin and then toggling `config.offline` which is unlike Kotlin which has direct access to `amplitude.config`. I think it's an extended feature and we should release this version without supporting custom offline logic first. If there is customer request, then we can plan for it later.

packages/analytics-core/test/plugins/destination.test.ts
```diff
@@ -140,6 +144,40 @@ describe('destination', () => {
       expect(flush).toHaveBeenCalledTimes(2);
     });
 
+    test('should not schedule a flush', async () => {
```
justin-fiedler, line 147:
```suggestion
    test('should not schedule a flush if offline', async () => {
```

packages/analytics-browser/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const setup = async (config: BrowserConfig, amplitude: BrowserClient) => {
+    config.offline = !navigator.onLine;
+
+    addNetworkListener('online', () => {
+      config.offline = false;
+      // Flush immediately will cause ERR_NETWORK_CHANGED
+      setTimeout(() => {
+        amplitude.flush();
+      }, config.flushIntervalMillis);
```
justin-fiedler, line 43: I feel this should be more immediate. Can we do next frame instead? `setTimeout(() => {...}, 1)`?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -117,6 +120,11 @@ export class Destination implements DestinationPlugin {
   }
 
   async flush(useRetry = false) {
+    // Skip flush if offline
+    if (this.config.offline) {
+      return;
```
justin-fiedler, line 125: I feel we should log here to help with debugging. e.g. `config.logger.debug("Skipping flush while offline.")

justin-fiedler (approved):
Overall looks good. Thanks @Mercy811. Really appreciate the test coverage.

See question about scheduling flush earlier on return to online.

packages/analytics-core/src/plugins/destination.ts
Mercy811: Good point! I added logging messages when goes to online and offline too in https://github.com/amplitude/Amplitude-TypeScript/pull/644/commits/c2c585304a4e6381c648fe471e9603cc5accd735

packages/analytics-browser/src/plugins/network-connectivity-checker.ts
justin-fiedler: Ok, I think that's ok. Can you please call this out in docs. Thanks

Mercy811: Definitely!

Mercy811: Docs PR: https://github.com/amplitude/amplitude-dev-center/pull/159

Multiply:
> It requires `setOffline()` to enable custom offline logic by first opting out the default plugin and then toggling `config.offline` which is unlike Kotlin which has direct access to `amplitude.config`. I think adding `setOffline()` is an extended feature and we should release this version without supporting custom offline logic first. If there is customer request, then we can plan for it later.

We just noticed this landing, and we'll most definitely use this functionality to pause sending events, and just let them queue up.
For the most part, we'll just do the same as the default plugin, but also check if we've paused sending events before setting it back to online.

Mercy811:
Hi @Multiply, thanks for choosing Amplitude and using the new feature! I will put it into our roadmap. Could you give me more context on the difference between the default `networkConnectivityCheckerPlugin` and the one you want to implement? We can see if that's what we can make it as default to support more customers with the same needs.

Multiply:
Currently, I'm trying to set offline to disabled, and then set it to true via our own plugin, but it seems to result in the default plugin being loaded anyway. (Our plugin is most likely added first, and setup also called first, which is why the default works fine)

What we're trying to achieve is storing all events in the storage provider, until a user gives consent, instead of simply dropping events.
If the user eventually consents to tracking, we'd like to push the whole queue at that point in time.

Opting out doesn't quite do what we want, but the offline variable would.

Essentially we just want to pause sending for a while.

Mercy811:
Hi @Multiply, 
> What we're trying to achieve is storing all events in the storage provider, until a user gives consent, instead of simply dropping events.
That sounds like another feature request. But glad to hear that we are unblocked now.

cgibson-emesent:
Is there going to be react-native support for offline mode when the network drops out and automatically start sending once the connection returns? I would have expected that if the attempt to send the event fails due to a network error, it's not removed from the _unset array in the local storage. I'm a bit puzzled as to what the purpose of the `offline` config flag is as you can then never send events.

Mercy811:
Hi @Multiply, I created a separate [issue](https://github.com/amplitude/Amplitude-TypeScript/issues/654) to track requests on `setOffline()`. Could you leave a comment with why you need this feature there? Thanks.

Mercy811:
Hi @cgibson-emesent, I also created a separate [issue](https://github.com/amplitude/Amplitude-TypeScript/issues/655) to track requests on supporting offline mode for React Native SDK. Please leave a comment there to help us better track customer requests.

PR #653, opened 2024-01-27
feat: session seplay debug mode

jackson-amplitude:
### Summary
Adds an optional configuration that turns on debug mode. 

When debug mode is turned on, additional debug information is captured in the `[Amplitude] Session Replay Debug` event property. Currently it only contains the hash of the app, can capture additional information such as errors in the future.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -18,8 +18,9 @@ import {
   SESSION_REPLAY_SERVER_URL,
   STORAGE_PREFIX,
   defaultSessionStore,
+  SESSION_REPLAY_DEBUG_PROPERTY,
 } from './constants';
-import { isSessionInSample, maskInputFn, getCurrentUrl, generateSessionReplayId } from './helpers';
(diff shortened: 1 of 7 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -103,6 +104,16 @@ export class SessionReplay implements AmplitudeSessionReplay {
     this.recordEvents();
   }
 
+  getSessionReplayDebugPropertyValue() {
+    let apiKeyHash = '';
+    if (this.config) {
+      apiKeyHash = generateHashCode(this.config.apiKey).toString();
```
bravecod, line 110: QQ: Is it necessary to convert it to a string value?

bravecod (commented):
Lgtm!

jxiwang (commented):
LGTM

packages/session-replay-browser/src/session-replay.ts
jxiwang: I think the hash can be negative

jackson-amplitude: the hash code returns an int, a string value is more generic.

PR #660, opened 2024-02-09
fix(deta plugin): type and error issues

liuyang1520:
### Summary
fix(deta plugin): type and error issues

- adjust the `@amplitude/analytics-types` version based on user installed version, so it won't display the type error like:
```
Argument of type 'BrowserEnrichmentPlugin' is not assignable to parameter of type 'Plugin<BrowserClient, BrowserConfig>'.
```
- in the case that `<html>` document doesn't have the `<body>` yet, it runs the scripts (sync) inside `<head>`, we mount the mutation observer later after page load finishes
- in some cases, `finder` might throw error, e.g., if an element is removed from the document tree, we fallback to use a simple strategy to get the selector

### Checklist

### DIFF
--- a/packages/plugin-default-event-tracking-advanced-browser/src/helpers.ts
+++ b/packages/plugin-default-event-tracking-advanced-browser/src/helpers.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-restricted-globals */
 import { finder } from './libs/finder';
 import * as constants from './constants';
+import { Logger } from '@amplitude/analytics-types';
 
 const SENTITIVE_TAGS = ['input', 'select', 'textarea'];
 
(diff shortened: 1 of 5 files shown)


packages/plugin-default-event-tracking-advanced-browser/src/helpers.ts
```diff
@@ -57,6 +58,39 @@ export const getText = (element: Element): string => {
...
 
+export const getSelector = (element: Element, logger?: Logger): string => {
+  let selector = '';
+  try {
+    selector = finder(element, {
+      className: (name: string) => name !== constants.AMPLITUDE_VISUAL_TAGGING_HIGHLIGHT_CLASS,
+    });
+    return selector;
+  } catch (error) {
+    if (logger && logger.warn) {
```
justin-fiedler, line 69: Why the check for logger.warn here?

justin-fiedler (approved):
Thanks for the fixes @liuyang1520!

packages/plugin-default-event-tracking-advanced-browser/src/helpers.ts
liuyang1520: Thanks for reviewing! Safe to remove, updated!

PR #661, opened 2024-02-12
feat(deta): pass the on track message for instrumentations

liuyang1520:
### Summary
feat(deta): pass the on track message for instrumentations

Also refactor the types and export, so we can use these types in Amplitude side potentially.

### Checklist

### DIFF
--- a/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
+++ b/packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
@@ -246,8 +246,7 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
         });
       }
     };
-    const allElements = querySelectUniqueElements(document.body, cssSelectorAllowlist);
-    allElements.forEach(addListener);
+
     if (typeof MutationObserver !== 'undefined') {
       observer = new MutationObserver((mutations) => {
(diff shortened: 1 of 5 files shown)


Mercy811 (approved):
LGTM Thanks Marvin

liuyang1520:
@Mercy811 , thank you for reviewing! Sorry, I just pushed another quick fix for the `document.body` issue. Could you please take another look at this one? Thanks!

packages/plugin-default-event-tracking-advanced-browser/src/default-event-tracking-advanced-plugin.ts
```diff
@@ -259,20 +257,27 @@ export const defaultEventTrackingAdvancedPlugin = (options: Options = {}): Brows
...
-        observer.observe(document.body, {
+    }
+    if (document.body) {
+      const allElements = querySelectUniqueElements(document.body, cssSelectorAllowlist);
+      allElements.forEach(addListener);
+      /* istanbul ignore next */
+      observer?.observe(document.body, {
+        subtree: true,
+        childList: true,
+      });
```
Mercy811, line 268: Nit: put these block into a function so that it can be reused in the else condition.

liuyang1520: Updated, thanks!

Issue #500, opened 2023-07-20
TypeError: globalScope.addEventListener is not a function


nabusman:
## Expected Behavior
Start the dev server without error

## Current Behavior
After installing I tried running `npm run dev` and got the following error:

TypeError: globalScope.addEventListener is not a function
    at Object.<anonymous> (/.../node_modules/@amplitude/plugin-page-view-tracking-browser/src/page-view-tracking.ts:78:21)
    at step (/.../node_modules/@amplitude/plugin-page-view-tracking-browser/node_modules/tslib/tslib.js:195:27)
    at Object.next (/.../node_modules/@amplitude/plugin-page-view-tracking-browser/node_modules/tslib/tslib.js:176:57)
    at /.../node_modules/@amplitude/plugin-page-view-tracking-browser/node_modules/tslib/tslib.js:169:75
    at new Promise (<anonymous>)
    at Object.__awaiter (/.../node_modules/@amplitude/plugin-page-view-tracking-browser/node_modules/tslib/tslib.js:165:16)
    at Object.setup (/.../node_modules/@amplitude/plugin-page-view-tracking-browser/src/page-view-tracking.ts:71:63)
    at Timeline.<anonymous> (/.../node_modules/@amplitude/analytics-core/src/timeline.ts:28:23)
    at step (/.../node_modules/@amplitude/analytics-core/node_modules/tslib/tslib.js:195:27)
    at Object.next (/.../node_modules/@amplitude/analytics-core/node_modules/tslib/tslib.js:176:57)

## Steps to Reproduce
1. `npm install @amplitude/analytics-browser`
2. `npm run dev`

## Environment
My dependencies in package.json:
  "@amplitude/analytics-browser": "^2.1.2",
  "next": "12.1.6",
  "nextjs-cors": "^2.1.2",
  "ngrok": "^4.3.1",
  "react": "^18.1.0",
  "react-dom": "18.1.0",
  "react-markdown": "^8.0.7",
  "sharp": "^0.30.7",
  "swr": "^1.3.0"

Mercy811:
Hi @nabusman, for NextJS we usually recommend using the Amplitude [Node SDK](https://www.docs.developers.amplitude.com/data/sdks/typescript-node/). This is because NextJS code may be run on both on the client (browser) and the server (SSR).

Hope this helps and let me know if it doesn't!

glung:
@Mercy811 I am also facing this issue. And I have some questions too.

[...]

PR #662, opened 2024-02-13
fix: should not install page view tracking plugin based on config

Mercy811:
### Summary

Workaround to https://github.com/amplitude/Amplitude-TypeScript/issues/500 where `TypeError: globalScope.addEventListener is not a function` happens in Next.js application. This PR will prevent page view tracking plugin from installing based on config when `config.defaultTracking = false` or `config.defaultTracking.pageViews = false`

### Checklist


justin-fiedler (approved):
Thanks @Mercy811

qingzhuozhen (approved):

PR #106, opened 2022-05-19
feat: adds support for multiple instance creation

kevinpagtakhan:
### Summary

Adds support for multiple instance creation. Each instance has its own API_KEY, plugins, configurations.

```typescript
// this is the default instance
amplitude.init('asdf');
amplitude.track('event-1');

// this is the second instance
const amp2 = amplitude.createInstance();
amp2.init('qwerty', undefined, {
  serverUrl: 'https://domain.com',
});
amp2.track('event-2');
```

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:  No

PR #673, opened 2024-03-01
fix: fix event flush process

yuhao900914:
### Summary
The issue was:

[In the setup](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/plugins/destination.ts#L62), we get all events from the storage, put them into a local storage and clean the cookies storage. 
However, some customers, as noted in [Jira 1](https://amplitude.atlassian.net/browse/AMP-93344) and [Jira 2](https://amplitude.atlassian.net/browse/AMP-93297), are missing session start/session end events when redirecting to another page. After investigation, we found that the event_id has been incremented, but the server did not receive this event. We think the original logic might have caused the events to be dropped after we cleared the storage.

Current logic:
While adding an event to the queue, we saved the event to the storage. In line #106.
After fulfillRequest, we removed the events from the storage. In line #309.

### Checklist

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -303,24 +303,28 @@ export class Destination implements DestinationPlugin {
   }
 
   /**
+   * This is called on
+   * 1) new events are added to queue; or
+   * 2) response comes back for a request
+   *
    * update the event storage
    */
-  async updateEventStorage(filterList: Context[], newContext?: Context[]) {
+  async updateEventStorage(eventsToRemove: Context[], eventsToAdd?: Context[]) {
     if (!this.config.storageProvider) {
       return;
     }
 
-    const filterEventInsertIdSet = filterList.reduce((filtered, context) => {
+    const filterEventInsertIdSet = eventsToRemove.reduce((filtered, context) => {
       if (context.event.insert_id) {
         filtered.add(context.event.insert_id);
       }
       return filtered;
     }, new Set<string>());
 
-    const savedEvent = await this.config.storageProvider.get(this.storageKey);
-    const updatedEvents: Event[] = newContext?.map((context) => context.event) || [];
+    const savedEvents = await this.config.storageProvider.get(this.storageKey);
+    const updatedEvents: Event[] = eventsToAdd?.map((context) => context.event) || [];
 
-    savedEvent?.forEach((event) => {
+    savedEvents?.forEach((event) => {
       if (event.insert_id && !filterEventInsertIdSet.has(event.insert_id)) {
         updatedEvents.push(event);
       }
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -53,6 +53,7 @@ export class Destination implements DestinationPlugin {
   config: Config;
   private scheduled: ReturnType<typeof setTimeout> | null = null;
   queue: Context[] = [];
+  operationQueue: Promise<any> = Promise.resolve();
 
   async setup(config: Config): Promise<undefined> {
     this.config = config;
@@ -79,21 +80,15 @@ export class Destination implements DestinationPlugin {
   }
 
   addToQueue(...list: Context[]) {
-    const dropEventSet = new Set<string>();
     const tryable = list.filter((context) => {
       if (context.attempts < this.config.flushMaxRetries) {
         context.attempts += 1;
 
         return true;
       }
-      //Remove the events more than max retries from the storage.
-      if (context.event.insert_id) {
-        dropEventSet.add(context.event.insert_id);
-      }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
     });
-    void this.filterEvent(dropEventSet);
 
     tryable.forEach((context) => {
       this.queue = this.queue.concat(context);
@@ -217,14 +212,6 @@ export class Destination implements DestinationPlugin {
 
   handleSuccessResponse(res: SuccessResponse, list: Context[]) {
     this.fulfillRequest(list, res.statusCode, SUCCESS_MESSAGE);
-    const dropEventSet = list.reduce(function (filtered, context) {
-      if (context.event.insert_id) {
-        filtered.add(context.event.insert_id);
-      }
-      return filtered;
-    }, new Set<string>());
-
-    void this.filterEvent(dropEventSet);
   }
 
   handleInvalidResponse(res: InvalidResponse, list: Context[]) {
@@ -240,13 +227,9 @@ export class Destination implements DestinationPlugin {
       ...res.body.silencedEvents,
     ].flat();
     const dropIndexSet = new Set(dropIndex);
-    const dropEventSet = new Set<string>();
 
     const retry = list.filter((context, index) => {
       if (dropIndexSet.has(index)) {
-        if (context.event.insert_id) {
-          dropEventSet.add(context.event.insert_id);
-        }
         this.fulfillRequest([context], res.statusCode, res.body.error);
         return;
       }
@@ -258,13 +241,11 @@ export class Destination implements DestinationPlugin {
       this.config.loggerProvider.warn(getResponseBodyString(res));
     }
 
-    void this.filterEvent(dropEventSet);
     this.addToQueue(...retry);
   }
 
   handlePayloadTooLargeResponse(res: PayloadTooLargeResponse, list: Context[]) {
     if (list.length === 1) {
-      void this.filterEvent(new Set(list[0].event.event_type));
       this.fulfillRequest(list, res.statusCode, res.body.error);
       return;
     }
@@ -283,16 +264,12 @@ export class Destination implements DestinationPlugin {
     const dropUserIdsSet = new Set(dropUserIds);
     const dropDeviceIdsSet = new Set(dropDeviceIds);
     const throttledIndexSet = new Set(throttledIndex);
-    const dropEventSet = new Set<string>();
 
     const retry = list.filter((context, index) => {
       if (
         (context.event.user_id && dropUserIdsSet.has(context.event.user_id)) ||
         (context.event.device_id && dropDeviceIdsSet.has(context.event.device_id))
       ) {
-        if (context.event.insert_id) {
-          dropEventSet.add(context.event.insert_id);
-        }
         this.fulfillRequest([context], res.statusCode, res.body.error);
         return;
       }
@@ -307,7 +284,6 @@ export class Destination implements DestinationPlugin {
       this.config.loggerProvider.warn(getResponseBodyString(res));
     }
 
-    void this.filterEvent(dropEventSet);
     this.addToQueue(...retry);
   }
 
@@ -322,17 +298,25 @@ export class Destination implements DestinationPlugin {
 
   fulfillRequest(list: Context[], code: number, message: string) {
     list.forEach((context) => context.callback(buildResult(context.event, code, message)));
+    void this.filterEvent(list);
   }
 
   /**
    * update the event storage
    */
-  async filterEvent(dropEventSet: Set<string>) {
+  async filterEvent(list: Context[]) {
+    const dropEventSet = list.reduce((filtered, context) => {
+      if (context.event.insert_id) {
+        filtered.add(context.event.insert_id);
+      }
+      return filtered;
+    }, new Set<string>());
     if (!this.config.storageProvider) {
       return;
     }
     const savedEvent = await this.config.storageProvider.get(this.storageKey);
     const reTriedEvents: Event[] = [];
+
     if (savedEvent) {
       savedEvent.filter((event) => {
         if (!dropEventSet.has(event.event_type)) {
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -87,11 +87,13 @@ export class Destination implements DestinationPlugin {
         return true;
       }
       //Remove the events more than max retries from the storage.
-      dropEventSet.add(context.event.event_type);
+      if (context.event.insert_id) {
+        dropEventSet.add(context.event.insert_id);
+      }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
     });
-    void this.updateEvent(dropEventSet);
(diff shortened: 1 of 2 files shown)


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -59,7 +59,6 @@ export class Destination implements DestinationPlugin {
 
     this.storageKey = `${STORAGE_PREFIX}_${this.config.apiKey.substring(0, 10)}`;
     const unsent = await this.config.storageProvider?.get(this.storageKey);
-    this.saveEvents(); // sets storage to '[]'
```
yuhao900914, line 62: In the Destination plugin, we originally clear the storage and copy all events into a local variable queue(line 62). And when we get a response, we put them back into the queue for retry and save the events back to the storage(line 105).

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -80,14 +79,19 @@ export class Destination implements DestinationPlugin {
...
         context.attempts += 1;
+
         return true;
       }
+      //Remove the events more than max retries from the storage.
+      dropEventSet.add(context.event.event_type);
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
     });
+    void this.updateEvent(dropEventSet);
```
yuhao900914, line 94: CASE 1: when an event exceeds the maximum retry limit, it should be deleted from the storage.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -213,6 +215,12 @@ export class Destination implements DestinationPlugin {
 
   handleSuccessResponse(res: SuccessResponse, list: Context[]) {
     this.fulfillRequest(list, res.statusCode, SUCCESS_MESSAGE);
+    const dropEventSet = list.reduce(function (filtered, context) {
```
yuhao900914, line 218: CASE 2: when events have been sent out successfully.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -241,11 +251,14 @@ export class Destination implements DestinationPlugin {
       // log intermediate event status before retry
       this.config.loggerProvider.warn(getResponseBodyString(res));
     }
+
+    void this.updateEvent(dropEventSet);
```
yuhao900914, line 255: CASE 3: when events have an invalid response.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -264,12 +277,14 @@ export class Destination implements DestinationPlugin {
...
     const dropDeviceIdsSet = new Set(dropDeviceIds);
     const throttledIndexSet = new Set(throttledIndex);
+    const dropEventSet = new Set<string>();
 
     const retry = list.filter((context, index) => {
       if (
         (context.event.user_id && dropUserIdsSet.has(context.event.user_id)) ||
         (context.event.device_id && dropDeviceIdsSet.has(context.event.device_id))
       ) {
+        dropEventSet.add(context.event.event_type);
```
yuhao900914, line 287: CASE 4: when events are filtered in the RateLimitResponse

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +313,25 @@ export class Destination implements DestinationPlugin {
...
 
   /**
-   * Saves events to storage
-   * This is called on
-   * 1) new events are added to queue; or
-   * 2) response comes back for a request
+   * update the event storage
    */
-  saveEvents() {
+  async updateEvent(dropEventSet: Set<string>) {
```
izaaz, line 322: This function name feels a little misleading. Looking at the function i'd expect it to take a list of events to save and not drop. Maybe we can call it `filterEvents`?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +313,25 @@ export class Destination implements DestinationPlugin {
...
     if (!this.config.storageProvider) {
       return;
     }
-    const events = Array.from(this.queue.map((context) => context.event));
-    void this.config.storageProvider.set(this.storageKey, events);
+    const savedEvent = await this.config.storageProvider.get(this.storageKey);
```
izaaz, line 326: How do we handle multiple threads trying to write to the storage at the same time? What happens in a case like this
1. We get the events to be saved
2. We are filtering events
3. While we are filtering events, another thread adds a new event to storage
4. We are done filtering events and we now save the retriedEvents, which does not include the new event that was saved in 3

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +313,25 @@ export class Destination implements DestinationPlugin {
...
+    const reTriedEvents: Event[] = [];
+    if (savedEvent) {
+      savedEvent.filter((event) => {
+        if (!dropEventSet.has(event.event_type)) {
```
izaaz, line 330: By doing this, won't we drop all events of that event_type? Shouldn't we be filtering by insert ids instead?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +313,25 @@ export class Destination implements DestinationPlugin {
...
+          reTriedEvents.push(event);
+        }
+      });
+    }
+    void this.config.storageProvider.set(this.storageKey, reTriedEvents);
```
qingzhuozhen, line 335: Seems fine, just to confirm there wouldn't be any event go to storage after we read and before write?

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Good call. Originally use insert_id, but it's optional, so changed to event_type. But this is the right concern, and the insert_id is be assigned in the context plugin, there must be a value in the destination plugin. Will update this.

packages/analytics-core/src/plugins/destination.ts
Mercy811: how about changing `dropEventSet` to `sentEventSet`? As they are removed from the storage not because of being dropped but being sent successfully.

packages/analytics-core/src/plugins/destination.ts
Mercy811: When invalid response, we drop events 
- [When body has missing field or invalid api key](https://github.com/amplitude/Amplitude-TypeScript/blob/e8216bc65399106214cc5c0ae4d7e8d053029af4/packages/analytics-core/src/plugins/destination.ts#L219-L222)
- [When events have invalid or missing fields](https://github.com/amplitude/Amplitude-TypeScript/blob/e8216bc65399106214cc5c0ae4d7e8d053029af4/packages/analytics-core/src/plugins/destination.ts#L233-L236) . 
Both cases will return early and not execute `updateEvent()` here.

If code is executed here, events should be retried.

packages/analytics-core/src/plugins/destination.ts
Mercy811: It seems that `updateEvent()` can be merged into `fulfillRequest()`. As they are always called together.

Mercy811 (changes requested):

packages/analytics-core/src/plugins/destination.ts
yuhao900914: it's not only the event that has been successfully sent. It's also including the event which missing some info. I think `filterEventSet` make more sense?

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Added a lock to prevent the concurrency issue. Please check.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Added a lock to prevent the concurrency issue. Please check.

qingzhuozhen: I guess this lock is not preventing the saving events though. Can we confirm if this is a concern in JS/TS? I recall JS is single threaded so just to confirm if we still have this race condition in JS.

Mercy811 (changes requested):
Could you put a summary of what the current logic of adding and removing events from storage? 

The biggest problem of current logic is that we fetch all events from storage, and if it's crashed, all events will not be set to storage and thus be lost. To fix this:
1. Storage events to storage every time when tracked. 
2. Remove events from storage when it's successfully sent or dropped. 

Please let me know where we do #1. I might miss it. So far, this PR achieve #2 but not #1. So newly tracked events will never be in storage and will be lost.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,42 @@ export class Destination implements DestinationPlugin {
...
 
   /**
-   * Saves events to storage
-   * This is called on
-   * 1) new events are added to queue; or
-   * 2) response comes back for a request
+   * update the event storage
    */
-  saveEvents() {
+  async filterEvent(list: Context[]) {
```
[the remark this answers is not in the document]
yuhao900914: It's not only for the successfully sent events, it's also include the dropped event.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,42 @@ export class Destination implements DestinationPlugin {
...
     }
-    const events = Array.from(this.queue.map((context) => context.event));
-    void this.config.storageProvider.set(this.storageKey, events);
+
+    while (this.lock) {
+      await new Promise((resolve) => setTimeout(resolve, 100)); // Wait for 100ms before retrying
+    }
+
+    this.lock = true;
+    const filterEventSet = list.reduce((filtered, context) => {
```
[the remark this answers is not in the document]
yuhao900914: changed to `filterEventInsertIdSet` because we have both inserted and eventId.

packages/analytics-core/src/plugins/destination.ts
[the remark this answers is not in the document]
yuhao900914: I like the updateStorage.

packages/analytics-core/test/plugins/destination.test.ts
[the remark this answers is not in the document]
yuhao900914: Sync offline, we already have the seprate storage operation test, we can skip this.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,42 @@ export class Destination implements DestinationPlugin {
...
-  saveEvents() {
+  async filterEvent(list: Context[]) {
     if (!this.config.storageProvider) {
       return;
     }
-    const events = Array.from(this.queue.map((context) => context.event));
-    void this.config.storageProvider.set(this.storageKey, events);
+
+    while (this.lock) {
+      await new Promise((resolve) => setTimeout(resolve, 100)); // Wait for 100ms before retrying
```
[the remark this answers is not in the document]
yuhao900914: Removed this logic.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Javascript has a single thread, it executes one instruction/sequence at a time. So there is no concept of concurrent execution or parallelism.
Asynchronous means the program can work or execute another sequence of instructions while one of the sequences waits. It is responsive and lets the other instructions manage instead of making them starve. 
In this case, the code in `updateEventStorage`, between the await this.config.storageProvider.get(this.storageKey) and await this.config.storageProvider.set(this.storageKey, xxx) won't be interrupted.

yuhao900914: Followup. The logic for the legacy SDK is as following:
In init, we load the unsent event from the storage, and we save them in a local queue (array).
Before sending an event,  we save the event in the local queue, overwrite the storage based on the local queue. 
While there have a response, we update the local queue, and update the storage with the local queue.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +298,33 @@ export class Destination implements DestinationPlugin {
...
-    void this.config.storageProvider.set(this.storageKey, events);
+
+    const filterEventInsertIdSet = filterList.reduce((filtered, context) => {
+      if (context.event.insert_id) {
+        filtered.add(context.event.insert_id);
+      }
+      return filtered;
+    }, new Set<string>());
+
+    const savedEvent = await this.config.storageProvider.get(this.storageKey);
```
justin-fiedler, line 320: nit: 
```suggestion
    const savedEvents = await this.config.storageProvider.get(this.storageKey);
```

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +298,33 @@ export class Destination implements DestinationPlugin {
...
 
   /**
-   * Saves events to storage
-   * This is called on
-   * 1) new events are added to queue; or
-   * 2) response comes back for a request
+   * update the event storage
    */
-  saveEvents() {
+  async updateEventStorage(filterList: Context[], newContext?: Context[]) {
```
justin-fiedler, line 308: nit: given the similarity of these params, and both are optional based on usage, it could be could to make these named to avoid accidentally setting the wrong value and prevent needing to pass in an empty filter list like on line 105.

```suggestion
  async updateEventStorage({ filterList?: Context[], newContext?: Context[] }) {
```

kevinpagtakhan, line 308: imo the names are confusing as it doesn't tell what they are used for. can we rename `filterList` to `eventsToRemove and `newContext` to `eventsToAdd`? if I'm understanding it correctly

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +298,33 @@ export class Destination implements DestinationPlugin {
...
     list.forEach((context) => context.callback(buildResult(context.event, code, message)));
+    void this.updateEventStorage(list);
   }
 
   /**
-   * Saves events to storage
-   * This is called on
-   * 1) new events are added to queue; or
-   * 2) response comes back for a request
+   * update the event storage
```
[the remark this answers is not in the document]
yuhao900914: it's actually the same, I added the description back.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Sync offline, updated the parameters to eventsToRemove and eventsToAdd instead of making the input an object.

PR #675, opened 2024-03-07
feat(session replay): improve initialization and event tagging

jackson-amplitude:
### Summary
This PR improves the general of the SR plugin by: 
1. By synchronizing Session Id between Browser SDK and SR plugin on each event. In the process also getting rid of the need to require start session event. 
2. Updates getSessionReplayProperties to ignore the focus state handler to enhance tagging of events. 

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -68,6 +68,24 @@ export class SessionReplayPlugin implements DestinationPlugin {
     this.client = client;
     this.config = config;
 
+    if (!this.options.disableSessionTracking) {
+      if (typeof config.defaultTracking === 'boolean') {
+        if (config.defaultTracking === false) {
+          config.defaultTracking = {
+            pageViews: false,
+            formInteractions: false,
+            fileDownloads: false,
+            sessions: true,
+          };
+        }
+      } else {
+        config.defaultTracking = {
+          ...config.defaultTracking,
+          sessions: true,
+        };
+      }
+    }
+
     await sessionReplay.init(config.apiKey, {
       instanceName: this.config.instanceName,
       deviceId: this.config.deviceId,
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -55,6 +55,10 @@ export class SessionReplayPlugin implements DestinationPlugin {
 
   constructor(options?: SessionReplayOptions) {
     this.options = { ...options };
+    // The user did not explicitly configure forceSessionTracking to false, default to true.
+    if (this.options.forceSessionTracking !== false) {
+      this.options.forceSessionTracking = true;
+    }
   }
 
   async setup(config: BrowserConfig, client?: BrowserClient) {
@@ -68,7 +72,7 @@ export class SessionReplayPlugin implements DestinationPlugin {
     this.client = client;
     this.config = config;
 
-    if (!this.options.disableSessionTracking) {
+    if (this.options.forceSessionTracking) {
       if (typeof config.defaultTracking === 'boolean') {
         if (config.defaultTracking === false) {
           config.defaultTracking = {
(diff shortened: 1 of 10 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -7,21 +7,27 @@ import {
...
+    this.config = _config;
   }
 
   async execute(event: Event) {
-    if (event.event_type === DEFAULT_SESSION_START_EVENT && event.session_id) {
-      sessionReplay.setSessionId(event.session_id);
+    // On event, synchronize the session id to the what's on the browserConfig (source of truth)
+    // Choosing not to read from event object here, concerned about offline/delayed events messing up the state stored
+    // in SR.
+    if (this.config.sessionId && this.config.sessionId !== sessionReplay.getSessionId()) {
```
jackson-amplitude, line 29: @justin-fiedler would like your thoughts on this chunk of initialization logic.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -89,9 +89,10 @@ export class SessionReplay implements AmplitudeSessionReplay {
       this.loggerProvider.error('Session replay init has not been called, cannot set session id.');
       return;
     }
-
-    if (sessionId && this.config.deviceId) {
-      this.config.sessionReplayId = generateSessionReplayId(sessionId, this.config.deviceId);
+    // use a consistent device id.
+    const deviceId = this.getDeviceId();
```
jackson-amplitude, line 93: this is what's used when sending SR events.

packages/plugin-session-replay-browser/src/session-replay.ts
```diff
@@ -62,22 +68,6 @@ export class SessionReplayPlugin implements DestinationPlugin {
     this.client = client;
     this.config = config;
 
-    if (typeof config.defaultTracking === 'boolean') {
```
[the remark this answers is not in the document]
jackson-amplitude: Good points. was thinking it's not a breaking change from an interface perspective, but it does change the underlying behavior. 

There is also feedback from customers that start/end session events is consuming too much event quota.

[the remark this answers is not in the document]
jackson-amplitude: I will add an optional param to turn off session tracking, and keep the current behavior by default.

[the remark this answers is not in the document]
jackson-amplitude: added the optional param!

packages/plugin-session-replay-browser/src/typings/session-replay.ts
```diff
@@ -5,4 +5,5 @@ export interface SessionReplayOptions {
   sampleRate?: number;
   privacyConfig?: SessionReplayPrivacyConfig;
   debugMode?: boolean;
+  disableSessionTracking?: boolean;
```
justin-fiedler, line 8: `disableSessionTracking` is somewhat confusing imo. It makes me think it would actively disable default tracking sessions.

What about `forceSessionTracking` with a default value of true instead?

jackson-amplitude: ooo good points. will make the update in a bit.

jackson-amplitude: hmm took a look at this.. it's actually not as easy to have a default value of true. because the param is part of an dictionary/object that's passed in. Also not a js expert, so maybe there is a way. 

What do you think about `disableForcedSessionTracking`? a bit verbose, but more descriptive.

jackson-amplitude: I guess i can override the value in the constructor, but logic feels brittle.

jackson-amplitude: ok nvm, figured out a clean to do the override in the constructor (was a bit hung up on how undefines/nulls work). the optional param is now `forceSessionTracking` that defaults to true (unless the user explicitly configures it to be false).

justin-fiedler (approved):

jxiwang (approved):

PR #680, opened 2024-03-13
fix(session replay): fix event tagging 

jackson-amplitude:
### Summary
Fixes the event tagging issues from last PR, mainly: 
1. we should not tag events that have mismatching session id with the current session (these are likely late/out of session events) 
2. only tag events when we are in focus/recording. The focus handler is only ignored when in debug mode, this can be used for people who are developing with the console open (which makes the page out of focus). 

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/test/session-replay.test.ts
+++ b/packages/plugin-session-replay-browser/test/session-replay.test.ts
@@ -245,11 +245,11 @@ describe('SessionReplayPlugin', () => {
       });
     });
 
-    test('should add event property for [Amplitude] Session Recorded', async () => {
+    test('should add event property for [Amplitude] Session Replay ID', async () => {
       const sessionReplay = sessionReplayPlugin();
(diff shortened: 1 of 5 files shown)


packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -245,18 +245,48 @@ describe('SessionReplayPlugin', () => {
       });
     });
 
-    test('should add event property for [Amplitude] Session Recorded', async () => {
+    test('should add event property for [Amplitude] Session Replay ID', async () => {
       const sessionReplay = sessionReplayPlugin();
       await sessionReplay.setup(mockConfig, mockAmplitude);
       getSessionReplayProperties.mockReturnValueOnce({
-        '[Amplitude] Session Recorded': true,
```
kwalker3690, line 252: Ooh, why was this passing before? Didn't this get changed in v1?

kwalker3690 (approved):
Looks great, thanks for the quick thinking on this!

PR #684, opened 2024-03-19
feat: pass iselementselectable to visual selector

liuyang1520:
### Summary
Related to https://github.com/amplitude/visual-tagging-selector/pull/15.

This PR:
- get `actionType` from the message
- pass the `isElementSelectable` to visual tagging selector

### Checklist

### DIFF
--- a/packages/plugin-default-event-tracking-advanced-browser/src/libs/messenger.ts
+++ b/packages/plugin-default-event-tracking-advanced-browser/src/libs/messenger.ts
@@ -7,6 +7,7 @@ import {
 } from '../constants';
 import { asyncLoadScript, getEventTagProps } from '../helpers';
 import { Logger } from '@amplitude/analytics-types';
+import { ActionType } from '../typings/default-event-tracking-advanced';
 
 export interface Messenger {
   logger?: Logger;
@@ -23,6 +24,10 @@ export type Action =
   | 'element-selected'
   | 'track-selector-mode-changed'
   | 'track-selector-moved';
+
+interface InitializeVisualTaggingSelectorData {
+  actionType: ActionType;
+}
 
 interface ElementSelectedData {
   '[Amplitude] Element Tag'?: string;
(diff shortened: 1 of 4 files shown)


packages/plugin-default-event-tracking-advanced-browser/src/libs/messenger.ts
```diff
@@ -91,14 +105,21 @@ export class WindowMessenger implements Messenger {
...
+        const actionData = eventData?.data as InitializeVisualTaggingSelectorData;
         asyncLoadScript(AMPLITUDE_VISUAL_TAGGING_SELECTOR_SCRIPT_URL)
           .then(() => {
             // eslint-disable-next-line
             amplitudeVisualTaggingSelectorInstance = (window as any)?.amplitudeVisualTaggingSelector?.({
-              visualHighlightClass: AMPLITUDE_VISUAL_TAGGING_HIGHLIGHT_CLASS,
               getEventTagProps,
+              isElementSelectable: (element: Element) => {
+                if (isElementSelectable) {
+                  return isElementSelectable(actionData?.actionType || 'click', element);
```
yuhao900914, line 116: use ActionType.click instead?

packages/plugin-default-event-tracking-advanced-browser/src/libs/messenger.ts
```diff
@@ -91,14 +105,21 @@ export class WindowMessenger implements Messenger {
...
+                }
+                return true;
```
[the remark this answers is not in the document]
liuyang1520: Thanks for reviewing! This is to get the same behavior as we have right now, if `isElementSelectable`, then every element is selectable, like we can select `div`, `svg`, and etc. now.

packages/plugin-default-event-tracking-advanced-browser/src/libs/messenger.ts
liuyang1520: The string literal union type here doesn't work like this, intended to use string literal type instead of enum as it is easier to deal with.

yuhao900914 (approved):

PR #689, opened 2024-03-25
fix: fix event drop issue

yuhao900914:
### Summary

So this PR should

- Fix the original losing events problem which has been reverted https://github.com/amplitude/Amplitude-TypeScript/pull/673
- Stop accumulating events which are caused by the https://github.com/amplitude/Amplitude-TypeScript/pull/673

### Checklist

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -95,8 +95,12 @@ export class Destination implements DestinationPlugin {
     });
   }
 
-  scheduleTryable(list: Context[]) {
+  scheduleTryable(list: Context[], shouldAddToQueue = false) {
     list.forEach((context) => {
+      // Only need to concat the queue for the first time
+      if (shouldAddToQueue) {
+        this.queue = this.queue.concat(context);
+      }
       if (context.timeout === 0) {
         this.schedule(this.config.flushIntervalMillis);
         return;
@@ -111,20 +115,7 @@ export class Destination implements DestinationPlugin {
 
   addToQueue(...list: Context[]) {
     const tryable = this.getTryableList(list);
-
-    tryable.forEach((context) => {
-      this.queue = this.queue.concat(context);
-      if (context.timeout === 0) {
-        this.schedule(this.config.flushIntervalMillis);
-        return;
-      }
-
-      setTimeout(() => {
-        context.timeout = 0;
-        this.schedule(0);
-      }, context.timeout);
-    });
-
+    this.scheduleTryable(tryable, true);
     this.saveEvents();
   }
 
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -67,6 +67,7 @@ export class Destination implements DestinationPlugin {
   }
 
   execute(event: Event): Promise<Result> {
+    //console.log("testt");
     return new Promise((resolve) => {
       const context = {
         event,
@@ -78,8 +79,8 @@ export class Destination implements DestinationPlugin {
     });
   }
 
-  addToQueue(...list: Context[]) {
-    const tryable = list.filter((context) => {
+  tryableList(list: Context[]) {
+    return list.filter((context) => {
       if (context.attempts < this.config.flushMaxRetries) {
         context.attempts += 1;
         return true;
@@ -87,6 +88,19 @@ export class Destination implements DestinationPlugin {
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
     });
+  }
+
+  scheduleTryable(list: Context[]) {
+    list.forEach((context) => {
+      setTimeout(() => {
+        context.timeout = 0;
+        this.schedule(0);
+      }, context.timeout);
+    });
+  }
+
+  addToQueue(...list: Context[]) {
+    const tryable = this.tryableList(list);
 
     tryable.forEach((context) => {
       this.queue = this.queue.concat(context);
@@ -100,6 +114,7 @@ export class Destination implements DestinationPlugin {
         this.schedule(0);
       }, context.timeout);
     });
+
     this.saveEvents();
   }
 
@@ -124,11 +139,16 @@ export class Destination implements DestinationPlugin {
       return;
     }
 
-    const list = this.queue.filter((context) => context.timeout === 0);
+    const list: Context[] = [];
+    const later: Context[] = [];
+    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
+
     if (this.scheduled) {
       clearTimeout(this.scheduled);
       this.scheduled = null;
     }
+
+    this.scheduleTryable(later);
 
     const batches = chunk(list, this.config.flushQueueSize);
 
@@ -199,6 +219,7 @@ export class Destination implements DestinationPlugin {
       default: {
         // log intermediate event status before retry
         this.config.loggerProvider.warn(`{code: 0, error: "Status '${status}' provided for ${list.length} events"}`);
+        this.handleOtherResponse(list);
         break;
       }
     }
@@ -234,6 +255,9 @@ export class Destination implements DestinationPlugin {
       // log intermediate event status before retry
       this.config.loggerProvider.warn(getResponseBodyString(res));
     }
+
+    const tryable = this.tryableList(retry);
+    this.scheduleTryable(tryable);
   }
 
   handlePayloadTooLargeResponse(res: PayloadTooLargeResponse, list: Context[]) {
@@ -246,9 +270,13 @@ export class Destination implements DestinationPlugin {
     this.config.loggerProvider.warn(getResponseBodyString(res));
 
     this.config.flushQueueSize /= 2;
+
+    const tryable = this.tryableList(list);
+    this.scheduleTryable(tryable);
   }
 
   handleRateLimitResponse(res: RateLimitResponse, list: Context[]) {
+    //console.log(list);
     const dropUserIds = Object.keys(res.body.exceededDailyQuotaUsers);
     const dropDeviceIds = Object.keys(res.body.exceededDailyQuotaDevices);
     const throttledIndex = res.body.throttledEvents;
@@ -274,6 +302,20 @@ export class Destination implements DestinationPlugin {
       // log intermediate event status before retry
       this.config.loggerProvider.warn(getResponseBodyString(res));
     }
+
+    const tryable = this.tryableList(retry);
+    this.scheduleTryable(tryable);
+  }
+
+  handleOtherResponse(list: Context[]) {
+    const later = list.map((context) => {
+      context.timeout = context.attempts * this.retryTimeout;
+      return context;
+    });
+
+    const tryable = this.tryableList(later);
+    this.scheduleTryable(tryable);
+    this.saveEvents();
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
@@ -302,12 +344,9 @@ export class Destination implements DestinationPlugin {
    */
   removeEvents(eventsToRemove: Context[]) {
     this.queue = this.queue.filter(
-      (queuedContext) =>
-        !(
-          queuedContext.event.insert_id &&
-          eventsToRemove.some((context) => context.event.insert_id === queuedContext.event.insert_id)
-        ),
+      (queuedContext) => !eventsToRemove.some((context) => context.event.insert_id === queuedContext.event.insert_id),
     );
+    //console.log(this.queue);
     this.saveEvents();
   }
 }
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -138,7 +138,6 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(list, this.config.flushQueueSize);
     await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
-    // should we wait to
   }
 
   async send(list: Context[], useRetry = true) {
@@ -232,7 +231,6 @@ export class Destination implements DestinationPlugin {
     const retry = list.filter(async (context, index) => {
       if (dropIndexSet.has(index)) {
         this.fulfillRequest([context], res.statusCode, res.body.error);
-        // we didn't await at here.
         return;
       }
       return true;
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -82,7 +82,6 @@ export class Destination implements DestinationPlugin {
     const tryable = list.filter((context) => {
       if (context.attempts < this.config.flushMaxRetries) {
         context.attempts += 1;
-
         return true;
       }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
@@ -228,6 +227,7 @@ export class Destination implements DestinationPlugin {
       ...res.body.silencedEvents,
     ].flat();
     const dropIndexSet = new Set(dropIndex);
+
     const retry = list.filter((context, index) => {
       if (dropIndexSet.has(index)) {
         this.fulfillRequest([context], res.statusCode, res.body.error);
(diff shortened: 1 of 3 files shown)


izaaz:
Currently, in flush, we empty the queue. We should not be emptying the queue there too right?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +296,33 @@ export class Destination implements DestinationPlugin {
...
+  updateEventStorage(eventsToRemove?: Context[]) {
     if (!this.config.storageProvider) {
       return;
     }
-    const events = Array.from(this.queue.map((context) => context.event));
-    void this.config.storageProvider.set(this.storageKey, events);
+
+    const eventsToRemoveInsertIdSet = eventsToRemove?.reduce((filtered, context) => {
+      if (context.event.insert_id) {
+        filtered.add(context.event.insert_id);
```
izaaz, line 317: In the event schema, why is insert_id nullable? Can we make it non-nullable so that we know that insert id is never going to be null or empty?

yuhao900914: I think insert_id in the backend is nullable. Only the event_type and user id or device id are required.

izaaz: Yes, it is nullable in the backend but with the current model, it feels like it can be nullable in the SDK too? If it can be null in the SDK, then using this field for filtering might have issues?

Mercy811 (changes requested):

Mercy811 (commented):
Could you update the PR description🙏? 

This PR is based on main branch with the revert of https://github.com/amplitude/Amplitude-TypeScript/pull/673.  So this PR should 
- Fix the original losing events problem in https://github.com/amplitude/Amplitude-TypeScript/pull/673 
- Drop bad events and update storage to fix https://amplitude.atlassian.net/browse/AMP-96244

Mercy811:
> Currently, in flush, we empty the queue. We should not be emptying the queue there too right?

```diff
  async flush(useRetry = false) {
    // Skip flush if offline
    if (this.config.offline) {
      this.config.loggerProvider.debug('Skipping flush while offline.');
      return;
    }
    const list: Context[] = [];
-    const later: Context[] = [];
-    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
-    this.queue = later;
    if (this.scheduled) {
      clearTimeout(this.scheduled);
      this.scheduled = null;
    }

    const batches = chunk(list, this.config.flushQueueSize);

    await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
  }
```

Mercy811 (commented):
No need to call `callToQueue()` in `handleXxxxReponse()` as `this.queue` contains all events.

yuhao900914:
> Reponse

This should be `const list = this.queue.filter((context) => context.timeout === 0);` instead of update the queue. We only want to send out the events timeout === 0.

yuhao900914:
> > Currently, in flush, we empty the queue. We should not be emptying the queue there too right?
> 
> ```diff
>   async flush(useRetry = false) {
>     // Skip flush if offline
>     if (this.config.offline) {
>       this.config.loggerProvider.debug('Skipping flush while offline.');
>       return;
>     }
>     const list: Context[] = [];
> -    const later: Context[] = [];
> -    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
> -    this.queue = later;
>     if (this.scheduled) {
>       clearTimeout(this.scheduled);
>       this.scheduled = null;
>     }
> 
>     const batches = chunk(list, this.config.flushQueueSize);
> 
>     await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
>   }
> ```

This should be const list = this.queue.filter((context) => context.timeout === 0); instead of update the queue. We only want to send out the events timeout === 0.

izaaz:
> @Mercy811
> 
> > > Currently, in flush, we empty the queue. We should not be emptying the queue there too right?
> > 
> > 
> > ```diff
> >   async flush(useRetry = false) {
> >     // Skip flush if offline
> >     if (this.config.offline) {
> >       this.config.loggerProvider.debug('Skipping flush while offline.');
> >       return;
> >     }
> >     const list: Context[] = [];
> > -    const later: Context[] = [];
> > -    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
> > -    this.queue = later;
> >     if (this.scheduled) {
> >       clearTimeout(this.scheduled);
> >       this.scheduled = null;
> >     }
> > 
> >     const batches = chunk(list, this.config.flushQueueSize);
> > 
> >     await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
> >   }
> > ```
> 
> This should be const list = this.queue.filter((context) => context.timeout === 0); instead of update the queue. We only want to send out the events timeout === 0.

One more thing to note - if there are 5 events in the queue and the first event was throttled, there is no point in sending the remaining 4 since we lose out on the ordered sending criteria. So if a request is throttled, we just need to wait and try again later.

Mercy811:
> @Mercy811
> 
> > > Currently, in flush, we empty the queue. We should not be emptying the queue there too right?
> > 
> > 
> > ```diff
> >   async flush(useRetry = false) {
> >     // Skip flush if offline
> >     if (this.config.offline) {
> >       this.config.loggerProvider.debug('Skipping flush while offline.');
> >       return;
> >     }
> >     const list: Context[] = [];
> > -    const later: Context[] = [];
> > -    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
> > -    this.queue = later;
> >     if (this.scheduled) {
> >       clearTimeout(this.scheduled);
> >       this.scheduled = null;
> >     }
> > 
> >     const batches = chunk(list, this.config.flushQueueSize);
> > 
> >     await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
> >   }
> > ```
> 
> This should be const list = this.queue.filter((context) => context.timeout === 0); instead of update the queue. We only want to send out the events timeout === 0.

Oh yes, we should keep the line to update `list`.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -126,17 +124,14 @@ export class Destination implements DestinationPlugin {
       return;
     }
 
-    const list: Context[] = [];
-    const later: Context[] = [];
-    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
-    this.queue = later;
-
+    const list = this.queue.filter((context) => context.timeout === 0);
```
izaaz, line 127: Will we fix this separately in a different PR where we'll make sure we send data in the right order?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,41 @@ export class Destination implements DestinationPlugin {
...
   /**
-   * Saves events to storage
    * This is called on
    * 1) new events are added to queue; or
    * 2) response comes back for a request
+   *
+   * Update the event storage based on the queue
    */
-  saveEvents() {
+  saveEvent() {
```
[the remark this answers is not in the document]
Mercy811: Updated.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,41 @@ export class Destination implements DestinationPlugin {
...
-    void this.config.storageProvider.set(this.storageKey, events);
+
+    const updatedEvents = this.queue.map((context) => context.event);
+    void this.config.storageProvider.set(this.storageKey, updatedEvents);
+  }
+
+  /**
+   * This is called on response comes back for a request
+   */
+  removeEvent(eventsToRemove?: Context[]) {
```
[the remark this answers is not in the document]
Mercy811: Updated.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,41 @@ export class Destination implements DestinationPlugin {
...
+      if (context.event.insert_id) {
+        filtered.add(context.event.insert_id);
+      }
+      return filtered;
+    }, new Set<string>());
+
+    const updatedEvents = this.queue.map((context) => context.event);
+    if (eventsToRemoveInsertIdSet) {
+      updatedEvents.filter((event) => !(event.insert_id && !eventsToRemoveInsertIdSet.has(event.insert_id)));
+    }
```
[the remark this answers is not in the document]
Mercy811: Updated.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -254,10 +269,13 @@ export class Destination implements DestinationPlugin {
...
 
     this.config.flushQueueSize /= 2;
-    this.addToQueue(...list);
+
+    const tryable = this.tryableList(list);
+    this.scheduleTryable(tryable);
   }
 
   handleRateLimitResponse(res: RateLimitResponse, list: Context[]) {
+    //console.log(list);
```
Mercy811, line 278: nit: delete it

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -284,34 +302,50 @@ export class Destination implements DestinationPlugin {
...
-      }),
-    );
+    const later = list.map((context) => {
+      context.timeout = context.attempts * this.retryTimeout;
+      return context;
+    });
+
+    const tryable = this.tryableList(later);
+    this.scheduleTryable(tryable);
+    this.saveEvents();
```
Mercy811, line 317: Seems no need to call `saveEvents()` here like other response handlers.

Mercy811 (changes requested):

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -79,15 +78,28 @@ export class Destination implements DestinationPlugin {
...
+  }
+
+  scheduleTryable(list: Context[]) {
+    list.forEach((context) => {
+      setTimeout(() => {
+        context.timeout = 0;
+        this.schedule(0);
+      }, context.timeout);
+    });
+  }
```
Mercy811, line 99:
```suggestion
  scheduleTryable(list: Context[]) {
  // Check context.timeou
  // Because timeout is unchanged when payload too large for example
  // if (context.timeout) == 0, retry this in config.flushIntervalMillis
  
    list.forEach((context) => {
      setTimeout(() => {
        context.timeout = 0;
        this.schedule(0);
      }, context.timeout);
    });
  }
```

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Will assign the UUID as insert_id if it's null before sending out.

packages/analytics-core/src/plugins/destination.ts
Mercy811: Deleted in https://github.com/amplitude/Amplitude-TypeScript/pull/689/commits/3b954b4c61f9715be3dfff3ed6fb7e9bdc7d4d35

packages/analytics-core/src/plugins/destination.ts
Mercy811: Deleted in https://github.com/amplitude/Amplitude-TypeScript/pull/689/commits/3b954b4c61f9715be3dfff3ed6fb7e9bdc7d4d35

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -79,15 +84,33 @@ export class Destination implements DestinationPlugin {
...
       if (context.attempts < this.config.flushMaxRetries) {
         context.attempts += 1;
         return true;
       }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
     });
+  }
+
+  scheduleTryable(list: Context[]) {
```
Mercy811, line 98: Can we reuse this function in `addToQueue()` for cleanliness?

packages/analytics-core/src/plugins/destination.ts
Mercy811: Updated.

packages/analytics-core/test/plugins/destination.test.ts
```diff
@@ -236,7 +270,6 @@ describe('destination', () => {
       ];
       const send = jest.spyOn(destination, 'send').mockReturnValueOnce(Promise.resolve());
       const result = await destination.flush();
-      expect(destination.queue).toEqual([]);
```
Mercy811, line 239: Why this and another one below are deleted?

packages/analytics-core/test/plugins/destination.test.ts
```diff
@@ -560,16 +599,48 @@ describe('destination', () => {
...
       const set = jest.spyOn(destination.config.storageProvider, 'set').mockResolvedValueOnce(undefined);
-      destination.saveEvents();
+      destination.queue = eventsToAdd;
+      const eventsToRemove = [eventsToAdd[1]];
+      destination.removeEvents(eventsToRemove);
       expect(set).toHaveBeenCalledTimes(1);
+      expect(set).toHaveBeenCalledWith('', expect.objectContaining([event1]));
     });
 
-    test('should be ok with no storage provider', () => {
```
Mercy811, line 568: Can we keep the test as we keep the code where it tests on?

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Those are different because in addToQueue we want to contact the queue, but for other retry we don't. 
But since most of the code is the same added a flag to enable/disable concat the queue. Please check later comment.

packages/analytics-core/test/plugins/destination.test.ts
yuhao900914: It has been moved into line 584, what do you mean?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -297,21 +297,41 @@ export class Destination implements DestinationPlugin {
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
-    this.saveEvents();
+    this.removeEvent(list);
```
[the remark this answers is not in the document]
yuhao900914: I prefer to keep them separate, it's much clearer.

packages/analytics-core/src/plugins/destination.ts
Mercy811: Update in https://github.com/amplitude/Amplitude-TypeScript/pull/689/commits/e2151480f4e997ce9dd3b65023cf58b8c29e203a

Mercy811: Update in https://github.com/amplitude/Amplitude-TypeScript/pull/689/commits/e2151480f4e997ce9dd3b65023cf58b8c29e203a

packages/analytics-core/test/plugins/destination.test.ts
yuhao900914: test `should filter dropped event and update the storage provider` is making sure the remove event from queue function.

Mercy811: I see. Can we check `destination.queue` in tests with mocked http response like [this](https://github.com/amplitude/Amplitude-TypeScript/blob/802da9c85c35c6e1309812aff704efa3d5304933/packages/analytics-core/test/plugins/destination.test.ts#L644-L687) to make sure that the queue behaves as expected? I'm thinking if we could have added this check in https://github.com/amplitude/Amplitude-TypeScript/pull/673, we would not have the events accumulating issue.

packages/analytics-core/test/plugins/destination.test.ts
Mercy811: I didn't see it was moved, I though this test was deleted somehow

packages/analytics-core/test/plugins/destination.test.ts
yuhao900914: I added the destination.queue check in the test which has mocked HTTP response.

Mercy811 (approved):
Thanks Alyssa for fixing this! 🙏

yuhao900914:
Other clean up will be covered in other pr.
1. Remove the timeout to await on the result before sending other events
2. Flush events should happen serially. Promise.all should be a for-loop
3. Clean up the logic to only schedule the event based on the queue.

PR #691, opened 2024-03-26
fix: have session and device IDs ready before plugin setup

Mercy811:
### Summary

This PR fixes the bug described at https://amplitude.atlassian.net/browse/AMP-94551 where `SessionReplayPlugin` throws an "session Id and device Id not found" error because `plugin.setup()` is executed before `client.config` is set. 

This PR is made in the consideration of not breaking the fix of https://github.com/amplitude/Amplitude-TypeScript/pull/426, specifically the session Id of the session end event should be the one fetched from storage rather than `options.sessionId`. That being said, we cannot simply solve this by 
- Change this [line](https://github.com/amplitude/Amplitude-TypeScript/blob/d9e6fc678e6d364f209f0a942ebd57143966992e/packages/analytics-browser/src/config.ts#L226) in `useBrowserConfig()` to be `const sessionId = options.sessionId ?? previousCookies?.sessionId ?? legacyCookies.sessionId;`

And have to make sure

```diff
  protected async _init(options: BrowserOptions & { apiKey: string }) {
    ...
    // When called, make sure config.sessionId is the previous session Id for session end event 
    this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
```

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -69,7 +69,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     // Step 2: Create browser config
     const browserOptions = await useBrowserConfig(options.apiKey, options, this);
-    await super._init(browserOptions);
+    this.config = browserOptions;
 
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
@@ -77,6 +77,8 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
     this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
+
+    await super._init(this.config);
 
     // Set up the analytics connector to integrate with the experiment SDK.
     // Send events from the experiment SDK and forward identifies to the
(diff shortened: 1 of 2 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -78,6 +78,8 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     // Session ID is handled differently than device ID and user ID due to session events
     this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
 
+    await super._init(this.config);
```
yuhao900914, line 81: Question: 
In the setSessionId function, we also sent the session event. 
Since we set the config after the setSessionId, will the session event have the right configuration?

Mercy811: Yes, the session event should have the right configuration as we have `this.config = browserOptions` before `setSessionId()`.

justin-fiedler (approved):
Thanks @Mercy811!

packages/analytics-browser/src/browser-client.ts
yuhao900914: We reset the timeline plugin in the super._init functions. Does that mean there might be some remaining plugins overwrite the session event?

Mercy811: Yes, session Id can be reset in `plugin.teardown()` as it's called in `super._init()`. But technically, `teardown()` is designed for releasing resources, etc. Users should not change configuration there. Current implementation also doesn't guarantee that configuration is unchanged in `teardown()`. In current implementation, if session Id is changed in `teardown()`, the session Id of session end event will be wrong.

yuhao900914: Thanks for checking this.

yuhao900914 (approved):

PR #692, opened 2024-03-26
fix: register download and form listeners after window is loaded

izaaz:
### Summary

When the amplitude script tag is added before the body tag, the file download listeners and form listeners weren't able to attach themselves to the body of the page (since the body was still not loaded). Fixing this by setting up the listener after the window is loaded

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/file-download-tracking.ts
+++ b/packages/analytics-browser/src/plugins/file-download-tracking.ts
@@ -30,75 +30,81 @@ export const fileDownloadTracking = (): EnrichmentPlugin => {
   const name = '@amplitude/plugin-file-download-tracking-browser';
   const type = 'enrichment';
   const setup = async (config: BrowserConfig, amplitude: BrowserClient) => {
-    /* istanbul ignore if */
-    if (!amplitude) {
-      // TODO: Add required minimum version of @amplitude/analytics-browser
-      config.loggerProvider.warn(
-        'File download tracking requires a later version of @amplitude/analytics-browser. File download events are not tracked.',
-      );
-      return;
-    }
-
-    /* istanbul ignore if */
-    if (typeof document === 'undefined') {
-      return;
-    }
-
-    const addFileDownloadListener = (a: HTMLAnchorElement) => {
-      let url: URL;
-      try {
-        // eslint-disable-next-line no-restricted-globals
-        url = new URL(a.href, window.location.href);
-      } catch {
-        /* istanbul ignore next */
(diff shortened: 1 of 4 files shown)


Mercy811 (approved):
Thanks Izaaz for fixing this. LGTM.

It looks like the PR only puts the original code in `setup()` into `window.addEventListener('load', function () {}`. But git shows it as the whole `setup()` is changed because of indents.

Mercy811:
For [failed tests](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/8443403606/job/23126953252?pr=692)  , I think you can pass them by mocking a window.load event by https://stackoverflow.com/questions/71220140/how-we-can-mock-or-trigger-window-onload-method-in-unit-test-case-using-jest

yuhao900914 (approved):

izaaz:
> Thanks Izaaz for fixing this. LGTM.
> 
> It looks like the PR only puts the original code in `setup()` into `window.addEventListener('load', function () {}`. But git shows it as the whole `setup()` is changed because of indents.

Yea, you can look at the non whitespace changes by doing this

image

izaaz:
> For [failed tests](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/8443403606/job/23126953252?pr=692) , I think you can pass them by mocking a window.load event by https://stackoverflow.com/questions/71220140/how-we-can-mock-or-trigger-window-onload-method-in-unit-test-case-using-jest.
> 
> **UPD:** Could you
> 
> * Update the failed tests to expect() that no default events are tracked if not `window.load`
> * Add new tests to expect() that default events are tracked if `window.load`

Updated tests too.

packages/analytics-browser/src/plugins/file-download-tracking.ts
```diff
@@ -30,75 +30,81 @@ export const fileDownloadTracking = (): EnrichmentPlugin => {
...
-      try {
-        // eslint-disable-next-line no-restricted-globals
-        url = new URL(a.href, window.location.href);
-      } catch {
-        /* istanbul ignore next */
+    // The form interaction plugin observes changes in the dom. For this to work correctly, the observer can only be setup
+    // after the body is built. When Amplitud gets initialized in a script tag, the body tag is still unavailable. So register this
+    // only after the window is loaded
+    // eslint-disable-next-line no-restricted-globals
+    window.addEventListener('load', function () {
```
kevinpagtakhan, line 37: this direct access to window might cause an issue with using this SDK in chrome extensions

PR #696, opened 2024-03-29
fix: fix web attribution identify and session start order

yuhao900914:
### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -27,7 +27,7 @@ import {
 import { convertProxyObjectToRealObject, isInstanceProxy } from './utils/snippet-helper';
 import { Context } from './plugins/context';
 import { useBrowserConfig, createTransport } from './config';
-import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser';
+import { WebAttribution } from './utils/web-attribution';
 import { pageViewTrackingPlugin } from '@amplitude/plugin-page-view-tracking-browser';
 import { formInteractionTracking } from './plugins/form-interaction-tracking';
 import { fileDownloadTracking } from './plugins/file-download-tracking';
@@ -41,6 +41,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
   config: BrowserConfig;
   previousSessionDeviceId: string | undefined;
   previousSessionUserId: string | undefined;
+  webAttribution: WebAttribution | undefined;
 
   init(apiKey = '', userIdOrOptions?: string | BrowserOptions, maybeOptions?: BrowserOptions) {
     let userId: string | undefined;
@@ -71,11 +72,19 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     const browserOptions = await useBrowserConfig(options.apiKey, options, this);
     this.config = browserOptions;
 
+    // Add web attribution plugin
+    if (isAttributionTrackingEnabled(this.config.defaultTracking)) {
+      const attributionTrackingOptions = getAttributionTrackingConfig(this.config);
+      this.webAttribution = new WebAttribution(attributionTrackingOptions, this, this.config);
+      await this.webAttribution.init();
+    }
+
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
     // Priority 2: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
+    console.log('before set sesion id');
     this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
 
     await super._init(this.config);
@@ -107,13 +116,6 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     if (isFormInteractionTrackingEnabled(this.config.defaultTracking)) {
       await this.add(formInteractionTracking()).promise;
-    }
-
-    // Add web attribution plugin
-    if (isAttributionTrackingEnabled(this.config.defaultTracking)) {
-      const attributionTrackingOptions = getAttributionTrackingConfig(this.config);
-      const webAttribution = webAttributionPlugin(attributionTrackingOptions);
-      await this.add(webAttribution).promise;
     }
 
     // Add page view plugin
@@ -187,9 +189,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     this.config.sessionId = sessionId;
     this.config.lastEventTime = undefined;
     this.config.pageCounter = 0;
+    console.log(this.config.sessionId);
 
     if (isSessionTrackingEnabled(this.config.defaultTracking)) {
       if (previousSessionId && lastEventTime) {
+        console.log('end session track');
         this.track(DEFAULT_SESSION_END_EVENT, undefined, {
           device_id: this.previousSessionDeviceId,
           event_id: ++lastEventId,
@@ -198,8 +202,16 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
           user_id: this.previousSessionUserId,
         });
       }
-
       this.config.lastEventTime = this.config.sessionId;
+    }
+
+    // fire web attribution events
+    console.log('last event time in setSessionId: ', this.config.lastEventTime);
+    this.webAttribution?.track();
+
+    if (isSessionTrackingEnabled(this.config.defaultTracking)) {
+      console.log('start session track');
+
       this.track(DEFAULT_SESSION_START_EVENT, undefined, {
         event_id: ++lastEventId,
         session_id: this.config.sessionId,
@@ -262,15 +274,28 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
   async process(event: Event) {
     const currentTime = Date.now();
+    console.log('currentTime: ', currentTime);
     const isEventInNewSession = isNewSession(this.config.sessionTimeout, this.config.lastEventTime);
-
+    await this.webAttribution?.fetchCampaign();
+    const shouldTrackNewCampaign = this.webAttribution?.shouldTrackNewCampaign();
+    const shouldSetSessionId = shouldTrackNewCampaign && this.webAttribution?.options.resetSessionOnNewCampaign;
+    console.log('isEventInNewSession: ', isEventInNewSession);
+    console.log('shouldTrackNewCampaign: ', shouldTrackNewCampaign);
+    console.log('shouldSetSessionId: ', shouldSetSessionId);
+    console.log('event type', event.event_type);
     if (
       event.event_type !== DEFAULT_SESSION_START_EVENT &&
       event.event_type !== DEFAULT_SESSION_END_EVENT &&
-      (!event.session_id || event.session_id === this.getSessionId()) &&
-      isEventInNewSession
+      (!event.session_id || event.session_id === this.getSessionId())
     ) {
-      this.setSessionId(currentTime);
+      if (isEventInNewSession || shouldSetSessionId) {
+        console.log('setting session id');
+        this.setSessionId(currentTime);
+      } else if (!isEventInNewSession && shouldTrackNewCampaign) {
+        console.log('in process identify track');
+        // web attribution should be track during the middle of the session if there has any new campaign
+        this.webAttribution?.track();
+      }
     }
 
     return super.process(event);
--- a/packages/analytics-browser/src/utils/web-attribution-helper.ts
+++ b/packages/analytics-browser/src/utils/web-attribution-helper.ts
@@ -1,0 +1,88 @@
+import { BASE_CAMPAIGN } from '@amplitude/analytics-client-common';
+import { AMPLITUDE_PREFIX, createIdentifyEvent, Identify } from '@amplitude/analytics-core';
+import { Campaign } from '@amplitude/analytics-types';
+
+export interface Options {
+  excludeReferrers?: (string | RegExp)[];
+  initialEmptyValue?: string;
+  resetSessionOnNewCampaign?: boolean;
+}
+
+export const getStorageKey = (apiKey: string, postKey = '', limit = 10) => {
+  return [AMPLITUDE_PREFIX, postKey, apiKey.substring(0, limit)].filter(Boolean).join('_');
+};
+
+const domainWithoutSubdomain = (domain: string) => {
+  const parts = domain.split('.');
+
+  if (parts.length <= 2) {
+    return domain;
+  }
+
+  return parts.slice(parts.length - 2, parts.length).join('.');
+};
+
+//Direct traffic mean no external referral, no UTMs, no click-ids, and no other customer identified marketing campaign url params.
+const isDirectTraffic = (current: Campaign) => {
+  return Object.values(current).every((value) => !value);
+};
+
+export const isNewCampaign = (
+  current: Campaign,
+  previous: Campaign | undefined,
+  options: Options,
+  isNewSession = true,
+) => {
+  const { referrer, referring_domain, ...currentCampaign } = current;
+  const { referrer: _previous_referrer, referring_domain: prevReferringDomain, ...previousCampaign } = previous || {};
+
+  if (isExcludedReferrer(options.excludeReferrers, current.referring_domain)) {
+    return false;
+  }
+
+  //In the same session, direct traffic should not override or unset any persisting query params
+  if (!isNewSession && isDirectTraffic(current) && previous) {
+    return false;
+  }
+
+  const hasNewCampaign = JSON.stringify(currentCampaign) !== JSON.stringify(previousCampaign);
+  const hasNewDomain =
+    domainWithoutSubdomain(referring_domain || '') !== domainWithoutSubdomain(prevReferringDomain || '');
+
+  return !previous || hasNewCampaign || hasNewDomain;
+};
+
+export const isExcludedReferrer = (excludeReferrers: (string | RegExp)[] = [], referringDomain = '') => {
+  return excludeReferrers.some((value) =>
+    value instanceof RegExp ? value.test(referringDomain) : value === referringDomain,
+  );
+};
+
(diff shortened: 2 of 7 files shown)


omares:
Hey @yuhao900914, since we're affected by this issue, I would like to understand the solution which has been applied. I would appreciate it if you could elaborate on that.
I'd also like to understand why the changes haven't been applied to the web attribution browser plugin. Instead, it seems the code has been duplicated.

packages/analytics-browser/src/utils/web-attribution.ts
```diff
@@ -0,0 +1,58 @@
+import { BrowserClient, BrowserConfig } from '@amplitude/analytics-types';
```
izaaz, line 1: I see all these functions added. Where is the code being removed from? I see no deletion in our code?

yuhao900914: Web attribution is a plugin that is released separately. I don't think we should remove it. We only need to remove the web attribution installation.

izaaz: Don't we add the web attribution plugin by default? Or do customers add that manually? Do we have any docs that ask customers to add the web attribution plugin by themselves anywhere?

I am worried that we have duplicate logic in multiple places and its getting harder to figure out which logic is the most accurate. And if we aren't removing the plugin, then the plugin should also use this helper at the very least to avoid code duplication right?

yuhao900914:
> Hey @yuhao900914, since we're affected by this issue, I would like to understand the solution which has been applied. I would appreciate it if you could elaborate on that. I'd also like to understand why the changes haven't been applied to the web attribution browser plugin. Instead, it seems the code has been duplicated.

Hi @omares , Thanks for your support. Previously, fire session end and session start events were bundled together, the web attribution was a plugin, and when to execute the web attribution plugin had a specific order. This makes it impossible to fire a web attribution to identify the call in the middle of the session end and session start event.

packages/analytics-browser/playground/index.html
```diff
@@ -7,8 +7,19 @@
     <title>Amplitude SDK Playground</title>
   </head>
   <script src="./amplitude.js"></script>
-	<script>
-    amplitude.init('API_KEY', 'playground.user@amplitude.com');
+	
+  <script>
+    amplitude.init('82b148f7211db7f9ccaff8048d0f7192', 'playground.user@amplitude.com',
```
izaaz, line 12: NIT: Unintentional?

packages/analytics-browser/src/utils/web-attribution-helper.ts
```diff
@@ -0,0 +1,88 @@
+import { BASE_CAMPAIGN } from '@amplitude/analytics-client-common';
```
izaaz, line 1: Curious: Why is this a separate file and not a part of the web-attribution.ts file?

yuhao900914: - Modify anything in the web attribution plugin will release a new version while we publish, this duplicate will not. 
- There is a case for using web attribution separately in browser 1.0. Copy over is for isolating purposes.  
Let me what do you think.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -263,14 +272,21 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
-      isEventInNewSession
+      (!event.session_id || event.session_id === this.getSessionId())
     ) {
-      this.setSessionId(currentTime);
+      if (isEventInNewSession || shouldSetSessionId) {
+        await this.setSessionId(currentTime, shouldTrackNewCampaign);
+      } else if (!isEventInNewSession && shouldTrackNewCampaign) {
+        // web attribution should be track during the middle of the session if there has any new campaign
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        await this.webAttribution!.track().promise;
```
yuhao900914, line 288: Make this as required(!) instead optional (?) because 
1. shouldTrackNewCampaign already make sure this.webAttribution exist
2. the test coverage needs to cover `?` 
Will revert this back to optional and try other ways to bypass the test coverage.

packages/analytics-browser/src/utils/web-attribution-helper.ts
izaaz: Why is it a concern that modifying the web attribution plugin will release a new version?

Copy over is fine but we don't want code to be duplicated in two places right? We can still keep the web attribution plugin and just call these util methods right?

izaaz:
@yuhao900914 on the summary, can you please also describe the problem that we are trying to solve?

packages/analytics-browser/src/browser-client.ts
```diff
@@ -169,12 +169,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     return this.config?.sessionId;
   }
 
-  setSessionId(sessionId: number) {
+  async setSessionId(sessionId: number, shouldTrackNewCampaign?: boolean) {
```
justin-fiedler, line 172: Instead of making this `async` we could make it return a "Promise object" like we do for other methods (e.g. `await amplitude.track().promise`).

This pattern allows users to choose whether or not them want to await the result without needing to always explicitly ignore the promise with `void setSessionId()`

packages/analytics-browser/src/browser-client.ts
```diff
@@ -190,21 +189,32 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     if (isSessionTrackingEnabled(this.config.defaultTracking)) {
       if (previousSessionId && lastEventTime) {
-        this.track(DEFAULT_SESSION_END_EVENT, undefined, {
+        await this.track(DEFAULT_SESSION_END_EVENT, undefined, {
```
justin-fiedler, line 192: Instead of waiting on this result here, I think it would be better to return this promise so users can choose to await or ignore.

packages/analytics-browser/src/utils/web-attribution.ts
justin-fiedler: What about making the web attribution util part of the WA plugin package and importing it in the client instead of the plugin? 

That way we can support both without duplicating the code

packages/analytics-browser/src/utils/web-attribution.ts
```diff
@@ -0,0 +1,58 @@
...
+   * This can be called when enable web attribution and either
+   * 1. set a new session
+   * 2. has new campaign and enable resetSessionOnNewCampaign
+   */
+  track(event_id?: number) {
+    const campaignEvent = createCampaignEvent(this.currentCampaign, this.options);
+    if (event_id) {
+      campaignEvent.event_id = event_id;
+    }
+    return this.amplitude.track(campaignEvent);
```
justin-fiedler, line 56: Rather than tracking directly here which requires the amplitude client, what about just returning the campaign event, and leave the tracking to the Amplitude client?

packages/analytics-browser/src/utils/web-attribution-helper.ts
yuhao900914: Make sense. 
I have put the web attribution logic in analytics-client-common.

packages/analytics-browser/src/browser-client.ts
yuhao900914: Thanks the sugesstion. I return the `returnWrapper` in the setSessionId.

packages/analytics-browser/src/utils/web-attribution-helper.ts
izaaz: I think Justin had a different suggestion. Please check that as well.

yuhao900914: I prefer my current way, instead of putting the util in the web attribution, because it can prevent installing the web attribution package since the plugin stuff is irrelevant. 
We already have something in the analytics-client-common for the attribution, put the web attribution logic in analytics-client-common and import the logic in web attribution plugin and browser-client can also prevent duplication code. 

@justin-fiedler Let me know what you think.

packages/analytics-browser/src/utils/web-attribution.ts
yuhao900914: Changed.

packages/analytics-browser/src/utils/web-attribution.ts
justin-fiedler: The reason I am thinking to put the logic into the plugin so that we don't need to worry about if the plugin is being used in 1.0 or 2.0, if we put the shared logic into the client/core, then the plugin will only work with that version of the Browser SDK.

packages/analytics-browser/src/utils/web-attribution-helper.ts
justin-fiedler: https://github.com/amplitude/Amplitude-TypeScript/pull/696#discussion_r1560102752

packages/analytics-browser/src/utils/web-attribution.ts
justin-fiedler: If the WA plugin already has 1.0 and 2.0 versions this may not matter.

yuhao900914: I think it would be fine, web attribution v1 [depended on ^1.x.x ](https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/plugin-web-attribution-browser/package.json#L39-L43), [v2 depended on ^2.x.x](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-web-attribution-browser/package.json#L39-L43). So there already have breaking change.

yuhao900914: Then I will leave the web attribution logic in analytics-client-common

packages/analytics-browser/src/browser-client.ts
yuhao900914: See [comment](https://github.com/amplitude/Amplitude-TypeScript/pull/696#discussion_r1560203134).

packages/analytics-browser/src/browser-client.ts
```diff
@@ -261,16 +274,27 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
   }
 
   async process(event: Event) {
+    let shouldTrackNewCampaign = false;
+    let shouldSetSessionId = false;
+    if (this.webAttribution) {
```
yuhao900914, line 279: This is the replacement of `.?` for the test coverage purpose. Jest test cannot catch the branch coverage for `.?`.

packages/analytics-browser/src/utils/web-attribution.ts
izaaz: packages/plugin-web-attribution-browser/src/web-attribution.ts

Is the duplication in this code also being fixed?

yuhao900914: @izaaz Yes, I updated the web-attribution to import from  analytics-client-common.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -27,7 +27,7 @@ import {
 import { convertProxyObjectToRealObject, isInstanceProxy } from './utils/snippet-helper';
 import { Context } from './plugins/context';
 import { useBrowserConfig, createTransport } from './config';
-import { webAttributionPlugin } from '@amplitude/plugin-web-attribution-browser';
+import { WebAttribution } from '@amplitude/analytics-client-common';
```
justin-fiedler, line 30: Since web attribution is web only, maybe it should be in `analytics-browser` package?

packages/analytics-browser/src/browser-client.ts
```diff
@@ -71,6 +72,14 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     const browserOptions = await useBrowserConfig(options.apiKey, options, this);
     this.config = browserOptions;
 
+    // Add web attribution plugin
```
justin-fiedler, line 75: nit
```suggestion
    // Enable web attribution
```

packages/analytics-browser/src/browser-client.ts
```diff
@@ -169,12 +171,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     return this.config?.sessionId;
   }
 
-  setSessionId(sessionId: number) {
+  setSessionId(sessionId: number, shouldTrackNewCampaign?: boolean) {
```
justin-fiedler, line 174: Do we need to add `shouldTrackNewCampaign?: boolean` to the method signature? We didn't have that before, and would like to avoid interface changes if possible

packages/analytics-browser/src/browser-client.ts
```diff
@@ -169,12 +170,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
```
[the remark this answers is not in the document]
justin-fiedler: I'm happy we don't need to make this async, but think it could still make sense to return a Promise object to allow users to await session event tracking if desired. e.g.
`setSessionId(sessionId: number): { promise: Promise }`

packages/analytics-browser/src/browser-client.ts
yuhao900914: Then in web-attribution plugin, we need to install analytics-browser? I don't think that's efficient.

yuhao900914: I saw we have campaign-tracker in the analytics-client-common which has similar logic as web-attribution. So it should not only for the web, it's also used in react-native-client.
I think it can be in '@amplitude/analytics-client-common' but should have a better name.
Maybe campaign-tracker-v2?

justin-fiedler: > web-attribution plugin, we need to install analytics-browser

I don't think that's too much of an issue since `plugin-web-attribution-browser` only works with `analytics-browser`, it will always have that dependency anyway.

Not a huge deal either way, but since this only works for browser it could help remove confusion by moving it out of core

yuhao900914: Sync offline. 
React native also has web, which makes more sense to keep the logic in `analytics-client-common`.
Since React native doesn't have 2.0.0 yet, we can wait the changes until we have react-native to 2.0.

packages/analytics-browser/src/browser-client.ts
yuhao900914: Removed the await and make setSessionId return a promise.

packages/analytics-browser/src/browser-client.ts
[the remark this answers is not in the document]
yuhao900914: Done.

packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -952,6 +973,33 @@ describe('browser-client', () => {
...
+          attribution: {
+            resetSessionOnNewCampaign: true,
+          },
+        },
+      }).promise;
+
+      expect(client.webAttribution).toBeDefined();
+
+      // Making sure client.webAttribution is not undefined
+      if (!client.webAttribution) {
```
justin-fiedler, line 990: Shouldn't this be impossible based on the expect().toBeDefined() above?

packages/analytics-browser/src/browser-client.ts
```diff
@@ -260,17 +276,40 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
       event.event_type !== DEFAULT_SESSION_END_EVENT &&
-      (!event.session_id || event.session_id === this.getSessionId()) &&
-      isEventInNewSession
+      (!event.session_id || event.session_id === this.getSessionId())
     ) {
-      this.setSessionId(currentTime);
+      if (isEventInNewSession || shouldSetSessionIdOnNewCampaign) {
+        this.setSessionId(currentTime);
+        if (shouldSetSessionIdOnNewCampaign) {
+          this.config.loggerProvider.log('Created a new session for new campaign.');
```
izaaz, line 307: How does our logging work? Will this show up always? If yes, should this be loggerProvider.debug?

packages/analytics-browser/test/browser-client.test.ts
yuhao900914: Yea, this bugs me a lot. the client.webAttribution is defined, but since client.webAttribution can be undefined, it won't let me spy on client.webAttribution.

izaaz (approved):
A couple of minor comments but the rest looks good. Thanks @yuhao900914

packages/analytics-browser/src/browser-client.ts
yuhao900914: If the logLevel is equal or more than 3, it means if it's either `Verbose, Debug`, we will log this. We do have logger.debug(). But we provide verbose for more possibility. 
[Reference](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/logger.ts#L20).
 
```
  log(...args: any[]): void {
    if (this.logLevel < LogLevel.Verbose) {
      return;
    }
    console.log(`${PREFIX}[Log]: ${args.join(' ')}`);
  }
```

yuhao900914: I kept it as what it was in the web attribution plugin. [Code reference](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-web-attribution-browser/src/web-attribution.ts#L35).

packages/analytics-browser/src/browser-client.ts
```diff
@@ -190,25 +194,37 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
-          user_id: this.previousSessionUserId,
-        });
+        promise.push(
+          this.track(DEFAULT_SESSION_END_EVENT, undefined, {
+            device_id: this.previousSessionDeviceId,
+            event_id: ++lastEventId,
+            session_id: previousSessionId,
+            time: lastEventTime + 1,
+            user_id: this.previousSessionUserId,
+          }),
```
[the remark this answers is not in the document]
yuhao900914: Yes, it should be promise. Done.

justin-fiedler (approved):

PR #698, opened 2024-04-02
refactor(session replay): split out destination logic into separate class

kwalker3690:
### Summary

In preparation for adding logic to support targeting, I am simplifying the core SessionReplay class. This class has grown large already as we've added functionality, and will continue to grow larger. This PR separates out all logic related to sending the sequence of events to the API into a new class, TrackDestination. There are no functional changes in this PR.

### Checklist


jxiwang (approved):
Besides comment LGTM

PR #702, opened 2024-04-02
feat(targeting): introduce new package for targeting

kwalker3690:
### Summary

This PR introduces the base of a new package called `targeting`, that will provide a wrapper around the experiment-core Evaluation Engine. This package will manage the storage of some event and user property data, but ultimately be a concise interface between Session Replay and future action based products to the Evaluation Engine.

### Checklist

### DIFF
--- a/packages/targeting/src/targeting.ts
+++ b/packages/targeting/src/targeting.ts
@@ -8,8 +8,9 @@ export class Targeting implements AmplitudeTargeting {
     this.evaluationEngine = new EvaluationEngine();
   }
 
-  evaluateTargeting({ event, userProperties, deviceId, flag }: TargetingParameters) {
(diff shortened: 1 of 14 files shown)


packages/targeting/src/targeting.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  evaluateTargeting({ event, userProperties, deviceId, flag }: TargetingParameters) {
+    const context = {
+      event,
+      user: {
+        device_id: deviceId,
+        user_properties: userProperties,
+      },
+    };
```
bgiori, line 18: In your test you're bucketing by session id, but that is not included in this context. Most likely you want to add that here.

kwalker3690: Ah thanks for pointing that out, will update

justin-fiedler (approved):

PR #704, opened 2024-04-03
refactor(session replay): move events storage logic in IDB to separate class

kwalker3690:
### Summary

Separate out the events storage logic for Session Replay into it's own class 

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -13,12 +13,12 @@ import {
   MIN_INTERVAL,
   SESSION_REPLAY_DEBUG_PROPERTY,
 } from './constants';
-import { SessionReplayEventsStorage } from './events-storage';
 import { generateHashCode, generateSessionReplayId, isSessionInSample, maskInputFn } from './helpers';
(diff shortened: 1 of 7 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -334,16 +334,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
       deviceId: this.getDeviceId(),
-      sampleRate: this.getSampleRate(),
+      sampleRate: this.config.sampleRate,
       serverZone: this.config.serverZone,
-      onComplete: this.cleanUpSessionEventsStore.bind(this),
+      onComplete: this.sessionIDBStore.cleanUpSessionEventsStore.bind(this),
     });
   }
 
-  getSampleRate() {
```
kwalker3690, line 343: I noticed this is an unnecessary wrapper, as we set the sample rate to the DEFAULT_SAMPLE_RATE in the config class constructor: https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/session-replay-browser/src/config.ts#L41-L41

jxiwang (approved):
LGTM

PR #705, opened 2024-04-04
refactor(session replay): separate out session identifiers from the rest of config

kwalker3690:
### Summary

This PR in the Session Replay SDK creates a distinction between config that is set once upon initialization, and identifiers that can be updated throughout the SDK lifecycle (session id, device id, and session replay id). This helps add clarity to which aspects of the configuration are changed in the lifecycle, and which we can expect to remain static.

### Checklist


jxiwang (approved):
lgmt

PR #709, opened 2024-04-08
build: fix the dependency issue

yuhao900914:
### Summary

While graduate the beta version, github action cannot recognize the latest change in React Native and Node. 

Screenshot 2024-04-08 at 5 03 54 PM

### Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,18 @@
 # yarn lockfile v1
 
 
+"@amplitude/analytics-browser@^2.3.6-beta.0":
+  version "2.6.1"
+  resolved "https://registry.yarnpkg.com/@amplitude/analytics-browser/-/analytics-browser-2.6.1.tgz#b127adbe1905ce2561a73ec01d5046df3a4d5b09"
+  integrity sha512-mGHPW5gGbYJhJOskYd9t4TPpf8v5jkDB2ndCmWipSm/OXHVCcQucRGUrJdmQIRrW+5aB4ertYCK0t2m/yzj2zw==
+  dependencies:
+    "@amplitude/analytics-client-common" "^2.1.2"
+    "@amplitude/analytics-core" "^2.2.3"
+    "@amplitude/analytics-types" "^2.5.0"
+    "@amplitude/plugin-page-view-tracking-browser" "^2.2.5"
+    "@amplitude/plugin-web-attribution-browser" "^2.1.6"
+    tslib "^2.4.1"
+
 "@amplitude/analytics-client-common@^2.0.0-beta.0":
   version "2.0.8"
   resolved "https://registry.yarnpkg.com/@amplitude/analytics-client-common/-/analytics-client-common-2.0.8.tgz#6b9169afc7059e8c64ea1971e194a12ffe146e4f"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -2,27 +2,58 @@
...
+  integrity sha512-mGHPW5gGbYJhJOskYd9t4TPpf8v5jkDB2ndCmWipSm/OXHVCcQucRGUrJdmQIRrW+5aB4ertYCK0t2m/yzj2zw==
+  dependencies:
+    "@amplitude/analytics-client-common" "^2.1.2"
+    "@amplitude/analytics-core" "^2.2.3"
+    "@amplitude/analytics-types" "^2.5.0"
+    "@amplitude/plugin-page-view-tracking-browser" "^2.2.5"
+    "@amplitude/plugin-web-attribution-browser" "^2.1.6"
+    tslib "^2.4.1"
+
+"@amplitude/analytics-client-common@^2.0.0", "@amplitude/analytics-client-common@^2.1.2":
```
izaaz, line 17: Why do we have two versions of analytics-client-common in here?

yarn.lock
```diff
@@ -2,27 +2,58 @@
...
 "@amplitude/analytics-connector@^1.4.8":
   version "1.4.8"
   resolved "https://registry.yarnpkg.com/@amplitude/analytics-connector/-/analytics-connector-1.4.8.tgz#dd801303db2662bc51be7e0194eeb8bd72267c42"
   integrity sha512-dFW7c7Wb6Ng7vbmzwbaXZSpqfBx37ukamJV9ErFYYS8vGZK/Hkbt3M7fZHBI4WFU6CCwakr2ZXPme11uGPYWkQ==
 
-"@amplitude/analytics-core@^2.0.0-beta.0", "@amplitude/analytics-core@^2.1.1":
-  version "2.1.1"
-  resolved "https://registry.yarnpkg.com/@amplitude/analytics-core/-/analytics-core-2.1.1.tgz#1b6105bd763b38e5633533c9917a78f64fde612a"
-  integrity sha512-2dHHiOnK7J/0Uk3gqu70JEvCKSgNBAXIUvw6u7bEHCQHBBW3ulpsVRSQomBeruyBBLKjgarwgawGs3yJrjIDkA==
+"@amplitude/analytics-core@^2.0.0", "@amplitude/analytics-core@^2.2.3":
```
izaaz, line 32: Same here.

yuhao900914:
Sync with Justin, why we use `2.0.0-beta.0` in node and react is because there is no 2.0 for node and react-native. And we have not published them yet.  So this shouldn't be changed. Will close this pr. 
However, it's preventing me to graduate the analytics-browser SDK. It might be the issue between this [pr](https://github.com/amplitude/Amplitude-TypeScript/pull/644) and the graduate process.

yarn.lock
yuhao900914: Because we installed the analytics-client-common in multiple packages, it has a different version. 
The node and react-native are private repos, run release won't bump the version. So the versions of those packages are not synced.

yarn.lock
yuhao900914: The same.

PR #711, opened 2024-04-10
feat(session replay): introduce remote config fetch class

kwalker3690:
### Summary

Add a RemoteConfigFetch class to the Session Replay SDK. This will allow us to fetch targeting config in the initialization of the SR SDK

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/remote-config-fetch.ts
+++ b/packages/session-replay-browser/src/remote-config-fetch.ts
@@ -1,0 +1,96 @@
+import { BaseTransport } from '@amplitude/analytics-core';
+import { Status } from '@amplitude/analytics-types';
+import { TargetingFlag } from '@amplitude/targeting';
+import { SESSION_REPLAY_SERVER_URL } from './constants';
+import { UNEXPECTED_ERROR_MESSAGE } from './messages';
+import {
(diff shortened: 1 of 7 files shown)


packages/session-replay-browser/src/remote-config-fetch.ts
```diff
@@ -0,0 +1,96 @@
...
+    const remoteConfig: SessionReplayRemoteConfig = (await res.json()) as SessionReplayRemoteConfig;
+    this.completeRequest({ success: SUCCESS_REMOTE_CONFIG });
+    this.remoteConfig = remoteConfig;
+    return remoteConfig;
+  };
+
+  completeRequest({ err, success }: { err?: string; success?: string }) {
+    this.attempts = 0; // Reset attempts back to 0 for restart
+    if (err) {
+      this.config.loggerProvider.warn(err);
```
jxiwang, line 91: nit: Think some customers suppress warnings. Do you think there's any value in showing an actual error log?

jxiwang (approved):
LGTM

packages/session-replay-browser/src/remote-config-fetch.ts
kwalker3690: This should only show up if the customer has set their log level high enough! Ie the customer has to opt into it. By default this won't show up

PR #712, opened 2024-04-12
feat(session replay): evaluate targeting in SR init

kwalker3690:
### Summary

Evaluate targeting in SR init, storing the result in IndexedDB. With this PR we are not actually doing anything with the result, that will come later. I decided to put the logic for evaluation in the RemoteConfigFetch class, as it felt like too much for the base SessionReplay class to have to know about, I may update the name of RemoteConfigFetch in a future PR to make it a little more inclusive of targeting logic.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/remote-config-fetch.ts
+++ b/packages/session-replay-browser/src/remote-config-fetch.ts
@@ -1,6 +1,6 @@
 import { BaseTransport } from '@amplitude/analytics-core';
 import { Status } from '@amplitude/analytics-types';
-import { TargetingFlag } from '@amplitude/targeting';
+import { TargetingFlag, TargetingParameters, evaluateTargeting } from '@amplitude/targeting';
 import { UNEXPECTED_ERROR_MESSAGE } from './messages';
 import {
   SessionReplayRemoteConfigFetch as AmplitudeSessionReplayRemoteConfigFetch,
@@ -19,6 +19,7 @@ export class SessionReplayRemoteConfigFetch implements AmplitudeSessionReplayRem
   sessionIDBStore: AmplitudeSessionReplaySessionIDBStore;
   retryTimeout = 1000;
   attempts = 0;
+  sessionTargetingMatch = false;
 
   constructor({
     config,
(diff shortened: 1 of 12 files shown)


packages/session-replay-browser/test/integration.test.ts
```diff
@@ -100,7 +104,7 @@ describe('module level integration', () => {
       expect(record).not.toHaveBeenCalled();
       expect(update).not.toHaveBeenCalled();
       await runScheduleTimers();
-      expect(fetch).not.toHaveBeenCalled();
+      expect(fetch).not.toHaveBeenCalledWith(SESSION_REPLAY_EU_SERVER_URL);
```
kwalker3690, line 107: I made these changes in a previous version of this PR, when I was fetching the remote config from the SR class itself, so these changes aren't really necessary, but they are more precise than the previous version of the test so I decided to leave them in

packages/session-replay-browser/src/remote-config-fetch.ts
```diff
@@ -103,7 +142,7 @@ export class SessionReplayRemoteConfigFetch implements AmplitudeSessionReplayRem
   completeRequest({ err, success }: { err?: string; success?: string }) {
     this.attempts = 0; // Reset attempts back to 0 for restart
     if (err) {
-      this.config.loggerProvider.warn(err);
+      throw new Error(err);
```
kwalker3690, line 145: I decided to change this to throw an error so that the consuming function can know that an API error occurred and behave accordingly

packages/session-replay-browser/src/remote-config-fetch.ts
```diff
@@ -31,15 +32,53 @@ export class SessionReplayRemoteConfigFetch implements AmplitudeSessionReplayRem
...
+      const targetingConfig = await this.getTargetingConfig(targetingParams.sessionId);
+      if (targetingConfig && Object.keys(targetingConfig).length > 0) {
+        const targetingResult = evaluateTargeting({ ...targetingParams, flag: targetingConfig });
+        this.sessionTargetingMatch =
+          this.sessionTargetingMatch === false && targetingResult.sr_targeting_config.key === 'on';
+      } else {
+        // If the targeting config is undefined or an empty object,
+        // assume the response was valid but no conditions were set,
+        // so all users match targeting
+        this.sessionTargetingMatch = true;
```
jxiwang, line 64: How come we assume true here if config is undefined?

jxiwang (approved):
LGTM

packages/session-replay-browser/src/remote-config-fetch.ts
kwalker3690: We want to provide backwards compatibility to the existing functionality. So that means that some customers won't have any targeting config defined (especially when we are in the beta phase). In that case, we'll set this to true, so that we can know to start sending events to the backend

PR #716, opened 2024-04-12
fix(session replay): always stop recording before starting anew

kwalker3690:
### Summary

We've observed that it is possible to generate many duplicate track calls, because there are code paths that do not stop the recording before starting it again. This change makes it so that the recording is always stopped right before it is started again, ensuring those duplicate track calls cannot be made.

### Checklist


jackson-amplitude (approved):
LGTM! Thanks for the fix!

PR #717, opened 2024-04-12
fix(session replay): remove event listeners on init

kwalker3690:
### Summary

Safety check in case users call init multiple times

### Checklist


jackson-amplitude (approved):
lgtm!

PR #719, opened 2024-04-15
feat(session replay): added ability to change maskTextFn

MYXOMOPX:
### Summary

Allow to change maskTextFn: passes maskTextFn deep down to rrweb

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -105,6 +105,7 @@ export class SessionReplayPlugin implements DestinationPlugin {
       sampleRate: this.options.sampleRate,
       privacyConfig: {
         blockSelector: this.options.privacyConfig?.blockSelector,
+        maskTextFn: this.options.privacyConfig?.maskTextFn,
       },
       debugMode: this.options.debugMode,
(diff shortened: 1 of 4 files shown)


lewgordon-amplitude:
Hey there! Thanks for the contribution. I'm implementing something related in (#756). I'd be curious if that solves what you're trying to do here?

MYXOMOPX:
@lewgordon-amplitude Hi! 
Actually my idea was to change maskTextFn and check there CSS variable. If on element css-variable value is 1 (f.e. --amplitude-mask: 1) - then mask text. Else - not.
With this logic we can use amp-mask, amp-unmask classes and only last one will affect element.
So in my example I will be able to do smth like this:

```html
<div class="amp-mask">
    <span>This text is masked</span>
    <div class="amp-unmask">
        This UNmasked
        <span class="amp-mask">Masked</span>
        <input placeholder="Unmasked input"/>
    </div>
</div>
```
And somewhere in CSS i will place
```css
.amp-mask {
  --amplitude-mask: 1;
}

.amp-unmask {
  --amplitude-mask: 0;
}
```

Not sure that mask-levels can help in using nested mask/unmask classes. But if you will implement it - i'll be thankful

PR #714, opened 2024-04-12
test: example improvement

yuhao900914:
### Summary

Improve the react example app
1. add multiple pages to fire page view without fresh the page
2. use the local script instead of using the npm dependency for debugging 

### Checklist

* [ ] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:

PR #720, opened 2024-04-15
fix: [AMP-95816] fix pageCounter bug

yuhao900914:
Related ticket: https://amplitude.atlassian.net/browse/AMP-95816

### Summary
The original issue is the pageCounter of the first page is not 1. It is the previous session pageCounter + 1. This is caused by the order of the page view event and the the setSessionId. The previous order is:
- fire the page view event with the current pageCounter
- setSessionId in `async process` clean the pageCounter. 

The fix is to check if it's in a new session while creating the page view event. 

Test: 
- integration test
- local example test with the react-app in  https://github.com/amplitude/Amplitude-TypeScript/pull/714

### Checklist


packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -20,7 +20,12 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
   let localConfig: BrowserConfig;
 
   const createPageViewEvent = async (): Promise<Event> => {
+    const isEventInNewSession = isNewSession(localConfig.sessionTimeout, localConfig.lastEventTime);
+    if (isEventInNewSession && amplitude) {
+      amplitude.setSessionId(Date.now());
```
izaaz, line 25: We now have another place where session id is getting reset? Isn't this also being done in the timeline when a new event is fired?

yuhao900914: We call the setSessionId at [here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-browser/src/browser-client.ts#L273).
We actually hijack the event while processing and set the sessionId. However, page view event at that time already set the pageCounter. We need to set the session first before firing the page view event.

izaaz: In a plugin, in the execute part is where we enrich an event right? Should the page counter be set in the execute method instead? THat way, the session id is already reset and the page counter is set correctly. 

Setting session id here is more of a fix for the page counter and not something that we are doing because thats the right thing to do. 

cc: @justin-fiedler for feedback as well.

yuhao900914: It's hard to catch the page view event in the plugin, because the event type can be customized for the page view event. 
And every event will go through any enrichment plugin. That means [here](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts#L108-L120) is not only for the page view events.

justin-fiedler: Do we need to `setSessionId()` here? Would it be enough to check if this event is in a new session and clear the page counter? And leave the session management to the core SDK when it processes the page view event?

yuhao900914: @justin-fiedler  then we need to remove clear page counter logic from the setSessionId function. For me, it makes more sense to have this logic while setting the session because this is related behavior in a new session. We also call setSessioId in web attribution plugins though.

izaaz: There are 2 places where we create a new session
1. When the time between sessions > session timeout. This happens in the browser-client execute method.
2. When UTM parameters are changed within a session, we force create a new session. This happens in the web attribution plugin.

There is no use case where the page view tracking plugin needs to forcefully reset the session id. It seems like we are only calling setSessionId here to indirectly reset the page counter which isn't all that obvious. 

Wrt the event name - its a part of the options object and the options is accessible in the execute method. So can we check the event name against that option and use that to populate page counter? 

Open to what @justin-fiedler is suggesting as well. 

Session management is already fairly complex and adding one more place where session can be reset is just exposing one more possibility of bugs coming up.

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
```diff
@@ -200,6 +199,31 @@ describe('pageViewTrackingPlugin', () => {
...
+        }),
+      });
+
+      // Make sure session expire
+      const plugin = pageViewTrackingPlugin();
+      await plugin.setup?.(config, amplitude);
+      mockWindowLocationFromURL(url);
+
+      expect(config.pageCounter).toBe(1);
+      expect(config.pageCounter).toBe(1);
```
justin-fiedler, line 225: nit: duplicate?

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
```diff
@@ -200,6 +199,31 @@ describe('pageViewTrackingPlugin', () => {
...
+      jest.spyOn(amplitude, 'track').mockReturnValue({
+        promise: Promise.resolve({
+          code: 200,
+          message: '',
+          event: {
+            event_type: '[Amplitude] Page Viewed',
+          },
+        }),
+      });
+
```
justin-fiedler, line 218: Maybe check the page counter in increased before it is reset?

packages/analytics-browser/src/browser-client.ts
```diff
@@ -186,7 +186,6 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     this.config.sessionId = sessionId;
     this.config.lastEventTime = undefined;
-    this.config.pageCounter = 0;
```
izaaz, line 189: So what is the expectation when someone calls setSessionId. Is it ok that the page counter will not go to 0 in that case?

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
yuhao900914: See [comment](https://github.com/amplitude/Amplitude-TypeScript/pull/720#discussion_r1568314094).

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -116,6 +116,15 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
           ...pageViewEvent.event_properties,
         };
       }
+
+      // Update the pageCounter for the page view event
+      if (localConfig && (event.event_type === options.eventType || event.event_type === defaultPageViewEvent)) {
```
justin-fiedler, line 121: Not sure we should continue to support the defaultPageViewEvent is it is overridden in Options. Could be better to set the Options.eventType = defaultEventType by default and then only use `options.eventType` in the plugin logic

```suggestion
      if (localConfig && event.event_type === options.eventType) {
```

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
```diff
@@ -34,6 +33,7 @@ describe('pageViewTrackingPlugin', () => {
     },
     pageCounter: 0,
   };
+  const defaultPageViewEvent = '[Amplitude] Page Viewed';
```
justin-fiedler, line 36: Reuse the existing const?

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
izaaz: +1.

packages/plugin-page-view-tracking-browser/test/page-view-tracking.test.ts
yuhao900914: export the defaultPageViewEvent from the plugin-page-view and reuse here.

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
yuhao900914: Give the default event  type at the beginning of the plugin

izaaz (approved):

PR #722, opened 2024-04-18
fix(session replay): prevent multiple of the same sequences from being added to send queue

kwalker3690:
### Summary

A customer discovered a race condition where its possible to add multiple of the same sequences to the send queue in specific circumstances, particularly upon blur. This adds a check to the queue to prevent the same sequence number from being added and sent

### Checklist


jackson-amplitude (approved):
lgtm!

PR #726, opened 2024-04-22
fix: add logic to escape campaign tracking for direct traffic in session

yuhao900914:
### Summary

The last PR (change web attribution plugin to util function) didn't consider the direct traffic in the session event. 
Adding that logic checking in the webAttribution.init().

### Checklist

### DIFF
--- a/packages/analytics-client-common/src/attribution/web-attribution.ts
+++ b/packages/analytics-client-common/src/attribution/web-attribution.ts
@@ -4,6 +4,7 @@ import { Options, getDefaultExcludedReferrers, createCampaignEvent, isNewCampaig
 import { getStorageKey } from '../storage/helpers';
 import { CampaignParser } from './campaign-parser';
 import { BASE_CAMPAIGN } from './constants';
+import { isNewSession } from '../session';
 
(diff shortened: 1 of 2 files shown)


packages/analytics-client-common/src/attribution/web-attribution.ts
```diff
@@ -23,13 +26,16 @@ export class WebAttribution {
     this.storage = config.cookieStorage as unknown as Storage<Campaign>;
     this.storageKey = getStorageKey(config.apiKey, 'MKTG');
     this.currentCampaign = BASE_CAMPAIGN;
+    this.sessionTimeout = config.sessionTimeout;
```
izaaz, line 29: Should we keep a reference to the config instead of the actual values? What happens when the last event time changed and we call init once again mid session?

yuhao900914: We call the init in the mid session, then we will get the latest lastEventTime? We are updating the this.config.lastEventTime while sending out the event.

izaaz (approved):

PR #730, opened 2024-04-25
test: web attribution integration test

yuhao900914:
### Summary

Jira [AMP-98581](https://amplitude.atlassian.net/browse/AMP-98581)
Add web attribution integration test
The test covers
1. Page reload. The test uses double init to mock hard page reload. 
   a. In a new session should track all UTMs and referrers
   b. In a new session should track no campaign if no UTMs and no referrer
   c. In a new session should not track all UTMs and referrers if the referrer is excluded referrer
   d. During a session should not fire campaign identify event for direct traffic
   e. During a session  should track all UTMs and referrers if any campaign changed and not direct traffic
   f.  During a session should start a new session with resetSessionCampaign enable and track campaign updates

2. Process event, when any events under process, which include SPA redirect 
  a.  in a new session Should track all UTMs and referrers while processing any event
  b.  in a new session  Should not track UTMs and referrers if the referrer is excluded referrer 
  c. during a session Should not track updated campaign during
  d. during a session  Should not track campaign with resetSessionCampaign enable

### Checklist

[AMP-98581]: https://amplitude.atlassian.net/browse/AMP-98581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser-test/test/helpers.ts
+++ b/packages/analytics-browser-test/test/helpers.ts
@@ -32,9 +32,9 @@ const getReferrerObject = (referrerString?: string) => {
   const referring_domain = referrerURL.hostname;
   return { referrer, referring_domain };
 };
+
 const generateAttributionUserProps = (campaignURL: string, referrerString?: string) => {
   const referrer = getReferrerObject(referrerString);
-
   const campaign = parseQueryString(campaignURL);
   const campaignObject: Campaign = {
     ...BASE_CAMPAIGN,
--- a/packages/analytics-browser-test/test/helpers.ts
+++ b/packages/analytics-browser-test/test/helpers.ts
@@ -25,11 +25,21 @@ const parseQueryString = (url: string) => {
   return params;
 };
 
-const generateAttributionUserProps = (campaignURL: string) => {
+const getReferrerObject = (referrerString?: string) => {
+  if (!referrerString) return {};
+  const referrerURL = new URL(referrerString);
+  const referrer = referrerURL.href.split('?')[0];
+  const referring_domain = referrerURL.hostname;
+  return { referrer, referring_domain };
+};
+const generateAttributionUserProps = (campaignURL: string, referrerString?: string) => {
+  const referrer = getReferrerObject(referrerString);
+
   const campaign = parseQueryString(campaignURL);
   const campaignObject: Campaign = {
     ...BASE_CAMPAIGN,
     ...campaign,
+    ...referrer,
   };
   const hasNoCampaign = Object.entries(campaign).length === 0;
   const hasAllCampaign = Object.entries(campaign).length === Object.entries(BASE_CAMPAIGN).length;
@@ -57,8 +67,8 @@ const generateAttributionUserProps = (campaignURL: string) => {
 export const generateEvent = (event_id: number, event_type: string): BaseEvent => {
   return {
     device_id: uuid,
-    event_id,
-    event_type,
+    event_id: event_id,
+    event_type: event_type,
     insert_id: uuid,
     ip: '$remote',
     language: 'en-US',
@@ -73,9 +83,9 @@ export const generateEvent = (event_id: number, event_type: string): BaseEvent =
   };
 };
 
-export const generateAttributionEvent = (event_id: number, campaignURL: string) => {
+export const generateAttributionEvent = (event_id: number, campaignURL: string, referrer?: string) => {
   const attributionEvent = generateEvent(event_id, '$identify');
-  attributionEvent.user_properties = generateAttributionUserProps(campaignURL);
+  attributionEvent.user_properties = generateAttributionUserProps(campaignURL, referrer);
   return attributionEvent;
 };
 
@@ -88,16 +98,7 @@ export const generateSessionEndEvent = (event_id: number) => {
 };
 
 const generatePageViewEventProps = (pageCounter: number, urlString: string, referrerString?: string) => {
-  let referrerObj = {};
-  if (referrerString) {
-    const referrerURL = new URL(referrerString);
-    const referrer = referrerURL.href.split('?')[0];
-    const referring_domain = referrerURL.hostname;
-    referrerObj = {
-      referrer,
-      referring_domain,
-    };
-  }
+  const referrer = getReferrerObject(referrerString);
 
   const url = new URL(urlString);
   const campaign = parseQueryString(urlString);
@@ -110,7 +111,7 @@ const generatePageViewEventProps = (pageCounter: number, urlString: string, refe
     '[Amplitude] Page Title': '',
     '[Amplitude] Page URL': url.href.split('?')[0],
     ...campaign,
-    ...referrerObj,
+    ...referrer,
   };
 };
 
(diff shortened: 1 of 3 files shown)


packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,2007 @@
...
+                min_id_length: undefined,
+              },
+            });
+            scope.done();
+            resolve();
+          }, 4000);
+        });
+      });
+
+      test('should track no campaign if no UTMs and no referrer', async () => {
```
izaaz, line 188:
```suggestion
      test('should track EMPTY campaign if no UTMs and no referrer', async () => {
```

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,2007 @@
...
+          sessionTimeout: 500,
+          flushIntervalMillis: 3000,
+        }).promise;
+
+        return new Promise<void>((resolve) => {
+          setTimeout(() => {
+            expect(payload).toEqual({
+              api_key: apiKey,
+              client_upload_time: event_upload_time,
+              events: [
```
izaaz, line 220: These tests are really nice. I'd hope we keep adding more such integration tests. To make that easy in the future, can we remove duplication here? We have a few tests and we are already at 2000 lines of code. The identify event is basically copied over 10 times so can we move those to methods instead?

Eg:
generateIdentifyEvent(utmCampaign) {
return {
    ... camapign identify event
}

so then you can use this method to remove all duplication.

Similarly methods like this will be helpful

```
function navigateTo(url: string, referrer: string) {
        Object.defineProperty(document, 'referrer', { value: referrer, configurable: true });
        const url = new URL(url);
        Object.defineProperty(window, 'location', {
          value: {
            hostname: url.hostname,
            href: url.href,
            pathname: url.pathname,
            search: url.search,
          },
          writable: true,
        });
}
```

yuhao900914: I created the helper functions in the integration test, but I feel we'd better create tests for those helper function?

packages/analytics-browser-test/test/web-attribution.test.ts
yuhao900914: Updated.

packages/analytics-browser-test/test/helpers.ts
```diff
@@ -0,0 +1,139 @@
...
+    return operationObject;
+  }, operationObject);
+
+  return user_properties;
+};
+
+export const generateEvent = (event_id: number, event_type: string): BaseEvent => {
+  return {
+    device_id: uuid,
+    event_id,
```
izaaz, line 60: can we add the property key like other fields here so we know its being set correctly?

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,603 @@
...
+        eventId = -1;
+      });
+
+      afterEach(() => {
+        cleanup();
+      });
+
+      test('should track all UTMs and referrers', async () => {
+        const url = 'https://www.example.com?utm_source=test_utm_source';
+        navigateTo(url);
```
izaaz, line 52: Should we also pass a referral here to make sure we test that referral is also tracked?

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,603 @@
...
+      let eventId = -1;
+      beforeEach(() => {
+        eventId = -1;
+      });
+
+      afterEach(() => {
+        cleanup();
+      });
+
+      test('should track all UTMs and referrers while processing any event', async () => {
```
izaaz, line 371: This is a check for tracking UTMs and referrers while processing any event and a new session right? WIthin the session, the process function should not trigger any campaign identifies

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,603 @@
...
+      let eventId = -1;
+      beforeEach(() => {
+        eventId = -1;
+      });
+
+      afterEach(() => {
+        cleanup();
+      });
+
+      test('should not track updated campaign during', async () => {
```
izaaz, line 490: Worth adding a comment here that this is for a SPA. Because when the page is reloaded, the init will track a campaign change

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,603 @@
...
+          defaultTracking,
+          sessionTimeout: 500,
+          flushIntervalMillis: 3000,
+        }).promise;
+
+        // update the url and fire the first event during the same session without refreshing the page
+        const newCampaignURL = 'https://www.example.com?utm_source=second_utm_source&utm_content=test_utm_content';
+        navigateTo(newCampaignURL);
+
+        client.track('test event after session timeout');
```
izaaz, line 513: This is after session timeout or before session timeout?

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -0,0 +1,603 @@
...
+              },
+            });
+            scope.done();
+            resolve();
+          }, 4000);
+        });
+      });
+    });
+
+    describe('during a session', () => {
```
izaaz, line 480: These all feel like SPA related tests. Can we rename this to during a SPA session?

izaaz (approved):
LGTM. Feel free to address the comments and merge the PR

packages/analytics-browser-test/test/web-attribution.test.ts
yuhao900914: yes, if the event comes from a new session, it should track all campaigns.  if the event still within the session, it should not track anything. that has been covered at `Web attribution -> process event -> during a session -> should not track updated campaign `  https://github.com/amplitude/Amplitude-TypeScript/pull/730/files#diff-bbd8238204219a8329ef6d342f9447527516495d50bb268c6eb8ee14d42caa26R490

izaaz: ah missed the line 361 where it says `in a new session`

packages/analytics-browser-test/test/web-attribution.test.ts
yuhao900914: It can also happen for not SPA, right? I will add a comment before `during a session` to mention that we are testing the SPA event specificly.

packages/analytics-browser-test/test/web-attribution.test.ts
yuhao900914: Nits. Fixed.

packages/analytics-browser-test/test/web-attribution.test.ts
yuhao900914: Added.

packages/analytics-browser-test/test/helpers.ts
yuhao900914: Fixed.

PR #731, opened 2024-04-25
feat(session replay): Fetch config from remote and apply to SDK for sampling

kwalker3690:
### Summary
This PR adds a set of new classes to fetch a config from remote, and join it to the local config. The remote config is fetched once at the start of each session, then stored in/read from IndexedDB. The remote config provides sample rate and whether capture is enabled. The joined config is generated upon SDK init, as well as when setSessionId is called.

### Checklist


packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.remoteConfigFetch = new SessionReplayRemoteConfigFetch({
+      localConfig: this.localConfig,
+      sessionIDBStore: this.sessionIDBStore,
+    });
+  }
+
+  async generateJoinedConfig(sessionId: number) {
+    const config: SessionReplayJoinedConfig = { ...this.localConfig };
+    // Special case here as optOut is implemented via getter/setter
+    config.optOut = this.localConfig.optOut;
```
jackson-amplitude, line 30: [nit] might consider a static method here that takes local and remote config. i.e something like
`generateJoinedConfig(localConfig, remoteConfig) -> config`. more explicit and easier to test. right now `config` is assumed to be local config.

packages/session-replay-browser/src/session-idb-store.ts
```diff
@@ -57,6 +58,35 @@ export class SessionReplaySessionIDBStore implements AmplitudeSessionReplaySessi
     }
   };
 
+  storeRemoteConfigForSession = async (sessionId: number, remoteConfig: SessionReplayRemoteConfig) => {
+    try {
+      await IDBKeyVal.update(this.storageKey, (sessionMap: IDBStore = {}): IDBStore => {
+        const session: IDBStoreSession = sessionMap[sessionId] || { ...defaultSessionStore };
```
jackson-amplitude, line 64: this might be the interface you are thinking about refactoring?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -37,13 +41,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+      loggerProvider: this.loggerProvider,
+      apiKey: apiKey,
     });
+    this.identifiers = new SessionIdentifiers(options, this.loggerProvider);
+    this.joinedConfigGenerator = new SessionReplayJoinedConfigGenerator(apiKey, options, this.sessionIDBStore);
+    if (this.identifiers.sessionId) {
+      this.config = await this.joinedConfigGenerator.generateJoinedConfig(this.identifiers.sessionId);
+    }
+    if (this.config) {
+      this.eventsManager = new SessionReplayEventsManager({
```
jackson-amplitude, line 55: thought/not blocking. 

Longer term, this makes me think we should decouple the config & events storage in the key. I.e. `<api-key>/sr/<session-id>/config` for the configs, and `<api-key>/sr/<session-id>/events` for sr events.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -81,7 +94,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
     this.stopRecordingAndSendEvents(this.identifiers.sessionId);
     this.identifiers.sessionId = sessionId;
     this.eventsManager && this.eventsManager.resetSequence();
-    this.recordEvents();
+    if (this.joinedConfigGenerator) {
```
jackson-amplitude, line 97: this kinda feels like we just want "re-initialize". would be nice if there is a way to just update config, and call initialize again.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -166,25 +189,28 @@ export class SessionReplay implements AmplitudeSessionReplay {
   }
 
   getShouldRecord(ignoreFocus = false) {
-    if (!this.identifiers || !this.config) {
+    if (!this.identifiers || !this.config || !this.identifiers.sessionId) {
       this.loggerProvider.error(`Session is not being recorded due to lack of config, please call sessionReplay.init.`);
       return false;
     }
+
+    if (!this.config.captureEnabled) {
```
jackson-amplitude, line 197: [nit] kinda wish "config" based decisions can be grouped together (i.e. captureEnabled + sampling being evaluated together) and separate from other things.

jackson-amplitude (approved):
lgtm! Mostly organizational comments, don't need to be addressed right away.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -166,25 +189,28 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
-    } else if (this.shouldOptOut()) {
-      if (this.identifiers.sessionId) {
-        this.loggerProvider.log(`Opting session ${this.identifiers.sessionId} out of recording due to optOut config.`);
-      }
-      return false;
-    } else if (!this.identifiers.sessionId) {
-      this.loggerProvider.warn(`Session is not being recorded due to lack of session id.`);
+    }
+
+    if (this.shouldOptOut()) {
```
jxiwang, line 212: If capture is enabled but the user opts out, then we don't capture at all?

jxiwang (approved):
LGTM

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Yep exactly!

packages/session-replay-browser/src/session-idb-store.ts
```diff
@@ -57,6 +58,35 @@ export class SessionReplaySessionIDBStore implements AmplitudeSessionReplaySessi
     }
   };
 
+  storeRemoteConfigForSession = async (sessionId: number, remoteConfig: SessionReplayRemoteConfig) => {
```
lewgordon-amplitude, line 61: How do we eventually clean these up? I'm thinking this eventually grows into a very large map.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -81,7 +94,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
     this.stopRecordingAndSendEvents(this.identifiers.sessionId);
     this.identifiers.sessionId = sessionId;
     this.eventsManager && this.eventsManager.resetSequence();
-    this.recordEvents();
+    if (this.joinedConfigGenerator) {
+      this.joinedConfigGenerator
+        .generateJoinedConfig(this.identifiers.sessionId)
+        .then((updatedConfig) => {
```
lewgordon-amplitude, line 100: Do we want to eventually refactor this to an async method?

packages/session-replay-browser/src/config/remote-config.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.attempts += 1;
+      const res = await fetch(server_url, options);
+      if (res === null) {
+        return this.completeRequest({ err: UNEXPECTED_ERROR_MESSAGE });
+      }
+      const parsedStatus = new BaseTransport().buildStatus(res.status);
+      switch (parsedStatus) {
+        case Status.Success:
+          return this.parseAndStoreConfig(sessionId, res);
+        case Status.Failed:
```
lewgordon-amplitude, line 91: Should we also retry on 429 or is that already built in?

packages/session-replay-browser/src/config/remote-config.ts
```diff
@@ -0,0 +1,121 @@
...
+      const options: RequestInit = {
+        headers: {
+          'Content-Type': 'application/json',
+          Accept: '*/*',
+        },
+        method: 'GET',
+      };
+      const server_url = `${this.getServerUrl()}?${urlParams.toString()}`;
+      this.attempts += 1;
+      const res = await fetch(server_url, options);
```
lewgordon-amplitude, line 83: Nit :)

```suggestion
      const serverUrl = `${this.getServerUrl()}?${urlParams.toString()}`;
      this.attempts += 1;
      const res = await fetch(serverUrl, options);
```

packages/session-replay-browser/src/config/remote-config.ts
```diff
@@ -0,0 +1,121 @@
...
+        default:
+          return this.completeRequest({ err: UNEXPECTED_NETWORK_ERROR_MESSAGE });
+      }
+    } catch (e) {
+      return this.completeRequest({ err: e as string });
+    }
+  };
+
+  retryFetch = async (sessionId: number): Promise<ISessionReplayRemoteConfig | void> => {
+    await new Promise((resolve) => setTimeout(resolve, this.attempts * this.retryTimeout));
```
lewgordon-amplitude, line 102: Usually it's good to add a bit of random jitter on retries, but it's very unlikely all clients will be retrying at the same time.

packages/session-replay-browser/src/config/remote-config.ts
```diff
@@ -0,0 +1,121 @@
...
+  SessionReplayRemoteConfig as ISessionReplayRemoteConfig,
+  SessionReplayRemoteConfigFetch as ISessionReplayRemoteConfigFetch,
+  SamplingConfig,
+  SessionReplayRemoteConfigAPIResponse,
+} from './types';
+
+const UNEXPECTED_NETWORK_ERROR_MESSAGE = 'Network error occurred, session replay remote config fetch failed';
+const SUCCESS_REMOTE_CONFIG = 'Session replay remote config successfully fetched';
+const MAX_RETRIES_EXCEEDED_MESSAGE = 'Session replay remote config fetch rejected due to exceeded retry count';
+export const REMOTE_CONFIG_SERVER_URL = 'https://sr-config.amplitude.com/config';
```
lewgordon-amplitude, line 16: We'll also want to use the CDN URL, but definitely TODO on the naming though I imagine this will stick.

```suggestion
export const REMOTE_CONFIG_SERVER_URL = 'https://sr-client-cfg.amplitude.com/config';
```

lewgordon-amplitude (approved):
LGTM! Nothing blocking. It'd be good to get a sense of how we eventually remove session ID remote config state.

packages/session-replay-browser/src/session-idb-store.ts
kwalker3690: We get rid of all session data for sessions older than 3 days here: https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/session-replay-browser/src/session-idb-store.ts#L111-L111, so this should get removed as a part of that process as well!

packages/session-replay-browser/src/config/remote-config.ts
kwalker3690: We made a decision with @jackson-amplitude a while ago to specifically only retry on 500, this was from a time of more uncertainty on our system stability, but I think it's ok to stick with for now. We could loosen up a bit if we decide we are more certain in our stability in upcoming weeks

packages/session-replay-browser/src/session-replay.ts
kwalker3690: https://amplitude.atlassian.net/browse/AMP-98766

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Definitely - will be doing this in my IDB refactor

packages/session-replay-browser/src/config/joined-config.ts
kwalker3690: True, what I was hoping though is that the consumer (session-replay.ts) would not have to know the difference between local and remote config - so it only has to deal with a single config that is generated by this method. The localConfig comes from the options passed to the SDK, and is created in this class, not exposed outside of this class

lewgordon-amplitude (approved):
LGTM!

PR #732, opened 2024-04-26
refactor(session replay) use custom IDB store for event management

kwalker3690:
### Summary

This PR migrates the SR SDK off of `idb-keyval` to a more robust IDB schema that will allow us more flexibility in the implementation of Targeted Replay Capture.

```
export interface SessionReplayDB extends DBSchema {
  sessionCurrentSequence: {
    key: number;
    value: {
      sessionId: number;
      events: Events;
    };
  };
  sequencesToSend: {
    key: number;
    value: SendingSequencesIDBInput;
    indexes: { sessionId: number };
  };
}
```

The current replay being captured will add events to `sessionCurrentSequence`. Once that list is large enough, as well as on blur, we will move that list to `sequencesToSend`, which will automatically provide it a sequence id. Then the list will be sent to the backend, upon a successful API response we will delete the list from `sequencesToSend`. In this model, IDB will be the source of truth for events (previously, they were also stored in memory).

https://www.loom.com/share/d3a5b4bc26b84f2d92d49458e9f7794f

### Checklist


packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -0,0 +1,308 @@
...
+    this.loggerProvider = loggerProvider;
+    this.apiKey = apiKey;
+  }
+
+  async initialize(sessionId?: number) {
+    const dbName = `${this.apiKey.substring(0, 10)}_amp_session_replay_events`;
+    this.db = await createStore(dbName);
+    this.timeAtLastSplit = Date.now(); // Initialize this so we have a point of comparison when events are recorded
+    await this.transitionFromKeyValStore(sessionId);
+  }
```
lewgordon-amplitude, line 100: My suggestion here is not to have a public constructor, but instead make it private and initialize is the true constructor that's async. Roughly, I'm thinking it'll look like this in the end.
```suggestion
db: IDBPDatabase<SessionReplayDB>;
  loggerProvider: ILogger;
  storageKey = '';
  maxPersistedEventsSize = MAX_EVENT_LIST_SIZE_IN_BYTES;
  interval = MIN_INTERVAL;
  timeAtLastSplit: number;

  private constructor({ loggerProvider, apiKey, db, timeAtLastSplit }: { loggerProvider: ILogger; apiKey: string; db: IDBPDatabase<SessionReplayDB>; timeAtLastSplit: number }) {
    this.loggerProvider = loggerProvider;
    this.apiKey = apiKey;
    this.db = db;
    this.timeAtLastSplit = timeAtLastSplit;
  }

  static async initialize({ loggerProvider, apiKey, sessionId }: { loggerProvider: ILogger; apiKey: string; sessionId?: number}): Promise<SessionReplayEventsIDBStore> {
    const dbName = `${apiKey.substring(0, 10)}_amp_session_replay_events`;
    const db = await createStore(dbName);
    const timeAtLastSplit = Date.now(); // Initialize this so we have a point of comparison when events are recorded
    const store = new SessionReplayEventsIDBStore({loggerProvider, apiKey, db, timeAtLastSplit})
    await store.transitionFromKeyValStore(sessionId);
    return store
  }
```

The reason I like this is that it can remove some possibly null fields and users of this class will never need to worry about the pattern of constructor then initialize.

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -0,0 +1,308 @@
...
+      let eventsToSend;
+      if (this.shouldSplitEventsList(sequenceEvents.events, event)) {
+        eventsToSend = sequenceEvents.events;
+        // set store to empty array
+        await tx.store.put({ sessionId, events: [event] });
+      } else {
+        // add event to array
+        const updatedEvents = sequenceEvents.events.concat(event);
+        await tx.store.put({ sessionId, events: updatedEvents });
+      }
```
lewgordon-amplitude, line 183: I think this should be fine for now, but I'm curious how well it scales long term. We may need a strategy where we need to write to multiple rows.

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -0,0 +1,308 @@
...
+            await Promise.all(
+              currentSessionSequences[sessionStore.currentSequenceId].events.map(async (event) => {
+                await this.addEventToCurrentSequence(numericSessionId, event);
+              }),
+            );
+          } else if (currentSessionSequences[numericSequenceId].status !== RecordingStatus.SENT) {
+            await this.storeSendingEvents(numericSessionId, currentSessionSequences[numericSequenceId].events);
+          }
+        }),
+      );
```
lewgordon-amplitude, line 249: It may be hard to get a sense of the upper bound of promises this may create, but it may be better to batch these. This is just in case this causes too many promises to fire.

lewgordon-amplitude (approved):
Looks good overall. Nothing hard blocking, but I think some suggestions might make the code trimmer.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -0,0 +1,108 @@
...
+      flushMaxRetries: config.flushMaxRetries,
+      apiKey: config.apiKey,
+      deviceId: deviceId,
+      sampleRate: config.sampleRate,
+      serverZone: config.serverZone,
+      onComplete: eventsIDBStore.cleanUpSessionEventsStore.bind(eventsIDBStore),
+    });
+  };
+
+  const sendCurrentSequenceEvents = ({ sessionId, deviceId }: { sessionId: number; deviceId: string }) => {
```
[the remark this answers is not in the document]
kwalker3690: And then just `void` it in the places it's called? Yeah thats doable - this and the other async fn you mentioned are called from places where we can't/shouldn't await on them (blur listener and rrweb record emit callback), but agreed that the await syntax is more readable

packages/session-replay-browser/src/events/events-idb-store.ts
kwalker3690: Like we need more than one current sequence? Or what do you mean by multiple rows?

packages/session-replay-browser/src/events/events-idb-store.ts
kwalker3690: Implementing it in a batchable way could also make it more readable... this code got so nested

packages/session-replay-browser/src/events/events-idb-store.ts
lewgordon-amplitude: I was thinking if we're doing a constant grab the array, add to it, then put it back there may be contention on the key. The idea was to potentially split the array across multiple rows since there's a bit of overhead in adding to that array.

packages/session-replay-browser/src/events/events-idb-store.ts
kwalker3690: I thought about this a lot too (probably too much haha), and tried an implementation like you're mentioning, based on the suggestion here: https://dev.to/somedood/the-proper-way-to-write-async-constructors-in-javascript-1o8c, sounds like the same approach. I ended up backing away from it earlier because at the time I found it harder to work with/wrap my head around than the create fn I exposed at the bottom of this file, but I'm definitely not against it. I'm going to leave it as is for now, but I'm totally open to this pattern moving forward and definitely happy to consider refactoring it down the line

PR #734, opened 2024-05-03
fix(core): correctly await promises called before init

kwalker3690:
### Summary

We've observed that the SR plugin is not correctly awaited when added before the client is initialized. This is because the SR plugin performs some async behavior in setup, and the client timeline was not awaiting on the `.promise` of the `AmplitudeReturn` of `client.add`.

See the unit test in this PR for a good representation of the issue - I confirmed that this test fails without this fix.

### Checklist


justin-fiedler (approved):
Thanks @kwalker3690. LGTM

PR #735, opened 2024-05-03
feat(session replay): add ability to wait for setSessionId process to complete

kwalker3690:
### Summary
Now that setSessionId has an async action, we should wait for it to complete before proceeding in the plugin. This PR adds an optional async interface to setSessionId via the AmplitudeReturn interface, `await setSessionId().promise`. This will allow us to properly tag start session events with the correct replay id, after fetching the remote config to ensure the replays shoul d be captured.

### Checklist


lewgordon-amplitude (approved):
LGTM!

PR #736, opened 2024-05-04
fix: sends events in order

yuhao900914:
### Summary
[Jira AMP-98809
](https://amplitude.atlassian.net/browse/AMP-98809)

This change is for making sure the events sent into the backend are in order.
Previously in the Typescript SDK, we 
- Sent different promises in parallel using Promise.all, which led to race conditions.
- Scheduled individual events based on their own timeout, which also resulted in events being sent out of order.

Current behavior:
- In the event of an error, we filter out problematic events while keeping all other events in the queue. These events will be retried with the next trigger.

### Checklist

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -91,11 +91,19 @@ export class Destination implements DestinationPlugin {
   filterTriableList() {
     this.queue = this.queue.filter((context) => {
       if (context.attempts < this.config.flushMaxRetries) {
-        context.attempts += 1;
         return true;
       }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
       return false;
+    });
+  }
+
+  increaseAttempts(list: Context[]) {
+    const updateEventSet = new Set(list.map((context) => context.event.insert_id));
+    this.queue.forEach((context) => {
+      if (updateEventSet.has(context.event.insert_id)) {
+        context.attempts += 1;
+      }
     });
   }
 
@@ -143,6 +151,7 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(this.queue, this.config.flushQueueSize);
     for (const batch of batches) {
+      this.increaseAttempts(batch);
       const responseStatus = await this.send(batch, useRetry);
       // To keep the order of events sending to the backend, stop sending the other requests while error happending.
       // All the events are still queued, and will be retried with next trigger.
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -52,6 +52,8 @@ export class Destination implements DestinationPlugin {
   config: Config;
   private scheduled: ReturnType<typeof setTimeout> | null = null;
   queue: Context[] = [];
+  throttled = false;
+  throttleTimeout = 30000;
 
   async setup(config: Config): Promise<undefined> {
     this.config = config;
@@ -106,7 +108,7 @@ export class Destination implements DestinationPlugin {
   }
 
   sendEventsIfReady() {
-    if (this.config.offline || this.scheduled) {
+    if (this.config.offline || this.scheduled || this.throttled) {
       return;
     }
 
@@ -124,6 +126,11 @@ export class Destination implements DestinationPlugin {
     // Skip flush if offline
     if (this.config.offline) {
       this.config.loggerProvider.debug('Skipping flush while offline.');
+      return;
+    }
+
+    if (this.throttled) {
+      this.config.loggerProvider.debug('Skipping flush while throttled. Will retry after 30 seconds.');
       return;
     }
 
@@ -158,6 +165,10 @@ export class Destination implements DestinationPlugin {
     if (!this.config.apiKey) {
       this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
       return true;
+    }
+
+    if (this.throttled) {
+      return false;
     }
 
     const payload = {
@@ -280,14 +291,16 @@ export class Destination implements DestinationPlugin {
   handleRateLimitResponse(res: RateLimitResponse, list: Context[]): boolean {
     const dropUserIds = Object.keys(res.body.exceededDailyQuotaUsers);
     const dropDeviceIds = Object.keys(res.body.exceededDailyQuotaDevices);
+    // Array of indexes in the events array indicating events whose user_id or device_id were throttled.
+    const throttledIndexSet = new Set(res.body.throttledEvents);
     const dropUserIdsSet = new Set(dropUserIds);
     const dropDeviceIdsSet = new Set(dropDeviceIds);
-
-    const retry = list.filter((context) => {
+    const retry = list.filter((context, index) => {
       if (
         (context.event.user_id && dropUserIdsSet.has(context.event.user_id)) ||
         (context.event.device_id && dropDeviceIdsSet.has(context.event.device_id))
       ) {
+        throttledIndexSet.delete(index);
         this.fulfillRequest([context], res.statusCode, res.body.error);
         return;
       }
@@ -298,6 +311,16 @@ export class Destination implements DestinationPlugin {
     if (retry.length > 0) {
       // log intermediate event status before retry
       this.config.loggerProvider.warn(getResponseBodyString(res));
+
+      // Has not dropped throttled event
+      if (throttledIndexSet.size > 0 && !this.throttled) {
+        this.throttled = true;
+        setTimeout(() => {
+          this.throttled = false;
+          this.sendEventsIfReady();
+        }, this.throttleTimeout);
+      }
+
       return false;
     }
 
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -97,13 +97,13 @@ export class Destination implements DestinationPlugin {
   }
 
   sendEventIfReady() {
-    this.filterTriableList();
-    if (this.scheduled || this.config.offline || this.queue.length === 0) {
+    if (this.scheduled || this.config.offline) {
       return;
     }
     // if batching enabled, check if min threshold met for batch size
     if (this.queue.length >= this.config.flushQueueSize) {
       void this.flush(true);
+      // return;
     }
 
     this.scheduled = setTimeout(() => {
@@ -135,6 +135,8 @@ export class Destination implements DestinationPlugin {
   }
 
   async send(list: Context[], useRetry = true) {
+    this.filterTriableList();
+
     if (!this.config.apiKey) {
       return this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
     }
@@ -155,7 +157,6 @@ export class Destination implements DestinationPlugin {
     try {
       const { serverUrl } = createServerConfig(this.config.serverUrl, this.config.serverZone, this.config.useBatch);
       const res = await this.config.transportProvider.send(serverUrl, payload);
-
       if (res === null) {
         this.fulfillRequest(list, 0, UNEXPECTED_ERROR_MESSAGE);
         return;
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -24,6 +24,8 @@ import { createServerConfig } from '../config';
 import { UUID } from '../utils/uuid';
 import { chunk } from '../utils/chunk';
 
+const retryStatus = new Set([Status.Invalid, Status.PayloadTooLarge, Status.RateLimit]);
+
 function getErrorMessage(error: unknown) {
   if (error instanceof Error) return error.message;
   return String(error);
@@ -83,7 +85,7 @@ export class Destination implements DestinationPlugin {
   addToQueue(...list: Context[]) {
     this.queue = this.queue.concat(list);
     this.saveEvents();
-    this.sendEventIfReady();
+    this.sendEventsIfReady();
   }
 
   filterTriableList() {
@@ -97,8 +99,8 @@ export class Destination implements DestinationPlugin {
     });
   }
 
-  sendEventIfReady() {
-    if (this.scheduled || this.config.offline) {
+  sendEventsIfReady() {
+    if (this.config.offline) {
       return;
     }
 
@@ -106,10 +108,14 @@ export class Destination implements DestinationPlugin {
       void this.flush(true);
     }
 
+    if (this.scheduled) {
+      return;
+    }
+
     this.scheduled = setTimeout(() => {
       void this.flush(true).then(() => {
         if (this.queue.length > 0) {
-          this.sendEventIfReady();
+          this.sendEventsIfReady();
         }
       });
     }, this.config.flushIntervalMillis);
@@ -137,13 +143,19 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(this.queue, this.config.flushQueueSize);
     for (const batch of batches) {
-      await this.send(batch, useRetry);
+      const responseStatus = await this.send(batch, useRetry);
+      // To keep the order of events sending to the backend, stop sending the other requests while error happending.
+      // All the events are still queued, and will be retried with next trigger.
+      if (retryStatus.has(responseStatus) && useRetry) {
+        break;
+      }
     }
   }
 
   async send(list: Context[], useRetry = true) {
     if (!this.config.apiKey) {
-      return this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
+      this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
+      return Status.Invalid;
     }
 
     const payload = {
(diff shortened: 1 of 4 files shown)


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -78,14 +74,19 @@ export class Destination implements DestinationPlugin {
...
       };
       void this.addToQueue(context);
     });
   }
 
-  getTryableList(list: Context[]) {
-    return list.filter((context) => {
+  addToQueue(...list: Context[]) {
+    this.queue = this.queue.concat(list);
+    this.sendEventIfReady();
```
izaaz, line 84: Should we save before we send? This is to make sure an error in `sendEventIfReady` does not drop data

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -271,55 +252,37 @@ export class Destination implements DestinationPlugin {
...
-
-  handleOtherResponse(list: Context[]) {
-    const later = list.map((context) => {
-      context.timeout = context.attempts * this.retryTimeout;
-      return context;
-    });
-
-    const tryable = this.getTryableList(later);
-    this.scheduleTryable(tryable);
+    // All the events are still queued, and will be retried with next trigger
```
izaaz, line 280: For rate limiting, shouldn't we reduce the flush queue size?
```
    this.config.flushQueueSize /= 2;
```

izaaz: What do the other SDKs do for rate limiting?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -220,7 +204,6 @@ export class Destination implements DestinationPlugin {
       default: {
         // log intermediate event status before retry
         this.config.loggerProvider.warn(`{code: 0, error: "Status '${status}' provided for ${list.length} events"}`);
-        this.handleOtherResponse(list);
```
izaaz, line 223: Why are we removing this?

yuhao900914: This is what we do for handle other response:
```
  handleOtherResponse(list: Context[]) {
    const later = list.map((context) => {
      context.timeout = context.attempts * this.retryTimeout;
      return context;
    });

    const tryable = this.getTryableList(later);
    this.scheduleTryable(tryable);
  }
```
This logic is:
1. setup the timeout for a single event
2. put the tryable event back

Right now we don't need those two, because 
1. schedule event based on a single event timeout will mess up the event order
2. the current logic is leaving those events in the queue and retrying later with the next trigger. 

So it's not needed anymore.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: old android -> do nothing
new android -> split file, triggerBackOff (increase the flush interval)
new ios -> di nothing
old javascript -> do nothing.

packages/analytics-core/src/plugins/destination.ts
izaaz: Nice! Thanks for simplifying this logic! :)

packages/analytics-core/src/plugins/destination.ts
yuhao900914: I checked the HTTP API [doc](https://www.docs.developers.amplitude.com/analytics/apis/http-v2-api/#429-response-toomanyrequestsfordeviceerror). It says `You should pause sending events for that user or device for a period of 30 seconds before retrying and continue retrying until you no longer receive a 429 response.`

I think only doing one thing, only updating the flushQueueSize or increasing the flush interval is not working.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Moved the saveEvents before sendEventIfReady.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -271,55 +252,37 @@ export class Destination implements DestinationPlugin {
...
       if (
         (context.event.user_id && dropUserIdsSet.has(context.event.user_id)) ||
         (context.event.device_id && dropDeviceIdsSet.has(context.event.device_id))
       ) {
         this.fulfillRequest([context], res.statusCode, res.body.error);
         return;
       }
-      if (throttledIndexSet.has(index)) {
-        context.timeout = this.throttleTimeout;
-      }
```
Mercy811, line 297: When 429, we should wait for 30 seconds before retry according to the [docs](https://www.docs.developers.amplitude.com/analytics/apis/http-v2-api/#429-response-toomanyrequestsfordeviceerror). If we want to enforce event order, we can just pause sending for the all events. Simply deleting it 
 would keep getting 429.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -95,65 +96,46 @@ export class Destination implements DestinationPlugin {
...
-    this.scheduleTryable(tryable, true);
-    this.saveEvents();
-  }
-
-  schedule(timeout: number) {
+  sendEventIfReady() {
     if (this.scheduled || this.config.offline) {
       return;
     }
+    // if batching enabled, check if min threshold met for batch size
```
Mercy811, line 103: when is batching not enabled and don't check min threshold? We can delete this comment if it's always enabled to ease confusion.

packages/analytics-core/test/plugins/destination.test.ts
Mercy811, line 1: Can we add a test to verify order is correct when retry?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -95,65 +96,46 @@ export class Destination implements DestinationPlugin {
...
-
-    await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
+    this.filterTriableList();
+    if (this.queue.length === 0) {
+      return;
+    }
 
-    this.scheduleTryable(later);
+    const currentFlushQueue = this.queue.slice(0, this.config.flushQueueSize);
+    await this.send(currentFlushQueue, useRetry);
```
Mercy811, line 138: `amplitude.flush()` now would only send `flushQueueSize` events.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Removed.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Yea, you are right. Put a for loop and await on this.send.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Thanks for brining this up. I also find that. However, pausing sending events for 30 minutes will blow the storage if there are a lot of requests per second, isn't it?  

This is how other SDK behave. 
old android -> do nothing
new android -> split file, triggerBackOff (increase the flush interval) -> this is wrong.
new ios -> di nothing
old javascript -> do nothing.

I make it align with others behavior.

Mercy811: Why pausing sending events for 30 **seconds** will blow the storage? 

> This is how other SDK behave.
old android -> do nothing
new android -> split file, triggerBackOff (increase the flush interval) -> this is wrong.
new ios -> di nothing
old javascript -> do nothing.

Amplitude-Swift just wait for the next flush [here](https://github.com/amplitude/Amplitude-Swift/blob/0cd014a332d098e70f760dfc2b0cb75f5ef74abe/Sources/Amplitude/Utilities/PersistentStorageResponseHandler.swift#L113-L114). But I do think this is a bug. @crleona wdyt?

packages/analytics-core/src/plugins/destination.ts
Mercy811: Thanks Alyssa for the change at https://github.com/amplitude/Amplitude-TypeScript/pull/736/commits/ae271bdd9bbcfc736de0565a0545acadbaca4b06. But we might need a more complex logic to enforce order when flush().

Edge cases: default configuration. 60 events in storage and manually call flush()
- batch 1 has an identify event
- batch 2 has only normal events
Request 1 with batch 1 failed but request 2 with batch 2 succeeded. So batch 2 is deleted from the storage while batch 1 will retry next time. 

We might need to check each response to decide whether to continue.

packages/analytics-core/src/plugins/destination.ts
crleona: So on a 429 the legacy iOS SDK does [exponential backoff](https://github.com/amplitude/Amplitude-iOS/blob/bfdec453a31fd35942a619ef472fcf2f09e2313a/Sources/Amplitude/Amplitude.m#L1172) on a 429, the new Android SDK looks like it just [drops](https://github.com/amplitude/Amplitude-Kotlin/blob/5c155e2b7c0c6b55ab9870469ce4a05e79275ce8/core/src/main/java/com/amplitude/core/utilities/FileResponseHandler.kt#L122) the file (which is definitely bad!).

I'm definitely not concerned with increasing storage size too much if we do backoff, we're probably talking thousands of events to get past a few MB in storage, which will easily fit in memory or a file.

That said, I'd still expect us to make forward progress uploading events without backoff as a rejected request shouldn't count toward our quota (though, our specific implementation may not match this assumption). We'd just expect more retries until one eventually goes through.

I would recommend leaving this in though (perhaps even moving to an exponential backoff model). It's a defense against misconfigured clients (perhaps with a too small flush interval), and it's not like an earlier request is going to do anything but 429 again.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -95,65 +97,48 @@ export class Destination implements DestinationPlugin {
...
-  }
-
-  addToQueue(...list: Context[]) {
-    const tryable = this.getTryableList(list);
-    this.scheduleTryable(tryable, true);
-    this.saveEvents();
-  }
-
-  schedule(timeout: number) {
+  sendEventIfReady() {
```
justin-fiedler, line 100:
```suggestion
  sendEventsIfReady() {
```

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -271,55 +255,37 @@ export class Destination implements DestinationPlugin {
...
-
-    const tryable = this.getTryableList(later);
-    this.scheduleTryable(tryable);
+    // All the events are still queued, and will be retried with next trigger
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
-    this.removeEvents(list);
     list.forEach((context) => context.callback(buildResult(context.event, code, message)));
+    this.removeEvents(list);
```
justin-fiedler, line 288: Why move this after callbacks? I think it would be better to finalize the state before calling callbacks.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: updated.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: That's not an intended change. Updated.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: That makes sense. I made `send` function return a status, and based on that stop sending other batches if there is an issue. Please check. Thanks.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Thanks for the feedback. Align offline. We determine to hold 30 seconds. I will add this in a separate PR.

packages/analytics-core/test/plugins/destination.test.ts
yuhao900914: Added several tests in packages/analytics-core/test/plugins/destination.test.ts.

packages/analytics-core/src/plugins/destination.ts
izaaz: Btw, we don't need to wait for 30 seconds before sending events again. Our default EPS is 30 events per second. So if sending an event fails, trying again within 1 second should be sufficient.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -78,87 +77,94 @@ export class Destination implements DestinationPlugin {
...
+    }
 
-    this.scheduleTryable(later);
+    const batches = chunk(this.queue, this.config.flushQueueSize);
+    for (const batch of batches) {
+      this.increaseAttempts(batch);
+      const responseStatus = await this.send(batch, useRetry);
+      // To keep the order of events sending to the backend, stop sending the other requests while error happending.
+      // All the events are still queued, and will be retried with next trigger.
+      if (retryStatus.has(responseStatus) && useRetry) {
```
izaaz, line 158: Would it make sense to check for success instead of checking for failure? There is only one success mode and many failure modes. If a new failure more is added later, if someone does not update this list, we'll end up with events being sent out of order.

izaaz: Why are we checking for retry when breaking here?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -78,87 +77,94 @@ export class Destination implements DestinationPlugin {
...
+        break;
+      }
+    }
   }
 
   async send(list: Context[], useRetry = true) {
     if (!this.config.apiKey) {
-      return this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
+      this.fulfillRequest(list, 400, MISSING_API_KEY_MESSAGE);
+      return Status.Invalid;
```
izaaz, line 167: Should we add the return type to the function's signature too?

packages/analytics-core/src/plugins/destination.ts
izaaz: nvm. just read the doc that Xinyi shared.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: > Why are we checking for retry when breaking here?
The previous logic is `useRetry = true` when SDK auto flush. And if the customer manually call `flush()`, `useRetry = false`. And the call will be directly `fulfill` without dealing with the response.
So here adding the `useRetry` to only break and resend later if useRetry = true.

yuhao900914: > Would it make sense to check for success instead of checking for failure? There is only one success mode and many failure modes. If a new failure more is added later, if someone does not update this list, we'll end up with events being sent out of order.

There are 2 more status we don't retry,   `Status.Failed` and `Status.Unknown`,  I think either way we need to have a list.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Added

packages/analytics-core/src/plugins/destination.ts
izaaz: There's an implicit dependency between the `send` function and this break statement here. Ideally, if the send function fulfilled the request, you can continue. If it did not fulfill it, then you cannot. Why can't that be returned and checked instead of statuses and useRetry?

For example, if the API key was missing, the send function will fulfill the request. Why should we break from the loop here and not send the rest? Similarly, if there was an unexpected error message, the items in the queue have been cleared. Why are we not moving forward?

Mercy811: Order of requests of batches should be enforced if we send the next batch once the previous one is fulfilled regardless of the response and whether events in previous request are deleted or not.

izaaz: @Mercy811 not sure if you're agreeing with me or disagreeing :D fulfill == message deleted from storage right?

Mercy811: Opinion 1:
> Would it make sense to check for success instead of checking for failure? 

Opinion 2:
> Ideally, if the send function fulfilled the request, you can continue. If it did not fulfill it, then you cannot. 

@izaaz I agree with your second opinion but not the first one

izaaz: Thanks @Mercy811. We are aligned. I had to look deeper to understand when we fulfill requests. So its not sufficient to just look at success vs failure.

yuhao900914: Thanks @izaaz @Mercy811 , 
Yes, the invalid status also includes unexpected errors and missing API, those shouldn't be retried.
It makes sense to continue if the request has been fulfilled. 
Fixed in the [last commit](https://github.com/amplitude/Amplitude-TypeScript/pull/736/commits/562735800bc7d54bc9e78b73864b7bd5e8425f56). Please check.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -179,61 +188,64 @@ export class Destination implements DestinationPlugin {
...
         // log intermediate event status before retry
         this.config.loggerProvider.warn(`{code: 0, error: "Status '${status}' provided for ${list.length} events"}`);
-        this.handleOtherResponse(list);
         break;
       }
     }
+    return isFullfilledRequest;
   }
 
   handleSuccessResponse(res: SuccessResponse, list: Context[]) {
```
Mercy811, line 240: nit: add return type. And the same for other `handleXXXresponse()` methods.
```suggestion
  handleSuccessResponse(res: SuccessResponse, list: Context[]): boolean {
```

Mercy811 (approved):
Thank you Alyssa. Overall looks good to me. See nits.

packages/analytics-core/src/plugins/destination.ts
yuhao900914: Added the rate limit handler in this [pr](https://github.com/amplitude/Amplitude-TypeScript/pull/754).

PR #738, opened 2024-05-06
fix: location is polyfilled to an empty object

Mercy811:
### Summary

https://amplitude.atlassian.net/browse/AMP-97294

If `window.location` is set to be an empty object in polyfills, 

```
global.window.location = global.window.location || {};
```
an error occurs "TypeError: Cannot read property 'split' of undefined".

This PR patches this edges cases with tests.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -301,11 +301,15 @@ export const createTransport = (transport?: TransportType) => {
 };
 
 export const getTopLevelDomain = async (url?: string) => {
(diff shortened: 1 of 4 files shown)


yuhao900914:
The issue is on React Natvie, which doesn't has version 2.0 yet. Should there be another pr to do the same fix for v1.x?

izaaz (approved):
Should we do the same thing in the v1.x branch as well for react native?

Mercy811:
Thanks @yuhao900914 @izaaz for reviewing and calling out. Here is the PR of cherry picked commits from this PR for v1.x.

justin-fiedler (approved):

PR #742, opened 2024-05-09
feat(analytics remote config): introduce generic remote config fetch and storage package

kwalker3690:
### Summary

Introduce a generic version of SR's remote config fetch package, to be shared across different products/plugins. This package handles fetching remote config based on passed configKeys, and storing the results in Indexeddb. Config is stored in it's own DB, with object stores for each config namespace (ie, Session Replay config will be retrievable independently from Browser SDK config).

### Checklist

### DIFF
--- a/packages/analytics-remote-config/src/remote-config-idb-store.ts
+++ b/packages/analytics-remote-config/src/remote-config-idb-store.ts
@@ -1,0 +1,87 @@
+import { Logger as ILogger } from '@amplitude/analytics-types';
+import { DBSchema, openDB } from 'idb';
+import {
+  ConfigNamespace,
+  RemoteConfig,
+  RemoteConfigAPIResponse,
+  RemoteConfigIDBStore,
+  SessionReplayRemoteConfig,
+} from './types';
+
+export interface RemoteConfigDB extends DBSchema {
+  [ConfigNamespace.SESSION_REPLAY]: {
+    key: string;
+    value: SessionReplayRemoteConfig;
+    indexes: { sessionId: number };
+  };
+  lastFetchedSessionId: {
+    key: string;
+    value: number;
+  };
+}
+
+export const createStore = async (dbName: string) => {
+  return await openDB<RemoteConfigDB>(dbName, 1, {
+    upgrade: (db) => {
+      if (!db.objectStoreNames.contains(ConfigNamespace.SESSION_REPLAY)) {
+        db.createObjectStore(ConfigNamespace.SESSION_REPLAY);
+      }
+      if (!db.objectStoreNames.contains('lastFetchedSessionId')) {
+        db.createObjectStore('lastFetchedSessionId');
+      }
+    },
+  });
+};
+
+export const createRemoteConfigIDBStore = async ({
+  loggerProvider,
+  apiKey,
+}: {
+  loggerProvider: ILogger;
+  apiKey: string;
+}): Promise<RemoteConfigIDBStore> => {
+  const dbName = `${apiKey.substring(0, 10)}_amp_config`;
+  const db = await createStore(dbName);
+
+  const storeRemoteConfig = async (remoteConfig: RemoteConfigAPIResponse, sessionId?: number) => {
+    try {
+      if (sessionId) {
+        await db.put<'lastFetchedSessionId'>('lastFetchedSessionId', sessionId, 'sessionId');
+      }
+      let configNamespace: ConfigNamespace;
+      for (configNamespace in remoteConfig.configs) {
+        await db.put(configNamespace, {
+          ...remoteConfig.configs[configNamespace],
+        });
+      }
+    } catch (e) {
+      loggerProvider.warn(`Failed to store remote config: ${e as string}`);
+    }
+  };
+
+  const getLastFetchedSessionId = async (): Promise<number | void> => {
+    try {
+      const lastFetchedSessionId = await db.get('lastFetchedSessionId', 'sessionId');
+      return lastFetchedSessionId;
+    } catch (e) {
+      loggerProvider.warn(`Failed to fetch lastFetchedSessionId: ${e as string}`);
+    }
+    return undefined;
+  };
+
+  const getRemoteConfig = async (configNamespace: ConfigNamespace, key: string): Promise<RemoteConfig | void> => {
+    try {
+      const config = await db.get(configNamespace, key);
+      return config;
+    } catch (e) {
+      loggerProvider.warn(`Failed to fetch remote config: ${e as string}`);
+    }
+    return undefined;
+  };
+
+  return {
+    storeRemoteConfig,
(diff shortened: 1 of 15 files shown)


packages/analytics-remote-config/src/remote-config-idb-store.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  ConfigNamespace,
+  RemoteConfig,
+  RemoteConfigAPIResponse,
+  RemoteConfigIDBStore,
+  SessionReplayRemoteConfig,
+} from './types';
+
+export interface RemoteConfigDB extends DBSchema {
+  [ConfigNamespace.SESSION_REPLAY]: {
```
izaaz, line 12: If we are making this generic, shouldn't the namespace be passed as a parameter to this class?

justin-fiedler: Agreed, I was thinking that the Remote config service would focus on loading a generic config (i.e. key value pairs) and then provide a callback for when the config has been updated. 
The SDK/plugins would be responsible for copying the loaded values (by generic keys) to it's own strongly typed local config.

kwalker3690: Yeah agreed with this, up until the definition of the IndexedDB store itself. I chose to have this package manage the IndexedDB store, rather than have each consumer manage their own (which would be pretty repetitive). I'm very hesitant to provide a generic interface to the definition of the IndexedDB store, as IndexedDB is a database that requires versioning when making updates, so I think we could run into trouble generalizing it.

Instead of thinking of this as wholly generic, I'm thinking of it as the client match to the [Remote Config Service](https://amplitude.atlassian.net/wiki/spaces/IG/pages/2412806453/Remote+Config+Service+Interface+Design+WIP). As a service, it knows about the existence of certain keys and schemas, and provides the typing for those.

packages/analytics-remote-config/src/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  sr_sampling_config: SessionReplaySamplingConfig;
+}
+
+export type RemoteConfig = SessionReplayRemoteConfig;
+
+export enum ConfigNamespace {
+  SESSION_REPLAY = 'sessionReplay',
+}
+
+export interface RemoteConfigAPIResponse {
```
kwalker3690, line 16: @izaaz @justin-fiedler Also if we make this package fully generic, where does a type like this live? This type is tightly coupled to the response from the Remote Config Service, so to me it makes the most sense to expose it through this package

packages/analytics-remote-config/src/remote-config-idb-store.ts
lewgordon-amplitude: I think I do agree with trying to make this a bit more generic. While I'm not certain how to provide a clean interface yet, I think it'll be easier to manage in the long run. This package provides a good amount of help with interfacing with the service and providing a caching mechanism.

However, if it starts to know more about the configs and we have stronger typing in this repo when new teams onboard they would need to:
1. Plan the config schema for their service and put that in the remote config service.
2. Come to this repo and adding the translation layer + typing. Hopefully they wouldn't need to know details of the caching mechanism (IndexedDB), since that adds room for error.
3. Change in their plugin/package/etc to utilize this new version with the new strict typing.

Now that would lead to this package being updated every time there is a model change, which can be fine, but seems like additional overhead if they're the only one using that new model.

Alternatively, if this could just focus on querying the service, providing a caching layer, and returning a generic object that the client would need to parse and strongly type. This way this package changes less often and clients of this package could operate independently without having two update two packages (this one and their own).

izaaz: > Alternatively, if this could just focus on querying the service, providing a caching layer, and returning a generic object that the client would need to parse and strongly type

I was kinda leaning towards what @lewgordon-amplitude mentioned too. 

Can the storage layer be something like this?
```
Namespace | Config json | sessionId
```

The remote config store takes in the namespace, version and session id as input, checks if the data is already exists for this combination and returns it. If not, fetches the latest data, persists it and returns it.

The remote config data is ephemeral. We should never have to do a migration. We can pretty much nuke it and fetch from the remote service again if there's any change in the schema.

izaaz: But @kwalker3690 , pretty sure you've thought about this and since you've been in the thick of the implementation, let us know if there's something that we are missing.

packages/analytics-remote-config/src/types.ts
izaaz: If this is super generic, I'd imagine the type lives closer to the code that cares about this type.

packages/analytics-remote-config/src/remote-config-idb-store.ts
izaaz: If a new namespace comes in, the caller needs to pass in the new namespace and this component can construct the API to call and fetch this data and store it in indexdb.

izaaz:
I am assuming this code with move to the main branch at some point (it can happen later if you prefer)

izaaz (approved):

kwalker3690:
Going to merge this as is to do some more exploration on the feedback in this PR, as I need to publish the package once to be able to consume it in the SR SDK

Mercy811:
Hi @izaaz, why do we need to move it to the main branch? Main branch is for v2 packages. We can just keep `@amplitude/analytics-remote-config` in v1.x branch and use it in `@amplitude/analytics-browser` v2.x.

izaaz:
> Hi @izaaz, why do we need to move it to the main branch? Main branch is for v2 packages. We can just keep `@amplitude/analytics-remote-config` in v1.x branch and use it in `@amplitude/analytics-browser` v2.x.

Its mostly hard to manage two different branches with production code. Given that its a new package, is there any harm in starting with version 2.x? When working with browser v2, having all the code it references in the same branch keeps things simple. 

Aside from having a version number v1.x, what other reason do we have to keep it in this branch?

kwalker3690:
@izaaz the issue is that it breaks the existing (very implicit) pattern defined here. This is the reason we were told to put the session replay packages in the v1.x branch as well, because they were not at v2 yet, and once they are at v2, they can be bumped up to the main branch. We can definitely revisit this decision as an overall guidance for this repo, but I hesitate to do it without a broader discussion (any some documentation on the decision would help too!)

Mercy811:
I agree it's hard to jumping between two different branches. But I'd like to keep it on branch v1.x until all packages (SR and SR plugin) depend on `@amplitude/analytics-remote-config` are moved it main branch.

izaaz:
Sounds good @Mercy811 and @kwalker3690. We can keep it as is then. Not a hard blocker to move it to main. As long as it works for y'all, being in v1.x is fine.

PR #744, opened 2024-05-10
fix(core): api calls should queue while awaiting queued promises

kwalker3690:
### Summary
I observed that it is possible to fire a track call while awaiting on a plugin to be setup, and in that case, the track call is dropped, as the call is not added to the queue. Instead it's processed immediately, but in the Browser SDK, the Destination plugin has not been registered, so the track call is not sent to the backend. 

This PR adds a property to core-client to determine if it has processed all functions in the queue. This property can be used to determine whether any API call should be added to the queue or processed immediately.

### Checklist

### DIFF
--- a/packages/analytics-core/test/core-client.test.ts
+++ b/packages/analytics-core/test/core-client.test.ts
@@ -9,6 +9,11 @@ async function runScheduleTimers() {
   await new Promise(process.nextTick);
   jest.runAllTimers();
 }
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 describe('core-client', () => {
   const success = { event: { event_type: 'sample' }, code: 200, message: Status.Success };
   const badRequest = { event: { event_type: 'sample' }, code: 400, message: Status.Invalid };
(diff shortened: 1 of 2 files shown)


packages/analytics-core/test/core-client.test.ts
```diff
@@ -156,6 +158,62 @@ describe('core-client', () => {
       expect(setupMock.mock.calls[0][0]).toEqual('firstPlugin');
       expect(setupMock.mock.calls[1][0]).toEqual('secondPlugin');
     });
+
+    test('should queue while promise is being awaited', async () => {
+      const client = new AmplitudeCore();
+      // const register = jest.spyOn(client.timeline, 'register');
```
Mercy811, line 164: Nit: delete?

packages/analytics-core/test/core-client.test.ts
```diff
@@ -156,6 +158,62 @@ describe('core-client', () => {
...
+      await setupMockResolve;
+      expect(process).not.toHaveBeenCalled();
+      await client.runQueuedFunctions('dispatchQ');
+      expect(process).toHaveBeenCalled();
+      expect(process).toHaveBeenCalledWith({ event_type: 'myEvent' });
+    });
+
+    test('should flush queue if new functions are queued while awaiting a promise', async () => {
+      const client = new AmplitudeCore();
+      // const register = jest.spyOn(client.timeline, 'register');
```
Mercy811, line 192: Nit: the same as above.

Mercy811 (approved):
Thanks Kelly for fixing it. It happens because configuration is ready in`core-client._init()` at the first step and functions in `q` is run at the third step. Adding a `isReady` flag can solve this issue.

kevinpagtakhan (approved):
lgtm! thank you kelly!

PR #745, opened 2024-05-13
refactor(remote config): consume remote config package in session replay

kwalker3690:
### Summary
This PR makes the remote config package more generic, as well as consumes it in Session Replay, replacing the internal package

### Checklist


packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -1,26 +1,24 @@
...
+  constructor(apiKey: string, options: SessionReplayOptions) {
     this.localConfig = new SessionReplayLocalConfig(apiKey, options);
-    this.remoteConfigFetch = new SessionReplayRemoteConfigFetch({
+  }
+
+  async initialize() {
+    this.remoteConfigFetch = await createRemoteConfigFetch<SessionReplayRemoteConfig>({
       localConfig: this.localConfig,
-      sessionIDBStore: this.sessionIDBStore,
+      configKeys: ['sessionReplay'],
```
justin-fiedler, line 21: Since config keys are shared across multiple SDKs and plugins I think it would be good to have a way to structure/track them. 

Before there was a ConfigNamespace in the Remote config package.
https://github.com/amplitude/Amplitude-TypeScript/pull/745/files#diff-b803bf7e9e25d71172f51d5b97ec5108f57dc7d115f57ac084e6cf3ad2504cb7L12-L14

I was thinking we could instead define a ConfigNamespace in the SDK/Plugin that uses it.

To enforce via the type system, what about making `ConfigNamspace` a tagging interface that should be implemented by users of the remote config? e.g.

~/remote-config/
```
export interface ConfigNamespace {
  key: string
}
```

~/session-replay-borwser/
```
export SessionReplayConfigNamespace extends ConfigNamespace {
  key: "sessionReplay"
}
```

packages/session-replay-browser/test/session-replay-factory.test.ts
```diff
@@ -1,11 +1,19 @@
...
+    getRemoteConfigMock = jest.fn();
+    jest.spyOn(RemoteConfigFetch, 'createRemoteConfigFetch').mockResolvedValue({
+      getRemoteConfig: getRemoteConfigMock,
+    });
+  });
   describe('getLogConfig', () => {
     test('return the log config if config defined on session replay', async () => {
       const sessionReplay = new SessionReplay();
-      await sessionReplay.init('apiKey', {});
+      sessionReplay.init('apiKey', {});
```
lewgordon-amplitude, line 16: Any reason why we wouldn't want to keep the await here such as?
```suggestion
      await sessionReplay.init('apiKey', {}).promise
```

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -29,7 +27,15 @@ export class SessionReplayJoinedConfigGenerator {
...
-      const samplingConfig = await this.remoteConfigFetch.getSamplingConfig(sessionId);
+      const sessionReplayConfig =
+        this.remoteConfigFetch &&
+        ((await this.remoteConfigFetch.getRemoteConfig(
+          'sessionReplay',
+          'sr_sampling_config',
+          sessionId,
+        )) as SessionReplayRemoteConfig['sr_sampling_config']);
+
+      const samplingConfig = sessionReplayConfig;
```
lewgordon-amplitude, line 38: Could we just use the one variable?

```suggestion
      const samplingConfig =
        this.remoteConfigFetch &&
        ((await this.remoteConfigFetch.getRemoteConfig(
          'sessionReplay',
          'sr_sampling_config',
          sessionId,
        )) as SessionReplayRemoteConfig['sr_sampling_config']);
```

lewgordon-amplitude (approved):
LGTM overall. Just a few questions/comments since I'm not as familiar with the codebase.

Mercy811 (approved):
Thanks Kelly for using `@amplitude/analytics-remote-config` in `@amplitude/session-replay-browser` and enabling TTL by storing the last fetched session Id as timeout in a separated meta DB.

packages/session-replay-browser/src/config/joined-config.ts
kwalker3690: I just tried this in code, and it ends up reading pretty confusingly, as I end up needing to refer to `ConfigNamespace['key'] in all places I'm using the ConfigNamespace type, as I really only need the string value. Then I tried just doing `export type ConfigNamespace = string`, but this can't be extended, and you can't create an interface from a primitive like string. 

Maybe I'm just getting blind to the right solution here in typescript at the moment (I've been wrestling with it all day), but I think this might overcomplicate things without a lot of payoff. I will say this is why I liked the enum I had before - we could add an enum for ConfigNamespace somewhere generic like the `@amplitude/analytics-types` package?

packages/session-replay-browser/src/config/joined-config.ts
kwalker3690: Whoops yes

packages/session-replay-browser/test/session-replay-factory.test.ts
kwalker3690: Yep can do that, that's a better idea, luckily for this test I don't think either will matter!

PR #748, opened 2024-05-15
docs: add comments and default value for configuration

yuhao900914:
### Summary

Add comments and the default value for configuration

Screenshot 2024-05-15 at 3 49 49 PM

### Checklist

### DIFF
--- a/packages/analytics-types/src/config/browser.ts
+++ b/packages/analytics-types/src/config/browser.ts
@@ -77,7 +77,7 @@ interface InternalBrowserConfig {
 
 export interface DefaultTrackingOptions {
   /**
-   * Enables/disables marketing attribution tracking or with detailed attribution options.
+   * Enables/disables marketing attribution tracking or config with detailed attribution options.
    * @defaultValue `true`
    */
   attribution?: boolean | AttributionOptions;
@@ -105,17 +105,17 @@ export interface DefaultTrackingOptions {
 
 export interface TrackingOptions {
   /**
-   * Enables/disables ip address.
+   * Enables/disables ip address tracking.
    * @defaultValue `true`
    */
   ipAddress?: boolean;
   /**
-   * Enables/disables the language.
+   * Enables/disables language tracking.
    * @defaultValue `true`
    */
   language?: boolean;
   /**
-   * Enables/disables plantform.
+   * Enables/disables plantform tracking.
    * @defaultValue `true`
    */
   platform?: boolean;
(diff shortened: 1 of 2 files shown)


izaaz (approved):

packages/analytics-types/src/config/browser.ts
```diff
@@ -7,17 +7,64 @@ import { PageTrackingOptions } from '../page-view-tracking';
...
+   */
   partnerId?: string;
+  /**
+   * The custom Session ID for the current session.
+   * @defaultValue `timestamp`
+   */
   sessionId?: number;
+  /**
+   * The period of inactivity from the last tracked event before a session expires in milliseconds.
+   * @defaultValue `1800000`
```
Mercy811, line 47: can we add delimiter for readability? 
```suggestion
   * @defaultValue `1,800,000` (30 minutes)
```

packages/analytics-types/src/config/browser.ts
```diff
@@ -26,34 +73,97 @@ interface InternalBrowserConfig {
   lastEventTime?: number;
   lastEventId?: number;
   transportProvider: Transport;
-  pageCounter?: number;
```
Mercy811, line 29: Is there any cases where customers need to set it at `init()`? If no, we should keep it as internal.

Mercy811 (approved):
Thanks Alyssa for adding in-editor comments. This is really useful.

We can also add dev doc links to each configuration for users need more details like [this](https://amplitude.slack.com/archives/C04HSDQA2VB/p1712607841384209?thread_ts=1712521462.343889&cid=C04HSDQA2VB).

packages/analytics-types/src/config/browser.ts
yuhao900914: Nope. However, this is bundled with page view plugin. We set the page counter in the page view plugin, we cannot access pageCounter if it's internal.

packages/analytics-types/src/config/browser.ts
yuhao900914: Updated.

yuhao900914:
> Thanks Alyssa for adding in-editor comments. This is really useful.
> 
> We can also add dev doc links to each configuration for users need more details like [this](https://amplitude.slack.com/archives/C04HSDQA2VB/p1712607841384209?thread_ts=1712521462.343889&cid=C04HSDQA2VB).


Added in this commit.

PR #753, opened 2024-05-20
feat: support sessionId from url for cross domain tracking

Mercy811:
### Summary
[AMP-93936]

This PR allows the Browser SDK to get and set sessionId from `ampSessionId` url parameter of the url (`window.location`). This allows the same sessionId and thus continuous user journey for cross domain tracking. 

This PR also changes how the SDK gets deviceId from the url.
- Previously, the SDK only supports `deviceId` url parameter
- Now, the SDK gets deviceId from `ampDeviceId` url parameter first and then falls back to `deviceId`

We add `amp` prefix because `deviceId` and `sessionId` are very generic names. 

For example, the SDK will automatically use the deviceId `my-device-id` and sessionId `1716245958483` from `https://www.example.com?ampDeviceId=my-device-id&ampSessionId=1716245958483`. 

It's recommended to follow the same sessionId format as the Browser SDK by using `Date.now()`. Because the SDK checks whether an event is in session every time an event is tracked.

```
# https://www.example.com?deviceId=my-device-id&sessionId=12345
amplitude.init(API_KEY)

# client.config.sessionId = 12345

amplitude.track("event")

# client.config.sessionId will set to Date.now() because "event" is not in the previous session 12345
# this logic is in browser-client.process()

```

#### Implementation
`useBrowserConfig()` should remain unchanged and return sessionId from cookie storage because `_init()` calls `setSessionId()` which starts a new session by comparing to the sessionId from storage. 

#### Note

Changing `packages/analytics-client-common/test/analytics-connector.test.ts` because the error below. This error happens even if running `yarn test` on the latest main branch. 
```
feat: support sessionId from url for cross domain tracking
Test suite failed to run

    test/analytics-connector.test.ts:36:78 - error TS2345: Argument of type '{ setUserId: () => ...; setDeviceId: () => ...; setUserProperties: () => ...; updateUserProperties: () => ...; commit: Mock<any, any>; }' is not assignable to parameter of type 'IdentityEditor'.
      Property 'setOptOut' is missing in type '{ setUserId: () => ...; setDeviceId: () => ...; setUserProperties: () => ...; updateUserProperties: () => ...; commit: Mock<any, any>; }' but required in type 'IdentityEditor'.

    36       jest.spyOn(instance.identityStore, 'editIdentity').mockReturnValueOnce(identityEditor);
                                                                                    ~~~~~~~~~~~~~~

      ../../node_modules/@amplitude/analytics-connector/dist/types/src/identityStore.d.ts:19:5
        19     setOptOut(optOut: boolean): IdentityEditor;
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        'setOptOut' is declared here.


```

### Checklist

[AMP-93936]: https://amplitude.atlassian.net/browse/AMP-93936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -89,16 +89,10 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
     const queryParams = getQueryParams();
-    this.setSessionId(
-      options.sessionId ??
-        (queryParams.ampSessionId
-          ? Number.isNaN(Number(queryParams.ampSessionId))
-            ? undefined
-            : Number(queryParams.ampSessionId)
-          : undefined) ??
-        this.config.sessionId ??
-        Date.now(),
-    );
+    const querySessionId = Number.isNaN(Number(queryParams.ampSessionId))
+      ? undefined
+      : Number(queryParams.ampSessionId);
+    this.setSessionId(options.sessionId ?? querySessionId ?? this.config.sessionId ?? Date.now());
 
     // Set up the analytics connector to integrate with the experiment SDK.
     // Send events from the experiment SDK and forward identifies to the
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -84,14 +84,18 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
-    // Priority 2: sessionId from url
+    // Priority 2: sessionId from url if it's Number
     // Priority 3: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
     const queryParams = getQueryParams();
     this.setSessionId(
       options.sessionId ??
-        (queryParams.ampSessionId ? Number(queryParams.ampSessionId) : undefined) ??
+        (queryParams.ampSessionId
+          ? Number.isNaN(Number(queryParams.ampSessionId))
+            ? undefined
+            : Number(queryParams.ampSessionId)
+          : undefined) ??
         this.config.sessionId ??
         Date.now(),
     );
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -13,6 +13,7 @@ import {
   isNewSession,
   isPageViewTrackingEnabled,
   WebAttribution,
+  getQueryParams,
 } from '@amplitude/analytics-client-common';
 import {
   BrowserClient,
@@ -83,10 +84,17 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
 
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
-    // Priority 2: last known sessionId from user identity storage
+    // Priority 2: sessionId from url
+    // Priority 3: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
-    this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
+    const queryParams = getQueryParams();
+    this.setSessionId(
+      options.sessionId ??
+        (queryParams.sessionId ? Number(queryParams.sessionId) : undefined) ??
+        this.config.sessionId ??
+        Date.now(),
+    );
 
     // Set up the analytics connector to integrate with the experiment SDK.
     // Send events from the experiment SDK and forward identifies to the
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -91,7 +91,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
     const queryParams = getQueryParams();
     this.setSessionId(
       options.sessionId ??
-        (queryParams.sessionId ? Number(queryParams.sessionId) : undefined) ??
+        (queryParams.ampSessionId ? Number(queryParams.ampSessionId) : undefined) ??
         this.config.sessionId ??
         Date.now(),
     );
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -219,7 +219,12 @@ export const useBrowserConfig = async (
 
   // Step 3: Reconcile user identity
   const deviceId =
-    options.deviceId ?? queryParams.deviceId ?? previousCookies?.deviceId ?? legacyCookies.deviceId ?? UUID();
+    options.deviceId ??
+    queryParams.ampDeviceId ??
+    queryParams.deviceId ??
+    previousCookies?.deviceId ??
+    legacyCookies.deviceId ??
+    UUID();
   const lastEventId = previousCookies?.lastEventId ?? legacyCookies.lastEventId;
   const lastEventTime = previousCookies?.lastEventTime ?? legacyCookies.lastEventTime;
   const optOut = options.optOut ?? previousCookies?.optOut ?? legacyCookies.optOut;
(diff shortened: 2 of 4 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -83,10 +84,17 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
-    // Priority 2: last known sessionId from user identity storage
+    // Priority 2: sessionId from url
+    // Priority 3: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
-    this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
+    const queryParams = getQueryParams();
+    this.setSessionId(
+      options.sessionId ??
+        (queryParams.sessionId ? Number(queryParams.sessionId) : undefined) ??
```
crleona, line 94: sessionId is a fairly generic name - is it possible that we may pick up unrelated sessionId parameters?

crleona: This would be good to take to team sync - deviceId probably needs to be less generic as well to reduce the chance of conflicts. An `amplitude` prefix on all of them would work, though we might want to go with something a bit shorter as urls have something like a 2083 character limit on chrome / ie.

yuhao900914: I remembered Zoom? or other customer has the issue, because, they use deviceId for internal purposes([security limitation](https://amplitude.slack.com/archives/C070W7YG9RR/p1714151781380809)). And, they are using GTM, so they set `amp_device_id=xxx` and call `setDeviceId` in the GTM tag.

izaaz: There are customers that use `deviceId` url query parameter already. So we should not change that. For consistency, can we use `sessionId` query parameter and provide an option to override these default query params with custom ones?

Mercy811: @izaaz I'm not very into custom sessionId query parameter as it introduces an extra configuration.

justin-fiedler: What about supporting `sessionId` and `ampSessionId`, and giving `ampSessionId` precedence?

This would allow us to keep the existing naming but have a workaround in case of naming conflicts.

Mercy811: @justin-fiedler I like `ampSessionId` but Chris has a good point of using `sessionId` might catch unwanted session Id parameter because it's generic.

justin-fiedler: @Mercy811 I was suggesting that, if set, we always use the "namespaced" session ID first (e.g. `ampSessionId`) and ignore `sessionId`. If `ampSessionId` is not set then we fallback to `sessionId`.

justin-fiedler: What about allowing `ampSessionId=` or `ampSessionId=-1` or something similar to allow preventing reading the `sessionId` if needed?

justin-fiedler: After the call, I get that we haven't supported `sessionId` before, in that cause I would say we name it `ampSessionId` and also add `ampDeviceId` (in addition to `deviceId`) as described [here](https://github.com/amplitude/Amplitude-TypeScript/pull/753#discussion_r1608762357). This way are docs can be clear that query params have a `amp` prefix, but existing customers can continue to use `deviceId`

Mercy811: Thanks team for discussion. I added the logic to use `ampDeviceId` and falls back to `deviceId` at https://github.com/amplitude/Amplitude-TypeScript/pull/753/commits/368f4cd0f9a4fc6821b90c0f88526e83d8d30bf4

packages/analytics-browser/src/browser-client.ts
```diff
@@ -83,10 +84,17 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
-    // Priority 2: last known sessionId from user identity storage
+    // Priority 2: sessionId from url
+    // Priority 3: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
-    this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
+    const queryParams = getQueryParams();
+    this.setSessionId(
+      options.sessionId ??
+        (queryParams.ampSessionId ? Number(queryParams.ampSessionId) : undefined) ??
```
justin-fiedler, line 94: Not 100% that this convert Number logic works as expected.

```
let ampSessionId = "hi"
let sessionId = (ampSessionId ? Number(ampSessionId) : undefined) ?? 42
console.log(sessionId) // 42
```

justin-fiedler (changes requested):

packages/analytics-browser/src/browser-client.ts
Mercy811: Updated at https://github.com/amplitude/Amplitude-TypeScript/pull/753/commits/c5b79601bf525d532afd157572c28617209db9ef

justin-fiedler (approved):
Thanks @Mercy811!

packages/analytics-browser/src/browser-client.ts
```diff
@@ -83,10 +84,21 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
...
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
-    // Priority 2: last known sessionId from user identity storage
+    // Priority 2: sessionId from url if it's Number
+    // Priority 3: last known sessionId from user identity storage
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
-    this.setSessionId(options.sessionId ?? this.config.sessionId ?? Date.now());
+    const queryParams = getQueryParams();
+    this.setSessionId(
```
yuhao900914, line 92: Can we add some helper function for this?
```suggestion
const querySessionId =   (queryParams.ampSessionId && isValidNumber(queryParams.ampSessionId) ? Number(queryParams.ampSessionId) : undefined;
          
 this.setSessionId(
      options.sessionId ??
      querySessionId ??
      this.config.sessionId ??
      Date.now(),
    );```

Comment (Mercy811) [reply packages/analytics-browser/src/browser-client.ts:92]: I had a sync with Justin on this 

option 1:
```
_init(){    
    this.setSessionId(
        options.sessionId ??
        (queryParams.ampSessionId ?
            (Number.isNaN(Number(queryParams.ampSessionId)) ?
                    undefined : Number(queryParams.ampSessionId)
            ) : undefined
        ) ??
        this.config.sessionId ??
        Date.now(),
    );
}
```

option 2:
```
_init(){
    this.setSessionId(this._getSessionId());
}

private _getSessionId(options: BrowserOptions) {
    if (options.sessionId) return options.sessionId;
    if (queryParams.ampSessionId && !Number.isNaN(queryParams.ampSessionId)) {
      return Number.isNaN(queryParams.ampSessionId)
    }
    return this.config.sessionId || Date.now(); 
  }
```

I go with option 1 to avoid introducing one more helper method and existing comments can also help on explaining what is going on. I also use indents to make the hierarchy more clear.

Any other pros of option 2 adding a helper function?

Comment (yuhao900914) [reply packages/analytics-browser/src/browser-client.ts:92]: I would prefer to at least extract all the following into a variable to make it easier to read.
```
        (queryParams.ampSessionId ?
            (Number.isNaN(Number(queryParams.ampSessionId)) ?
                    undefined : Number(queryParams.ampSessionId)
            ) : undefined
        ) ?
```

yuhao900914 (approved):

packages/analytics-browser/src/browser-client.ts
Mercy811: I see what you mean. Thanks Alyssa for pointing it out. That's a better solution! I updated at https://github.com/amplitude/Amplitude-TypeScript/pull/753/commits/5eb9495326287700ee6aca664ea87525b3b0591d.

PR #756, opened 2024-05-24
feat(session-replay-browser): additional privacy configs

lewgordon-amplitude:
### Summary

**DO NOT MERGE** I still want to do additional manual testing before cutting a release.

This introduces the concept of a mask level, which currently provides three levels: light, medium, and conservative. Roughly they're defined as such:

* **Light** - only mask a subset of inputs that’s deemed sensitive - password, credit card, telephone #, email. These are information we never want to capture.
* **Medium (_default_)** - mask _all_ inputs.
* **Conservative** - mask all inputs _and_ all texts.

This is set via privacyConfig:

```ts
{
  privacyConfig: {
    defaultMaskLevel: MaskLevel.Medium
  }
}
```

This also provides a way to mask input and text fields individually based on CSS selector. These are configured in two parallel arrays `includeSelectors` and `maskSelectors`. If there is a matcher in both the `includeSelectors` and `maskSelectors` we will err on masking the element.

```ts
This is set via privacyConfig:

```ts
{
  privacyConfig: {
    includeSelectors: ['.search-field'],
    maskSelectors: ['.sensitive-field', 'div > .another-sensitive-field']
  }
}
```

### Checklist


packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -89,16 +88,10 @@ export class SessionReplayJoinedConfigGenerator {
...
-      if (!config.privacyConfig) {
-        config.privacyConfig = {};
-      }
-
-      for (const key in privacyConfig) {
-        const k = key as keyof PrivacyConfig;
-        config.privacyConfig[k] = privacyConfig[k];
-      }
-    }
+    config.privacyConfig = {
```
jackson-amplitude, line 91: if remote privacyConfig is empty, we should keep the privacyConfig set through the SDK for backwards compat.

packages/session-replay-browser/src/helpers.ts
```diff
@@ -1,13 +1,102 @@
...
+ *
+ * Priority is:
+ *  1. [In code] Element/class based masking/unmasking
+ *  2. [Config based] Selector based masking/unmasking
+ *  3. Use app defaults
+ */
+const isMasked = (element: HTMLElement, elementType: 'input' | 'text', config?: PrivacyConfig): boolean => {
+  // Element is explicitly instrumented to mask/unmask
+  if (element.classList) {
+    if (element.classList.contains(MASK_TEXT_CLASS)) {
```
jackson-amplitude, line 60: i think the initial proposal was the most restrictive would win (out of 1&2).

packages/session-replay-browser/src/helpers.ts
```diff
@@ -1,13 +1,102 @@
...
+  if (!config) {
+    return isMaskedForLevel(element, elementType, DEFAULT_MASK_LEVEL);
+  }
+
+  // Config has override for mask/unmask
+  const shouldMask = (config.maskSelector ?? []).some((selector) => element.matches(selector));
+  if (shouldMask) {
+    return true;
+  }
+  const shouldInclude = (config.includeSelector ?? []).some((selector) => element.matches(selector));
```
jackson-amplitude, line 77: maybe we should use "unmask" terminology instead of include here? since in the UI exclude translates to block, this adds more confusion.

packages/session-replay-browser/src/config/joined-config.ts
lewgordon-amplitude: Yeah I thought that was happening here, did I miss something?

packages/session-replay-browser/src/helpers.ts
lewgordon-amplitude: Did you mean in the variable name or the config field?

packages/session-replay-browser/src/config/joined-config.ts
jackson-amplitude: no, i just don't know enough about js =).

packages/session-replay-browser/src/helpers.ts
lewgordon-amplitude: That's right. Just to confirm though do you mean 1&2 are similar priority, but prefer masking or unmasking that is?

packages/session-replay-browser/src/helpers.ts
jackson-amplitude: probably good to update both.

packages/session-replay-browser/src/helpers.ts
jackson-amplitude: Yea, mostly thinking about the case where the engineer and admin configures something conflicting. We err on masking and let them figure it out.

lewgordon-amplitude: Yep, makes sense!

packages/session-replay-browser/src/config/types.ts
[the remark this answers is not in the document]
lewgordon-amplitude: I think that's out of scope for this change. Is that true, @jackson-amplitude ?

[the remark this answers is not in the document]
jackson-amplitude: Yea, let's do that as a follow up and focus on making sure the privacy features are in a good state. @bravecod from your side, if you already have it in the UI, feel free to ff it and we can choose to turn it on in the future.

packages/session-replay-browser/src/helpers.ts
```diff
@@ -1,13 +1,102 @@
...
-import { UNMASK_TEXT_CLASS } from './constants';
+import { MASK_TEXT_CLASS, UNMASK_TEXT_CLASS } from './constants';
+import { DEFAULT_MASK_LEVEL, MaskLevel, PrivacyConfig } from './config/types';
+import { getInputType } from '@amplitude/rrweb-snapshot';
 
-export const maskInputFn = (text: string, element: HTMLElement) => {
-  if (element.classList?.contains(UNMASK_TEXT_CLASS)) {
-    return text;
+// compiler guard
+/* istanbul ignore next */
```
jackson-amplitude, line 7: can we add a todo here and else where? SDK code does have the general standard of 100% test coverage (recognizing that's not a perfect measure, and it's sometimes a pain to upkeep), but would prefer not to start a pattern of exceptions.

jackson-amplitude (approved):
LGTM!

PR #757, opened 2024-05-24
fix: response with non-json format

Mercy811:
### Summary

[AMP-100725]

This PR fixes Amplitude transports so that The SDK will not throw an error when response has either empty body or has non-json format body, which triggers `Unexpected end of JSON input` error. Now in these two cases, the SDK will build the response based on the response code. 

### Checklist

[AMP-100725]: https://amplitude.atlassian.net/browse/AMP-100725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser/src/transports/xhr.ts
+++ b/packages/analytics-browser/src/transports/xhr.ts
@@ -17,15 +17,13 @@ export class XHRTransport extends BaseTransport implements Transport {
       xhr.open('POST', serverUrl, true);
       xhr.onreadystatechange = () => {
         if (xhr.readyState === this.state.done) {
-          try {
-            const responsePayload = xhr.responseText;
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-            const parsedResponsePayload: Record<string, any> = JSON.parse(responsePayload);
-            const result = this.buildResponse(parsedResponsePayload);
-            resolve(result);
-          } catch (e) {
-            reject(e);
-          }
+          const responsePayload = xhr.responseText;
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+          const parsedResponsePayload: Record<string, any> = responsePayload
+            ? JSON.parse(responsePayload)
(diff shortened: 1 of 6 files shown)


justin-fiedler (approved):

packages/analytics-browser/src/transports/xhr.ts
```diff
@@ -17,14 +17,12 @@ export class XHRTransport extends BaseTransport implements Transport {
...
         if (xhr.readyState === this.state.done) {
+          const responseText = xhr.responseText;
           try {
-            const responsePayload = xhr.responseText;
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-            const parsedResponsePayload: Record<string, any> = JSON.parse(responsePayload);
-            const result = this.buildResponse(parsedResponsePayload);
-            resolve(result);
-          } catch (e) {
-            reject(e);
```
yuhao900914, line 27: This is something different with v2.x.
Should we reject it when there have an error?

Mercy811: main and v1.x branches should be the same. Please see the other [PR](https://github.com/amplitude/Amplitude-TypeScript/pull/758/files) in review where this logic is the same. 

We should not reject it because it's possible that this is a success response with no or non-json format body.

yuhao900914: If an error occurs in the try block, will we lose the useful information about why the error occurred?

Mercy811: We use try catch block here because of `JSON.parse()`. I don't see anywhere else errors can occur.

yuhao900914: Got it.

yuhao900914 (approved):

PR #761, opened 2024-06-04
fix(session replay): add error handling for indexeddb

jxiwang:
### Summary

Adding error handling for [AMP-101119](https://amplitude.atlassian.net/browse/AMP-101119) to defer issues as warnings and not trigger errors in client. 

https://amplitude.atlassian.net/browse/DDI-489

### Checklist

[AMP-101119]: https://amplitude.atlassian.net/browse/AMP-101119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/session-replay-browser/src/events/events-idb-store.ts
+++ b/packages/session-replay-browser/src/events/events-idb-store.ts
@@ -39,7 +39,7 @@ export const keyValDatabaseExists = function (): Promise<IDBDatabase | void> {
     }
 
     if (!globalScope.indexedDB) {
-      return reject(new Error('IndexedDB is not available in the global scope'));
+      return reject(new Error('Session Replay: cannot find indexedDB'));
     }
 
     try {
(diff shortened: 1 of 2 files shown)


lewgordon-amplitude (commented):
The change makes sense if I follow it correctly that it causes us to fail on `initialize`. Is there a way we could change the messaging to say something along the lines of session replay is available because we couldn't find indexedDB? Maybe also we could point them at a doc page with more help for this issue?

jxiwang:
@lewgordon-amplitude Does making the errors less generic about Global Scope sound better in this case? I agree with adding some docs, but the only documentation I can think of is maybe something security related [thread](https://amplitude.slack.com/archives/C05LPGKQW4W/p1717379625676929?thread_ts=1716932297.993479&cid=C05LPGKQW4W) I'm trying to mitigate unhandled errors that are propagating to a client's monitoring service.

lewgordon-amplitude:
Yeah I think that makes sense.

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -33,8 +33,16 @@ export interface SessionReplayDB extends DBSchema {
...
   const globalScope = getGlobalScope();
-  return new Promise((resolve) => {
-    if (globalScope) {
+  return new Promise((resolve, reject) => {
+    if (!globalScope) {
+      return reject(new Error('Global scope not found'));
+    }
+
+    if (!globalScope.indexedDB) {
+      return reject(new Error('IndexedDB is not available in the global scope'));
```
jackson-amplitude, line 42: I don't remember if this was discussed, but what were the reasons for this to not be available? is iframe the main one?

jxiwang: There are a few security cases where indexeddb might be blocked from being used like CSP, iframe, browser extensions, etc. I think it's good to have the error handling here and provide some context in the docs on why it could be the case.

jackson-amplitude: SG. and just for context, when this fails, the SDK does not capture right (at least on this specific page)?

jxiwang: In this context, this function is called to transition the old IDB store to the new store. I believe this is for the case where a user might open a new tab that has the newer version of the SDK and the stored events will transition over.

lewgordon-amplitude (approved):
Looks great! 🚢

jackson-amplitude (approved):

PR #763, opened 2024-06-06
fix(session replay): rethrow external errors and fix block selectors null state

jackson-amplitude:
### Summary
* for errors that are marked external (as in triggered by clients outside of rrweb), rethrow those errors. Frameworks like styled components depend on these errors being thrown. 
* also fixed an issue I noticed while looking at safari and the default state of block selectors. 
WIP: needs the new version of rrweb

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -241,6 +241,15 @@ export class SessionReplay implements AmplitudeSessionReplay {
   }
 
   getBlockSelectors(): string | string[] | undefined {
+    // For some reason, this defaults to empty array ([]) if undefined in the compiled script.
+    // Empty arrays cause errors when being evaluated in Safari.
+    // Force the selector to be undefined if it's an empty array.
+    if (
+      Array.isArray(this.config?.privacyConfig?.blockSelector) &&
+      this.config?.privacyConfig?.blockSelector.length === 0
+    ) {
+      return undefined;
+    }
     return this.config?.privacyConfig?.blockSelector;
   }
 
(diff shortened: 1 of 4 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -230,16 +239,24 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
-      blockSelector: this.getBlockSelectors() as string,
+      blockSelector: this.getBlockSelectors() as string | undefined,
       maskInputFn,
       recordCanvas: false,
       errorHandler: (error) => {
-        const typedError = error as Error;
+        const typedError = error as Error & { _external_?: boolean };
 
         // styled-components relies on this error being thrown and bubbled up, rrweb is otherwise suppressing it
         if (typedError.message.includes('insertRule') && typedError.message.includes('CSSStyleSheet')) {
```
jackson-amplitude, line 249: leaving this around for now to avoid potential regressions.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -204,6 +204,15 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
   getBlockSelectors(): string | string[] | undefined {
+    // For some reason, this defaults to empty array ([]) if undefined in the compiled script.
+    // Empty arrays cause errors when being evaluated in Safari.
+    // Force the selector to be undefined if it's an empty array.
+    if (
+      Array.isArray(this.config?.privacyConfig?.blockSelector) &&
+      this.config?.privacyConfig?.blockSelector.length === 0
+    ) {
+      return undefined;
+    }
```
lewgordon-amplitude, line 215: NIt: I'd rework this slightly.
```suggestion
    const blockSelector = this.config?.privacyConfig?.blockSelector ?? [];
    if (blockSelector.length === 0) {
      return undefined;
    }
    return blockSelector;
```

This is mostly because the JS generated has a smaller footprint: https://gist.github.com/lewgordon-amplitude/b84c69d0518053ee68ed90218a7c349b

lewgordon-amplitude (approved):
LGTM!

jxiwang (approved):
LGTM

packages/session-replay-browser/src/session-replay.ts
jackson-amplitude: good point. Was concerned because block selector could be string, but `.length` also works on strings =)

PR #764, opened 2024-06-06
fix: don't use global window directly

izaaz:
### Summary

Avoid using window global variable directly and instead use `getGlobalScope()`

### Checklist


Mercy811 (approved):
Thanks Izaaz. LGTM.

PR #766, opened 2024-06-07
fix(plugin-autocapture-browser): export AutocaptureOptions

Mercy811:
### Summary

Export `AutocaptureOptions` for remote config type.
Use `fix` to bump the version. 

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/index.ts
+++ b/packages/plugin-autocapture-browser/src/index.ts
@@ -3,6 +3,6 @@ export {
   autocapturePlugin,
   DEFAULT_CSS_SELECTOR_ALLOWLIST,
   DEFAULT_DATA_ATTRIBUTE_PREFIX,
-  Options as AutocaptureOptions,
+  AutocaptureOptions,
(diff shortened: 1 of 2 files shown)


packages/plugin-autocapture-browser/src/index.ts
```diff
@@ -3,5 +3,6 @@ export {
   autocapturePlugin,
   DEFAULT_CSS_SELECTOR_ALLOWLIST,
   DEFAULT_DATA_ATTRIBUTE_PREFIX,
+  Options as AutocaptureOptions,
```
justin-fiedler, line 6: nit: I think it might be better to rename Options -> `AutocaptureOptions` and export it as-is so that it has the same name across the entire codebase.

Mercy811: I agree it's less confusing. Renamed at https://github.com/amplitude/Amplitude-TypeScript/pull/766/commits/7fe6fe54f968d3b4bc3210e3635c2910961f4ec6

PR #769, opened 2024-06-10
feat(analytics-browser): consume remote config

Mercy811:
### Checklist


packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -0,0 +1,42 @@
+import { BrowserConfig as IBrowserConfig } from '@amplitude/analytics-types';
+import { createRemoteConfigFetch, RemoteConfigFetch } from '@amplitude/analytics-remote-config';
+import { BrowserRemoteConfig } from './types';
+
+export class BrowserJoinedConfigGenerator {
+  // Local config before generateJoinedConfig is called
+  // Joined config after generateJoinedConfig is called
+  config: IBrowserConfig;
```
kwalker3690, line 8: This API makes me a little nervous since a consumer might not notice the distinction in the lifecycle when the values in config can change. For this reason in the session replay joined config class I explictly [set localConfig on the class](https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/session-replay-browser/src/config/joined-config.ts#L12), and only return the joined config as a [return value of generateJoinedConfig](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/config/joined-config.ts#L26C23-L26C23)

Mercy811: Thanks Kelly for the comment. I did refer to session replay joined class when writing this. The reasons why I want to update the config instead of keeping it as a copy of local config are that
- The config in the generator is only used to merge with the remote config. It makes more sense for me to preserve and use the last joined config.
- When new remote config comes, just merge it with the last joined config. It won't change the result.
- The last and the most import one is that `BrowserConfig` contains protected properties which cannot be copied by shallow copy like using the spread operator `this.localConfig = {...localConfig}`. I just don't like how we need to introduce a deep copy function only for this and the function is only used here which doesn't seem to be very efficient. But I'm open to discussion.

```
const deepCopy = <T>(obj: T): T => {
  if (typeof obj !== 'object' || obj === null) {
    return obj;
  }
  const res: any = Array
    .isArray(obj) ? [] : {};
  for (const key in obj) {
    // eslint-disable-next-line no-prototype-builtins
    if (obj.hasOwnProperty(key)) {
      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
      res[key] = deepCopy(obj[key]);
    }
  }
  return Object.assign(res, obj) as T;
}
```

packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  async generateJoinedConfig(): Promise<IBrowserConfig> {
+    const remoteConfig =
+      this.remoteConfigFetch &&
+      (await this.remoteConfigFetch.getRemoteConfig('analyticsSDK', 'browserSDK', this.config.sessionId));
+    this.config.loggerProvider.debug('Remote configuration:', JSON.stringify(remoteConfig, null, 2));
+    if (remoteConfig && remoteConfig.defaultTracking) {
+      this.config.defaultTracking = remoteConfig.defaultTracking;
+    }
```
Mercy811, line 32: Another solution is to patch the joined config with whatever in the remote config

packages/analytics-browser/src/config/joined-config.ts
justin-fiedler: I agree with Kelly, I would vote for preserving the original local config.

Mercy811: Hi @justin-fiedler, why would you vote for preserving the original local config? I'm asking because updating the config doesn't change the final result of the joined config and consumers are thus not affected.

kwalker3690: Yeah the deep copy is not ideal, for SR we only had to deal with that for one property so I [worked around it for just the single property](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/config/joined-config.ts#L29C9-L29C9) rather than a full deep copy. I actually wasn't sure why optOut was implemented with a getter/setter, it seemed unnecessary? Are there many other protected properties in browser sdk config?

justin-fiedler: I agree, I think a deep copy seems too heavy here.

Some options

1. We could add an internal `shallowCopy()` method to the config that handles transferring protected properties. This seems risky as we would need to remember to add new properties as they are added, but maybe we could make a unit test?
2. Since the remote config is currently only loaded once at startup it *should* be ok to modify a single config instance as we never need to apply remote changes multiple times. If we add more complex remote config logic in the future we modify as needed.

I think option 1 is better in the long run but am ok with 2 for now.

justin-fiedler (approved):

packages/analytics-browser/src/config/joined-config.ts
Mercy811: Thanks team for the discussion. I created a ticket for this https://amplitude.atlassian.net/browse/AMP-102396.

kwalker3690: Agreed that 2 is ok for now, thanks for creating the ticket

kwalker3690 (approved):

PR #771, opened 2024-06-11
feat(session replay): add ability to set inline stylesheets to false

kwalker3690:
### Summary

Add option to allow customers to set inline stylesheets to false, so css stylesheets are fetched via link tag instead

### Checklist


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -74,10 +74,6 @@ export class SessionReplay implements AmplitudeSessionReplay {
     this.joinedConfigGenerator = await createSessionReplayJoinedConfigGenerator(apiKey, options);
     this.config = await this.joinedConfigGenerator.generateJoinedConfig(this.identifiers.sessionId);
 
-    this.loggerProvider.debug(
```
kwalker3690, line 77: This seemed like internal debugging, rather than something that should be customer facing, but lmk if this was meant to be customer facing!

lewgordon-amplitude: I was thinking it might be useful for customers with our now complex joining logic, otherwise it's hard to tell what the end config should look like. Let me know if there's a better way to expose this, though.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -53,7 +53,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
         try {
           fragment.querySelector(selector);
         } catch {
-          console.warn(`[session-replay-browser] omitting selector "${selector}" because it is invalid`);
+          this.loggerProvider.warn(`[session-replay-browser] omitting selector "${selector}" because it is invalid`);
```
lewgordon-amplitude, line 56: Good catch!

lewgordon-amplitude (approved):
LGTM!

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Got it, I'll just update the message to be a little more user friendly then!

kwalker3690: Oh that looks much different in the console than I expected, I see its purpose now, ok I'll just prepend the name with session replay

PR #775, opened 2024-06-13
refactor(session replay): reorganize targeting logic and get test coverage to 100%

kwalker3690:
### Summary
*note: this will be merged into a feature branch, some changes in this PR are fixing previous errors in rebasing*
This PR gets the sr-targeting branch up to 100% test coverage, as well as reworks the organization of targeting code in session replay. It also introduces a block on recording until targeting has been evaluated.

### Checklist


packages/session-replay-browser/src/targeting/targeting-manager.ts
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    if (config.targetingConfig && Object.keys(config.targetingConfig).length) {
+      const targetingResult = evaluateTargetingPackage({
+        ...targetingParams,
+        flag: config.targetingConfig,
+        sessionId: sessionId,
+      });
+      sessionTargetingMatch = targetingResult.sr_targeting_config.key === 'on';
+    }
+    void TargetingIDBStore.storeTargetingMatchForSession({
```
lewgordon-amplitude, line 36: Why `void` here and not `await`?

lewgordon-amplitude (approved):
LGTM. I do have one question and will give a more thorough look once it's ready to be merged to `v1.x`.

packages/session-replay-browser/src/targeting/targeting-manager.ts
kwalker3690: I was thinking we don't need to wait on the targeting match update - but on second thought that could cause issues, and will make little difference to await here, so I'll update!

kwalker3690: Will update this in the next PR i'm working on

PR #776, opened 2024-06-14
feat(session-replay-browser): adds capability to track click events in session replay

lewgordon-amplitude:
### Summary

This utilizes RRweb's native click interactions and adds additional metadata to be sent to Amplitude. Events will be stored in a separate indexed db database and batched similar to RRweb events are today. Features:

* Adds `interactionConfig` which gives the ability to enable/disable click tracking and also gives control on how frequent we send data.
* Stores click events in a similar way to RRweb to persist events in the case of a browser crash/error.

### Checklist


lewgordon-amplitude:
@kwalker3690 this is ready for review. I need to do a last minute change to fix `@medv/finder` once the new package is published.

packages/session-replay-browser/src/config/types.ts
```diff
@@ -31,6 +31,11 @@ export type PrivacyConfig = {
   unmaskSelector?: string[];
 };
 
+export type InteractionConfig = {
+  trackEveryNms?: number;
+  enabled?: boolean; // defaults to false
```
kwalker3690, line 36: If it has a default, should it not be marked as optional?

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -102,16 +103,33 @@ export class SessionReplayEventsIDBStore implements AmplitudeSessionReplayEvents
...
+    this.maxInterval = maxInterval ?? MAX_INTERVAL;
+    this.minInterval = minInterval ?? MIN_INTERVAL;
+    this.interval = 0;
   }
 
-  async initialize(sessionId?: number) {
-    const dbName = `${this.apiKey.substring(0, 10)}_amp_session_replay_events`;
+  async initialize(type: EventType, sessionId?: number) {
+    const dbSuffix = type === 'rrweb' ? '' : `_${type}`;
+    const dbName = `${this.apiKey.substring(0, 10)}_amp_session_replay_events${dbSuffix}`;
```
kwalker3690, line 132: Ah I just realized, the change to add `rrweb` will break existing stores on end user's version update, and leave a dangling version of the old `events` db. Instead of using suffix for each type, what about just leaving the existing events store as is, and adding the suffix for click events? Otherwise we'll need to add in a migration here

kwalker3690: Wow, rereading and I see that's what you have here already! Sorry, misread it as appending `rrweb` the first time I looked at this

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -83,10 +90,31 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
     this.removeInvalidSelectors();
 
-    this.eventsManager = await createEventsManager({
+    const managers: { name: EventType; manager: SessionReplayEventsManager<string, string> }[] = [];
```
kwalker3690, line 93: Nit: would be nice to have a shared type here used in MultiManager args as well

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -285,10 +308,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+          deviceId &&
+          this.eventsManager.addEvent({ event: { type: 'rrweb', data: eventString }, sessionId, deviceId });
       },
       packFn: pack,
       inlineStylesheet: this.config.shouldInlineStylesheet,
+      hooks: {
+        mouseInteraction:
+          this.eventsManager &&
+          clickHook({
+            getGlobalScopeFn: getGlobalScope,
```
kwalker3690, line 321: Why pass this as an option? Is it for testing reasons? If so I can point you to where I've mocked global scope for other testing purposes

packages/session-replay-browser/src/typings/session-replay.ts
```diff
@@ -29,9 +31,16 @@ export interface SendingSequencesIDBInput {
 export type SendingSequencesIDBReturn = Required<SendingSequencesIDBInput>;
 
 export interface SessionReplayEventsIDBStore {
-  initialize(): Promise<void>;
+  initialize(type: EventType): Promise<void>;
   getSequencesToSend(): Promise<SendingSequencesIDBReturn[] | undefined>;
+  /**
```
kwalker3690, line 36: Thank you for the docstrings 🙏

kwalker3690 (approved):
Overall looks good, a few minor things that are fine to address in a follow up if easier!

packages/session-replay-browser/src/config/types.ts
lewgordon-amplitude: Yeah I think now it makes sense to mark it as non optional since it defaults to `false`. I had a previous iteration where it was defaulted to `true` and I wanted to make sure if it wasn't specified it'd be `true`.

packages/session-replay-browser/src/session-replay.ts
lewgordon-amplitude: Good callout. I had thought this was an instance function is probably why I added it as a parameter.

packages/session-replay-browser/src/config/types.ts
```diff
@@ -41,6 +46,7 @@ export interface SessionReplayLocalConfig extends Config {
   debugMode?: boolean;
   configEndpointUrl?: string;
   shouldInlineStylesheet?: boolean;
+  interactionConfig?: InteractionConfig;
```
jackson-amplitude, line 49: I forgot which config interface is external, but ideally we don't allow configuration of this through SDK?

packages/session-replay-browser/src/events/multi-manager.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  SessionReplayEventsManager as AmplitudeSessionReplayEventsManager,
+  EventsManagerWithType,
+} from '../typings/session-replay';
+
+/**
+ * "Registers" events managers internally. When an event is added this class routes the event to the correct
+ * manager. For all send or flush methods this will invoke the event for all registered managers.
+ */
+export class MultiEventManager<EventType, EventDataType>
```
jackson-amplitude, line 10: nice!

packages/session-replay-browser/test/session-replay.test.ts
```diff
@@ -195,15 +195,24 @@ describe('SessionReplay', () => {
...
+      });
+      jest.spyOn(RemoteConfigFetch, 'createRemoteConfigFetch').mockResolvedValue({
+        getRemoteConfig: getRemoteConfigMock,
+        fetchTime: 0,
+      });
       await sessionReplay.init(apiKey, {
         ...mockOptions,
         sampleRate: 0.5,
-        interactionConfig,
+        // interactionConfig,
```
kwalker3690, line 211: Nit: can remove this

kwalker3690 (approved):

packages/session-replay-browser/package.json
```diff
@@ -28,6 +28,7 @@
     "lint": "yarn lint:eslint & yarn lint:prettier",
     "lint:eslint": "eslint '{src,test}/**/*.ts'",
     "lint:prettier": "prettier --check \"{src,test}/**/*.ts\"",
+    "postinstall": "rimraf ../../node_modules/@medv/finder/finder.ts # This is required until the package is fixed upstream",
```
roderickhsiao, line 31: It seems like we also having error when installing the lastest reply plugin via npm for throwing rimraf not found error.

lewgordon-amplitude: Thanks for letting me know. Let me take a look.

lewgordon-amplitude: I just published a change for:
```
Successfully published:
 - @amplitude/plugin-session-replay-browser@1.6.1
 - @amplitude/session-replay-browser@1.9.1
```
Can you check to see if these fix your issue? If not, would you mind opening an issue with more details?

roderickhsiao: yes, it works now :) Thanks for the quick update!

PR #779, opened 2024-06-17
feat(targeting and session replay): Add support for targeting by single event and multiple events

kwalker3690:
### Summary

This PR adds support for targeting via single or multiple event conditions. A new method is exposed, evaluateTargetingAndRecord, which accepts an event as a parameter. That event is passed to the targeting package, which updates an internal IDB store of events for the current session. The single event and the list of stored events are passed to the evaluation package to be compared against the targeting flag config.

### Checklist

### DIFF
--- a/packages/session-replay-browser/README.md
+++ b/packages/session-replay-browser/README.md
@@ -47,13 +47,19 @@ sessionReplay.init(API_KEY, {
 });
 ```
 
-### 3. Evaluate targeting and get session replay event properties
-Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. It must also be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties
+### 3. Evaluate targeting (optional)
+Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. This should be done *before* step 4, getting the event properties. If you are not using the targeting condition logic provided via the Amplitude UI, this step is not required.
 ```typescript
 const sessionTargetingMatch = sessionReplay.evaluateTargetingAndRecord({ event: {
   event_type: EVENT_NAME,
+  time: EVENT_TIMESTAMP,
   event_properties: eventProperties
 } });
+```
+
+### 4. Get session replay event properties
+Any event must be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties.
+```typescript
 const sessionReplayProperties = sessionReplay.getSessionReplayProperties();
 track(EVENT_NAME, {
   ...eventProperties,
@@ -61,7 +67,7 @@ track(EVENT_NAME, {
 })
 ```
 
-### 4. Update session id
+### 5. Update session id
 Any time that the session id for the user changes, the session replay SDK must be notified of that change. Update the session id via the following method:
 ```typescript
 sessionReplay.setSessionId(UNIX_TIMESTAMP)
@@ -71,7 +77,7 @@ You can optionally pass a new device id as a second argument as well:
 sessionReplay.setSessionId(UNIX_TIMESTAMP, deviceId)
 ```
 
-### 5. Shutdown (optional)
+### 6. Shutdown (optional)
 If at any point you would like to discontinue collection of session replays, for example in a part of your application where you would not like sessions to be collected, you can use the following method to stop collection and remove collection event listeners.
 ```typescript
 sessionReplay.shutdown()
--- a/packages/session-replay-browser/README.md
+++ b/packages/session-replay-browser/README.md
@@ -47,9 +47,13 @@ sessionReplay.init(API_KEY, {
 });
 ```
 
-### 3. Get session replay event properties
-Any event that occurs within the span of a session replay must be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties
+### 3. Evaluate targeting and get session replay event properties
+Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. It must also be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties
 ```typescript
+const sessionTargetingMatch = sessionReplay.evaluateTargetingAndRecord({ event: {
+  event_type: EVENT_NAME,
+  event_properties: eventProperties
+} });
 const sessionReplayProperties = sessionReplay.getSessionReplayProperties();
 track(EVENT_NAME, {
   ...eventProperties,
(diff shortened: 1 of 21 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -171,26 +171,37 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
       sessionId: this.identifiers.sessionId,
       config: this.config,
-      targetingParams: { userProperties },
+      targetingParams: { userProperties, event: eventForTargeting },
     });
 
-    if (sessionTargetingMatch) {
+    if (this.sessionTargetingMatch) {
       this.recordEvents();
     }
```
lewgordon-amplitude, line 202: Is `recordEvents()` idempotent? Could this function potentially be called more than once causing record events to be triggered multiple times?

lewgordon-amplitude (changes requested):
Looks good to me overall, but I just wanted to clarify things before approval.

packages/targeting/test/targeting-idb-store.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      await TargetingIDBStore.storeEventTypeForSession({
+        eventType: 'Add to Cart',
+        sessionId: 123,
+        apiKey: 'static_key',
+        loggerProvider: mockLoggerProvider,
+      });
+      expect(mockLoggerProvider.warn).toHaveBeenCalled();
+    });
+
+    test('should handle undefinted transaction', async () => {
```
lewgordon-amplitude, line 87:
```suggestion
    test('should handle undefined transaction', async () => {
```

packages/session-replay-browser/src/targeting/targeting-idb-store.ts
```diff
@@ -72,3 +73,29 @@ export const storeTargetingMatchForSession = async ({
...
+  loggerProvider,
+  apiKey,
+  currentSessionId,
+}: {
+  loggerProvider: ILogger;
+  apiKey: string;
+  currentSessionId: number;
+}) => {
+  try {
+    const db = await openOrCreateDB(apiKey);
```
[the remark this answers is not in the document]
kwalker3690: Yeah that's fair, I'll make this change! I also don't really know about the performance implications here, I've tried to research about different aspects of IndexedDB performance and haven't been able to find good resources for nuances like this - if you ever come across any, please let me know!

packages/targeting/src/targeting.ts
```diff
@@ -8,10 +9,30 @@ export class Targeting implements AmplitudeTargeting {
...
+    const eventTypes =
+      event &&
+      (await storeEventTypeForSession({
+        loggerProvider: loggerProvider,
+        apiKey: apiKey,
+        sessionId,
+        eventType: event.event_type,
+      }));
+
+    const eventStrings = eventTypes && new Set(eventTypes.map((eventTypeObj) => eventTypeObj.event_type));
```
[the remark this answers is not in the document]
kwalker3690: Yeah agreed, originally I had the store data as a set. However, Jackson and I were looking at the UI that we might use for building targeting conditions, and I noticed that there is a UI for 'count', ie how many times an event has happened. I could see in future iterations us wanting to include the event count feature for TRC (we'd need to update evaluation package logic to support it, so it's out of scope for now). So I thought that leaving the stored events as an array, with potential duplicates, gives us the most flexibility, without needing to introduce breaking changes to the store down the road. However for the current capability of the experiment evaluation engine, we need to deduplicate events before passing them along.

An alternative option would be to store the events as an object keyed by event type, in which case we could add a 'count' property to the store objects when it made sense in the future. The reason I didn't do that is that I thought we may someday want to support a condition like ```event_type = "Add to Cart" AND count = 3 AND event_properties include "price = 4"```. To support that, we'd need to be able to distinguish between the different fired events and their property values.

We're likely a long way off from supporting that type of condition however, so let me know if you think that the performance benefits of not having to transform the data on each function call outweigh the forward compatibility of the data structure.

packages/session-replay-browser/src/session-replay.ts
kwalker3690: It can be called more than once, the first thing this function does is call rrweb's cancel recording callback: https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/session-replay-browser/src/session-replay.ts#L320C19-L320C19. So there can never be more than one recording/MutationObserver observing at once

lewgordon-amplitude: I wonder if that leads to any janky behavior though? That is if we're constantly restarting recording.

packages/targeting/src/targeting-idb-store.ts
```diff
@@ -3,110 +3,148 @@ import { DBSchema, IDBPDatabase, IDBPTransaction, openDB } from 'idb';
...
+  openOrCreateDB = async (apiKey: string) => {
+    if (this.dbs && this.dbs[apiKey]) {
+      return this.dbs[apiKey];
+    }
+    const dbName = `${apiKey.substring(0, 10)}_amp_targeting`;
+    const db = await this.createStore(dbName);
+    this.dbs = {
+      ...this.dbs,
+      [apiKey]: db,
+    };
```
lewgordon-amplitude, line 47: Nit: Why not?
```suggestion
    this.dbs[apiKey] = db;
```

packages/session-replay-browser/src/targeting/targeting-idb-store.ts
```diff
@@ -12,90 +12,103 @@ export interface SessionReplayTargetingDB extends DBSchema {
...
+  openOrCreateDB = async (apiKey: string) => {
+    if (this.dbs && this.dbs[apiKey]) {
+      return this.dbs[apiKey];
+    }
+    const dbName = `${apiKey.substring(0, 10)}_amp_session_replay_targeting`;
+    const db = await this.createStore(dbName);
+    this.dbs = {
+      ...this.dbs,
+      [apiKey]: db,
+    };
```
lewgordon-amplitude, line 39: Same as the other comment

lewgordon-amplitude (approved):
LGTM! It looks really great now, thanks for making those changes.

packages/targeting/src/targeting-idb-store.ts
kwalker3690: I think I did this in case `this.dbs` wasn't defined, however I'm realizing I can just provide a base empty object for it - will update

packages/session-replay-browser/README.md
```diff
@@ -47,8 +47,18 @@ sessionReplay.init(API_KEY, {
 });
 ```
 
-### 3. Get session replay event properties
-Any event that occurs within the span of a session replay must be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties
+### 3. Evaluate targeting (optional)
+Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. This should be done *before* step 4, getting the event properties. If you are not using the targeting condition logic provided via the Amplitude UI, this step is not required.
+```typescript
+const sessionTargetingMatch = sessionReplay.evaluateTargetingAndRecord({ event: {
```
kwalker3690, line 53: This should have an `await` before it

PR #783, opened 2024-06-19
fix(analytics-client-common): adding ampIntegrationsContext to getGlobalScope

zacbennett:
### Summary

Adding another option to get the getGlobalScope function to assist with our integration

### Checklist

* Does your PR have a breaking change?:  NO

### DIFF
--- a/packages/analytics-client-common/src/global-scope.ts
+++ b/packages/analytics-client-common/src/global-scope.ts
@@ -1,11 +1,12 @@
 /* eslint-disable no-restricted-globals */
 /* Only file allowed to access to globalThis, window, self */
 
-import { ampIntegrationContext } from './types/global';
-
 export const getGlobalScope = (): typeof globalThis | undefined => {
-  if (typeof ampIntegrationContext !== 'undefined') {
-    return ampIntegrationContext;
+  // This should only be used for integrations with Amplitude that are not running in a browser environment
+  //   We need to specify the name of the global variable as a string to prevent it from being minified
+  const ampIntegrationContextName = 'ampIntegrationContext' as keyof typeof globalThis;
+  if (typeof globalThis[ampIntegrationContextName] !== 'undefined') {
+    return globalThis[ampIntegrationContextName] as typeof globalThis;
   }
   if (typeof globalThis !== 'undefined') {
     return globalThis;
--- a/packages/analytics-client-common/src/global-scope.ts
+++ b/packages/analytics-client-common/src/global-scope.ts
@@ -2,6 +2,9 @@
 /* Only file allowed to access to globalThis, window, self */
 
 export const getGlobalScope = (): typeof globalThis | undefined => {
+  if (typeof ampIntegrationContext !== 'undefined') {
+    return ampIntegrationContext;
+  }
   if (typeof globalThis !== 'undefined') {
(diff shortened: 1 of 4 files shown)


Mercy811:
Hi @zacbennett, `.d.ts` file will not be compiled that's why the CI is failing. You can bypass it by
- Change `global.d.ts` to `global.ts`
- Replace `declare` with `export`
- Import `ampIntegrationContext` in `global-scope.ts`

Mercy811 (approved):

zacbennett:
Unfortunately, this is not working as expected when I point Shopify to this version. Will take another whack at it later

zacbennett:
So this was the issue we had before, where the name of the global variable was being minified, and thus not picked up:
image

But was able to fix this, just needed to put the global variable name in a string
image

packages/analytics-client-common/src/global-scope.ts
```diff
@@ -2,6 +2,12 @@
 /* Only file allowed to access to globalThis, window, self */
 
 export const getGlobalScope = (): typeof globalThis | undefined => {
+  // This should only be used for integrations with Amplitude that are not running in a browser environment
+  //   We need to specify the name of the global variable as a string to prevent it from being minified
+  const ampIntegrationContextName = 'ampIntegrationContext' as keyof typeof globalThis;
+  if (typeof globalThis[ampIntegrationContextName] !== 'undefined') {
```
izaaz, line 8: should you also check to make sure `globalThis` is also not undefined?

Mercy811: Can we also add some tests for this? I was surprised that we didn't have tests for this file at atll but the CI passes which requires 100% coverage.

zacbennett: Hmm maybe the commons folder is excluded? Will write tests on Monday!

Mercy811: It's not included in test coverage because it has no corresponding test file. I'm ok with no tests as I didn't find an easy way to mock `globalThis` to be undefined. 

Here is what I tried `global-scope.test.ts`:
```
import { getGlobalScope } from '../src/global-scope'

describe('getGlobalScope', () => {
  test('should return globalThis if defined', () => {
    expect(globalThis).toBeDefined();
    expect(getGlobalScope()).toBe(globalThis);
  });

  test('should return window if globalThis is undefined and window is defined', () => {
    const originalGlobalThis: typeof globalThis = globalThis;
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    globalThis = undefined;

    expect(globalThis).toBeUndefined();
    expect(window).toBeDefined();
    expect(getGlobalScope()).toBe(window);

    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    globalThis = originalGlobalThis;
  });
});
```

The error suggests that something in the Jest setup or the mocking process is interfering with Jest's internal state or its matchers.
```
Test suite failed to run
TypeError: Cannot read properties of undefined (reading 'Symbol($$jest-matchers-object)')
```

Mercy811:
Thanks @zacbennett for adding the tests. LGTM.

PR #786, opened 2024-06-21
feat(autocapture): Add element hierarchy tracking and improve attribute filtering

Dogfalo:
### Summary
Adds element hierarchy tracking to autocapture events and improves attribute filtering for better privacy

### Changes
- Element hierarchy tracking provides more context for autocaptured events
- Long attribute values will be trimmed to 128 chars to allow for capture of more ancestor nodes
- Improved privacy by filtering sensitive attributes and limiting data collection on sensitive elements
- Limit hierarchy node capture to not exceed 1024 chars


packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -0,0 +1,135 @@
...
+  'onchange',
+  'oninput',
+  'onblur',
+  'onsubmit',
+  'onfocus',
+  'onkeydown',
+  'onkeyup',
+  'onkeypress',
+
+  // React specific
```
liuyang1520, line 26: Should we also consider adding a allow list to combine, since other frameworks might introduce attributes too

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const tagName = element.tagName.toLowerCase();
+  const properties: HierarchyNode = {
+    tag: tagName,
+  };
+
+  const siblings = Array.from(element.parentElement?.children ?? []);
+  if (siblings.length) {
+    properties.index = siblings.indexOf(element);
+    properties.indexOfType = siblings.filter((el) => el.tagName === element.tagName).indexOf(element);
```
liuyang1520, line 51: Curious what are these 2 properties used for? Is it for the query engine to match back the node?

liuyang1520 (approved):

packages/plugin-autocapture-browser/src/hierarchy.ts
Dogfalo: The information from these are used to enable nth-child and nth-of-type queries

packages/plugin-autocapture-browser/src/hierarchy.ts
Dogfalo: Yes, good point. As a follow up diff, I'll investigate options to expose for customizing data collection

PR #787, opened 2024-06-21
fix(session replay): fix logger for standalone

kwalker3690:
### Summary

The logger was not representing the correct log level when used with the standalone SDK, this fixes it so we can observe proper logging outside of the plugin

### Checklist


jackson-amplitude (approved):
nice catch!

lewgordon-amplitude (approved):
LGTM, good catch!

PR #686, opened 2024-03-22
fix: #664 Handle URL-encoded cookie values

dajinchu:
### Summary

Fixes #664

Cookie values are often URL-encoded. In Rails, for example, there is no way to opt out of url-encoding cookies. So [this code in the docs](https://www.docs.developers.amplitude.com/experiment/sdks/ruby-sdk/#access-amplitude-cookies) does not work.

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?: no

PR #790, opened 2024-06-24
fix(analytics-browser): handle double URL encoded cookie values for R…

Mercy811:
…uby Rails

### Summary

[AMP-101996]

Make up https://github.com/amplitude/Amplitude-TypeScript/pull/686.

This PR supports getting cookie values which are encoded by the following order: URL encoding, base64, URL encoding for Ruby on Rails. 

### Checklist

[AMP-101996]: https://amplitude.atlassian.net/browse/AMP-101996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-client-common/src/storage/cookie.ts
+++ b/packages/analytics-client-common/src/storage/cookie.ts
@@ -31,18 +31,18 @@ export class CookieStorage<T> implements Storage<T> {
   }
 
   async get(key: string): Promise<T | undefined> {
-    let value = await this.getRaw(key);
+    const value = await this.getRaw(key);
     if (!value) {
       return undefined;
     }
     try {
-      try {
-        value = decodeURIComponent(atob(value));
-      } catch {
+      const decodedValue = decodeCookiesAsDefault(value) ?? decodeCookiesForRuby(value);
+      if (decodedValue === undefined) {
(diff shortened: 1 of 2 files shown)


packages/analytics-client-common/src/storage/cookie.ts
```diff
@@ -101,3 +101,21 @@ export class CookieStorage<T> implements Storage<T> {
     return;
   }
 }
+
+const decodeCookiesAsDefault = (value: string): string | undefined => {
```
izaaz, line 105: NIT: `decodeCookieWithoutURLEncoding` and `decodeCookieWithURLEncoding`?

default is a little too generic and the other one is not just for ruby but for any service that URL encodes cookies

Mercy811: I'd like to keep `decodeCookiesAsDefault` because in this function value is still url decoded once. For `decodeCookiesForRuby`, how about `decodeCookiesByDoubleUrl`?

izaaz (approved):

PR #791, opened 2024-06-24
fix(session-replay-browser): remove @medv/finder as a dependency

lewgordon-amplitude:
### Summary

`@medv/finder` is not packaged to dual support ESM and CJS. We could own this repo or try to make things work upstream, but it doesn't seem like it's worth the effort of "vendoring" the file. Our build process will appropriately create ESM and CJS versions.

### Checklist


jxiwang (approved):
LGTM

PR #792, opened 2024-06-25
fix(remote config): prevent extra fetches of remote config

kwalker3690:
### Summary

2 changes in this PR:
1. Prevent remote config package from making multiple requests if response has been returned without certain keys - for example, if config response has `sampling_config` but not `targeting_config`, we should not keep refetching when package tries to access `targeting_config`
2. Type export and log level change in session replay

### Checklist

### DIFF
--- a/packages/analytics-remote-config/src/remote-config-idb-store.ts
+++ b/packages/analytics-remote-config/src/remote-config-idb-store.ts
@@ -95,9 +95,20 @@ export const createRemoteConfigIDBStore = async <RemoteConfig extends { [key: st
     return undefined;
   };
 
+  const remoteConfigHasValues = async <K extends keyof RemoteConfig>(configNamespace: string): Promise<boolean> => {
+    try {
+      const config = (await remoteConfigDB.getAll(configNamespace)) as RemoteConfig[K][];
+      return !!config.length;
+    } catch (e) {
+      loggerProvider.warn(`Failed to fetch remote config: ${e as string}`);
+    }
+    return false;
+  };
+
   return {
     storeRemoteConfig,
     getRemoteConfig,
     getLastFetchedSessionId,
+    remoteConfigHasValues,
   };
 };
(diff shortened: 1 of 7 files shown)


packages/analytics-remote-config/src/remote-config-idb-store.ts
```diff
@@ -95,9 +95,20 @@ export const createRemoteConfigIDBStore = async <RemoteConfig extends { [key: st
     return undefined;
   };
 
+  const remoteConfigHasValues = async <K extends keyof RemoteConfig>(configNamespace: string): Promise<boolean> => {
+    try {
+      const config = (await remoteConfigDB.getAll(configNamespace)) as RemoteConfig[K][];
```
lewgordon-amplitude, line 100: TODO it'd be nice to avoid the type assertion here.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -233,7 +233,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
   getShouldRecord(ignoreFocus = false) {
     if (!this.identifiers || !this.config || !this.identifiers.sessionId) {
-      this.loggerProvider.error(`Session is not being recorded due to lack of config, please call sessionReplay.init.`);
+      this.loggerProvider.warn(`Session is not being recorded due to lack of config, please call sessionReplay.init.`);
```
lewgordon-amplitude, line 236: Why the warn here instead of the error?

lewgordon-amplitude (approved):
LGTM!

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Because it's possible, especially when used as a standalone, to setup the SR SDK without a session id, and add the session id later, at which point recording starts. In this case, it's not an error to use it that way, so this log seemed confusing as an error

PR #796, opened 2024-06-26
test(targeting): add unit tests for different flag condition types

kwalker3690:
### Summary

Add unit tests to track different types of supported conditions

### Checklist

### DIFF
--- a/packages/targeting/test/flag-config-data/multiple-events.ts
+++ b/packages/targeting/test/flag-config-data/multiple-events.ts
@@ -1,0 +1,41 @@
+import { TargetingFlag } from '../../src/typings/targeting';
+
+export const flagConfigMultipleEvents: TargetingFlag = {
+  key: 'sr_targeting_config',
+  variants: {
+    on: { key: 'on' },
+    off: { key: 'off' },
+  },
+  segments: [
+    {
+      metadata: { segmentName: 'multiple event trigger' },
+      bucket: {
+        selector: ['context', 'session_id'],
+        salt: 'xdfrewd',
+        allocations: [
+          {
+            range: [0, 99],
+            distributions: [
+              {
+                variant: 'on',
+                range: [0, 42949673],
+              },
+            ],
+          },
+        ],
+      },
+      conditions: [
+        [
+          {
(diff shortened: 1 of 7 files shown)


packages/targeting/test/flag-config-data/multiple-events.ts
```diff
@@ -0,0 +1,41 @@
...
+          {
+            selector: ['context', 'event_types'],
+            op: 'set contains',
+            values: ['Add to Cart', 'Sign In'],
+          },
+        ],
+      ],
+    },
+    {
+      variant: 'off',
```
jackson-amplitude, line 38: it doesn't look like this is added to all segments. is this only for specific segment types?

packages/targeting/test/targeting.test.ts
```diff
@@ -71,8 +75,103 @@ describe('targeting', () => {
           session_id: 123,
           event_types: ['Add to Cart', 'Purchase'],
         },
-        [flagConfig],
+        [flagCatchAll],
       );
     });
   });
+
+  describe('condition tests', () => {
```
lewgordon-amplitude, line 83: It'd be nice to create a fixture here instead. Something like I did [here](https://github.com/amplitude/Amplitude-TypeScript/blob/584e1a2aefaea14d344dcb607cc77ec518c9567c/packages/session-replay-browser/test/hooks/click.test.ts#L153). That way we can keep most of the fields that are common and we don't care too much about for a particular test the same. It makes it easier to write smaller tests more quickly. Let me know what you think!

lewgordon-amplitude (approved):
LGTM!

jackson-amplitude (approved):
lgtm! Thanks for adding these to make sure we are aligned!

packages/targeting/test/targeting.test.ts
kwalker3690: For sure, I definitely like that approach too, although in some cases I like to see the entirety of a config in a test data file (that was helpful to me when understanding the entirety of the flag config in the first place). I'm going to leave these as is for now but happy to iterate in that direction in a future pr

PR #797, opened 2024-07-01
feat(session-replay-browser): support higher granularity sample rate

lewgordon-amplitude:
### Summary

Supports higher granularity sample rate, up from 2 to 6 decimal places. Performance impact here should be negligible: https://jsperf.app/xeditu/2

This would give different results for the same session, but because sessions are short lived it should be OK.

### Checklist


jxiwang (approved):
lgtm

jackson-amplitude (approved):
Good catch! does this mean <1% sample rates were not working before?

lewgordon-amplitude:
Yeah, exactly.

PR #798, opened 2024-07-03
feat(session-replay-browser): scroll events support

lewgordon-amplitude:
### Summary

Adds scroll event support. This tries its best to send scroll events which should be small on a page leave. It uses the Beacon API for reliability and performance with a fallback to XHR. I decided to add a few new classes:

1. ScrollWatcher -- Watches for max page scrolls and updates its state. It includes a hook for RRweb to trigger it to update its state and a BeaconTracker to send a Beacon API event for either 'pagehide' or 'beforeunload', whichever is supported by the browser.
2. BeaconTracker -- Wraps the Beacon API and provides an automatic fallback to XHR on failure.

There was a slight refactor to pull out common functions and make testing easier.

#### Testing

I tested this using the tasks-app in Chrome and Firefox though I may need to do a bit more testing. I'll especially need to do end to end testing once we are ready to ingest events.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/hooks/beacon.ts
+++ b/packages/session-replay-browser/src/hooks/beacon.ts
@@ -20,19 +20,11 @@ export class BeaconTransport<T> {
 
   constructor(context: Omit<SessionReplayDestinationSessionMetadata, 'deviceId'>, config: SessionReplayJoinedConfig) {
     const globalScope = getGlobalScope();
-    if (
-      globalScope &&
-      globalScope.navigator &&
-      typeof globalScope.navigator.sendBeacon === 'function' &&
-      typeof globalScope.Blob === 'function'
-    ) {
-      this.sendBeacon = (pageUrl, payload, contentType) => {
-        const blobData = new globalScope.Blob([JSON.stringify(payload)], {
-          type: contentType,
-        });
-
+    if (globalScope && globalScope.navigator && typeof globalScope.navigator.sendBeacon === 'function') {
+      this.sendBeacon = (pageUrl, payload, _) => {
         try {
-          if (globalScope.navigator.sendBeacon(pageUrl, blobData)) {
+          // something seems to be broken if we try to send the content type header in chrome
+          if (globalScope.navigator.sendBeacon(pageUrl, JSON.stringify(payload))) {
             return true;
           }
         } catch (e) {
--- a/packages/session-replay-browser/src/hooks/beacon.ts
+++ b/packages/session-replay-browser/src/hooks/beacon.ts
@@ -54,10 +54,7 @@ export class BeaconTransport<T> {
       type: String(type),
     });
 
-    void urlParams;
-
     this.pageUrl = `${pageUrl}?${urlParams.toString()}`;
-    // this.pageUrl = pageUrl;
   }
 
   send(payload: T) {
--- a/packages/session-replay-browser/src/hooks/beacon.ts
+++ b/packages/session-replay-browser/src/hooks/beacon.ts
@@ -3,27 +3,35 @@ import { SessionReplayJoinedConfig } from '../config/types';
 import { SessionReplayDestinationSessionMetadata } from '../typings/session-replay';
 import { getServerUrl } from '../helpers';
 
-type BeaconSendFn<T> = (pageUrl: string, payload: T, contentType: 'application/json') => boolean;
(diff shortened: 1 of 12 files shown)


packages/session-replay-browser/src/hooks/beacon.ts
```diff
@@ -0,0 +1,67 @@
...
+  private readonly basePageUrl: string;
+  private readonly context: Omit<SessionReplayDestinationSessionMetadata, 'deviceId'>;
+  private static readonly contentType = 'application/json';
+
+  constructor(context: Omit<SessionReplayDestinationSessionMetadata, 'deviceId'>, config: SessionReplayJoinedConfig) {
+    const globalScope = getGlobalScope();
+    if (globalScope && globalScope.navigator && typeof globalScope.navigator.sendBeacon === 'function') {
+      this.sendBeacon = (pageUrl, payload, _) => {
+        try {
+          // something seems to be broken if we try to send the content type header in chrome
```
kwalker3690, line 26: What do you mean by this comment?

lewgordon-amplitude: For some reason in testing it always sent OPTIONS instead of POST to my test server.

lewgordon-amplitude: Yep:

From: https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API
> Starting in Chrome 59, this method cannot send a Blob whose type is not CORS safelisted. This is a temporary change until a mitigation can be found for the security issues that this creates. For more information, see [bug 720283](https://crbug.com/720283).

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -74,6 +83,27 @@ export class SessionReplay implements AmplitudeSessionReplay {
     this.config.privacyConfig.unmaskSelector = dropInvalidSelectors(this.config.privacyConfig.unmaskSelector);
   }
 
+  private setupEventListeners = (teardown: boolean) => {
+    const globalScope = getGlobalScope();
```
kwalker3690, line 87: I'd rather we have a `teardownEventListeners` and a `setupEventListeners` function (which would call `teardownEventListeners`) than pass `teardown` as a param here, using `setupEventListeners` in the `shutdown` method looks confusing to me, wdyt?

lewgordon-amplitude: Yeah that SGTM, I had some troubles with naming here.

packages/session-replay-browser/src/hooks/beacon.ts
```diff
@@ -0,0 +1,67 @@
+import { getGlobalScope } from '@amplitude/analytics-client-common';
```
kwalker3690, line 1: Nit: this itself isn't a hook - could it go in a helpers folder within the hooks folder?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -89,6 +119,19 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
 
+    if (options.sessionId && this.config.interactionConfig?.enabled) {
+      const transport = new BeaconTransport<ScrollEvent>(
+        {
+          sessionId: options.sessionId,
+          type: 'interaction',
+        },
+        this.config,
+      );
+      const scrollWatcher = new ScrollWatcher(transport);
```
kwalker3690, line 130: With a goal of keeping this top level class as simple as possible, ie only knowing what it needs to know about, wdyt about moving the instantiation of BeaconTransport to the constructor of ScrollWatcher? so that ScrollWatcher takes in the sessionId, type, and config, and calls `new BeaconTransport` within itself

lewgordon-amplitude: I'm thinking about doing this with another constructor. Being able to pass in the transport optionally makes it really easy to test.

packages/session-replay-browser/src/hooks/beacon.ts
kwalker3690: Yeah that sounds good!

packages/session-replay-browser/src/session-replay.ts
kwalker3690: That works for me

kwalker3690 (approved):

jxiwang (approved):
LGTM, like the usage of the Beacon API here!

PR #808, opened 2024-07-11
feat(session replay): support targeting via passing user properties

kwalker3690:
### Summary

The current implementation of targeting via user properties relied on a Browser SDK instance being present (it used the analytics connector functionality). This PR updates the logic in the standalone to accept user properties as config and arguments to setSessionId and to evaluateTargetingAndRecord, and updates the plugin to pass user properties when appropriate.

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/helpers.ts
+++ b/packages/plugin-session-replay-browser/src/helpers.ts
@@ -1,0 +1,25 @@
+import { Event, IdentifyOperation } from '@amplitude/analytics-types';
+import { PROPERTY_ADD_OPERATIONS } from './constants';
+
+export const parseUserProperties = (event: Event) => {
+  if (!event.user_properties) {
+    return;
+  }
+  let userPropertiesObj = {};
+  const userPropertyKeys = Object.keys(event.user_properties);
+
+  userPropertyKeys.forEach((identifyKey) => {
+    if (
+      Object.values(IdentifyOperation).includes(identifyKey as IdentifyOperation) &&
+      PROPERTY_ADD_OPERATIONS.includes(identifyKey as IdentifyOperation)
+    ) {
+      const typedUserPropertiesOperation =
+        event.user_properties && (event.user_properties[identifyKey as IdentifyOperation] as Record<string, any>);
+      userPropertiesObj = {
+        ...userPropertiesObj,
+        ...typedUserPropertiesOperation,
+      };
+    }
+  });
+  return userPropertiesObj;
+};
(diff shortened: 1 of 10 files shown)


packages/plugin-session-replay-browser/src/session-replay.ts
lewgordon-amplitude, line 1: When do we decide to use the event for user properties rather grabbing it from the store? Is this something we can make into a separate method, since I see it used a couple of times.

lewgordon-amplitude (commented):
LGTM overall, but I just wanted to understand a little more about when we grab properties from the event or store.

packages/plugin-session-replay-browser/src/session-replay.ts
kwalker3690: Yeah that's a good question, here's the logic we need:

1. On init, grab any previously set user properties from the store
2. On session id update, grab any previously set user properties from the store
3. On $identify event occurring, grab the user properties from the event operation - this is necessary because this event hasn't been processed yet, so the properties are not yet available from the store. There is a future improvement we could make here to make this cleaner and more accurate. I'd like to update the plugin interface to have a special hook for identify events that would perform the identify operation on the user properties, then make the updated user properties available to the execute function. But that's a larger change I'll need to discuss with the DX team, I wanted to get out a simpler solution now!

lewgordon-amplitude (approved):
LGTM!

packages/plugin-session-replay-browser/src/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  let userPropertiesObj = {};
+  const userPropertyKeys = Object.keys(event.user_properties);
+
+  userPropertyKeys.forEach((identifyKey) => {
+    if (
+      Object.values(IdentifyOperation).includes(identifyKey as IdentifyOperation) &&
+      PROPERTY_ADD_OPERATIONS.includes(identifyKey as IdentifyOperation)
+    ) {
+      const typedUserPropertiesOperation =
+        event.user_properties && (event.user_properties[identifyKey as IdentifyOperation] as Record<string, any>);
```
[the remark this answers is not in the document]
kwalker3690: We're checking to see if its defined outside of the function, so TS is saying "yeah you checked if it was defined at the time this function was created, but when this function is running, the value of event.user_properties could've changed since the function was created". I'd need to add an ```if (!event.user_properties) {
    return;
  }```
at the top of the forEach function to not have to do the check here as well.

There is also a typing issue with the `user_properties` object - I know at this point that it will definitely be a key that is one of the IdentifyOperations, but the type is written so that it could just be a generic string key as well, with a value of any. I believe this is actually not possible based on the documentation for the $identify event, but I'm not going to change the type definition right now, and our checks ensure that we are looking for one of the IdentifyOperations anyways

kwalker3690:
Something is out of sync on this PR, I'm going to close and try to reopen again

PR #810, opened 2024-07-15
test(session replay plugin): Add Browser SDK 2.0 integration tests in SR Plugin

kwalker3690:
### Summary

Setting up a few integration tests between the Browser SDK 2.0 and the SR plugin. I've left a few commented out that I'd still like to implement, but this is a start!

### Checklist


packages/plugin-session-replay-browser/test/jsdom-extended.js
```diff
@@ -0,0 +1,16 @@
+const JSDOMEnvironment = require('jest-environment-jsdom').default; // or import JSDOMEnvironment from 'jest-environment-jsdom' if you are using ESM modules
```
lewgordon-amplitude, line 1: Why not use the ESM syntax here instead?

lewgordon-amplitude (approved):
LGTM!

packages/plugin-session-replay-browser/package.json
```diff
@@ -46,9 +46,13 @@
     "tslib": "^2.4.1"
   },
   "devDependencies": {
+    "@amplitude/analytics-browser": ">=2 <3",
```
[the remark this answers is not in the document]
kwalker3690: It won't for end customers, since they will not install this package when they install the plugin (that's why theres a difference b/t dependencies and devDependencies). I haven't observed any issues when developing locally with this, but let's observe if there are any!

packages/plugin-session-replay-browser/test/jsdom-extended.js
kwalker3690: Idk this was a copy/paste from a stack overflow post, i think this file doesnt get transpiled

PR #811, opened 2024-07-15
fix: should disable autocapture plugin by default

Mercy811:
### Summary

- Deprecate defaultTracking with autocapture and rename to userInteractions
   - autocapture has all options from defaultTracking and plus userInteractions
   - if autocapture is set, it will override defaultTracking
- Disable autocapture.userInteractions by default. The autocapture plugin will only be installed if 
   - autocapture === true
   - autocapture.userInteractions === true
   - autocapture.userInteractions === object
- Add config.fetchRemoteConfig
- Fix: should take remote config values even when it's value is `false` by using `in` operator
- CI: use branch name as tag name and append it to the prerelease version

Note:
- For changes in `packages/analytics-browser-test/`, I reverted and deleted request metadata as fetch remote config is disable by default now
- I also added some tests when fetch remote config is enabled. However, due to unclear reasons, Github actions doesn't build the package correctly and get the following error so that I have to remove them. Those tests are running correctly locally. I will try to add them in a separate PR.
```
Module '"@amplitude/analytics-core"' has no exported member 'RequestMetadata'.
@amplitude/analytics-browser-test:     10 import { RequestMetadata, UUID } from '@amplitude/analytics-core';

Object literal may only specify known properties, and 'fetchRemoteConfig' does not exist in type 'Omit<Partial<ExternalBrowserConfig>, HiddenOptions>'.
```

### Checklist


packages/analytics-browser/src/config/types.ts
yuhao900914, line 1: Should we move this to packages/analytics-types/src/config/browser.ts?

packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -30,13 +30,11 @@ export class BrowserJoinedConfigGenerator {
         this.remoteConfigFetch &&
         (await this.remoteConfigFetch.getRemoteConfig('analyticsSDK', 'browserSDK', this.config.sessionId));
       this.config.loggerProvider.debug('Remote configuration:', JSON.stringify(remoteConfig, null, 2));
-      if (remoteConfig && remoteConfig.defaultTracking) {
-        this.config.defaultTracking = remoteConfig.defaultTracking;
-      }
-      if (remoteConfig && remoteConfig.autocapture) {
+      if (remoteConfig && 'autocapture' in remoteConfig) {
         this.config.autocapture = remoteConfig.autocapture;
+        this.config.defaultTracking = this.config.autocapture;
```
yuhao900914, line 35: What will happen if autocapture has userInteraction value but it's not matching the type of defaultTracking?

Mercy811: I will also pass userInteraction to config.defaultTracking because `DefaultTrackingOptions` is an interface not a type. The test `should disable userInteractions if remote config sets it to false` should cover this case and make sure there is not run time error.

packages/analytics-browser/src/config/types.ts
Mercy811: Good catch! I move it to there at https://github.com/amplitude/Amplitude-TypeScript/pull/811/commits/66610df58ad5a6e171154cc35963900aaa05345c

yuhao900914 (approved):

PR #816, opened 2024-07-23
fix(session-replay-browser): remote config failure handling

lewgordon-amplitude:
### Summary

We used to record on remote config fetch failure, however with privacy config added it's important that if we fail to fetch any config that we do not record. Otherwise, we risk exposing unwanted text.

### Checklist


packages/session-replay-browser/test/integration/sampling.test.ts
```diff
@@ -117,10 +117,8 @@ describe('module level integration', () => {
           // Ensure that sample rate matches what's passed in the options
```
kwalker3690, line 117: Nit: can you update the test label to show that its on purpose now that replays aren't captured?

kwalker3690 (approved):

PR #817, opened 2024-07-23
fix(remote config): return empty config from indexeddb

kwalker3690:
### Summary

It's a valid case for the remote config call to return no configs. In this case, we don't want to keep fetching the remote config, it should apply by the same rules as if configs were returned, and we should read from IndexedDB instead. We can leverage the lastFetchedSessionId matching the current to know if we've successfully fetched the config already.

### Checklist


lewgordon-amplitude (approved):
LGTM!

PR #818, opened 2024-07-24
fix(session replay): set sample rate if it is 0 as well

kwalker3690:
### Summary
I just observed that a sample rate of 0 wasn't being respected when sent via remote config, due to an improper falsy check. This is quite an edge case but figured I'd fix it upon observation anyways.

### Checklist


lewgordon-amplitude (approved):
LGTM!

PR #819, opened 2024-07-24
fix(session replay): ignore focus should allow recording to continue as well

kwalker3690:
### Summary
We should always record if debugMode is true, otherwise the replay will be a confusing experience.

### Checklist


lewgordon-amplitude (approved):
LGTM!

PR #821, opened 2024-07-24
fix(session replay): remove focus/blur handling

kwalker3690:
### Summary
I believe we can now remove our focus/blur handling to stop/start recording when moving between tabs. Many customers have run into issues around this part of our logic, so much that we've had to add a `debugMode` option to our sdk to ease their pain. Reflecting on this recently, I realized that we no longer need this logic for the following reasons:

- Because of the IndexedDB refactor I did a few months ago, we now use IndexedDB as our source of truth for event lists. So the cross tab communication already works seamlessly without doing needing to sync the IndexedDB state to memory (this can be observed by looking at the sequenceId numbers in the track calls when switching back and forth between tabs, the sequenceIds automatically jump ahead to match the state from the previous tab)
- RRweb likely won't emit any events when the user is not on the tab (RRweb is powered via MutationObserver, mutations likely won't occur when a user is not on the tab). This is the riskier part to me, in that if RRweb is emitting events simultaneously in two tabs, the replay could have a choppy/confusing experience. However I've reviewed a few of our competitor's implementations of RRweb and noticed none of them have tab handling logic, which makes me think this is not a large scale issue.

Here is an example replay taken without the focus/blur handler: https://app.amplitude.com/analytics/org/36958/project/408823/search/amplitude_id%3D950783228955?collapsed=true&sessionHandle=ZDmIW0Y_ZDmIW0Y_N1fGLwb_Bjz3_c77fdc1a-3ee5-4f57-82c8-f105f61dee3a&sessionReplayEventId=63af4c88-11bf-496a-ac8e-ede8df0fb820

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -86,9 +86,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
     const globalScope = getGlobalScope();
     if (globalScope) {
       globalScope.removeEventListener('blur', this.blurListener);
-      globalScope.removeEventListener('focus', this.focusListener);
       !teardown && globalScope.addEventListener('blur', this.blurListener);
-      !teardown && globalScope.addEventListener('focus', this.focusListener);
       // prefer pagehide to unload events, this is the standard going forward. it is not
       // 100% reliable, but is bfcache-compatible.
       if (globalScope.self && 'onpagehide' in globalScope.self) {
@@ -158,13 +156,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
     this.teardownEventListeners(false);
 
-    const globalScope = getGlobalScope();
-    // If the user is in debug mode, ignore the focus handler when tagging events.
-    // this is a common mishap when someone is developing locally and not seeing events getting tagged.
-    const ignoreFocus = !!this.config.debugMode;
-    if (!ignoreFocus && globalScope && globalScope.document && globalScope.document.hasFocus()) {
-      this.initialize(true);
-    }
+    this.initialize(true);
   }
 
   setSessionId(sessionId: number, deviceId?: string) {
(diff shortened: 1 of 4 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -347,9 +323,10 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
     this.recordCancelCallback = record({
       emit: (event) => {
-        const globalScope = getGlobalScope();
-        if ((globalScope && globalScope.document && !globalScope.document.hasFocus()) || !this.getShouldRecord()) {
-          this.stopRecordingAndSendEvents();
+        if (this.shouldOptOut()) {
```
lewgordon-amplitude, line 326: Should we have been doing this before? I wasn't sure if there was part of this refactor that caused us to add this now? Do we expect the opt out to happen during the middle of a session so that we need to check this on every emit? It seems like this is hitting some store that may affect performance?

lewgordon-amplitude (commented):
LGTM overall, but I had a quick question first

kwalker3690:
> So for this PR this is just removing the focus handling it seems. In a later PR are we eventually removing the blur handler?

I modified the blur handle to only send the events now, I thought that seemed like a decent thing to do still since some people may never come back to an open tab (or not for a long time), so I thought we should still be proactive about sending events in that case

packages/session-replay-browser/src/session-replay.ts
kwalker3690: We actually were doing this before, a little more implicitly. We check for shouldOptOut within `getShouldRecord`, so the previous code path would've done the same behavior as this does. This is likely not how customers would use our SDK, but I added this logic just in case they want to start recording anonymous users, then stop recording them once they've done something like clicked 'don't track me' on a cookie acceptance modal. I think it's more likely that customers would only start recording after a user accepts cookies to be on the safer side, but that's the background on this anyways. I do think we could leave it out if there is concern about the logic here - as far as performance, I don't have a concern there, the identity store we look for the optOut value in is just an in memory store: https://github.com/amplitude/experiment-js-client/blob/main/packages/analytics-connector/src/identityStore.ts

lewgordon-amplitude (approved):
🚢

PR #824, opened 2024-07-25
fix(session replay): restart recording on focus

kwalker3690:
### Summary

Jackson pointed out that we will attribute incremental snapshots incorrectly if we don't take a full snapshot on a tab change. This updates the SDK to restart recording on focus (which will cause a full snapshot to be taken).

Example incorrect replay capture without this change: https://app.amplitude.com/analytics/amplitude/project/408823/search/amplitude_id%3D951490210053

Example correct replay capture with this change:
https://app.amplitude.com/analytics/amplitude/project/408823/search/amplitude_id%3D951502713658

### Checklist


jackson-amplitude (approved):
lgtm! at some point, we should look into adding tab awareness to the backend.

PR #825, opened 2024-07-29
feat(session-replay-browser): support debug info in session replay SDK

lewgordon-amplitude:
### Summary

* This sends RRweb events that include the config for internal debugging purposes.
* I also fixed the version being incorrect in the SDK.

Tested this locally.

### Checklist


kwalker3690 (approved):
This adds a single event upon calling record, right? Just confirming its not adding a ton of extra events here

lewgordon-amplitude:
@kwalker3690 correct, but I did notice in testing record can be called multiple times. It shouldn't be a ton of events in general though.

PR #832, opened 2024-07-30
feat: remote config

Mercy811:
### Summary

This PR adds the feature remote config.

Note that this PR doesn't merge from the feature branch `feature-remote-config` as it has release commits which introduce conflicts. So I cherry pick all 6 commits from the feature branch `feature-remote-config` without the release commits.

I pushed another commit https://github.com/amplitude/Amplitude-TypeScript/pull/832/commits/637940680f61d060156ccb20fb5c6a9051aa9a40 to upgrade to the latest autocapture plugin 1.x

### Checklist

### DIFF
--- a/packages/analytics-types/src/element-interactions.ts
+++ b/packages/analytics-types/src/element-interactions.ts
@@ -27,7 +27,7 @@ export const DEFAULT_CSS_SELECTOR_ALLOWLIST = [
  */
 export const DEFAULT_DATA_ATTRIBUTE_PREFIX = 'data-amp-track-';
 
-export interface UserInteractionsOptions {
+export interface ElementInteractionsOptions {
   /**
    * List of CSS selectors to allow auto tracking on.
    * When provided, allow elements matching any selector to be tracked.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,9 +7,6 @@ on:
       - v1.x
   pull_request:
     types: [opened, synchronize]
-    branches:
-      - main
-      - v1.x
 
 jobs:
   build:
--- a/.github/workflows/feature-branch-prerelease.yml
+++ b/.github/workflows/feature-branch-prerelease.yml
@@ -89,6 +89,9 @@ jobs:
           echo "PREID=$(echo '${{ steps.branch-name.outputs.current_branch }}' | tr -cd '[:lower:]')" >> $GITHUB_ENV
 
       # Use --no-push to prevent pushing to remote
+      # Example:
+      #   - feature-branch-name -> featurebranchname
+      #   - minor: 1.0.0 -> 1.1.0-featurebranchname.0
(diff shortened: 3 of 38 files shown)


yuhao900914 (approved):

packages/analytics-types/src/element-interactions.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Default CSS selectors to define which elements on the page to track.
+ * Extend this list to include additional elements to track. For example:
+ * ```
+ * autocapturePlugin({
+ *    cssSelectorAllowlist: [...DEFAULT_CSS_SELECTOR_ALLOWLIST, ".my-class"],
+ * })
+ * ```
+ */
+export const DEFAULT_CSS_SELECTOR_ALLOWLIST = [
```
Dogfalo, line 14: Can this be exported from the `plugin-autocapture-browser` instead? That way we don't have to update it in two places.

packages/analytics-types/src/element-interactions.ts
```diff
@@ -0,0 +1,78 @@
...
+  '[data-amp-default-track]',
+  '.amp-default-track',
+];
+
+/**
+ * Default prefix to allo the plugin to capture data attributes as an event property.
+ */
+export const DEFAULT_DATA_ATTRIBUTE_PREFIX = 'data-amp-track-';
+
+export interface ElementInteractionsOptions {
```
Dogfalo, line 30: Same with this interface

Dogfalo: Recently we had to add one additional property inside this called `actionClickAllowlist`, I'm curious if anything needs to change other than this to support it?

packages/analytics-types/src/element-interactions.ts
Mercy811: `analytics-types` is a dependency of `plugin-autocapture-browser`. It's a circular dependency if we import these from `plugin-autocapture-browser` in `analytics-types`.

packages/analytics-types/src/element-interactions.ts
Mercy811: Once this version is released, we should import these `const` and `interface` from `analytics-types` in `plugin-autocapture-browser`. I copied these from v0.9.2, so we need to all changes after it

packages/analytics-types/src/element-interactions.ts
Dogfalo: Sounds good, we can think of some way to resolve this is the next version

Dogfalo (approved):

packages/analytics-types/src/element-interactions.ts
Mercy811: If we keep those in analytics-types, we don’t have circular dependency and this is the pattern other packages, for example page view tracking plugin.

PR #834, opened 2024-08-01
feat(session replay): add logging to SR SDK

kwalker3690:
### Summary

This PR adds extra logs to the SR SDK lifecycle, as well as adds storage size data to the custom event we are sending along with our rrweb events.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/helpers.ts
+++ b/packages/session-replay-browser/src/helpers.ts
@@ -141,13 +141,10 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone): string => {
 export const getStorageSize = async (): Promise<StorageData> => {
   const globalScope = getGlobalScope();
   if (globalScope) {
-    const storeEstimate: ChromeStorageEstimate = await globalScope.navigator.storage.estimate();
-    const totalStorageSize = storeEstimate.usage ? Math.round(storeEstimate.usage / KB_SIZE) : 0;
-    const percentOfQuota =
-      storeEstimate.usage && storeEstimate.quota
-        ? Math.round((storeEstimate.usage / storeEstimate.quota + Number.EPSILON) * 1000) / 1000
-        : 0;
-    return { totalStorageSize, percentOfQuota, usageDetails: JSON.stringify(storeEstimate.usageDetails) };
+    const { usage, quota, usageDetails }: ChromeStorageEstimate = await globalScope.navigator.storage.estimate();
+    const totalStorageSize = usage ? Math.round(usage / KB_SIZE) : 0;
+    const percentOfQuota = usage && quota ? Math.round((usage / quota + Number.EPSILON) * 1000) / 1000 : 0;
+    return { totalStorageSize, percentOfQuota, usageDetails: JSON.stringify(usageDetails) };
   }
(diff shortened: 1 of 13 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -406,19 +405,23 @@ export class SessionReplay implements AmplitudeSessionReplay {
     const { apiKey } = config;
     config.apiKey = `****${apiKey.substring(apiKey.length - 4)}`;
 
+    const storageSizeData = await getStorageSize();
+
     return {
       config,
       version: VERSION,
+      ...storageSizeData,
     };
```
lewgordon-amplitude, line 414:
```suggestion
    return {
      ...storageSizeData,
      config,
      version: VERSION,
    };
```

This is probably unlikely, but I wouldn't want `storageSizeData` to potentially override config/version

packages/session-replay-browser/src/helpers.ts
```diff
@@ -130,3 +137,17 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone): string => {
...
 };
+
+export const getStorageSize = async (): Promise<StorageData> => {
+  const globalScope = getGlobalScope();
+  if (globalScope) {
+    const storeEstimate: ChromeStorageEstimate = await globalScope.navigator.storage.estimate();
+    const totalStorageSize = storeEstimate.usage ? Math.round(storeEstimate.usage / KB_SIZE) : 0;
+    const percentOfQuota =
+      storeEstimate.usage && storeEstimate.quota
+        ? Math.round((storeEstimate.usage / storeEstimate.quota + Number.EPSILON) * 1000) / 1000
```
lewgordon-amplitude, line 148: Why the use of `Number.EPSILON` here?

packages/session-replay-browser/src/helpers.ts
```diff
@@ -130,3 +137,17 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone): string => {
...
+        : 0;
+    return { totalStorageSize, percentOfQuota, usageDetails: JSON.stringify(storeEstimate.usageDetails) };
+  }
```
lewgordon-amplitude, line 151:
```suggestion
    const {usage, quota, usageDetails}: ChromeStorageEstimate = await globalScope.navigator.storage.estimate();
    const totalStorageSize = usage ? Math.round(usage / KB_SIZE) : 0;
    const percentOfQuota =
      usage && quota
        ? Math.round((usage / quota + Number.EPSILON) * 1000) / 1000
        : 0;
    return { totalStorageSize, percentOfQuota, usageDetails: JSON.stringify(usageDetails) };
```

Nit: I think this is slightly cleaner?

lewgordon-amplitude (approved):
Nice! I think this is going to be really useful to help debug 💪

packages/session-replay-browser/src/helpers.ts
kwalker3690: It's a surprisingly controversial way of helping JS to round a number with floating points: https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary

PR #835, opened 2024-08-01
fix(session replay): add version and sdk type

jxiwang:
### Summary

Adding some version data to the headers.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/track-destination.ts
+++ b/packages/session-replay-browser/src/track-destination.ts
@@ -111,7 +111,7 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
       seq_number: `${context.sequenceId}`,
       type: `${context.type}`,
     });
-
+    const sessionReplayLibrary = `${context.version?.type || 'standalone'}/${context.version?.version || version}`;
     const payload = this.payloadBatcher({
       version: 1,
       events: context.events,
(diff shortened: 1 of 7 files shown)


packages/session-replay-browser/src/track-destination.ts
```diff
@@ -122,13 +122,18 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
     try {
       const options: RequestInit = {
         headers: {
           'Content-Type': 'application/json',
           Accept: '*/*',
           Authorization: `Bearer ${apiKey}`,
-          'X-Client-Version': version,
+          'X-Client-SR-Browser-SDK-Version': version,
+          'X-Client-SR-Type': sessionReplayType,
+          'X-Client-SR-Version': sessionReplayVersion,
```
jxiwang, line 136: @jackson-amplitude Does this work for tracking versions of SDK?

jackson-amplitude: it might be better to keep the `X-Client-Version` if that's the contract with mobile too. The type and version can be captured in the value `<type>/<version>`.  @weibrian @justin-fiedler, do you guys remember the format of the value?

jxiwang: Okay, I'll keep the same name for the standalone SDK. For the Plugin/Segment SDK version one field in the form of <type>/<version> is best?

jackson-amplitude: Sorry for the back and forth. I think this is what we added/used for mobile is `X-Client-Library`: https://amplitude.sourcegraph.com/github.com/amplitude/nova/-/blob/src/main/java/com/amplitude/sessionreplay/ingestion/SessionReplayServletV2.java?L123:47-123:63

That contains the library type and the version in the format of `<type>/<version>`: https://amplitude.sourcegraph.com/github.com/amplitude/session-replay-android@dfc122cc305762fb55c91e145fc6878f52897e9b/-/blob/session-replay/src/test/java/com/amplitude/android/sessionreplay/server/SessionReplayServerTest.kt?L29

For us, probably something like `standalone/1.0.0`, `segment/1.0.0`, `plugin/1.0.0` would be reasonable (the type name can be whatever you have in this PR as long as it's clear)

jxiwang: Oh thanks! Forgot the Java headers are already set. Will change it over to `X-Client-Library`

jackson-amplitude (approved):
lgtm!

PR #837, opened 2024-08-02
chore(session replay): add custom rrweb event method and fire on getSessionReplayProperties

kwalker3690:
### Summary

Add a tool to fire custom rrweb events when recording, and add an event when getSessionReplayProperties is called.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -7,6 +7,7 @@ import { createSessionReplayJoinedConfigGenerator } from './config/joined-config
 import { SessionReplayJoinedConfig, SessionReplayJoinedConfigGenerator } from './config/types';
 import {
   BLOCK_CLASS,
+  CustomRRwebEvent,
   DEFAULT_SESSION_REPLAY_PROPERTY,
   INTERACTION_MAX_INTERVAL,
   INTERACTION_MIN_INTERVAL,
@@ -172,9 +173,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
       if (this.config.debugMode) {
         eventProperties[SESSION_REPLAY_DEBUG_PROPERTY] = this.getSessionReplayDebugPropertyValue();
       }
+      void this.addCustomRRWebEvent(CustomRRwebEvent.GET_SR_PROPS, {
+        shouldRecord,
+        eventProperties: eventProperties,
+      });
       return eventProperties;
     }
 
+    void this.addCustomRRWebEvent(CustomRRwebEvent.GET_SR_PROPS, {
+      shouldRecord,
+      eventProperties: {},
+    });
     return {};
   }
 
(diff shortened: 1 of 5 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -205,9 +205,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
       if (this.config.debugMode) {
         eventProperties[SESSION_REPLAY_DEBUG_PROPERTY] = this.getSessionReplayDebugPropertyValue();
       }
+      void this.addCustomRRWebEvent('get-sr-props', {
+        shouldRecord,
+        eventProperties: eventProperties,
+      });
```
lewgordon-amplitude, line 211: The only thing I'm wondering about here is if this should also check the storage size estimate every time. It is async, but I wonder if there's a lot of overhead for checking that on every Amplitude event.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -205,9 +205,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
       return eventProperties;
     }
 
+    void this.addCustomRRWebEvent('get-sr-props', {
```
lewgordon-amplitude, line 215: I think we should probably enum our custom event types `'get-sr-props'`. Creating our own mini taxonomy here :)

lewgordon-amplitude (commented):
This is going to be great to have, but I have a question on performance implications before we merge it.

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Hmm I'm conflicted, we don't necessarily need it for every event, but if we don't include it in events we may only capture it at the beginning of the recording and miss out on valuable data from further in the recording. From a quick google search I don't see any notes about performance impact from this method. We are not blocking the execution of the event on this (hence the void), so it will allow the event to process through as normal. The other option would be to throttle it, to send it say only every 10 events or so, wdyt?

lewgordon-amplitude: >The other option would be to throttle it, to send it say only every 10 events or so, wdyt?

Yeah I think that works. An alternative would be if it's possible to determine at key points, but I think the sampling should be fine.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -164,18 +166,30 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
-    return {};
+    void this.addCustomRRWebEvent(
+      CustomRRwebEvent.GET_SR_PROPS,
+      {
+        shouldRecord,
+        eventProperties: eventProperties,
+      },
+      this.eventCount === 10,
+    );
+    if (this.eventCount === 10) {
```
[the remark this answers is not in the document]
kwalker3690: LOL yes, knew I was missing a piece here, thanks for catching that!

lewgordon-amplitude (approved):
🚢 !

PR #841, opened 2024-08-06
feat(session replay): correct timing of replay capture, and use targeting config or sampling config

kwalker3690:
### Summary

This PR covers the following changes:
- Choose either targeting config or sampling config (if no targeting config exists) to determine if users should be captured
- Ensure the timing of replay capture is as follows:
  - on init, stop capture and start anew (in the case a user is already being captured) - matches existing behavior
  - on session id change, stop capture and start anew - matches existing behavior
  - when evaluateTargetingAndCapture is called, only start capture if it hasn't been started yet - this is because `evaluateTargetingAndCapture` is called on every event - we don't want to cause a full dom snapshot on every event.
- Changed the names of any methods that have `record` in them to `capture`

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -197,7 +197,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
     // Restart recording on focus to ensure that when user
     // switches tabs, we take a full snapshot
     this.stopRecordingEvents();
-    this.captureEventsIfShould();
+    this.captureEvents();
   };
 
   /**
@@ -239,7 +239,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
       });
     }
 
-    this.captureEventsIfShould();
+    this.captureEvents();
   };
 
   sendEvents(sessionId?: number) {
@@ -288,12 +288,10 @@ export class SessionReplay implements AmplitudeSessionReplay {
           `Not capturing replays for session ${this.identifiers.sessionId} due to not matching targeting conditions.`,
         );
         return false;
-      } else {
-        this.loggerProvider.log(
-          `Capturing replays for session ${this.identifiers.sessionId} due to matching targeting conditions.`,
-        );
-        return true;
       }
+      this.loggerProvider.log(
+        `Capturing replays for session ${this.identifiers.sessionId} due to matching targeting conditions.`,
+      );
     } else {
       const isInSample = isSessionInSample(this.identifiers.sessionId, this.config.sampleRate);
       if (!isInSample) {
@@ -334,14 +332,15 @@ export class SessionReplay implements AmplitudeSessionReplay {
     return maskSelector as unknown as string;
   }
 
-  captureEventsIfShould() {
-    const shouldRecord = this.getShouldCapture();
-    const sessionId = this.identifiers?.sessionId;
-    if (!shouldRecord || !sessionId || !this.config) {
-      return;
-    }
+  captureEvents() {
     if (this.recordCancelCallback) {
       this.loggerProvider.debug('captureEvents method fired - Session Replay capture already in progress.');
+      return;
+    }
+
+    const shouldRecord = this.getShouldCapture();
+    const sessionId = this.identifiers && this.identifiers.sessionId;
+    if (!shouldRecord || !sessionId || !this.config) {
       return;
     }
 
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -128,7 +128,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
     this.teardownEventListeners(false);
 
-    await this.evaluateTargetingAndRecord({ userProperties: options.userProperties });
+    this.stopRecordingEvents();
+    await this.evaluateTargetingAndCapture({ userProperties: options.userProperties });
   }
 
   setSessionId(sessionId: number, deviceId?: string, options?: { userProperties?: { [key: string]: any } }) {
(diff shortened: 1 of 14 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -318,20 +334,24 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    const shouldRecord = this.getShouldCapture();
     const sessionId = this.identifiers?.sessionId;
     if (!shouldRecord || !sessionId || !this.config) {
       return;
     }
-    this.stopRecordingEvents();
+    if (this.recordCancelCallback) {
+      this.loggerProvider.debug('captureEvents method fired - Session Replay capture already in progress.');
+      return;
+    }
```
lewgordon-amplitude, line 346: Maybe this is worth putting at the top of the method? It seems like we'll always want to exit early if we're already capturing?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -318,20 +334,24 @@ export class SessionReplay implements AmplitudeSessionReplay {
     return maskSelector as unknown as string;
   }
 
-  recordEvents() {
-    const shouldRecord = this.getShouldRecord();
+  captureEventsIfShould() {
```
lewgordon-amplitude, line 337: Can we instead call this `captureEvents` or add a doc string here, since at first glance it's not immediately obvious what if should is in this context.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -281,14 +274,37 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        this.loggerProvider.log(
+          `Not capturing replays for session ${this.identifiers.sessionId} due to not matching targeting conditions.`,
+        );
+        return false;
+      } else {
+        this.loggerProvider.log(
+          `Capturing replays for session ${this.identifiers.sessionId} due to matching targeting conditions.`,
+        );
+        return true;
+      }
```
lewgordon-amplitude, line 296: WDYT about this?

```suggestion
      }
      this.loggerProvider.log(
        `Capturing replays for session ${this.identifiers.sessionId} due to matching targeting conditions.`,
      );
```

It seems more inline with the else which only returns early if it needs to.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -193,9 +196,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
   focusListener = () => {
     // Restart recording on focus to ensure that when user
     // switches tabs, we take a full snapshot
-    if (this.sessionTargetingMatch) {
-      this.recordEvents();
-    }
+    this.stopRecordingEvents();
+    this.captureEventsIfShould();
```
lewgordon-amplitude, line 200: Depending on the site, this may cause a lot of full snapshots. For example try Google Search and click on the nine dots next to your profile picture in the top right. For me that causes two focus events without switching tabs. I wonder if there is a better way to support tab switching.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -209,44 +211,35 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        sessionId: this.identifiers.sessionId,
+        targetingConfig: this.config.targetingConfig,
+        loggerProvider: this.loggerProvider,
+        apiKey: this.config.apiKey,
+        targetingParams: { userProperties: targetingParams?.userProperties, event: eventForTargeting },
+      });
     }
 
-    return this.sessionTargetingMatch;
+    this.captureEventsIfShould();
```
lewgordon-amplitude, line 242: I wonder if there's a longer term way to refactor this so that the methods aren't co-dependent on knowing what each other is doing. Maybe more specifically this knows that captureEventsIfShould will use sessionTargetingMatch.

packages/session-replay-browser/src/session-replay.ts
lewgordon-amplitude: Can we wrap this for targeting changes specifically or move to another PR? I'm worried about the potential performance impact here.

kwalker3690: This is actually how it already works! So this is preserving current functionality here: https://github.com/amplitude/Amplitude-TypeScript/blob/v1.x/packages/session-replay-browser/src/session-replay.ts#L188

And this has been how it's worked for a long while - we used to explicitly stop recording on blur, and restart on focus, I recently changed it to be stop and restart on focus only.

The only change happening in this PR is that this is more explicit than it used to be, that implicit behavior was happening within `recordEvents`. I need to make this change in this PR to allow `evaluateTargetingAndCapture` to call `recordEvents` _without_ causing a full dom snapshot, and without restarting recording.

Maybe an overexplanatory loom here: https://www.loom.com/share/57e2abca82714b85bf1f4b64b52309b4

lewgordon-amplitude: Gotcha. Thanks for the clarification!

lewgordon-amplitude (approved):
LGTM!

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Yeah I agree this is awkward. I wrote it this way to have a single source of truth for making the decision about capturing, the `getShouldCapture` fn. This is used for both the decision to start the rrweb record method, and to determine whether replay event properties are returned. My worry with passing around these state variables is that we may lose other important context - for example, `getShouldCapture` also takes into account the `optOut` and `captureEnabled` configs, to make a wholistic decision about replay capture. I'd love ideas about how to make the flow of logic here both bulletproof and easier to reason about though!

packages/session-replay-browser/src/session-replay.ts
kwalker3690: Yeah sounds good, agreed the `ifShould` is confusing to read

PR #843, opened 2024-08-07
fix(session replay): set targeting match to false when session id is changed

kwalker3690:
### Summary
We need to reset the targeting match state when the session id turns over. The unfortunate aspect about this change is it highlights the fact that we are keeping the state of targeting match in two places - both in component state and in IDB (within the targeting manager class). This is necessary because we need to be able to access the value of sessionTargetingMatch synchronously, as it's used by synchronous fns (getShouldCapture, which is used by captureEvents and getSessionReplayProperties), and accessing a value from IDB is an asynchronous action. The downside here is that it's possible for the component state to be out of sync with the IDB state if we're not careful to update it.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -140,6 +140,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
   async asyncSetSessionId(sessionId: number, deviceId?: string, options?: { userProperties?: { [key: string]: any } }) {
     this.stopRecordingEvents();
+    this.sessionTargetingMatch = false;
 
     const previousSessionId = this.identifiers && this.identifiers.sessionId;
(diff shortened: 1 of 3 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -460,7 +464,9 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
   stopRecordingEvents = () => {
     try {
-      this.loggerProvider.log('Session Replay capture stopping.');
+      if (this.recordCancelCallback) {
```
kwalker3690, line 467: I may make this change on v1.x separately as well, this log is confusing/misleading when this method is called but we're not actually recording

lewgordon-amplitude (approved):
Makes sense, LGTM!

PR #848, opened 2024-08-12
feat: support remote config for each autocapture field

Mercy811:
### Summary

This PR contains two changes

### Support enable remote config for each autocapture field by merging remote and local config. 
[AMP-106699]
Merging logic:
   - `options.autocapture` should override `options.defaultTracking` if it’s set. This happens at `useBrowserConfig()` when creating browser config based on options customers passed in 
   - In `joined-config.ts`, `remoteConfig.autocapture` should merge with `localConfig.autocapture` which means that if a field exists in `remoteConfig.autocapture`, it should add to mergedConfig, however, if a filed doesn’t exist, we should check the value in `localConfig.autocapture`

Examples
```
client.init(apiKey, userId, {
  autocapture: false
}).promise;

// remote config is enabled only for clicks
{
  autocapture: {
    elementInteractions: true
  }
}

// result
{
  fileDownloads: false,
  formInteractions: false,
  pageViews: false,
  attribution: false,
  sessions: false,
  elementInteractions: true
}
```
```
client.init(apiKey);

// remote config is enabled only for clicks
{
  autocapture: {
    elementInteractions: true
  }
}

// result
{
  elementInteractions: true
}

all of the other fields in autocapture will be default value
```

### Separate fetch time into 3 at a finer granularity
[AMP-106700]
   - fetch from indexedDB
   - fetch from API and succeed
   - fetch from API but fail

Below is an example of tracking the time of successful fetching from API 
Screenshot 2024-08-12 at 12 48 16
Screenshot 2024-08-12 at 12 48 29

### Checklist

[AMP-106699]: https://amplitude.atlassian.net/browse/AMP-106699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AMP-106700]: https://amplitude.atlassian.net/browse/AMP-106700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser/test/config/joined-config.test.ts
+++ b/packages/analytics-browser/test/config/joined-config.test.ts
@@ -80,16 +80,26 @@ describe('joined-config', () => {
           mockRemoteConfigFetch,
         );
 
+        const expectedAutocapture = {
+          sessions: true,
+          fileDownloads: true,
+          formInteractions: true,
+          pageViews: true,
+          attribution: true,
+          elementInteractions: false,
+        };
+
         await generator.initialize();
         expect(generator.config.autocapture).toBe(true);
         const joinedConfig = await generator.generateJoinedConfig();
-        expect(joinedConfig.autocapture).toBe(remoteConfig.autocapture);
-        expect(joinedConfig.defaultTracking).toBe(remoteConfig.autocapture);
+        expect(joinedConfig.autocapture).toStrictEqual(expectedAutocapture);
+        expect(joinedConfig.defaultTracking).toStrictEqual(expectedAutocapture);
       });
 
-      test('should not disable defaultTracking if remote config sets it to false', async () => {
+      test('should disable defaultTracking if remote config sets it to false', async () => {
         localConfig = createConfigurationMock(
           createConfigurationMock({
+            autocapture: true,
             defaultTracking: true,
           }),
         );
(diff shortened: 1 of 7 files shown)


packages/analytics-core/src/config.ts
```diff
@@ -123,7 +123,7 @@ export class RequestMetadata implements IRequestMetadata {
   constructor() {
     this.sdk = {
       metrics: {
-        histogram: new HistogramOptions(),
+        histogram: {},
```
yuhao900914, line 126: Is this intentional?

packages/analytics-browser/test/config/joined-config.test.ts
```diff
@@ -105,17 +117,98 @@ describe('joined-config', () => {
...
+          fileDownloads: false,
+          formInteractions: false,
+          pageViews: false,
+          attribution: false,
+          sessions: false,
+          elementInteractions: true,
+        };
+
         await generator.initialize();
         expect(generator.config.defaultTracking).toBe(true);
```
yuhao900914, line 137: why the `generator.config.defaultTracking` is true?

packages/analytics-browser/test/config/joined-config.test.ts
```diff
@@ -105,17 +117,98 @@ describe('joined-config', () => {
...
+      test.each([
+        {
+          sessions: false,
+          fileDownloads: false,
+          formInteractions: false,
+          attribution: false,
+          pageViews: false,
+        },
+        false,
+      ])('should only enable elementInteractions if remote config only sets it to true', async (localAutocapture) => {
```
yuhao900914, line 153: Can we change the description?  `elementInteractions: false,` in the remote config. You are actually test, the 
elementInteractions is false when  remote config only sets it to false.

packages/analytics-browser/test/config/joined-config.test.ts
```diff
@@ -105,17 +117,98 @@ describe('joined-config', () => {
...
+        };
+
+        await generator.initialize();
+        expect(generator.config.autocapture).toBe(localAutocapture);
+        const joinedConfig = await generator.generateJoinedConfig();
+        expect(joinedConfig.autocapture).toStrictEqual(expectedJoinedConfig);
+        expect(joinedConfig.defaultTracking).toStrictEqual(expectedJoinedConfig);
+      });
+
+      test('should use remote autocapture if local autocapture is undefined', async () => {
```
yuhao900914, line 190: Do we have a test for remoteConfig.autocapture is boolean and this.config.autocapture === 'boolean'?

packages/analytics-core/src/config.ts
Mercy811: Yes. `new HistogramOptions(),` will set every filed to undefined while `{}` won't.

packages/analytics-browser/test/config/joined-config.test.ts
Mercy811: https://github.com/amplitude/Amplitude-TypeScript/blob/e9e623419f257678acac7f70402c4683d1c8107a/packages/analytics-browser/src/config/joined-config.ts#L6-L7

packages/analytics-browser/test/config/joined-config.test.ts
Mercy811: I don't see much difference here. What will be the description you are suggestion here?

packages/analytics-browser/test/config/joined-config.test.ts
Mercy811: > this.config.autocapture === 'boolean'
Do you mean `expect(autocapture in the merged config is boolean)`?

Mercy811: I don't have test for it because `remoteConfig.autocapture` can only set through the frontend UI and cannot be boolean. But it definitely makes sense to add a test for this scenario as boolean is a valid type of remote config autocapture. I added at https://github.com/amplitude/Amplitude-TypeScript/pull/848/commits/9c8aacf88f8c2819fce2d6d3ee3d875bde66ffdc

yuhao900914: If it's not possible to be boolean. Why we consider the boolean type [at here](https://github.com/amplitude/Amplitude-TypeScript/blob/9c8aacf88f8c2819fce2d6d3ee3d875bde66ffdc/packages/analytics-browser/src/config/joined-config.ts#L37).

packages/analytics-browser/test/config/joined-config.test.ts
yuhao900914: `should only disable elementInteractions if remote config only sets it to false'`

packages/analytics-browser/test/config/joined-config.test.ts
yuhao900914: Got it `generator` is before merged.

packages/analytics-browser/test/config/joined-config.test.ts
Mercy811: Nice catch! I change the remote config value to match the description at https://github.com/amplitude/Amplitude-TypeScript/pull/848/commits/22de814075cde2d7ed9cadee444e04aa9cab5f4d

packages/analytics-browser/test/config/joined-config.test.ts
Mercy811: I defined the remote autocapture type to be `AutocaptureOptions | boolean` at  https://github.com/amplitude/Amplitude-TypeScript/blob/c415f792a98253ac60885eb1dc7e53b78ca47dcb/packages/analytics-types/src/config/browser.ts#L225-L229 because back to that time we didn't have the UI design yet and I thought we were going to have a toggle to turn autocapture on/off but it turned out we didn't go that path. I'd still prefer to keep it there as it's feature that can be added later in the UI.

yuhao900914 (approved):

PR #852, opened 2024-08-14
fix(uuid): set UUID's using crypto libraries instead of Math.random

visiodeibc:
### Summary

Hi, this is Jake from the Technical Support team.

There was an issue of user properties being mapped incorrectly due to colliding UUIDs set by the Amplitude TS SDK[due to math.random function]. 

There was a fix in JS SDK but not carried over to TS SDK
- https://github.com/amplitude/Amplitude-JavaScript/blob/v8.x/src/uuid.js

The issue has been stale, so I am making PR for the issue!

### Checklist

* [v] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?

### DIFF
--- a/packages/analytics-core/src/utils/uuid.ts
+++ b/packages/analytics-core/src/utils/uuid.ts
@@ -1,32 +1,19 @@
 /**
- * Source: [jed's gist]{@link https://gist.github.com/982883}.
+ * Source: [jed's gist's comment]{@link https://gist.github.com/jed/982883?permalink_comment_id=3223002#gistcomment-3223002}.
  * Returns a random v4 UUID of the form xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx,
  * where each x is replaced with a random hexadecimal digit from 0 to f, and
  * y is replaced with a random hexadecimal digit from 8 to b.
  * Used to generate UUIDs for deviceIds.
  * @private
  */
-export const UUID = function (a?: any): string {
-  return a // if the placeholder was passed, return
-    ? // a random number from 0 to 15
-      (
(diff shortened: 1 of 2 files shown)


Mercy811:
Hi @visiodeibc, TS’s `analytics-core` package is shared by both browser (`analytics-browser` package) and node (`analytics-node` package) environments in which crypto should be imported differently. It's predefined in browser environment but it requires to be imported in node environment.

daniel-graham-amplitude:
I've committed a change to this.

NodeJS and Web both export "crypto" to their global scope (global.crypto and window.crypto, respectively) so it doesn't need to imported in NodeJS, it can just be executed from the global scope.

Screenshot 2025-04-09 at 5 12 28 PM

I added a change that if "global.crypto.getRandomValues" isn't there, then fall back to the legacy UUID.

daniel-graham-amplitude:
@Mercy811 turns out you don't need to import "crypto" in a NodeJS environment. Crypto is exposed on the global scope (`global.crypto` or `globalThis.crypto`). Similar to how `setTimeout`, `setInterval`, etc... don't need to be imported as an NPM module.

Screenshot 2025-04-09 at 5 12 28 PM

Mercy811 (approved):
Thanks @daniel-graham-amplitude for fixing it.

visiodeibc:
@Mercy811 @daniel-graham-amplitude 
thank you so much for the fix!
is this available in the latest SDK version?!

daniel-graham-amplitude:
@visiodeibc it is yeah, it's been out for a few weeks. Thanks for the contribution!

PR #855, opened 2024-08-22
feat(analytics-remote-config): remove idb

lewgordon-amplitude:
### Summary

This is removing the use of IDB from remote config in favor of relying on browser caching to help support remote config fetching. There were definitely options to use IDB in other ways, but currently it's a bit broken when trying to support multiple namespaces. Rather than trying fix it, we'll see if just relying on the cache-control header and the browser cache to solve the problem.

### Checklist


kwalker3690 (approved):
Are there changes that need to be made on the service side to support the reading of  session id?

lewgordon-amplitude:
Nope! It's already in place.

PR #858, opened 2024-08-23
fix(session-replay-browser): always get device id from the current session identifiers

junjie-amplitude:
### Summary

The fix for bug: https://amplitude.atlassian.net/browse/AMP-104886

We used to get device ID from analyticsConnector which turned out can be changed in the middle of an ongoing replay capturing and broke it. This PR changed to read from the current session identifiers that are stored in the replay object.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -416,13 +416,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
   };
 
   getDeviceId() {
-    let identityStoreDeviceId: string | undefined;
-    if (this.config?.instanceName) {
-      const identityStore = getAnalyticsConnector(this.config.instanceName).identityStore;
-      identityStoreDeviceId = identityStore.getIdentity().deviceId;
-    }
-
-    return identityStoreDeviceId || this.identifiers?.deviceId;
+    return this.identifiers?.deviceId;
   }
 
   getSessionId() {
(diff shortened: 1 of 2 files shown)


lewgordon-amplitude (approved):

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -416,13 +416,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
 
   getDeviceId() {
-    let identityStoreDeviceId: string | undefined;
-    if (this.config?.instanceName) {
-      const identityStore = getAnalyticsConnector(this.config.instanceName).identityStore;
-      identityStoreDeviceId = identityStore.getIdentity().deviceId;
-    }
-
-    return identityStoreDeviceId || this.identifiers?.deviceId;
+    return this.identifiers?.deviceId;
```
jackson-amplitude, line 419: does this handle changes to device ids? or i guess for plugin we don't really expect device id to change.

junjie-amplitude: Kelly and I agreed that we shouldn't expect the device id to change in the middle of the ongoing replay capture. Callers can still change it through `sessionSessionId(sessionId, deviceId)`. But that will only happen when the previous session ends.

jxiwang (approved):

jackson-amplitude (approved):

PR #863, opened 2024-09-04
fix(autocapture): prevent potential memory exhaustion when generating combinations

Dogfalo:
### Summary

This change mirrors this PR in the upstream library: https://github.com/antonmedv/finder/pull/84
This prevents a scenario where a very large number of combinations are generated and causes memory exhaustion.

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/libs/finder.ts
+++ b/packages/plugin-autocapture-browser/src/libs/finder.ts
@@ -272,7 +272,7 @@ function notEmpty<T>(value: T | null | undefined): value is T {
   return value !== null && value !== undefined;
 }
 
-export function* combinations(stack: Knot[][], path: Knot[] = []): Generator<Knot[]> {
+function* combinations(stack: Knot[][], path: Knot[] = []): Generator<Knot[]> {
   if (stack.length > 0) {
     for (const node of stack[0]) {
       yield* combinations(stack.slice(1, stack.length), path.concat(node));
--- a/packages/plugin-autocapture-browser/src/libs/finder.ts
+++ b/packages/plugin-autocapture-browser/src/libs/finder.ts
@@ -132,10 +132,13 @@ function bottomUpSearch(
 }
 
 function findUniquePath(stack: Knot[][], fallback?: () => Path | null): Path | null {
-  const paths = sort(combinations(stack));
-  if (paths.length > config.threshold) {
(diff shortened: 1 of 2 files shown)


packages/plugin-autocapture-browser/src/libs/finder.ts
```diff
@@ -132,10 +132,13 @@ function bottomUpSearch(
 }
 
 function findUniquePath(stack: Knot[][], fallback?: () => Path | null): Path | null {
-  const paths = sort(combinations(stack));
-  if (paths.length > config.threshold) {
+  // Check first the total number of combinations first since generating the combinations can cause memory exhaustion
+  const numCombinations = stack.reduce((acc, i) => acc * i.length, 1);
```
cyhk, line 136: would you mind clarifying why we're multiplying by length? I think I'm not entirely clear on what a `Knot[]` (aka `Path`?) represents.

Dogfalo: stack could look something like this: 
```javascript
 const stack = [
   [
     { name: 'a', penalty: 1, level: 1 },
     { name: 'b', penalty: 1, level: 1 },
   ],
   [
     { name: 'c', penalty: 1, level: 2 },
     { name: 'd', penalty: 1, level: 2 },
   ],
 ];
```

and then the output of combinations would give:
```javascript
 [
   [
     { level: 1, name: 'a', penalty: 1 },
     { level: 2, name: 'c', penalty: 1 },
   ],
   [
     { level: 1, name: 'a', penalty: 1 },
     { level: 2, name: 'd', penalty: 1 },
   ],
   [
     { level: 1, name: 'b', penalty: 1 },
     { level: 2, name: 'c', penalty: 1 },
   ],
   [
     { level: 1, name: 'b', penalty: 1 },
     { level: 2, name: 'd', penalty: 1 },
   ],
 ]
```

cyhk: I took a look at what `combinations()` does - so each of these combinations is considered a valid path. I think I get it now, thanks for clarifying!

cyhk (approved):

packages/plugin-autocapture-browser/src/libs/finder.ts
Dogfalo: Yep, its a potential css selector that identifies the element.

PR #869, opened 2024-09-11
fix: fix the pathOnly option

yuhao900914:
### Summary

[Related ticket
](https://amplitude.atlassian.net/browse/AMP-108564)

### Checklist

### DIFF
--- a/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
+++ b/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
@@ -161,13 +161,11 @@ export const shouldTrackHistoryPageView = (
   switch (trackingOption) {
     case 'pathOnly': {
       if (oldURLStr == '') return true;
-      else {
-        const newURL = new URL(newURLStr);
-        const oldURL = new URL(oldURLStr);
-        const newHREF = newURL.protocol + '//' + newURL.host + newURL.pathname;
-        const oldHREF = oldURL.protocol + '//' + oldURL.host + oldURL.pathname;
-        return newHREF !== oldHREF;
-      }
+      const newURL = new URL(newURLStr);
+      const oldURL = new URL(oldURLStr);
+      const newBaseStr = newURL.protocol + '//' + newURL.host + newURL.pathname;
+      const oldBaseStr = oldURL.protocol + '//' + oldURL.host + oldURL.pathname;
+      return newBaseStr !== oldBaseStr;
     }
     default:
       return newURLStr !== oldURLStr;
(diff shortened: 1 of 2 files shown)


packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -155,13 +155,19 @@ const isCampaignEvent = (event: Event) => {
...
 ): boolean => {
   switch (trackingOption) {
-    case 'pathOnly':
-      return newURL.split('?')[0] !== oldURL.split('?')[0];
+    case 'pathOnly': {
+      if (oldURLStr == '') return true;
+      const newURL = new URL(newURLStr);
+      const oldURL = new URL(oldURLStr);
+      const newBaseStr = newURL.protocol + '//' + newURL.host + newURL.pathname;
+      const oldBaseStr = oldURL.protocol + '//' + oldURL.host + oldURL.pathname;
```
Mercy811, line 167: origin contains protocol + host https://web.dev/articles/url-parts#url_analyzer
```suggestion
      const newBaseStr = newURL.origin + newURL.pathname;
      const oldBaseStr = oldURL.origin + oldURL.pathname;
```

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -155,13 +155,19 @@ const isCampaignEvent = (event: Event) => {
...
-  newURL: string,
-  oldURL: string,
+  newURLStr: string,
+  oldURLStr: string,
 ): boolean => {
   switch (trackingOption) {
-    case 'pathOnly':
-      return newURL.split('?')[0] !== oldURL.split('?')[0];
+    case 'pathOnly': {
+      if (oldURLStr == '') return true;
```
Mercy811, line 163: `previousURL` is null at first https://github.com/amplitude/Amplitude-TypeScript/blob/812713c51b0b69909897d8ee0b1d72032266e2a1/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts#L46C69-L46C86. Could you also update the test?
```suggestion
      if (oldURLStr == null) return true;
```

Mercy811 (approved):
Thanks Alyssa for fixing it. `pathOnly` should track page views only when origin + pathname changes. [URL()](https://developer.mozilla.org/en-US/docs/Web/API/URL#constructor) can throw an error if the string is not a valid URL string, but it seems fine here without a catch as the inputs are always `location.href`

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
yuhao900914: We are passing un empty string while previousURL is null.

` shouldTrackHistoryPageView(trackHistoryChanges, newURL, previousURL || '') && shouldTrackOnPageLoad();`

https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts#L51

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
yuhao900914: Updated.

PR #870, opened 2024-09-12
feat: decode URI for page view properties

yuhao900914:
### Summary

[Related tickets 
](https://amplitude.atlassian.net/browse/AMP-107632)

### Checklist

### DIFF
--- a/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
+++ b/packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
@@ -28,26 +28,26 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
       decodedLocationStr = decodeURI(locationStr);
     } catch (e) {
       /* istanbul ignore next */
-      loggerProvider?.log('Malformed URI sequence: ', e);
+      loggerProvider?.error('Malformed URI sequence: ', e);
     }
 
     return decodedLocationStr;
   };
 
   const createPageViewEvent = async (): Promise<Event> => {
-    const locationHref = getDecodeURI(location.href);
+    /* istanbul ignore next */
+    const locationHREF = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
     return {
       event_type: eventType,
       event_properties: {
         ...(await getCampaignParams()),
         '[Amplitude] Page Domain':
           /* istanbul ignore next */ (typeof location !== 'undefined' && location.hostname) || '',
-        '[Amplitude] Page Location': /* istanbul ignore next */ (typeof location !== 'undefined' && locationHref) || '',
+        '[Amplitude] Page Location': locationHREF,
         '[Amplitude] Page Path':
           /* istanbul ignore next */ (typeof location !== 'undefined' && getDecodeURI(location.pathname)) || '',
         '[Amplitude] Page Title': /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
-        '[Amplitude] Page URL':
-          /* istanbul ignore next */ (typeof location !== 'undefined' && locationHref.split('?')[0]) || '',
+        '[Amplitude] Page URL': locationHREF.split('?')[0],
       },
     };
   };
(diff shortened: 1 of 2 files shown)


packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -22,20 +22,32 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
   let localConfig: BrowserConfig;
   const { trackOn, trackHistoryChanges, eventType = defaultPageViewEvent } = options;
 
+  const getDecodeURI = (locationStr: string): string => {
+    let decodedLocationStr = locationStr;
+    try {
+      decodedLocationStr = decodeURI(locationStr);
```
yuhao900914, line 28: Use decodeURI instead of decodeURIComponent since we are not decoding special characters like ?, #, etc. In the browser's navigation bar, these special characters are also not decoded.

For example, if the link provided is www.example.com/home-shell%3Fx%3Dtest, %3Fx%3Dtest represents ?x=test. The window.location.href will still be www.example.com/home-shell%3Fx%3Dtest, which matches how it appears in the navigation bar.

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -22,20 +22,32 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
...
+    } catch (e) {
+      /* istanbul ignore next */
+      loggerProvider?.log('Malformed URI sequence: ', e);
+    }
+
+    return decodedLocationStr;
+  };
+
   const createPageViewEvent = async (): Promise<Event> => {
+    const locationHref = getDecodeURI(location.href);
```
Mercy811, line 38: location can be undefined given the current code,  even though I don't get it why as location always exist in a browser environment. 
```suggestion
    const locationHref = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
```

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -22,20 +22,32 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
...
-        '[Amplitude] Page Location':
-          /* istanbul ignore next */ (typeof location !== 'undefined' && location.href) || '',
+        '[Amplitude] Page Location': /* istanbul ignore next */ (typeof location !== 'undefined' && locationHref) || '',
         '[Amplitude] Page Path':
-          /* istanbul ignore next */ (typeof location !== 'undefined' && location.pathname) || '',
+          /* istanbul ignore next */ (typeof location !== 'undefined' && getDecodeURI(location.pathname)) || '',
         '[Amplitude] Page Title': /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
         '[Amplitude] Page URL':
-          /* istanbul ignore next */ (typeof location !== 'undefined' && location.href.split('?')[0]) || '',
+          /* istanbul ignore next */ (typeof location !== 'undefined' && locationHref.split('?')[0]) || '',
```
Mercy811, line 50: don't need to check location here too

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
```diff
@@ -22,20 +22,32 @@ export const pageViewTrackingPlugin: CreatePageViewTrackingPlugin = (options: Op
   let localConfig: BrowserConfig;
   const { trackOn, trackHistoryChanges, eventType = defaultPageViewEvent } = options;
 
+  const getDecodeURI = (locationStr: string): string => {
+    let decodedLocationStr = locationStr;
+    try {
+      decodedLocationStr = decodeURI(locationStr);
+    } catch (e) {
+      /* istanbul ignore next */
+      loggerProvider?.log('Malformed URI sequence: ', e);
```
Mercy811, line 31: Use warn() or error() as it's an error?

Mercy811 (changes requested):

packages/plugin-page-view-tracking-browser/src/page-view-tracking.ts
yuhao900914: update to `error`

Mercy811 (approved):

PR #874, opened 2024-09-16
fix(session-replay-browser): handle exceptions when creating an events manager

junjie-amplitude:
### Summary

A Jenkins error was reported in https://amplitude.atlassian.net/browse/AMP-108769. It was caused by the unhandled idb open error. This PR handles it when creating the events managers. The events manager creation fails when the [idb store creation](https://amplitude.sourcegraph.com/github.com/amplitude/Amplitude-TypeScript@v1.x/-/blob/packages/session-replay-browser/src/events/events-idb-store.ts?L95) fails, and it skips adding the events manager so no replays will be captured then.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -97,24 +97,35 @@ export class SessionReplay implements AmplitudeSessionReplay {
     }
 
     const managers: EventsManagerWithType<EventType, string>[] = [];
-    const rrwebEventManager = await createEventsManager<'replay'>({
-      config: this.config,
-      sessionId: this.identifiers.sessionId,
-      type: 'replay',
-    });
-    managers.push({ name: 'replay', manager: rrwebEventManager });
+
+    try {
+      const rrwebEventManager = await createEventsManager<'replay'>({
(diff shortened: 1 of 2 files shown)


jxiwang (approved):
LGTM

jackson-amplitude (approved):
Do we know when this happens?

junjie-amplitude:
> Do we know when this happens?

@jackson-amplitude For example, when there is no enough disk space to open a new idb store.

jackson-amplitude:
> > Do we know when this happens?
> 
> @jackson-amplitude For example, when there is no enough disk space to open a new idb store.

our storage footprint should be small. maybe worth a quick check to see if idb has any diagnostics that we can log (don't spend too much time though).

junjie-amplitude:
> https://amplitude.atlassian.net/browse/AMP-108769

@jackson-amplitude Out of disk space is one of the possible reasons. Other reasons could be opening idb in incompatible browsers or simply an idb bug. People usually report this issue with no other diagnostic information. The solutions are often to restart the browser or so. So I think catching and logging the error in the console is the best we can do for it.

PR #721, opened 2024-04-18
fix(analytics-browser): prevent error being thrown when server-side rendering

johnf:
### Summary

Fixes #710 - I'm not sure why, but this library gets pulled into the SSR part of my next.js project.
Likely a treeshaking issue. It then throws and error because navigtator is missing.

### Checklist

* [X] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:  No

PR #876, opened 2024-09-17
fix: disable network checker if navigator is unavailable

izaaz:
### Summary

Check if navigator is available before using it for network connectivity checks

### Checklist

### DIFF
--- a/packages/analytics-browser/src/plugins/network-connectivity-checker.ts
+++ b/packages/analytics-browser/src/plugins/network-connectivity-checker.ts
@@ -33,6 +33,14 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
   };
 
   const setup = async (config: BrowserConfig, amplitude: BrowserClient) => {
+    if (typeof navigator === 'undefined') {
+      config.loggerProvider.debug(
+        'Network connectivity checker plugin is disabled because navigator is not available.',
+      );
+      config.offline = true;
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/analytics-browser/src/plugins/network-connectivity-checker.ts
```diff
@@ -33,6 +33,14 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
   };
 
   const setup = async (config: BrowserConfig, amplitude: BrowserClient) => {
+    if (typeof navigator === 'undefined') {
+      config.loggerProvider.debug(
+        'Network connectivity checker plugin is disabled because navigator is not available.',
+      );
+      config.offline = false;
```
johnf, line 40: As per your comment in my PR I think you wanted this to be `true`?

johnf (approved):
Looks good

packages/analytics-browser/src/plugins/network-connectivity-checker.ts
izaaz: The comment was to make it false by default. We want to be online if the network connectivity checker plugin was not available.

https://github.com/amplitude/Amplitude-TypeScript/pull/721/files#r1603984757

Mercy811 (approved):
Thanks @izaaz for fixing when `navigator` is not available when analytics-browser is pulled in the sever side rendering part of a next.js all.

PR #877, opened 2024-09-17
fix(analytics-types): update Element in element interactions

Mercy811:
### Summary

This PR fixes the issue in https://github.com/amplitude/Amplitude-TypeScript/pull/780 that `Element` type is resolved to the custom defined type because typescript looks up local scope first before global scope. 

Create a custom type `DomElement` if `Element` is available (`dom` library is added) otherwise set to `never`.
- `analytics-browser`, `plugin-autocapture-browser` will use `Element`
- `analytics-node` doesn't have access to `Element` and will use `never` instead

### Checklist


Dogfalo (approved):

sabin10:
FYI this update is breaking: https://github.com/amplitude/Amplitude-TypeScript/issues/883

PR #885, opened 2024-09-23
feat(session replay): delay compresion 

jxiwang:
### Summary

Delays the compression of events emitted by rrrweb record using requestIdleCallback.

### Checklist


lewgordon-amplitude (changes requested):
This is really cool. I had no idea about `requestIdleCallback` before the PR, it seems like a good use case for this. I'm wondering though if we want to change this a little bit. The things that come to mind for me:

1. The hard coded timeout seems like it'll fire a lot of events just slightly delayed. Therefore not immediately impacting performance, but just having it delayed. I think it being hard coded is to going to be slightly problematic since that number will vary from app to app.
2. It seems like we're creating a new callback for each event. I'm not entirely certain it's a problem, but for a high number of events this could be a lot of callbacks.

I'm thinking to help mitigate some of these things maybe we could instead of a queue that we work through during idle periods and get through as many as possible. We could probably use the [IdleDeadline](https://developer.mozilla.org/en-US/docs/Web/API/IdleDeadline) object to know when to quit. The idea being that we register a single callback that processes through the queue as much as possible during idle periods. Getting this exactly right and not going over time may be tough, but I think this would be more performant.

I think in either case before we merge this we should add an option so that we can test this ourselves for now and make sure we add some performance charts with a sample app or preferably our site so we're confident in the change.

I'm definitely open to feedback/other thoughts on this!

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -309,6 +310,36 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    const deviceId = this.getDeviceId();
+    this.eventsManager &&
+      deviceId &&
+      this.eventsManager.addEvent({ event: { type: 'replay', data: compressedEvent }, sessionId, deviceId });
+  };
+
+  deferEventCompression = (event: eventWithTime, sessionId: number) => {
+    const globalScope = getGlobalScope();
+    // In case the browser does not support requestIdleCallback, we will compress the event immediately
+    if (globalScope && 'requestIdleCallback' in globalScope) {
```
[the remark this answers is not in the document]
jxiwang: Ah true, the browser definitely wouldn't change. Let me see if I can move the check out of this function, maybe just check prior to recording.

jxiwang:
@lewgordon-amplitude 

Thanks for the feedback! 

1. For the timeout, I considered not including one but am also concerned with the case that the compression doesn't get fired. Even if the timeout is reached, the compression _should_ get pushed to the back of the event queue in cases like when the page takes > 2s to load. Although I'm not entirely sure that doesn't have performance implications still. I could remove the timeout, and we could monitor the effects. 

2. Good call on the callbacks being generated per event. I'll add an event queue based on the `timeRemaining`, and I actually did not know `idleDeadline` could provide that information. Pretty cool that it does do that! 

Agreed on adding an option so we can have control over testing this, and we can let customers know this is something they could turn on.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -312,12 +315,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
   recordEvents() {
     const shouldRecord = this.getShouldRecord();
     const sessionId = this.identifiers?.sessionId;
+    // const canDelayCompression = globalScope && 'requestIdleCallback' in globalScope;
```
lewgordon-amplitude, line 318:
```suggestion
```

No longer needed?

packages/session-replay-browser/src/events/event-compressor.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // Schedule processing during idle time
+  public scheduleIdleProcessing(): void {
+    if (!this.isProcessing) {
+      this.isProcessing = true;
+      requestIdleCallback(
+        (idleDeadline) => {
+          this.processQueue(idleDeadline);
+        },
+        { timeout: 2000 },
```
lewgordon-amplitude, line 40: I think this should be a parameter in the config, too.

packages/session-replay-browser/src/events/event-compressor.ts
```diff
@@ -0,0 +1,95 @@
...
+      );
+    } else {
+      this.isProcessing = false;
+    }
+  }
+
+  compressEvents = (event: eventWithTime) => {
```
lewgordon-amplitude, line 79:
```suggestion
  compressEvent = (event: eventWithTime) => {
```

Nit: :)

packages/session-replay-browser/src/events/event-compressor.ts
```diff
@@ -0,0 +1,95 @@
...
+  sessionId: number;
+}
+
+export class EventCompressor {
+  taskQueue: TaskQueue[] = [];
+  isProcessing = false;
+  eventsManager?: SessionReplayEventsManager<'replay' | 'interaction', string>;
+  config: SessionReplayJoinedConfig;
+  deviceId: string | undefined;
+  canUseIdleCallback: boolean | undefined;
```
lewgordon-amplitude, line 18: I think the code could be cleaner if these instance variables were not possibly undefined. It seems like all are set in the constructor, so I don't think we need to recheck them in the class methods.

lewgordon-amplitude (approved):
LGTM! I think there are minor changes we can make, but with the config in place and the current approach I don't see anything majorly blocking here.

lewgordon-amplitude (approved):

PR #886, opened 2024-09-23
fix(session replay): add version option into local config generation

jxiwang:
### Summary

Fixes issue where the library wasn't propagating to the track call header. 

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/config/local-config.ts
+++ b/packages/session-replay-browser/src/config/local-config.ts
@@ -3,7 +3,12 @@ import { Config, Logger } from '@amplitude/analytics-core';
 import { LogLevel } from '@amplitude/analytics-types';
 import { DEFAULT_SAMPLE_RATE, DEFAULT_SERVER_ZONE } from '../constants';
 import { SessionReplayOptions } from '../typings/session-replay';
-import { SessionReplayLocalConfig as ISessionReplayLocalConfig, InteractionConfig, PrivacyConfig } from './types';
+import {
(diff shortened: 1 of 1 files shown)


lewgordon-amplitude (approved):
LGTM! I'm thinking we should fix the typing in the future so that the version is enforced somewhat.

packages/session-replay-browser/src/config/local-config.ts
```diff
@@ -38,6 +44,7 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
     this.serverZone = options.serverZone || DEFAULT_SERVER_ZONE;
     this.configEndpointUrl = options.configEndpointUrl;
     this.shouldInlineStylesheet = options.shouldInlineStylesheet;
+    this.version = options.version;
```
izaaz, line 47: Curious - what is session replay version? 

cc: @Mercy811 in case we need to do something similar with remote config for analytics too.

Mercy811: It looks like it's type (enum) + version (string) https://github.com/amplitude/Amplitude-TypeScript/blob/e33f45a4f496cb0c01bec3d91137fc62d969ff67/packages/session-replay-browser/src/config/types.ts#L70-L75

jxiwang: This is for us to send the library type (standalone, plugin, wrapper) and the version of that library.

PR #895, opened 2024-10-07
feat(session-replay-browser): in memory events storage

lewgordon-amplitude:
### Summary

~~**Do not merge, still testing with the tasks app**~~

Currently, we store events using indexedDB (IDB). However, it's not possible in all cases where we want to capture replays that IDB is present. This PR adds an in-memory version of the event store that can either be the default store or will be the fallback if IDB is not present. The storage API is not quite stable yet and the implementation to make both stores unified needs a bit of work. This is small holdover to help in certain situations and to prevent a very large initial PR.

_Sorry for the long PR! This was an attempt at the minimal amount of changes with a few quality of life things thrown in._

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/events/events-idb-store.ts
+++ b/packages/session-replay-browser/src/events/events-idb-store.ts
@@ -4,13 +4,7 @@ import { Logger as ILogger } from '@amplitude/analytics-types';
 import { DBSchema, IDBPDatabase, openDB } from 'idb';
 import { MAX_EVENT_LIST_SIZE_IN_BYTES, MAX_INTERVAL, MIN_INTERVAL } from '../constants';
 import { STORAGE_FAILURE } from '../messages';
-import {
-  SessionReplayEventsIDBStore as AmplitudeSessionReplayEventsIDBStore,
-  EventType,
-  Events,
-  SendingSequencesIDBInput,
-  SendingSequencesIDBReturn,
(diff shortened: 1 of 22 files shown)


packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -96,80 +86,75 @@ export const createStore = async (dbName: string) => {
...
+    return allEvents;
+  }
+
+  getSequencesToSend = async (): Promise<SendingSequencesReturn<number>[] | undefined> => {
     try {
-      const sequencesToSend = (await this.db?.getAll<'sequencesToSend'>(
-        sequencesToSendKey,
-      )) as SendingSequencesIDBReturn[];
+      const sequences: SendingSequencesReturn<number>[] = [];
+      let cursor = await this.db.transaction('sequencesToSend').store.openCursor();
```
jxiwang, line 146: Is this so that we can use a transaction here?

jxiwang (approved):
Overall LGTM

packages/session-replay-browser/src/events/events-idb-store.ts
lewgordon-amplitude: Good question! I should have left a comment, but I wanted to build out a proper list here rather than doing an unsafe type assertion. Using the cursor key I can guarantee that it'll be defined.

PR #903, opened 2024-10-24
feat: adds support to \n separated request body for ga forwarding

kevinpagtakhan:
### Summary

When this feature was originally built, GA sent bulk events in a single request using the request payload. The request payload is a list of valid query parameter formatted string separated by the `\\r\\` character. A customer reported that their GA-forwarded events are not being formatted correctly. After some investigating, we are seeing the list of valid query parameter formatted stings are separated by `\n` character. The changes here include adding support to `\n` separated strings and ignoring events that are formatted outside the supported formats.

### Checklist

### DIFF
--- a/packages/plugin-ga-events-forwarder-browser/src/helpers.ts
+++ b/packages/plugin-ga-events-forwarder-browser/src/helpers.ts
@@ -33,9 +33,11 @@ export const parseGA4Events = (url: URL, data?: BodyInit | null): GA4Event[] => 
     return [sharedProperties];
   }
 
+  const separator = data.toString().includes('\n') ? '\n' : '\\r\\';
+
   return data
     .toString()
-    .split('\\r\\')
+    .split(separator)
     .map<GA4Event>((event) => {
       return {
         ...sharedProperties,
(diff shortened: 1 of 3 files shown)


packages/plugin-ga-events-forwarder-browser/src/helpers.ts
```diff
@@ -53,25 +55,29 @@ export const parseGA4Events = (url: URL, data?: BodyInit | null): GA4Event[] =>
...
-    ),
-    // NOTE: Unable to pass an event to track() with custom library value because an internal plugin will overwrite `event.library` value.
-    // Instead, since an enrichment plugin's execute function is performed at a later time. pass an event to track() with library info in `event.extra`,
-    // then enrich `event.library` in this plugin's execute function.
-    extra: {
-      library: AMPLITUDE_EVENT_LIBRARY,
-    },
-  }));
+  ga4Events
+    .filter((ga4Event) => Boolean(ga4Event[GA_PAYLOAD_EVENT_NAME_KEY]))
```
kevinpagtakhan, line 59: added a filter to guarantee that event_type will be defined

packages/plugin-ga-events-forwarder-browser/src/helpers.ts
```diff
@@ -33,9 +33,11 @@ export const parseGA4Events = (url: URL, data?: BodyInit | null): GA4Event[] =>
     return [sharedProperties];
   }
 
+  const separator = data.toString().includes('\n') ? '\n' : '\\r\\';
```
kevinpagtakhan, line 36: uses `\n` if it exists

packages/plugin-ga-events-forwarder-browser/src/helpers.ts
```diff
@@ -53,25 +55,29 @@ export const parseGA4Events = (url: URL, data?: BodyInit | null): GA4Event[] =>
...
-      library: AMPLITUDE_EVENT_LIBRARY,
-    },
-  }));
+  ga4Events
+    .filter((ga4Event) => Boolean(ga4Event[GA_PAYLOAD_EVENT_NAME_KEY]))
+    .map<BaseEvent>((ga4Event) => ({
+      event_type: String(ga4Event[GA_PAYLOAD_EVENT_NAME_KEY]),
+      ...(ga4Event[GA_PAYLOAD_USER_ID_KEY] !== null && ga4Event[GA_PAYLOAD_USER_ID_KEY] !== undefined
+        ? { user_id: String(ga4Event[GA_PAYLOAD_USER_ID_KEY]) }
+        : undefined),
```
Mercy811, line 64: It's not a valid Amplitude then event if user id and device id are missing. Do we insert either of them somewhere else? If not, we should just drop the event.

Mercy811 (commented):
Thanks Kevin for adding this. Overall looks good to me. See the comment about the change to user id.

packages/plugin-ga-events-forwarder-browser/src/helpers.ts
kevinpagtakhan: the browser SDK will generate the device ID. the forwarder does not use GA's device ID, only the user ID.

Mercy811 (approved):

PR #906, opened 2024-10-25
chore(session replay): rrweb changes require a few modifications to the SDK

jxiwang:
### Summary

rrweb upgrades resulted in some modifications needed to how certain functions are imported. 

### Checklist


packages/session-replay-browser/src/hooks/click.ts
```diff
@@ -96,8 +96,8 @@ export const clickHook: (options: Options) => mouseInteractionCallBack =
     const { left: scrollX, top: scrollY } = utils.getWindowScroll(globalScope as unknown as Window);
 
     const event: ClickEvent = {
-      x: x + scrollX,
-      y: y + scrollY,
+      x: ( x ?? 0 ) + scrollX,
+      y: ( y ?? 0 ) + scrollY,
```
jxiwang, line 100: I default to 0, but we can also not send the event if clicks are undefined.

lewgordon-amplitude: Yeah, good call. Let's opt to not sent if the x/y is undefined.

lewgordon-amplitude (approved):
LGTM!

PR #911, opened 2024-10-31
fix: allow passing sample rate and remote config fetching

izaaz:
### Summary

Ability to customize session rate and enable remote config in the react native SDK

### Checklist


packages/plugin-session-replay-react-native/example/src/App.tsx
```diff
@@ -239,11 +238,15 @@ const Stack = createNativeStackNavigator<RootStackParamList>();
 
 function App(): React.JSX.Element {
   useEffect(() => {
+    const config: SessionReplayConfig = {
+      enableRemoteConfig: false,
+      sampleRate: 1
+    };
     (async () => {
-      await init('YOUR-API-KEY', 'example_user_id', {
+      await init('6151e18fcbce1f94010e8791964d2a71', 'example_user_id', {
```
crleona, line 246: probably don't hardcode an api key?

crleona (approved):

packages/plugin-session-replay-react-native/example/src/App.tsx
izaaz: yea just reverted that.

PR #913, opened 2024-11-04
fix: make table non nullable

izaaz:
### Summary

Android SDK version 35 makes the table property non nullable in sqlite. 

### Checklist


Mercy811 (approved):
Thanks Izaaz for fixing it.

PR #932, opened 2024-12-12
feat(session-replay-browser): WebWorker support for compression

lewgordon-amplitude:
### Summary

Adds experimental support for WebWorkers to offload compression. This is an initial first pass to see performance with hopefully later iterations being more complex and offloading not only compression, but other things such as networking calls.

To enable add to your local config:

```javascript
{
  experimental: {
    useWebWorker: true
  }
}
```

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/rollup.config.js
+++ b/packages/plugin-session-replay-browser/rollup.config.js
@@ -1,65 +1,14 @@
 import { iife, umd } from '../../scripts/build/rollup.config';
-
-import resolve from '@rollup/plugin-node-resolve';
-import replace from '@rollup/plugin-replace';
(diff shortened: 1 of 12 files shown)


jxiwang (approved):
LGTM since this is behind a config flag, some questions - 

- When the worker fails for whatever reason, the events would essentially get dropped? 
- Do we need to add a limit to the amount of workers that can get spawned?

lewgordon-amplitude:
> When the worker fails for whatever reason, the events would essentially get dropped?

Yeah I think as we investigate this we should see if there are ways to make it more resilient.

> Do we need to add a limit to the amount of workers that can get spawned?

The goal I was thinking is 1 worker for 1 instance of session replay. So hopefully there are only a handful of these.

PR #935, opened 2024-12-17
fix(analytics-browser): should send batches sequentially

Mercy811:
### Summary

[AMP-118975]

Fix the issue that events are not received by order by the server. The root cause is that `Promise.all()` executes calls concurrently thus resolve order is not guaranteed. Fix it by sending batches sequentially, in other words, only send the next batch when the preceding one resolves. 

See [comment](https://amplitude.atlassian.net/browse/AMP-118975?focusedCommentId=301158) for additional notes.

### Checklist

[AMP-118975]: https://amplitude.atlassian.net/browse/AMP-118975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -151,7 +151,12 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(list, this.config.flushQueueSize);
 
-    await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
+    // Promise.all() doesn't guarantee resolve order.
+    // Sequentially resolve to make sure backend receives events in order
+    await batches.reduce(async (promise, batch) => {
+      await promise;
+      return await this.send(batch, useRetry);
+    }, Promise.resolve());
 
     this.scheduleTryable(later);
   }
--- a/packages/analytics-core/test/plugins/destination.test.ts
+++ b/packages/analytics-core/test/plugins/destination.test.ts
@@ -321,6 +321,57 @@ describe('destination', () => {
       expect(result).toBe(undefined);
       expect(send).toHaveBeenCalledTimes(0);
     });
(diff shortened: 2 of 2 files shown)


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -151,7 +151,12 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(list, this.config.flushQueueSize);
 
-    await Promise.all(batches.map((batch) => this.send(batch, useRetry)));
+    // Promise.all() doesn't guarantee resolve order.
+    // Sequentially resolve to make sure backend receives events in order
+    await batches.reduce(async (promise, batch) => {
```
izaaz, line 156: interesting way to use `batches.reduce` :)

izaaz (approved):

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -151,7 +151,12 @@ export class Destination implements DestinationPlugin {
...
+      await promise;
+      return await this.send(batch, useRetry);
+    }, Promise.resolve());
```
Mercy811, line 159: I just noticed that it's auto changed by Intellij 

I had 
```
await batches.reduce(
      (promise, batch) => {
        return promise.then(() => this.send(batch, useRetry))
      },
      Promise.resolve()
    );
```
After clicking that "convert to async function", it updated to what we have here. 
image

I was expecting to have
```
await batches.reduce(
      async (promise, batch) => {
        await promise;
        return this.send(batch, false);
      },
      Promise.resolve()
    );
```

It's still working although a little bit redundant. Should clean it in [AMP-120671](https://amplitude.atlassian.net/browse/AMP-120671).

PR #936, opened 2024-12-17
fix: amplitude.remove should only remove if plugin was already registered

izaaz:
### Summary

Currently, calling amplitude.remove(badPluginName) will remove the last registered plugin. Fixing that issue

### Checklist

### DIFF
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -41,8 +41,12 @@ export class Timeline {
     this.plugins.push(plugin);
   }
 
-  async deregister(pluginName: string) {
+  async deregister(pluginName: string, config: Config) {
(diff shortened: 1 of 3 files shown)


packages/analytics-core/src/timeline.ts
```diff
@@ -41,8 +41,12 @@ export class Timeline {
     this.plugins.push(plugin);
   }
 
-  async deregister(pluginName: string) {
+  async deregister(pluginName: string, config: Config) {
     const index = this.plugins.findIndex((plugin) => plugin.name === pluginName);
+    if (index == -1) {
```
Mercy811, line 46: Nit: I'd prefer `===` as `==` does auto type casting.

Mercy811 (approved):
Thanks Izaaz for fixing it. See nit.

PR #939, opened 2024-12-19
chore: update script library

Mercy811:
### Summary

- Change library to `amplitude-ts-sdk-script` in amplitude-min.js which is renamed as `analytics-browser-x.x.x-min.js.gz`and uploaded to s3 at https://github.com/amplitude/Amplitude-TypeScript/blob/cbb0afa63034959559216d4ea5217b50ce6f4cf1/scripts/publish/upload-to-s3.js#L9
- The main goal is to be able to distinguish installation methods: script vs npm/yarn which remains to be `amplitude-ts`
```
// install by script example
<script src="https://cdn.amplitude.com/libs/analytics-browser-2.11.10-min.js.gz"></script>
```
- Keep this as part of rollup instead of updating amplitude-min.js to avoid amplitude-min.js is used before library updated
- library in lib/esm and lib/cjs remains `amplitude-ts`
image

### Checklist

### DIFF
--- a/scripts/build/rollup.config.js
+++ b/scripts/build/rollup.config.js
@@ -72,11 +72,12 @@ const updateLibPrefix = (isUndo) => {
   }
 
   fs.writeFileSync(path, updatedContent, 'utf-8');
+  // Supported in rollup 4, we're currently rollup 2
   // this.info(`File updated: ${path}`);
 }
 
 
-const updateLibPrefixPlugin = (isUndo) => {
+const updateLibPrefixPlugin = () => {
   return {
     name: 'update-lib-prefix',
     buildStart() {
--- a/scripts/build/rollup.config.js
+++ b/scripts/build/rollup.config.js
@@ -5,6 +5,7 @@ import { terser } from 'rollup-plugin-terser';
 import gzip from 'rollup-plugin-gzip';
 import execute from 'rollup-plugin-execute';
 import { exec } from 'child_process';
+import fs from 'fs';
 
 // The paths are relative to process.cwd(), which are packages/*
 const base = '../..';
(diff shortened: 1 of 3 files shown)


scripts/build/rollup.config.js
```diff
@@ -54,6 +55,39 @@ export const umd = {
   ],
 };
 
+const updateLibPrefix = (isUndo) => {
+  const path = 'src/lib-prefix.ts'
+  if (!fs.existsSync(path)) {
+    // Supported in rollup 4, we're currently rollup 2
+    // this.error(`File not found: ${path}`);
```
liuyang1520, line 62: Minor, do we want to keep this logging comment?

scripts/build/rollup.config.js
```diff
@@ -54,6 +55,39 @@ export const umd = {
...
+  let content = fs.readFileSync(path, 'utf-8');
+  let updatedContent;
+  if (isUndo) {
+    updatedContent = content.replace(/amplitude-ts-sdk-script/g, 'amplitude-ts');
+  } else {
+    updatedContent = content.replace(/amplitude-ts/g, 'amplitude-ts-sdk-script');
+  }
+
+  fs.writeFileSync(path, updatedContent, 'utf-8');
+  // this.info(`File updated: ${path}`);
```
liuyang1520, line 75: Ditto

scripts/build/rollup.config.js
```diff
@@ -63,6 +97,7 @@ export const iife = {
     sourcemap: true,
   },
   plugins: [
+    updateLibPrefixPlugin(),
```
liuyang1520, line 100: I think another possible solution is using https://www.npmjs.com/package/rollup-plugin-inject-process-env, so we can dynamically inject environment variable here, then read in the build process.
Manipulating the library file solution LGTM too.

liuyang1520 (approved):

scripts/build/rollup.config.js
Mercy811: I'd like to keep it here so that when we migrate to rollup 4, we can just uncomment this line. wdyt?

scripts/build/rollup.config.js
Mercy811: the same as above https://github.com/amplitude/Amplitude-TypeScript/pull/939#discussion_r1896979850

scripts/build/rollup.config.js
Mercy811: Great solution! I'll make it as a backup if the plugin solution doesn't work well.

scripts/build/rollup.config.js
```diff
@@ -54,6 +55,39 @@ export const umd = {
...
+  } else {
+    updatedContent = content.replace(/amplitude-ts/g, 'amplitude-ts-sdk-script');
+  }
+
+  fs.writeFileSync(path, updatedContent, 'utf-8');
+  // this.info(`File updated: ${path}`);
+}
+
+
+const updateLibPrefixPlugin = (isUndo) => {
```
chungdaniel, line 79: Doesn't look like `isUndo` is used here

scripts/build/rollup.config.js
chungdaniel: Nit: Should we also add the comment `// Supported in rollup 4, we're currently rollup 2` here for clarity

scripts/build/rollup.config.js
```diff
@@ -54,6 +55,39 @@ export const umd = {
   ],
 };
 
+const updateLibPrefix = (isUndo) => {
```
chungdaniel, line 58: Nit: Should function name give a bit more context? Not sure if I'm just unfamiliar with rollup/context of the PR. `isUndo` makes a bit more sense with some context.

chungdaniel (approved):

scripts/build/rollup.config.js
Mercy811: deleted

PR #940, opened 2024-12-26
test: fix flaky test

Mercy811:
### Summary

[AMP-102589]

Fix flaky test by mocking Date.now().

### Checklist

[AMP-102589]: https://amplitude.atlassian.net/browse/AMP-102589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser/test/browser-client.test.ts
+++ b/packages/analytics-browser/test/browser-client.test.ts
@@ -824,8 +824,12 @@ describe('browser-client', () => {
       expect(eventTime1 > 0).toBeTruthy();
 
       // wait for session to almost expire, then extend it
-      jest.advanceTimersByTime(15);
-      client.extendSession();
+      await new Promise<void>((resolve) =>
+        setTimeout(() => {
+          client.extendSession();
+          resolve();
+        }, 15),
+      );
 
       // assert session id is unchanged
       expect(client.config.sessionId).toBe(firstSessionId);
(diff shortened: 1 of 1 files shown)


packages/analytics-browser/test/browser-client.test.ts
```diff
@@ -839,13 +839,16 @@ describe('browser-client', () => {
       expect(extendedLastEventTime > eventTime1).toBeTruthy();
 
       // send another event just before session expires (again)
+      // Mock Date.now() because isNewSession() depends on it
+      const dateNowMocked = jest.spyOn(Date, 'now').mockImplementation(() => extendedLastEventTime + 15);
```
chungdaniel, line 843: Is the +15 using the same value from L831? If so does it make sense for this to be put into a variable rather than using a magic number? Perhaps something like `sessionExtensionDelay`
Or changing `sessionTimeout` to a variable that can be used with +/- 1 as necessary

Mercy811: It is and a good suggestion. But I'd prefer to leave it unchanged since it works as-is and is not hard to understand.

Mercy811: Another reason is that this PR aims to fix the flakiness rather than to refactor.

chungdaniel (approved):

PR #944, opened 2025-01-13
fix(plugin-session-replay-browser): default forceSessionTracking to false

lewgordon-amplitude:
### Summary

The goal of the Session Replay plugin is to always have it added and control configuration through remote config. However, the plugin currently affects the configuration such that it's impossible to override. Rather than have users specify `forceSessionTracking` explicitly to false we'll have it default to false and have the downside that potentially Session Start and Session End events are missing.

### Checklist

* Does your PR have a breaking change?:  No, but a configuration change which should be inconsequential.


Mercy811 (approved):
Thanks @lewgordon-amplitude. LGTM

PR #948, opened 2025-01-14
fix: missing user id in ga4 forwarder plugin

liuyang1520:
### Summary

fix: missing user id in ga4 forwarder plugin

Tested with a local setup:
image

### Checklist

### DIFF
--- a/packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
+++ b/packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
@@ -185,7 +185,7 @@ export const gaEventsForwarderPlugin = ({ measurementIds = [] }: Options = {}): 
     // });
     if (userId !== lastSeenGAUserId) {
       // 2a. Set current event's user ID as Amplitude SDK's current user ID.
-      amplitude.setUserId(event.user_id);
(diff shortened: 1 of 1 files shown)


Mercy811 (commented):
I think the problem is that the if condition is always false so that setUserId() is never called. Please see https://discourse.amplitude.com/t/ga4-events-forwarder-how-can-i-set-user-id-for-forwarded-event/15319/2.

liuyang1520:
> I think the problem is that the if condition is always false so that setUserId() is never called. Please see https://discourse.amplitude.com/t/ga4-events-forwarder-how-can-i-set-user-id-for-forwarded-event/15319/2.

@Mercy811 Thanks for reviewing! I think the `event.user_id` will always be `undefined` since it is deleted in previous line, my understanding is when we see a new `userId`, we `setUserId` and update the `lastSeenGAUserId`, so I don't think `amplitude.setUserId(event.user_id)` is correct. `lastSeenGAUserId` should be set during the iterations, when we see a new `userId` but no `lastSeenGAUserId`. Let me know if I miss anything, haven't touch the SDK code for a long time 😄

kevinpagtakhan (approved):

Mercy811 (approved):
Thank you @liuyang1520

PR #951, opened 2025-02-03
feat(plugin-session-replay-browser): support custom device id

lewgordon-amplitude:
### Summary

In some cases, customers aren't able to get the right device id configured. This could be due to an issue with the browser SDK or multiple instances of the browser SDK with the plugin. In order to support issues where this is a problem, instead of requiring users to move to the standalone SDK we can allow them to override the device ID.

### Checklist


jxiwang (approved):
LGTM!

PR #961, opened 2025-02-18
feat(session replay);

jxiwang:
### Checklist

### DIFF
--- a/packages/analytics-core/src/logger.ts
+++ b/packages/analytics-core/src/logger.ts
@@ -5,8 +5,15 @@ const PREFIX = 'Amplitude Logger ';
 export class Logger implements ILogger {
   logLevel: LogLevel;
 
+  private readonly defaultLog: typeof console.log;
+  private readonly defaultWarn: typeof console.warn;
+  private readonly defaultError: typeof console.error;
+
   constructor() {
     this.logLevel = LogLevel.None;
+    this.defaultLog = Logger.getDefaultConsoleMethod(console.log);
+    this.defaultWarn = Logger.getDefaultConsoleMethod(console.warn);
+    this.defaultError = Logger.getDefaultConsoleMethod(console.error);
   }
 
   disable(): void {
@@ -21,21 +28,22 @@ export class Logger implements ILogger {
     if (this.logLevel < LogLevel.Verbose) {
       return;
     }
-    console.log(`${PREFIX}[Log]: ${args.join(' ')}`);
+
+    this.defaultLog(`${PREFIX}[Log]: ${args.join(' ')}`);
   }
 
   warn(...args: any[]): void {
     if (this.logLevel < LogLevel.Warn) {
       return;
     }
-    console.warn(`${PREFIX}[Warn]: ${args.join(' ')}`);
+    this.defaultWarn(`${PREFIX}[Warn]: ${args.join(' ')}`);
   }
 
   error(...args: any[]): void {
     if (this.logLevel < LogLevel.Error) {
       return;
     }
-    console.error(`${PREFIX}[Error]: ${args.join(' ')}`);
+    this.defaultError(`${PREFIX}[Error]: ${args.join(' ')}`);
   }
 
   debug(...args: any[]): void {
@@ -43,6 +51,10 @@ export class Logger implements ILogger {
       return;
     }
     // console.debug output is hidden by default in chrome
-    console.log(`${PREFIX}[Debug]: ${args.join(' ')}`);
+    this.defaultLog(`${PREFIX}[Debug]: ${args.join(' ')}`);
+  }
+
+  private static getDefaultConsoleMethod<T extends (...args: any[]) => void>(method: T): T {
+    return (method as { __rrweb_original__?: T })['__rrweb_original__'] || method;
   }
 }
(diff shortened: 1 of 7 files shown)


packages/analytics-core/src/logger.ts
```diff
@@ -21,28 +28,33 @@ export class Logger implements ILogger {
...
     if (this.logLevel < LogLevel.Debug) {
       return;
     }
     // console.debug output is hidden by default in chrome
-    console.log(`${PREFIX}[Debug]: ${args.join(' ')}`);
+    this.defaultLog(`${PREFIX}[Debug]: ${args.join(' ')}`);
+  }
+
+  private static getDefaultConsoleMethod<T extends (...args: any[]) => void>(method: T): T {
+    return (method as { __rrweb_original__?: T })['__rrweb_original__'] || method;
```
jxiwang, line 58: This is specifically to avoid the error related to `Uncaught RangeError: Maximum call stack size exceeded`. Open to suggestions if this is not the best process here. The place that this error can occur is within the `emit()` [function](https://github.com/amplitude/Amplitude-TypeScript/blob/9161289e3c7e82bbd02a3df57f1d685d05f95e49/packages/session-replay-browser/src/session-replay.ts#L356) here that uses `LoggerProvider`.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -317,6 +318,16 @@ export class SessionReplay implements AmplitudeSessionReplay {
     return maskSelector as unknown as string;
   }
 
+  getRecordingPlugins(loggingConfig: LoggingConfig | undefined) {
```
jxiwang, line 321: Created in case there are more plugins we need in the future related to the loggingConfig (i.e. network logging).

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -317,6 +318,16 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    const plugins = [];
+
+    if (loggingConfig?.console.enabled) {
+      plugins.push(getRecordConsolePlugin({ level: loggingConfig.console.levels }));
```
jxiwang, line 325: This will be defaulted like so - 

{
      level: ['info', 'log', 'warn', 'error'],
      lengthThreshold: 10000,
      stringifyOptions: {
        stringLengthLimit: undefined,
        numOfKeysLimit: 50,
        depthOfLimit: 4,
      },
      logger: window.console,
    }

packages/analytics-core/src/logger.ts
```diff
@@ -5,8 +5,15 @@ const PREFIX = 'Amplitude Logger ';
 export class Logger implements ILogger {
   logLevel: LogLevel;
 
+  private readonly defaultLog: typeof console.log;
```
izaaz, line 8: This is where I believe our branching makes life difficult. Changes made here will only affect browser SDK v1.0. Aren't y'all using browser sdk v2.0?

jxiwang: Would we need to make this change to Browser v2.0 branch then? Yes, we prefer Browser SDK 2.0.

izaaz: yes. This process is going to get better in the next few months but until then you'll have to make a change in main.  For any changes to the core package, please make sure you also get an approval from the SDK team cc: @Mercy811 

Curious - if you're just listening to console logs and sending it only via session replay, why modify the core library?

jxiwang: Okay, thank you! Let me pull this out then. Unfortunately, I need to make a change to the shared `LoggerProvider` class since we want to avoid an error - https://github.com/amplitude/Amplitude-TypeScript/pull/961#discussion_r1960746241.

lewgordon-amplitude (approved):
LGTM, though you may want to consider moving some of the Github comments to code comments.

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -74,6 +74,13 @@ export class SessionReplayJoinedConfigGenerator {
         sessionId,
       );
 
+      // This is intentionally forced to only be set through the remote config.
+      config.loggingConfig = await this.remoteConfigFetch.getRemoteConfig(
+        'sessionReplay',
+        'sr_logging_config',
```
izaaz, line 80: Does this have to be a new key? Can it be a subkey of an existing config?

izaaz: nvm. i see how its used below.

packages/analytics-core/src/logger.ts
izaaz: I am not sure I understand what causes this stack overflow. This feels like a workaround to an issue and generally, workarounds should be closer to the problem and not at a deeper level within the SDK. Can we try out alternate solutions?

jxiwang: Here's an [overview](https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/console.md) of the issue. This is to prevent us from getting into a recursive loop logging console information that we don't need.

Mercy811: When will `console.log['__rrweb_original__']` be available? I'm getting undefined. In this case, it will be the original console.log all the time.
image

jxiwang: Removed and replaced here: https://github.com/amplitude/Amplitude-TypeScript/pull/965

packages/analytics-core/src/logger.ts
jxiwang: Removed and replaced here: https://github.com/amplitude/Amplitude-TypeScript/pull/965

Issue #963, opened 2025-02-19
Network race condition causing duplicate events


albertxing:
## Expected Behavior
We expect the Amplitude client to wait for pending network requests to complete before sending more events, guaranteeing that the same event is not sent twice.

## Current Behavior
The client sends both new and existing events on every batch interval, resulting in duplicate events being reported. Events are not removed from the queue until the network request resolves, so if multiple requests are made in parallel, the same event can be sent multiple times.

## Possible Solution
I've been able to fix this locally with a patch that moves the [clearing of `this.scheduled` in `flush`](https://github.com/amplitude/Amplitude-TypeScript/blob/d8f0e2f45827d6f65c621a1f232a3b897cff8d3c/packages/analytics-core/src/plugins/destination.ts#L147-L150) _after_ the batch sends resolve (at the end of the function). In the deprecated JS version of the library, a separate flag was used to track pending request state, so that could also be a solution.

## Steps to Reproduce

1. In devtools, throttle network to 3G (or something slow)
2. Trigger some events
3. The client will send multiple analytics requests in parallel (without waiting for previous requests to complete)

## Environment
- JS SDK Version: @amplitude/analytics-browser v2.11.9
- Installation Method: NPM
- Browser and Version: Chrome 133

Mercy811:
Hi @albertxing, it should be resolved in https://github.com/amplitude/Amplitude-TypeScript/pull/935. Could you update to the latest version? Thanks

albertxing:
@Mercy811 that issue is similar, but I don't think we're running into a batching issue. In our case, even when events aren't split up into separate batches, Amplitude will send more requests containing the _same events_ before the previous request completes. I can confirm this still repros with the latest version of the library.

I made a simple repro JSFiddle here: https://jsfiddle.net/mo7Lyvz1/ 

![Image](https://github.com/user-attachments/assets/3fb333a2-379d-425f-a347-cdd93d0c6735)

If you set your network throttling to 3G and click the button every second or so, you'll see simultaneous requests piling up. Later requests will contain events that have already been included in previous requests.

albertxing:
We're using the NPM module, but I'm using the script in the JSFiddle for simplicity 😄

albertxing:
In either case, the core issue isn't that the requests are made in parallel, it's that the requests contain duplicate events. For example, these two pending requests from the same session are both simultaneously attempting to send events 3, 4, 5 (and more)

[...]

PR #964, opened 2025-02-20
fix(analytics-core): should not flush until previous request resolves

Mercy811:
##  Fix: should not flush until previous request resolves
This PR fixes https://github.com/amplitude/Amplitude-TypeScript/issues/963

When network is bad, for example, 3G, a network request takes time greater than flush interval to resolve, the SDK shouldn't keep sending events before the previous request.

Current behavior:
0       -> event1, schedule(1000)
1000    -> flush(event1)
1050    -> event2, schedule
**2000    -> flush(event1 + event2)**
2100    -> request1 resolve

Expected behavior:
0       -> event1, schedule(1000)
1000    -> flush(event1)
1050    -> event2, schedule(1000)
2050 (1050 + 1000)   -> no flush(event2) because request1 has not resolved
2200(1000 + 1200)    -> request1 resolved, schedule(1000)
3200    -> flush(event2)

## Refactor destination 

image

- Introduce 2 schedule indicators in blue
  - scheduleTimeout: the timeout of the current schedule. If there is a schedule with flush interval (by default 1s), but the SDK gets a rate limit, then the SDK should not flush the current schedule, but instead should respect the throttled timeout (30). 
  - scheduleId: whether there is a flush scheduled.
  - They are reset in flush() to indicate that the schedule job is done
- Introduce 1 flush indicator in pink
  - flushId: whether there is a flush in progress
  - reset when it's finished
- Now Every function has a clear functionality. Detangle the mess of timers. 

### Checklist

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -122,6 +122,10 @@ export class Destination implements DestinationPlugin {
     }
 
     if (this.scheduleId === null || (this.scheduleId && timeout > this.scheduledTimeout)) {
+      if (this.scheduleId) {
+        clearTimeout(this.scheduleId);
+      }
+      this.scheduledTimeout = timeout;
       this.scheduleId = setTimeout(() => {
         this.queue = this.queue.map((context) => {
           context.timeout = 0;
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -146,14 +146,14 @@ export class Destination implements DestinationPlugin {
 
     const batches = chunk(list, this.config.flushQueueSize);
 
+    this.scheduleTryable(later);
+
     // Promise.all() doesn't guarantee resolve order.
     // Sequentially resolve to make sure backend receives events in order
     await batches.reduce(async (promise, batch) => {
       await promise;
       return await this.send(batch, useRetry);
     }, Promise.resolve());
-
-    this.scheduleTryable(later);
 
     if (this.scheduled) {
       clearTimeout(this.scheduled);
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -52,7 +52,16 @@ export class Destination implements DestinationPlugin {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   config: Config;
-  private scheduled: ReturnType<typeof setTimeout> | null = null;
+  // Indicator of whether events that are scheduled (but not flushed yet).
+  // When flush:
+  //   1. assign `scheduleId` to `flushId`
+  //   2. set `scheduleId` to null
+  scheduleId: ReturnType<typeof setTimeout> | null = null;
+  // Timeout in milliseconds of current schedule
+  scheduledTimeout = 0;
+  // Indicator of whether current flush resolves.
+  // When flush resolves, set `flushId` to null
+  flushId: ReturnType<typeof setTimeout> | null = null;
   queue: Context[] = [];
 
   async setup(config: Config): Promise<undefined> {
@@ -80,14 +89,16 @@ export class Destination implements DestinationPlugin {
         callback: (result: Result) => resolve(result),
         timeout: 0,
       };
-      void this.addToQueue(context);
+      this.queue.push(context);
+      this.schedule(this.config.flushIntervalMillis);
+      this.saveEvents();
     });
   }
 
-  getTryableList(list: Context[]) {
+  removeEventsExceedFlushMaxRetries(list: Context[]) {
     return list.filter((context) => {
+      context.attempts += 1;
       if (context.attempts < this.config.flushMaxRetries) {
-        context.attempts += 1;
         return true;
       }
       void this.fulfillRequest([context], 500, MAX_RETRIES_EXCEEDED_MESSAGE);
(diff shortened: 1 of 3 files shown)


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -159,6 +154,11 @@ export class Destination implements DestinationPlugin {
     }, Promise.resolve());
 
     this.scheduleTryable(later);
+
+    if (this.scheduled) {
```
izaaz, line 158: Shouldn't this come before schedule tryable?

Mercy811: Good catch, but scheduleTryable() should be before sending batches. See newly added test "should schedule another flush after the previous resolves when pending events have timeout" for example.

izaaz: I don't think I follow this. We schedule the later batch, send all current batches and then if something is scheduled, we clear the scheduled batch? Then how does the later batch get scheduled again?

Mercy811: if later batch timeouts before current batch resolves, the later batch should not be flushed, and scheduled should not be cleared. 
if later batch timeouts after current batch resolves, scheduled should be cleared and later batch should be flushed.

Mercy811: Note that we await current batches, which means that scheduled won't be cleared until current batches are resolved.

Mercy811: #1 your solution
1. await send current batches
2. clear scheduled
3. schedule later batch

current batch resolves in 1500, later batch timeout is 600, then later batch is sent at 1500 + 600 = 2100
current batch resolves in 300, later batch timeout is 600, then later batch is send at 300 + 600 = 900

#2 current PR
1. schedule later batch
2. await send current batches
3. clear scheduled
current batch resolves in 1500, later batch timeout is 600, then later batch is sent at next flush interval, 2000
current batch resolves in 300, later batch timeout is 600, then later batch is send at 600

izaaz (approved):

packages/analytics-core/src/plugins/destination.ts
crleona: 1/ Moving up the schedule later batch step makes sense to keep the timing consistent.

2/ What is this later batch logic used for? To me the most consistent behavior would be to immediately send the later batch after the initial batch completes if it is over time, vs waiting for the next flush interval.

3/ What happens if I call `flush();flush();` ? are there any protections against this running in parallel?

Mercy811: @izaaz @crleona Thanks for review. I refactor the destination plugin please see updated description for details. The destination plugin was a mess of timers. I was planning to do improve it in the unified script. But did it now to avoid context switching.

albertxing (commented):
Raising a couple of concerns:
- It looks like we are replacing `scheduleId` with a new timer without clearing the previous timer, which might lead to confusing behaviour
- In the schedule timer, we set the timeout for all events in the queue to 0, which may incorrectly cause some events to be retried earlier than scheduled

Ideally, I'd like to see guarantees around only having a single active timer at a time. For example:
1. when an event is queued, schedule a timer (if none exists) for the lowest timeout in the queue (+ flushIntervalMillis)
2. in the timer, decrement all timeout values in the queue by the timeout that was scheduled
3. flush all events with a timeout of 0
4. after the flush, go to step 1

In this case, only one timer is scheduled at a time, and we won't be scheduling timers during an active flush. I think there are some variations on this, but this is the simplest system I could come up with. Just a suggestion, but I'd definitely like to see the two concerns at the top addressed 🙇‍♂️

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -80,75 +89,78 @@ export class Destination implements DestinationPlugin {
...
-      void this.flush(true).then(() => {
-        if (this.queue.length > 0) {
-          this.schedule(timeout);
-        }
-      });
-    }, timeout);
+    if (this.scheduleId === null || (this.scheduleId && timeout > this.scheduledTimeout)) {
+      this.scheduleId = setTimeout(() => {
+        this.queue = this.queue.map((context) => {
+          context.timeout = 0;
```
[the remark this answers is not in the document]
Mercy811: The sdk only sets timeout when 429 rate limit and other responses (responses other than 200 <= code < 300, 400 <= code < 500) and timeout are set to be greater than default. 

Since it's guaranteed now that the sdk won't send another request until previous request resolves, events added after the timer is scheduled can only be
1. by destination.execute(event), which is by amplitude.track(event) and timeout is 0
2. by retry of previous request with timeout

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -80,75 +89,78 @@ export class Destination implements DestinationPlugin {
...
 
-    this.scheduled = setTimeout(() => {
-      void this.flush(true).then(() => {
-        if (this.queue.length > 0) {
-          this.schedule(timeout);
-        }
-      });
-    }, timeout);
+    if (this.scheduleId === null || (this.scheduleId && timeout > this.scheduledTimeout)) {
+      this.scheduleId = setTimeout(() => {
```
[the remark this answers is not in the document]
Mercy811: Thanks for reviewing. yes, should clear the previous timer and set scheduledTimeout. I updated at https://github.com/amplitude/Amplitude-TypeScript/commit/aecd15e191f4758278d16053f4e85a2635c91849

packages/analytics-core/src/plugins/destination.ts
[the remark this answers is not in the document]
albertxing: This is a little difficult to reason about, but I think there's still a race here where we'll be setting the timeout for an event to 0 too early:
1. `t = 0ms`: A regular flush runs, we send the network request
2. `t = 100ms`: While the flush is pending, we schedule another flush with a 1000ms timeout (or whatever the default timeout is configured to), which would execute at `1100ms`
3. `t = 500ms`: Flush resolves, but is a 429. We re-queue the retryable events with the default timeout of `30000ms`, which means they should retry at `30500ms` at the earliest
4. `t = 1100ms`: The scheduled flush from step 2 runs, which will set all timeouts in the queue to 0 and call flush — this is too early! We'll be clearing the timeout values for the re-queued ratelimited events 
5. `t = 1100ms`: Second flush runs, the previous flush already resolved, so we'll send off all the events in the queue, including the ones that were ratelimited from the first flush — they'll be sent more than 29s before their scheduled time

I might have missed some logic here that prevents this, please let me know if I've gotten it wrong!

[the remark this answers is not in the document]
Mercy811: step4 `t=1100ms` won't flush because when a new timer is scheduled, the previous timer is cleared at https://github.com/amplitude/Amplitude-TypeScript/pull/964/files#r1972539272. This comply with 429, wait `30000ms` until next flush.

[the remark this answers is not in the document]
albertxing: Ah you're right! My bad 😅

PR #965, opened 2025-02-21
fix(session replay): mitigate risk of stack size exceeded error

jxiwang:
### Summary

Wrapping the `loggerProvider` at the SR SDK level to safely track console when emitting. 

### Checklist


packages/session-replay-browser/src/index.ts
```diff
@@ -1,3 +1,4 @@
 import sessionReplay from './session-replay-factory';
 export const { init, setSessionId, getSessionId, getSessionReplayProperties, flush, shutdown } = sessionReplay;
 export { SessionReplayOptions, StoreType } from './typings/session-replay';
+export { SafeLoggerProvider } from './logger';
```
Mercy811, line 4: I don't see where `SafeLoggerProvider` is used outside of session replay package. We should only export things we want consumers to use.

Mercy811 (approved):
Thanks @jxiwang. Overall looks good to me, see comment.

PR #970, opened 2025-02-28
feat: moves groups to eventoptions type

kevinpagtakhan:
### Summary

* Moves `groups` to `EventOptions` from `BaseEvent`
* Should not be a breaking change because `BaseEvent` still has `groups`

### Checklist


Mercy811 (approved):
Thanks @kevinpagtakhan. 

Quote from slack
> w/o it, i can't instrument group level events using ampli cause eventoptions don't have the groups key and baseevent isn't part of the params :grimacing: so for example

```
ampli.pageViewed(
  { path: 'home' },
  { groups: { 'org id': 'orgId' }}
)
```

PR #981, opened 2025-03-05
fix(autocapture): remove time assignment for element interactions tracking

Dogfalo:
### Summary
[AMP-122086](https://amplitude.atlassian.net/browse/AMP-122086)

Previously time was assigned to the event based on the exact timestamp of the event when it was fired. This was meant to ensure that any delay in calling track() would not impact the ordering of events.
However it seems like this potentially has the opposite effect where it causes these events to appear in the wrong order in the user timeline

### Checklist

[AMP-122086]: https://amplitude.atlassian.net/browse/AMP-122086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Mercy811 (approved):
Thanks @Dogfalo for fixing it. Rebase on the latest master should solve failing CIs.

PR #983, opened 2025-03-05
fix(session replay): dynamically import plugin-console-record

jxiwang:
### Summary

Break our `rrweb-console-plugin` import into a separate chunk.

Screenshot 2025-03-05 at 1 55 29 PM

We are now dynamically importing our Session Replay Plugin in order to prevent the `rrweb-console-plugin` package from adding ~40KB to our size. 

Implications -

- To take advantage of code-splitting, the main `session-replay-browser` SDK is now an esm build.
- A new legacy build for non-esm compatible browsers.

https://amplitude.atlassian.net/browse/SR-983?linkSource=email

https://amplitude.atlassian.net/wiki/spaces/IG/pages/3082682708/Session+Replay+SDK+Bundling

### Checklist


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "inlineSources": true,
     "importHelpers": true,
     "lib": ["es6"],
-    "module": "commonjs",
+    "module": "es2020",
```
Mercy811, line 10: Can you keep the top level `tsconfig.json` unchanged cause it's like a template of all packages. It should be enough to change package level `tsconfig.json`.

scripts/build/rollup.config.js
Mercy811, line 1: Can we keep the top level `rollup.config.js` to use `es6` so that modern syntax can be transpiled into what older version browsers support? We can change it at package level `rollup.config.js`

jxiwang: For this one, it throws a warning when at es6. It will still build though

jxiwang: Screenshot 2025-03-05 at 2 19 37 PM

Mercy811: I misunderstood [target](https://www.typescriptlang.org/tsconfig/#target) and [module](https://www.typescriptlang.org/tsconfig/#module) fields in tsconfig.
- target affects javascript syntax and features like async/await, let, and const.
- module affects how typescript handles imports/exports. 
So setting module to es2020 makes sense to me now because dynamic import is added in es2020. I was worried about the output file is not compatible with older version browsers but that only happens if target is changed.

tsconfig.json
Mercy811: the [typescript rollup plugin](https://github.com/rollup/plugins/tree/master/packages/typescript#options) in `rollup.config.js` overrides options in tsconfig. So if `rollup.config.js` is changed to use es2020 [here](https://github.com/amplitude/Amplitude-TypeScript/pull/983#discussion_r1983750113). It doesn't matter which module version we use here.

packages/session-replay-browser/rollup.config.js
```diff
@@ -1,15 +1,37 @@
...
 import { terser } from 'rollup-plugin-terser';
 import typescript from '@rollup/plugin-typescript';
 import { rollup } from 'rollup';
 import path from 'node:path';
+import commonjs from '@rollup/plugin-commonjs';
 
 iife.input = umd.input;
 iife.output.name = 'sessionReplay';
+iife.output.inlineDynamicImports = true;
+umd.output.inlineDynamicImports = true;
```
Mercy811, line 14: > This will inline dynamic imports instead of creating new chunks to create a single bundle. https://rollupjs.org/configuration-options/#output-inlinedynamicimports

umd and iffe outputs are what we host on `https://cdn.amplitude.com/libs/session-replay-browser-xxx-min.js.gz`. Setting true will not help to reduce the size of the CDN file. Ideally, we should have two chunks with https://rollupjs.org/configuration-options/#output-chunkfilenames. One for main SR SDK and the other for console record library.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -334,13 +333,18 @@ export class SessionReplay implements AmplitudeSessionReplay {
     //   logger: window.console,
     //   }
     if (loggingConfig?.console?.enabled) {
-      plugins.push(getRecordConsolePlugin({ level: loggingConfig.console.levels }));
+      try {
+        const { getRecordConsolePlugin } = await import('@amplitude/rrweb-plugin-console-record');
```
lewgordon-amplitude, line 337: I'm not familiar with this pattern so I'm curious if this will just work in general, or if folks that have different setups this would be incompatible with?

lewgordon-amplitude (commented):
I'd probably want to sync with you to better understand if there's any impact to how customers would install this and maybe how it'll fit into all the one line scripts we have.

Mercy811 (approved):
Thanks @jxiwang LGTM

PR #997, opened 2025-03-19
feat(analytics-core): remote config client

Mercy811:
### Summary

[AMP-124564]

Remote config client
- Have the same design and APIs as the iOS one https://github.com/amplitude/AmplitudeCore-Swift/pull/2
- A remote config client based on request-on-demand and subscription + callback modes. 
- The remote config client has 3 public APIs. See code comment for more details. 
- `subscribe()` is very similar to `addEventListener()`. Callback is called when the remote config fetched. 
- Analytics SDK will initialize a remote config client and plugins can use it for request remote config on their needs.
```
export interface IRemoteConfigClient {
  subscribe(key: string | undefined, deliveryMode: DeliveryMode, callback: RemoteConfigCallback): string;

  unsubscribe(id: string): boolean;

  updateConfigs(): void;
}
```

Remote config storage

- Default to local storage. Ideally we can take advantages of browser storage instead of using local storage, but it requires extra response headers ([AMP-126646]) and work. 

### Checklist

[AMP-124564]: https://amplitude.atlassian.net/browse/AMP-124564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AMP-126646]: https://amplitude.atlassian.net/browse/AMP-126646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
+++ b/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
@@ -1,4 +1,4 @@
-import { RemoteConfigStorage, RemoteConfigInfo, RemoteConfig } from './remote-config';
+import { RemoteConfigStorage, RemoteConfigInfo } from './remote-config';
 import { ILogger } from '../logger';
 
 export class RemoteConfigLocalstorage implements RemoteConfigStorage {
@@ -20,11 +20,11 @@ export class RemoteConfigLocalstorage implements RemoteConfigStorage {
       });
     } else {
       try {
-        const remoteConfig: RemoteConfig = JSON.parse(result) as RemoteConfig;
+        const remoteConfigInfo: RemoteConfigInfo = JSON.parse(result) as RemoteConfigInfo;
         this.logger.debug('Remote config localstorage get successfully.');
         return Promise.resolve({
-          remoteConfig: remoteConfig,
-          lastFetch: new Date(),
+          remoteConfig: remoteConfigInfo.remoteConfig,
+          lastFetch: new Date(remoteConfigInfo.lastFetch),
         });
       } catch (error) {
         this.logger.debug('Remote config localstorage failed to get: ', error);
--- a/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
+++ b/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
@@ -1,0 +1,42 @@
+import { RemoteConfigStorage, RemoteConfigInfo, RemoteConfig } from './remote-config';
+import { ILogger } from '../logger';
+
+export class RemoteConfigLocalstorage implements RemoteConfigStorage {
+  private readonly key: string;
+  private readonly logger: ILogger;
+
+  constructor(apiKey: string, logger: ILogger) {
+    this.key = `AMP_remote_config_${apiKey.substring(0, 10)}`;
+    this.logger = logger;
+  }
+
+  fetchConfig(): Promise<RemoteConfigInfo> {
+    const result = localStorage.getItem(this.key);
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const remoteConfig: RemoteConfig | null = result ? (JSON.parse(result) as RemoteConfig) : null;
+      this.logger.debug('Remote config localstorage get successfully.');
+      return Promise.resolve({
+        remoteConfig: remoteConfig,
+        lastFetch: new Date(),
+      });
+    } catch (error) {
+      this.logger.debug('Remote config localstorage failed to get: ', error);
+      return Promise.resolve({
+        remoteConfig: null,
+        lastFetch: new Date(),
+      });
+    }
+  }
+
+  setConfig(config: RemoteConfigInfo): Promise<boolean> {
+    try {
+      localStorage.setItem(this.key, JSON.stringify(config));
+      this.logger.debug('Remote config localstorage set successfully.');
+      return Promise.resolve(true);
+    } catch (error) {
+      this.logger.debug('Remote config localstorage failed to set: ', error);
+    }
+    return Promise.resolve(false);
+  }
+}
--- a/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
+++ b/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
@@ -1,7 +1,7 @@
 import { RemoteConfigStorage, RemoteConfigInfo } from './remote-config';
 import { ILogger } from '../logger';
 
-export class RemoteConfigLocalstorage implements RemoteConfigStorage {
+export class RemoteConfigLocalStorage implements RemoteConfigStorage {
   private readonly key: string;
   private readonly logger: ILogger;
 
@@ -31,13 +31,13 @@ export class RemoteConfigLocalstorage implements RemoteConfigStorage {
 
     try {
       const remoteConfigInfo: RemoteConfigInfo = JSON.parse(result) as RemoteConfigInfo;
-      this.logger.debug('Remote config localstorage get successfully.');
+      this.logger.debug('Remote config localstorage parsed successfully.');
       return Promise.resolve({
         remoteConfig: remoteConfigInfo.remoteConfig,
         lastFetch: new Date(remoteConfigInfo.lastFetch),
       });
     } catch (error) {
-      this.logger.debug('Remote config localstorage failed to get: ', error);
+      this.logger.debug('Remote config localstorage failed to parse: ', error);
       localStorage.removeItem(this.key);
       return Promise.resolve(failedRemoteConfigInfo);
     }
--- a/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
+++ b/packages/analytics-core/src/remote-config/remote-config-localstorage.ts
@@ -12,20 +12,27 @@ export class RemoteConfigLocalstorage implements RemoteConfigStorage {
 
   fetchConfig(): Promise<RemoteConfigInfo> {
     const result = localStorage.getItem(this.key);
-    try {
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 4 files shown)


packages/analytics-core/src/remote-config/remote-config-localstorage.ts
```diff
@@ -0,0 +1,49 @@
+import { RemoteConfigStorage, RemoteConfigInfo, RemoteConfig } from './remote-config';
+import { ILogger } from '../logger';
+
+export class RemoteConfigLocalstorage implements RemoteConfigStorage {
```
lewgordon-amplitude, line 4: I'll need to take a closer look later, but quick question. Would there be a possibility of an in memory storage here? I know some customers may not want to use local storage by default or at least have options here.

izaaz: In memory would be the equivalent of no storage right? We load the remote config at startup and use that until the page reloads. So we'd almost never query the in memory storage?

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
```diff
@@ -0,0 +1,49 @@
...
+        remoteConfig: null,
+        lastFetch: new Date(),
+      });
+    } else {
+      try {
+        const remoteConfig: RemoteConfig = JSON.parse(result) as RemoteConfig;
+        this.logger.debug('Remote config localstorage get successfully.');
+        return Promise.resolve({
+          remoteConfig: remoteConfig,
+          lastFetch: new Date(),
```
izaaz, line 27: last fetch should be the time it was fetched from the server. This should also be persisted in local storage right? That way, a blade could know if the data is stale and force refresh the data.

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
lewgordon-amplitude: Yeah no storage works too. Just some option to not have a persisted config.

Mercy811: @izaaz I am typing the same. We'd only fetch from in memory storage if the page doesn't reload. And as a result, all requests go to remote.

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(apiKey: string, logger: ILogger) {
+    this.key = `AMP_remote_config_${apiKey.substring(0, 10)}`;
+    this.logger = logger;
+  }
+
+  fetchConfig(): Promise<RemoteConfigInfo> {
+    const result = localStorage.getItem(this.key);
+    if (result === null) {
+      this.logger.debug('Remote config localstorage gets null because the key does not exist');
+      return Promise.resolve({
```
izaaz, line 17: Don't we also need to return the `source` from this function?

Mercy811: No, we set source in remote config client, if storage is used, then source is `cache`.

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
Mercy811: Nice catch! I fixed it at https://github.com/amplitude/Amplitude-TypeScript/pull/997/commits/0416922e84da686c16866cec0f11a58277266269

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      try {
+        const remoteConfigInfo: RemoteConfigInfo = JSON.parse(result) as RemoteConfigInfo;
+        this.logger.debug('Remote config localstorage get successfully.');
+        return Promise.resolve({
+          remoteConfig: remoteConfigInfo.remoteConfig,
+          lastFetch: new Date(remoteConfigInfo.lastFetch),
+        });
+      } catch (error) {
+        this.logger.debug('Remote config localstorage failed to get: ', error);
```
izaaz, line 30: Should we also delete it from local storage in this case so that the next call gets the correct data?

Mercy811: Thanks. It's also possible that localstorage is not available. Fixed both of them at https://github.com/amplitude/Amplitude-TypeScript/pull/997/commits/3f5e5fa2dd8b4c7d43092ee452a66b5f2186b6a2

crleona (approved):

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
```diff
@@ -0,0 +1,56 @@
...
+        lastFetch: new Date(remoteConfigInfo.lastFetch),
+      });
+    } catch (error) {
+      this.logger.debug('Remote config localstorage failed to get: ', error);
+      localStorage.removeItem(this.key);
+      return Promise.resolve(failedRemoteConfigInfo);
+    }
+  }
+
+  setConfig(config: RemoteConfigInfo): Promise<boolean> {
```
Dogfalo, line 46: Just something I was thinking about, it could be useful to save some additional information alongside the remote config. Metadata regarding the server response time (time it was saved) or params used to fetch the remote config. These could be useful when invalidating the cache

Mercy811: Yes! We can expand RemoteConfigInfo in the future. But for now we want minimum APIs to start with.

izaaz: Actually, that is a good point. What happens when the request info change? For example, lets say the fetched keys change. We would need to discard whats in the DB and fetch again right?

Mercy811: the remote config client will always fetch and store the whole remote config, it then filters based keys. See details in `sendCallback()`

packages/analytics-core/src/remote-config/remote-config.ts
```diff
@@ -0,0 +1,302 @@
...
+  deliveryMode: DeliveryMode;
+  callback: RemoteConfigCallback;
+  lastCallback?: Date;
+}
+
+/**
+ * Callback used in `RemoteConfigClient.subscribe()`.
+ * This function is called when the remote config is fetched.
+ */
+type RemoteConfigCallback = (remoteConfig: RemoteConfig | null, source: Source, lastFetch: Date) => void;
```
lewgordon-amplitude, line 74: It seems like we should never get a null remoteConfig for a cache source. If so, can we be stricter on the typing here? This is to save folks from needing to handle an additional case of null config and cache source.

packages/analytics-core/src/remote-config/remote-config.ts
```diff
@@ -0,0 +1,302 @@
...
+        },
+      });
+
+      // Handle unsuccessful fetch
+      if (!res.ok) {
+        const body = await res.text();
+        this.logger.debug(`Remote config client fetch with retry time ${retries} failed with ${res.status}: ${body}`);
+        retries -= 1;
+        if (retries > 0) {
+          return this.fetch(retries);
```
lewgordon-amplitude, line 270: Nit: Any chance this could be refactored to a for loop? The typical retry count is small, but it's usually better not to have this recursion if possible.

lewgordon-amplitude, line 270: I know we're trying to resolve the config as fast as possible, but it'd be good to have a bit of [jitter and at least a linear backoff policy](https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/) here.

Mercy811: Thanks Lew for sharing this doc. Backoff can prevent increasing traffic when error starts. To build a more solid retry logic, also have to know which error we should retry. Do you have a list of error code and response body for retries? My question is how much can we benefit from backoff retry. Can we try to estimate the reduced traffic in our 1:1?

Mercy811: for a reasonable backoff time, I'm thinking timeout / retry times

packages/analytics-core/src/remote-config/remote-config-localstorage.ts
Mercy811: goal: to keep config the same across a session

pros: easy to debug

cons: customers expect remote config change take effect immediately 

more of a design question. not a blocker for now

packages/analytics-core/src/remote-config/remote-config.ts
Mercy811: `subscribeWaitForRemote()` is equivalent to current remote config client. It only call callback once.

packages/analytics-core/src/remote-config/remote-config.ts
Mercy811: Thanks @lewgordon-amplitude I updated at https://github.com/amplitude/Amplitude-TypeScript/pull/997/commits/017b1731d5f012f4f2cc4f6d080e28ab43c2e82a

lewgordon-amplitude: Looks good. I'd just add some jitter like this:

```javascript
function getJitterDelay(baseDelay) {
  return baseDelay + Math.floor(Math.random() * baseDelay);
}
```

Mercy811: added at https://github.com/amplitude/Amplitude-TypeScript/pull/997/commits/6b4bc5bfced9748d6c5ea90988b82503a69f1705

PR #1001, opened 2025-03-21
fix(autocapture): removed deprecated Element Selector event property …

Dogfalo:
…and improved click performance

### Summary
https://amplitude.atlassian.net/browse/AMP-126627

- Removed the deprecated `[Amplitude] Element Selector` property. This was problematic because finding a unique CSS selector for an element on the page is a expensive calculation.
- Improved performance on click tracking. This was mostly a bug in the previous version where the observable didn't multicast itself using share, so duplicate work was being done to grab proprties about the event.

### Checklist


Mercy811 (approved):
Thanks Alvin. LGTM

PR #1002, opened 2025-03-21
feat(session replay rn): add log level config to session replay rn plugin

sojingle:
### Summary

Add a log-level config param to remove hardcoded debug logging in PluginSessionReplayReactNative

### Checklist

### DIFF
--- a/packages/plugin-session-replay-react-native/src/session-replay-config.tsx
+++ b/packages/plugin-session-replay-react-native/src/session-replay-config.tsx
@@ -1,11 +1,15 @@
+import { LogLevel } from "@amplitude/analytics-types";
+
 export interface SessionReplayConfig {
     sampleRate?: number;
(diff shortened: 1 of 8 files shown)


packages/plugin-session-replay-react-native/src/session-replay-config.tsx
```diff
@@ -1,11 +1,15 @@
+import { LogLevel } from "@amplitude/analytics-types";
+
 export interface SessionReplayConfig {
     sampleRate?: number;
     enableRemoteConfig?: boolean;
+    logLevel?: LogLevel;
```
izaaz, line 6: logLevel is nullable. Where does the null check happen before calling specific platforms? Eg, what if i create a SessionReplayConfig object without using the `getDefaultConfig()` function?

PR #1008, opened 2025-03-26
feat(analytics-browser): set default for fetchRemoteConfig option to true

Dogfalo:
### Summary

This PR sets the default for the `fetchRemoteConfig` option to be `true`.
The goal of this is to allow more users, excluding those who have set it to false explicitly, to be able to control their SDK configuration through the Amplitude platform.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -88,7 +88,7 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     },
     public transport: 'fetch' | 'xhr' | 'beacon' = 'fetch',
     public useBatch: boolean = false,
-    public fetchRemoteConfig: boolean = false,
+    public fetchRemoteConfig: boolean = true,
     userId?: string,
     pageCounter?: number,
     debugLogsEnabled?: boolean,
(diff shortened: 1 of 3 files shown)


Mercy811 (changes requested):
Can we also update the default value and add a url link (for example, `https://app.amplitude.com/data/amplitude/{your org name}/settings/autocapture`) to the app or a doc to where customers can edit remote config? https://github.com/amplitude/Amplitude-TypeScript/blob/823b695c3a67b4c44f4a2ba5a1f7f766845e9e55/packages/analytics-core/src/types/browser-config.ts#L79-L82.

Mercy811:
> > Can we also update the default value and add a url link (for example, `https://app.amplitude.com/data/amplitude/{your org name}/settings/autocapture`) to the app or a doc to where customers can edit remote config?
> > https://github.com/amplitude/Amplitude-TypeScript/blob/823b695c3a67b4c44f4a2ba5a1f7f766845e9e55/packages/analytics-core/src/types/browser-config.ts#L79-L82
> 
> @Mercy811 Updated the value in the comment and added an example link. Should this also be updated? `packages/analytics-types/src/config/browser.ts`

The latest version with turn remote config on by default only depends on the core package. So I don't think we need to update types package.

Mercy811 (approved):
Thanks @Dogfalo!

PR #1009, opened 2025-03-28
fix(analytics-core): cookie storage should delete duplicate

Mercy811:
### Summary

[AMP-125768]

The SDK, by default, creates cookies using a domain with a leading dot (e.g., .amplitude.com). However, some customers have reported cases where two cookies with the same name exist—one with a leading dot in the domain and one without. See an example screenshot in the linked ticket comment.

This PR fix the issue by detecting and deleting the duplicate cookie that does not have the leading dot, ensuring only one cookie remains.

Note: We use document.cookie here, which does not expose domain information. While the [CookieStore API](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) would provide full access to domain-specific cookies, it is not yet widely supported across modern browsers.

### Checklist

[AMP-125768]: https://amplitude.atlassian.net/browse/AMP-125768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-core/src/storage/cookie.ts
+++ b/packages/analytics-core/src/storage/cookie.ts
@@ -1,12 +1,17 @@
 import { Storage, CookieStorageOptions } from '../types/storage';
 import { getGlobalScope } from '../global-scope';
+import { ILogger, Logger } from '../logger';
 
 export class CookieStorage<T> implements Storage<T> {
   options: CookieStorageOptions;
+  // Cookie storage is initialized before the global logger in BrowserConfig.
+  // So have to crate a local logger.
+  logger: ILogger = new Logger();
   private static testValue: undefined | string;
 
   constructor(options?: CookieStorageOptions) {
     this.options = { ...options };
+    this.logger?.debug(`Cookie storage is created with options ${JSON.stringify(this.options, null, 2)}`);
   }
 
   async isEnabled(): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


Dogfalo:
Do we have a good understanding of how the cookies with different domains were created? Is it due to the customer setting certain options in the SDK?

Also are there cases where we want to set a different value on a subdomain `subdomain1.example.com` that overrides the wildcard one at `.example.com`?

izaaz:
> Do we have a good understanding of how the cookies with different domains were created? Is it due to the customer setting certain options in the SDK?
> 
> Also are there cases where we want to set a different value on a subdomain `subdomain1.example.com` that overrides the wildcard one at `.example.com`?

Yea - it would be good to understand what causes this in the first place.

packages/analytics-core/src/storage/cookie.ts
```diff
@@ -51,12 +56,35 @@ export class CookieStorage<T> implements Storage<T> {
...
+    // If only one cookie, return its value
+    if (cookies.length === 1) {
+      return cookies[0].substring(key.length + 1);
+    } else {
+      // If two cookies, remove the one without the leading dot.
+      // For example: .amplitude.com stays while amplitude.com is deleted
+      this.logger?.debug(`Two cookies with the same key are found. Deleting  ${cookies.join(' ')}`);
+      const originalDomain = this.options.domain;
+
+      const removalOptions = {
```
izaaz, line 76: Shouldn't we be deleting the cookie with the other domain?

Mercy811:
> > Do we have a good understanding of how the cookies with different domains were created? Is it due to the customer setting certain options in the SDK?
> > Also are there cases where we want to set a different value on a subdomain `subdomain1.example.com` that overrides the wildcard one at `.example.com`?
> 
> Yea - it would be good to understand what causes this in the first place.

We haven't find out why the cookie without leading dot is created for customers use default cookie options. The customer does not use experiment (looked up in datadog and admin-v2 page). This it's also ruled out that the cookie is set by experiment server-side sdk. They claimed they don't create the cookie either and have always used the default cookie options. By default, the sdk creates cookies on the top level domain so `.example.com` will always be used over `subdomain1.example.com` unless it's specified in config.

packages/analytics-core/src/storage/cookie.ts
Mercy811: Yes. I fixed it at https://github.com/amplitude/Amplitude-TypeScript/pull/1009/commits/1ec9fbfbc3c8460ab3ad7c02a337bc17b12161ca

PR #1017, opened 2025-04-04
DRAFT: fix(analytics-browser): add try/catch when accessing localStorage

daniel-graham-amplitude:
### Checklist

### DIFF
--- a/packages/analytics-browser/src/storage/local-storage.ts
+++ b/packages/analytics-browser/src/storage/local-storage.ts
@@ -10,8 +10,19 @@ export class LocalStorage<T> extends BrowserStorage<T> {
   loggerProvider?: ILogger;
 
   constructor(config?: LocalStorageOptions) {
-    super(getGlobalScope()?.localStorage);
-    this.loggerProvider = config?.loggerProvider;
+    let isUnavailable = false;
+    const loggerProvider = config?.loggerProvider; // Extract loggerProvider
+    let localStorage;
+    try {
+      localStorage = getGlobalScope()?.localStorage;
+    } catch {
+      isUnavailable = true;
+    }
+
+    super(localStorage);
+    this.loggerProvider = loggerProvider; // Assign after super()
+
+    this.isUnavailable = isUnavailable;
   }
 
   async set(key: string, value: T): Promise<void> {
--- a/packages/analytics-browser/src/storage/browser-storage.ts
+++ b/packages/analytics-browser/src/storage/browser-storage.ts
@@ -5,7 +5,7 @@ export class BrowserStorage<T> implements AmplitudeStorage<T> {
   constructor(private storage?: Storage) {}
 
   async isEnabled(): Promise<boolean> {
-    if (!this.isUnavailable) {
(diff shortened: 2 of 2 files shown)


packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -10,8 +10,17 @@ export class LocalStorage<T> extends BrowserStorage<T> {
...
+    let localStorage;
+    try {
+      loggerProvider = config?.loggerProvider;
+      localStorage = getGlobalScope()?.localStorage;
+    } catch {
+      localStorage = undefined;
+    }
+
+    super(localStorage);
+    this.loggerProvider = loggerProvider;
```
Mercy811, line 23: Can we set loggerProvider first and then in catch loggerProvider.debug()?

packages/analytics-browser/test/storage/local-storage.test.ts
```diff
@@ -35,4 +35,27 @@ describe('local-storage', () => {
...
+          // simulates security error: https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document/
+          const err = `SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document`;
+          throw new Error(err);
+        },
+        configurable: true,
+      });
+      const localStorage = new LocalStorage();
+      expect(await localStorage.isEnabled()).toBe(false);
+    } finally {
+      console.log('restoring localStorage');
```
Mercy811, line 54: nit: clean console.log()

Mercy811 (changes requested):

packages/analytics-browser/src/storage/local-storage.ts
daniel-graham-amplitude: I just pushed a fix. How about if I just take `this.loggerProvider = config?.loggerProvider;` outside of the "try --> catch" scope? It was never being caught to begin with, and it's not a cause of the bug, so best I just leave it as it was.

packages/analytics-browser/src/storage/local-storage.ts
```diff
@@ -10,7 +10,14 @@ export class LocalStorage<T> extends BrowserStorage<T> {
...
-    super(getGlobalScope()?.localStorage);
+    let localStorage;
+
+    try {
+      localStorage = getGlobalScope()?.localStorage;
+    } catch {
+      localStorage = undefined;
+    }
+    super(localStorage);
     this.loggerProvider = config?.loggerProvider;
```
Mercy811, line 21:
```suggestion
    this.loggerProvider = config?.loggerProvider;
    let localStorage;

    try {
      localStorage = getGlobalScope()?.localStorage;
    } catch {
    this.loggerProvider.debug('xxxxx');
      localStorage = undefined;
    }
    super(localStorage);

```

Mercy811: thanks `config.loggerProvider` should work.

PR #1025, opened 2025-04-15
feat(analytics-core): new plugin interfaces onXXXchanged()

Mercy811:
### Summary

[AMP-126972](https://amplitude.atlassian.net/browse/AMP-126972)

Add three APIs to plugin. They are called when the corresponding property is changed. 
- onSessionIdChanged()
- onOptOutChanged()
- onIdentityChanged()

Adopt SR plugin to the first two.
- onSessionIdChanged()
  - update session replay instance's session id
- onOptOutChanged()
  - call session replay shutdown() or init() according to optOut

 

### Checklist

[AMP-126972]: https://amplitude.atlassian.net/browse/AMP-126972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -188,6 +188,8 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, Pl
     this.config.loggerProvider.debug('function setUserId: ', userId);
     if (userId !== this.config.userId || userId === undefined) {
       this.config.userId = userId;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      this.timeline.onIdentityChanged({ userId: userId });
       setConnectorUserId(userId, this.config.instanceName);
     }
   }
@@ -202,8 +204,12 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, Pl
       return;
     }
     this.config.loggerProvider.debug('function setDeviceId: ', deviceId);
-    this.config.deviceId = deviceId;
-    setConnectorDeviceId(deviceId, this.config.instanceName);
+    if (deviceId !== this.config.deviceId) {
+      this.config.deviceId = deviceId;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      this.timeline.onIdentityChanged({ deviceId: deviceId });
+      setConnectorDeviceId(deviceId, this.config.instanceName);
+    }
   }
 
   reset() {
@@ -229,6 +235,11 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, Pl
     this.config.loggerProvider.debug('function setSessionId: ', sessionId);
 
     const previousSessionId = this.getSessionId();
+    if (previousSessionId !== sessionId) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      this.timeline.onSessionIdChanged(sessionId);
+    }
+
     const lastEventTime = this.config.lastEventTime;
     let lastEventId = this.config.lastEventId ?? -1;
 
--- a/packages/analytics-core/src/core-client.ts
+++ b/packages/analytics-core/src/core-client.ts
@@ -2,8 +2,8 @@ import { Plugin } from './types/plugin';
 import { IConfig } from './config';
 import { BaseEvent, EventOptions } from './types/event/base-event';
 import { Result } from './types/result';
-import { Event } from './types/event/event';
-import { IIdentify } from './identify';
+import { Event, IdentifyOperation, SpecialEventType, UserProperties } from './types/event/event';
+import { IIdentify, OrderedIdentifyOperations } from './identify';
 import { IRevenue } from './revenue';
 import { CLIENT_NOT_INITIALIZED, OPT_OUT_MESSAGE } from './types/messages';
 import { Timeline } from './timeline';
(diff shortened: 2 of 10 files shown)


packages/analytics-browser/src/browser-client.ts
```diff
@@ -188,6 +188,8 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, Pl
     this.config.loggerProvider.debug('function setUserId: ', userId);
     if (userId !== this.config.userId || userId === undefined) {
       this.config.userId = userId;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      this.timeline.onIdentityChanged({ userId: userId });
```
daniel-graham-amplitude, line 192: _Optional_: `{userId: userId}` can be shortened to just `{userId}`.

Mercy811: I would prefer to keep it verbose for clarity

lewgordon-amplitude (approved):
SR changes LGTM!

packages/analytics-core/src/core-client.ts
```diff
@@ -287,13 +286,94 @@ export class AmplitudeCore implements CoreClient {
...
+        case IdentifyOperation.SET:
+          Object.assign(updatedProperties, opProperties);
+          break;
+      }
+    });
+
+    // Merge non-operation properties.
+    // Custom properties should not be affected by operations.
+    // https://github.com/amplitude/nova/blob/343f678ded83c032e83b189796b3c2be161b48f5/src/main/java/com/amplitude/userproperty/model/ModifyUserPropertiesIdent.java#L79-L83
+    Object.assign(updatedProperties, nonOpProperties);
```
crleona, line 360: this is additional behavior from what we had discussed earlier?

crleona (approved):

packages/analytics-core/src/core-client.ts
```diff
@@ -287,13 +286,94 @@ export class AmplitudeCore implements CoreClient {
...
+
+    OrderedIdentifyOperations.forEach((operation) => {
+      if (!Object.keys(userProperties).includes(operation)) return;
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const opProperties = userProperties[operation];
+
+      switch (operation) {
+        case IdentifyOperation.CLEAR_ALL:
+          for (const prop in updatedProperties) {
```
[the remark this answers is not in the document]
daniel-graham-amplitude: `forEach` can be used too: 

`Object.keys(updatedProperties).forEach((prop) => delete updatedProperties[prop])`

packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -271,6 +272,55 @@ describe('SessionReplayPlugin', () => {
...
+
+    test('should re init when optOut is changed to false', async () => {
+      const customDeviceId = randomUUID();
+      const sessionReplay = new SessionReplayPlugin({
+        deviceId: customDeviceId,
+      });
+      await sessionReplay.setup?.(mockConfig, mockAmplitude);
+      await sessionReplay.onOptOutChanged?.(false);
+
+      expect(init).toHaveBeenCalledTimes(2);
```
daniel-graham-amplitude, line 312: Why is "init" called twice? Shouldn't it have only been called once (when the optOut status changed to "false")?

Seems like one of the "init" calls is from a previous run and should be reset, unless I'm wrong and this is expected.

packages/analytics-core/src/core-client.ts
Mercy811: Yes. I add this because the user properties type allows custom properties (no operations) https://github.com/amplitude/Amplitude-TypeScript/blob/43441594b6dd5e23bb09159a5bac8294a8948034/packages/analytics-core/src/types/event/event.ts#L88-L92.

packages/plugin-session-replay-browser/test/session-replay.test.ts
Mercy811: the first time is called in setup()

daniel-graham-amplitude (commented):
I left one non-optional comment (for..in) and then a few other optional comments.

Has all the testing been done from inside the unit tests (in a JSDom env) or were you able to test some of these out in a real browser sandbox?

packages/plugin-session-replay-browser/test/session-replay.test.ts
daniel-graham-amplitude: I see. If "init" was already called, how come it need to be called a second time when the opt out status changes to "false"? Shouldn't it just resume the session as usual?

packages/analytics-core/src/core-client.ts
crleona: ok, thanks - will add to iOS as well

packages/analytics-core/src/core-client.ts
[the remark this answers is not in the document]
Mercy811: Thanks @daniel-graham-amplitude, I updated it at https://github.com/amplitude/Amplitude-TypeScript/pull/1025/commits/a411e85fea69f39a7d220f01d9320e08bd26388d

daniel-graham-amplitude (approved):

packages/plugin-session-replay-browser/test/session-replay.test.ts
Mercy811: A more real example in the unified sdk would be that both analytics and SR SDKs are initialized with optOut and then opt out is set to false. The problem is that SR plugin doesn't hold a SR instance or SR has `getOptOut()` thus cannot know the current optOut status. Here just try to be respectful for each optOut changes. I will improve this in my next PR.

Mercy811:
Thanks @daniel-graham-amplitude for the review. We only have unit tests in CI/CD. It would be great if we set up a real browser sandbox and automate it. I've used datadog synthetic tests, but it requires manual setup so not very ideal.

PR #1035, opened 2025-04-18
feat(analytics-browser): autocapture Network Events

daniel-graham-amplitude:
### Summary

Added a "network observable" (along with the other observables like "mutation observable"). This observable subscribes to "NetworkObserver.subscribe" to detect whenever there is a "Network Request Event". When there's is, it will check if the request event matches the rules that determine if it should be captured as an Amplitude Event, or not. If it does match, it will send an event to Amplitude that looks like this:

```
{
  ...
 "event_type": "[Amplitude] Network Request",
 "event_properties": {
  "[Amplitude] URL": "https://httpstat.us/400",
  "[Amplitude] Request Method": "GET",
  "[Amplitude] Status Code": 400,
  "[Amplitude] Start Time": 1745355416452,
  "[Amplitude] Completion Time": 1745355416513,
  "[Amplitude] Duration": 61,
  "[Amplitude] Response Body Size": 15
 },
 ...
}
```

### Checklist


packages/analytics-core/src/index.ts
```diff
@@ -60,4 +60,7 @@ export {
 } from './types/browser-config';
 export { BrowserClient } from './types/browser-client';
 
-export { NetworkObserver, NetworkRequestEvent, NetworkEventCallback } from './network-observer';
+import NetworkObserver from './network-observer';
+export { NetworkObserver };
+export { NetworkRequestEvent, NetworkEventCallback } from './network-observer';
```
Mercy811, line 65: if `./network-observer` default export is an instance `export default new NetworkObserver();`, can we not capitalize it? It looks like a exporting a class.

And if we are exporting more than one things from one file, why do we still use default export?

Mercy811 (approved):
Thanks @daniel-graham-amplitude. LGTM. Just one nit please see comment

packages/analytics-core/src/index.ts
daniel-graham-amplitude: > It looks like a exporting a class

Yeah good thinking, I didn't think of that.

daniel-graham-amplitude: > And if we are exporting more than one things from one file, why do we still use default export?

That's how I've done it in the past (mixing named + default), but if it's a convention that we follow to not mix them, I can just change it to a named export.

PR #1036, opened 2025-04-21
chore: add "networkTracking" to autocapture config

daniel-graham-amplitude:
### Summary

Gets the configuration ready for "[autocapture.networkTracking](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3013804985/Error+Tracking+Frustration+SDK+Design+Proposal+Network+Tracking#Configuration)".

### Checklist


packages/analytics-core/src/types/browser-config.ts
```diff
@@ -81,6 +82,10 @@ export interface ExternalBrowserConfig extends IConfig {
    * @defaultValue `true`
    */
   fetchRemoteConfig?: boolean;
+  /**
+   * Captures network requests and responses.
+   */
```
Mercy811, line 87: Could you add the default value here like other fields? Thanks

daniel-graham-amplitude:
> Thanks @daniel-graham-amplitude. LGTM.
> 
> Why do you update analytics-types if the new feature will only be available on the latest version?

I just did it because I was copying the pattern of the other configs. I took it out though. So to confirm for future reference does this mean "analytics-types" is deprecated and only used for the old sdk version?

Mercy811:
> > Thanks @daniel-graham-amplitude. LGTM.
> 
> > 
> 
> > Why do you update analytics-types if the new feature will only be available on the latest version?
> 
> 
> 
> I just did it because I was copying the pattern of the other configs. I took it out though. So to confirm for future reference does this mean "analytics-types" is deprecated and only used for the old sdk version?

Yes. analytics-types is no longer a dependency of analytics-browser and is no longer in package.json.

PR #1037, opened 2025-04-21
fix(analytics-browser): default values for config were not being used for fetchRemoteConfig

Dogfalo:
### Summary
A previous change (#1008) set the default value of fetchRemoteConfig to be `true`, but this default was not being used which resulted in fetchRemoteConfig defaulting to `undefined`.

This change fixes it by referencing the correct browserOptions variable which includes the default values

### Checklist


Mercy811 (approved):
Thanks @Dogfalo for fixing it.

PR #1039, opened 2025-04-21
chore(unified): create a new package @amplitude/unified

Mercy811:
### Summary

- Create a new package for `@amplitude/unified`. `"private": true,` in `package.json` so it won't be published to npm for now

Usage
```ts
import * as amplitude from '@amplitude/unified';

amplitude.initAll('5b9a9510e261f9ead90865bbc5a7ad1d', {
  serverZone: 'US',
  instanceName: 'test-instance',
  analytics: {
    autocapture: true
  },
  sr: {
    sampleRate: 0.5
  },
  experiment: {
    deploymentKey: 'deployment-key'
  },
}).then(() => {
  void amplitude.experiment.fetch();
  const variant = amplitude.experiment.variant('FLAG_KEY');
});
```

### Checklist


packages/unified/src/unified.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export type UnifiedOptions = UnifiedSharedOptions & {
+  analytics?: BrowserOptions;
+  sr?: Omit<SessionReplayOptions, keyof UnifiedSharedOptions>;
+  experiment?: Omit<ExperimentConfig, keyof UnifiedSharedOptions>;
+};
+
+export interface UnifiedClient extends BrowserClient {
+  initAll(apiKey: string, unifiedOptions?: UnifiedOptions): Promise<void>;
```
Mercy811, line 36: Technically, we can use [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to keep the same name `init()`. For example
```ts
  init(apiKey: string, unifiedOptions?: UnifiedOptions): Promise<void>;
  init(apiKey?: string, userIdOrOptions?: string | BrowserOptions, maybeOptions?: BrowserOptions): ReturnType<AmplitudeBrowser['init']>;
  init(
    apiKey?: string,
    userIdOrOptions?: string | BrowserOptions,
    maybeOptions?: BrowserOptions,
  ): Promise<void> | ReturnType<AmplitudeBrowser['init']>;
```

But i'd like to keep this as `initAll()` so that the implementation is simpler for now. Please let me know if you have a strong opinion to not use `initAll()` here.

packages/unified/src/unified.ts
```diff
@@ -0,0 +1,110 @@
...
+  ExperimentPlugin,
+  experimentPlugin,
+} from '@amplitude/plugin-experiment-browser';
+import { BrowserClient, BrowserOptions, LogLevel } from '@amplitude/analytics-core';
+
+export interface UnifiedSharedOptions {
+  serverZone?: 'US' | 'EU';
+  instanceName?: string;
+  /**
+   * If logLevel is debug:
```
crleona, line 20: Do these debug modes just enable additional logging? If not, I'm hesitant to have them controlled by log level. There's no equivalent to debug mode on the mobile sdks that I'm aware of, but maybe we make `debug` a top level property, at least on ts for now?

crleona (approved):

packages/unified/src/unified.ts
Mercy811: experiment's debug seems to be just logs https://github.com/amplitude/experiment-js-client/blob/29b1640f54ac0cab1ac1795d0ffd104a1fbf44d5/packages/experiment-browser/src/config.ts#L13-L17. But `sr.debugMode` adds additionally event properties https://github.com/amplitude/Amplitude-TypeScript/blob/7a665d55fff89092ed5f2bb94caa1eb2c7efe5b1/packages/session-replay-browser/src/config/types.ts#L81-L86. 

How about not setting `sr.config.debugMode` here and keep the rest?
``` 
  /**
   * If logLevel is debug:
   * - analytics.config.logLevel = debug
   * - sr.config.logLevel = debug
   * - sr.config.debugMode = true // Remove this
   * - experiment.config.debug = true
   */
```

crleona: Both of these flags warn not to turn them on in prod environments, so I don't think setting them implicitly from log level (which should be fine to set to any level in prod environments) is a good idea.

Mercy811: Thanks @crleona, I remove logLevel at https://github.com/amplitude/Amplitude-TypeScript/pull/1039/commits/b0f3ea8a18273edc07f4cc023746fc1f89924b12

PR #1049, opened 2025-04-28
fix(analytics-browser): fix TS2742 by re-exporting types

daniel-graham-amplitude:
### Summary

When attempting to install "analytics-browser" using PNPM, we were getting this error in our postinstall script

`postinstall: @amplitude/instrumentation:build:lib: src/lightning/analytics-shared/src/ampli/index.ts(1661,3): error TS2742: The inferred type of 'event_type' cannot be named without a reference to '.pnpm/@amplitude+analytics-core@2.10.1/node_modules/@amplitude/analytics-core'. This is likely not portable`

The issue is that we migrated from using "analytics-types" to "analytics-core" by re-exporting from "analytics-core" instead of "analytics-types".

There was an issue with the re-exporting that meant that it wasn't a proper re-export.

Here's a simplified example:
```
import {
  BaseEvent as BaseEventType
} from '@amplitude/analytics-core';
export namespace Types {
  export type BaseEvent as BaseEventType
}
```

Because it actually creates a new variable (BaseEventType) whenever any code references the original `BaseEvent` in "analytics-core" it will look for the type definition in "analytics-core" and won't find it in "analytics-browser", because it's technically not present in "analytics-browser". This isn't a problem in NPM and Yarn because it can resolve "analytics-core". But PNPM this is a problem due to it's "content addressable memory" format whereby "analytics-core" isn't a sibling of "analytics-browser".

What this fix does is makes a new file called "./types.ts", re-exports all of the required public types, and then "index.ts" subsequently re-exports it like this `export * as types from './types'`.

### Checklist


packages/analytics-browser/src/index.ts
```diff
@@ -27,96 +27,8 @@ export { AmplitudeBrowser } from './browser-client';
...
-  Campaign as CampaignType,
-  CampaignParser as CampaignParserType,
-  ClickIdParameters as ClickIdParametersType,
-  ReferrerParameters as ReferrerParametersType,
-  UTMParameters as UTMParametersType,
-} from './attribution/campaign';
-
 // Export the following types to maintain backward compatibility with `analytics-types`.
 // In the next major version, only export customer-facing types to reduce the public API surface.
 // eslint-disable-next-line @typescript-eslint/no-namespace
```
ShMcK, line 32: This disable next line (line 32) may not be needed.

Mercy811 (approved):
Thanks @daniel-graham-amplitude for fixing it!

PR #1054, opened 2025-04-29
refactor: make "plugin-network-capture-browser" a new package

daniel-graham-amplitude:
### Summary
Before this PR, the "network capture" logic was mixed in with "plugin-autocapture-browser" along with element interactions. This proved to be a bit unruly because there was a lot of conditional type logic to make them play nicely. What this PR does is moves "network capture" into it's own package altogether and then removes any network capturing logic from "plugin-autocapture-browser".

The new "plugin-network-capture-browser" package doesn't do anything yet. The next PR will be to have it be conditionally used by "analytics-browser" (ie: have it be used if `autocapture.networkTracking` is `true`).

### Checklist


jxiwang:
Does this have any implications with the shared network observer? Such as if Session Replay wanted to re-use the network observer instance.

daniel-graham-amplitude:
> Does this have any implications with the shared network observer? Such as if Session Replay wanted to re-use the network observer instance.

It doesn't have any implications, no. The singleton "networkObserver" is exported from "analytics-core" still. The network capture plugin is a consumer of it. And then after this is done, session replay will also be a consumer.

packages/plugin-network-capture-browser/src/network-capture-plugin.ts
```diff
@@ -0,0 +1,132 @@
...
+  BrowserConfig,
+  EnrichmentPlugin,
+  ElementInteractionsOptions,
+  NetworkRequestEvent,
+  networkObserver,
+  NetworkEventCallback,
+  NetworkTrackingOptions,
+} from '@amplitude/analytics-core';
+import * as constants from './constants';
+import { Observable, Subscription } from 'rxjs';
```
Dogfalo, line 13: I would rewrite this without the use of observables and rxjs since overall it looks like a simple handler. We're later going to rework the autocapture-plugin to use a much lighter weight implementation to reduce bundle size

daniel-graham-amplitude: I can do that, but do you mind if I do this in a later pull request, since the scope of this PR is to move the location of the network capture plugin without meddling with the implementation.

jxiwang (approved):
LGTM

packages/analytics-browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@amplitude/analytics-browser",
-  "version": "2.16.1",
+  "version": "2.16.0",
```
Dogfalo, line 3: This looks unintentional?

daniel-graham-amplitude: Yep

PR #1055, opened 2025-04-30
feat(analytics-browser): add "autocapture.networkTracking" plugin

daniel-graham-amplitude:
### Summary
* Add the network capture functionality to the SDK
* Network Capture is done when user has set it on "config.autocapture" and they have set "config.networkTrackingOptions" (note: that the second condition of "networkTrackingOptions" being present is temporary, planning on doing a quiet release).
* 

### Checklist


packages/analytics-browser/src/browser-client.ts
```diff
@@ -154,6 +157,13 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient {
       await this.add(autocapturePlugin(getElementInteractionsConfig(this.config))).promise;
     }
 
+    // TODO: the "this.config.networkTrackingOptions" should be taken out once we have
+    // beta tested network tracking
+    if (isNetworkTrackingEnabled(this.config.autocapture) && !!this.config.networkTrackingOptions) {
```
jxiwang, line 162: If this is taken out, then we all users that turn auto-capture on will start tracking network requests?

jxiwang (approved):
LGTM, just one question

packages/analytics-browser/src/browser-client.ts
daniel-graham-amplitude: If it is taken out, then all users who have `autocapture=true` or `autocapture.networkTracking=true` will start tracking network requests.

jxiwang: Will they be able to set `autocapture.networkTracking = false` at the minimum?

daniel-graham-amplitude: Yeah, setting `autocapture.networkTracking` to `false` will disable it.

jxiwang: Awesome thanks for confirming!

PR #1056, opened 2025-05-01
chore: add dev server for testing

daniel-graham-amplitude:
### PROBLEM
While working on the network error capturing, I was a bit frustrated by the local testing process. Using the sandbox had a few problems. The two particular bottlenecks are: a build step that needed to be triggered on every change; the bundles are minified and the source maps don't map back to the original source code.

### SOLUTION
Create a development HTTP server for testing the SDK locally. It serves HTML files from a directory called "/test-server" and those HTML files can reference TS scripts from inside of the Amplitude-Typescript project.

For example, this is what a script tag looks like

```
<script type='module'>
  window.amplitude = amplitude;
  amplitude.init( ... )
</script>
```

It directly references the "analytics-browser" in your local filesystem and that gets served.

### USAGE
Run `yarn dev` and it will start the server and open the home page in your browser automatically

### VIDEO DEMO

This video demonstrates how to: start the server, make edits to the source, see the edits hot reload, find source files in Chrome Devtools, set breakpoints.

https://drive.google.com/file/d/19LJNQZV_MSSwX63Y5aTmINjV-VUYUZ3b/view?usp=sharing

### FEATURES
* While running "yarn dev", a watch script is run that rebuilds the TS files as they are edited
* With Vite's hot reloading in place, a change to a TS file will be immediately reflected in your browser. So as you make edits to files, the browser will update immediately with your changes
* Because of source maps, the JS files in your Chrome Devtools source match your local files exactly. No having to guess how compiled code maps back to source code.
* When using Workspaces, you can edit source files from within Chrome Devtools
* By adding a `.env` file to your root (see `.env.example` as a template), you can add variables like your Amplitude API Key so that you don't need to hardcode your API key in your code snippets. The API key can be referenced as `import.meta.env.VITE_AMPLITUDE_API_KEY`


daniel-graham-amplitude:
> Nice! Our team has a basic app (a bit more functionality than this though) that we use for testing. It's definitely good to have one setup though.

Yeah I would expect SR would need a pretty comprehensive server for testing. Do you have a link to that test server?

PR #1060, opened 2025-05-02
fix(analytics-browser): use performance.now in network capture

daniel-graham-amplitude:
### Summary

also, make it so `autocapture.networkTracking` defaults to `false`.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/default-tracking.ts
+++ b/packages/analytics-browser/src/default-tracking.ts
@@ -22,10 +22,12 @@ const isTrackingEnabled = (
     return autocapture;
   }
 
-  if (autocapture?.[event] === false) {
-    return false;
-  } else if (autocapture && !autocapture?.[event]) {
-    return defaultValue;
+  if (typeof autocapture === 'object') {
+    if (autocapture[event] === false) {
+      return false;
+    } else if (!autocapture[event]) {
+      return defaultValue;
+    }
   }
 
   return true;
--- a/packages/analytics-browser/src/default-tracking.ts
+++ b/packages/analytics-browser/src/default-tracking.ts
@@ -13,32 +13,38 @@ import {
  * Returns false if autocapture === false or if autocapture[event],
  * otherwise returns true
  */
-const isTrackingEnabled = (autocapture: AutocaptureOptions | boolean | undefined, event: keyof AutocaptureOptions) => {
+const isTrackingEnabled = (
+  autocapture: AutocaptureOptions | boolean | undefined,
+  event: keyof AutocaptureOptions,
+  defaultValue = false,
+) => {
   if (typeof autocapture === 'boolean') {
     return autocapture;
   }
 
   if (autocapture?.[event] === false) {
     return false;
+  } else if (autocapture && !autocapture?.[event]) {
+    return defaultValue;
   }
 
   return true;
 };
 
 export const isAttributionTrackingEnabled = (autocapture: AutocaptureOptions | boolean | undefined) =>
-  isTrackingEnabled(autocapture, 'attribution');
+  isTrackingEnabled(autocapture, 'attribution', true);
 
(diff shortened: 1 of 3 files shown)


packages/analytics-browser/src/default-tracking.ts
```diff
@@ -13,32 +13,40 @@ import {
  * Returns false if autocapture === false or if autocapture[event],
  * otherwise returns true
  */
```
Dogfalo, line 15: Update comment with new logic

packages/analytics-browser/src/default-tracking.ts
```diff
@@ -13,32 +13,40 @@ import {
...
+  event: keyof AutocaptureOptions,
+  defaultValue = false,
+) => {
   if (typeof autocapture === 'boolean') {
     return autocapture;
   }
 
-  if (autocapture?.[event] === false) {
-    return false;
+  if (typeof autocapture === 'object') {
```
Dogfalo, line 25: Although it's not somethign we expect, the below can error if autocapture is set to null since null is an object

packages/analytics-browser/src/default-tracking.ts
```diff
@@ -13,32 +13,40 @@ import {
...
+    if (autocapture[event] === false) {
+      return false;
+    } else if (!autocapture[event]) {
```
Dogfalo, line 28: Maybe simpler to read if we update to 
```
else if (autocapture[event] === undefined) {
```

packages/analytics-browser/src/default-tracking.ts
```diff
@@ -13,32 +13,40 @@ import {
...
 export const isPageViewTrackingEnabled = (autocapture: AutocaptureOptions | boolean | undefined) =>
-  isTrackingEnabled(autocapture, 'pageViews');
+  isTrackingEnabled(autocapture, 'pageViews', true);
 
 export const isSessionTrackingEnabled = (autocapture: AutocaptureOptions | boolean | undefined) =>
-  isTrackingEnabled(autocapture, 'sessions');
+  isTrackingEnabled(autocapture, 'sessions', true);
 
 export const isNetworkTrackingEnabled = (autocapture: AutocaptureOptions | boolean | undefined) =>
   isTrackingEnabled(autocapture, 'networkTracking');
```
Dogfalo, line 52: nit: Add false as default value since it's more explicit

packages/analytics-core/src/network-observer.ts
```diff
@@ -133,7 +133,7 @@ export class NetworkObserver {
     const originalFetch = this.globalScope.fetch;
 
     this.globalScope.fetch = async (input: RequestInfo | URL, init?: RequestInit) => {
-      const startTime = Date.now();
+      const startTime = performance.now();
```
Dogfalo, line 136: unfortunately performance.now isn't a drop-in replacement for Date.now().

I think performance.timeOrigin + performance.now() should be equivalent though. Some research and testing is needed here to confirm

Dogfalo (changes requested):

packages/analytics-core/src/network-observer.ts
Dogfalo: Not sure if the extra complexity is worth it overall to avoid issues with Date which relies on system clock, but I think the best way to do it is this:

Use Date.now for initial start time, but then use performance.now to measure the duration of the request. This is important because system clock updates won't affect this.

Then use start_time + duration to find the end time.

Note: this is all to avoid cases where a users system clock is changed during a request, either by the user or during automatic NTP clock synchronization

packages/analytics-browser/src/default-tracking.ts
daniel-graham-amplitude: Sounds good. Maybe just better to make it a non-optional parameter then.

packages/analytics-core/src/network-observer.ts
daniel-graham-amplitude: Yikes, good catch thanks.

PR #1062, opened 2025-05-05
feat(session-replay-browser): ugc removal poc

bravecod:
### Summary
This pull request introduces a new feature to support User-Generated Content (UGC) filtering in the Session Replay library. It includes changes to add UGC filter rules, apply them to page URLs, and validate their usage. The changes span across configuration, helper methods, hooks, and tests.

### UGC Filtering Feature Implementation

* **Configuration Updates**:
  - Added a new `UGCFilterRule` type to define UGC filter rules with `selector` and `replacement` properties. (`packages/session-replay-browser/src/config/types.ts`, [packages/session-replay-browser/src/config/types.tsR53-R66](diffhunk://#diff-3a86afe4928999add3708316114f2e24431702518dd9f124bee6b7df56299beaR53-R66))
  - Updated `SessionReplayLocalConfig` to include an optional `ugcFilterRules` property. (`packages/session-replay-browser/src/config/local-config.ts`, [[1]](diffhunk://#diff-b3985e000147ff08ef5d558589a5324db4905464ccf76e265ccf8eec81be9355R34) [[2]](diffhunk://#diff-b3985e000147ff08ef5d558589a5324db4905464ccf76e265ccf8eec81be9355R58)

* **Helper Functionality**:
  - Added `getPageUrl` helper to apply UGC filter rules to page URLs, with validation for rule structure and glob patterns. (`packages/session-replay-browser/src/helpers.ts`, [packages/session-replay-browser/src/helpers.tsR146-R193](diffhunk://#diff-0d86e594a4ecc24dccffd0a2a979b0904467a8a7c11d17b1dc9451c1b28a3530R146-R193))

* **Integration in Hooks**:
  - Updated `click` and `scroll` hooks to use `getPageUrl` for applying UGC filter rules to event data. (`packages/session-replay-browser/src/hooks/click.ts`, [[1]](diffhunk://#diff-d5419d6c240af4edb6c41bd0f11a1608663fa25f6fea9e3c4e9b1182848407cfR108-R117); `packages/session-replay-browser/src/hooks/scroll.ts`, [[2]](diffhunk://#diff-6701c003a923ff4315b6319297ddf6e8bfdecc5e60d8316e9c5ecb36397138b2L113-R119)

### Testing Enhancements

* **Unit Tests for `getPageUrl`**:
  - Validated scenarios such as no rules, invalid rules, and correct application of single/multiple rules. (`packages/session-replay-browser/test/helpers.test.ts`, [packages/session-replay-browser/test/helpers.test.tsR274-R373](diffhunk://#diff-654113a2278e3df354e180fdf4a8cf8872331058159f6c24c97e72850c969080R274-R373))

* **Integration Tests for Hooks**:
  - Ensured `click` and `scroll` events correctly apply UGC filter rules in various scenarios. (`packages/session-replay-browser/test/hooks/click.test.ts`, [[1]](diffhunk://#diff-b8dce375174146eab2ab8f528a6fbdecec905971bdffcf63205d0f3b5d07ac57R217-R248); `packages/session-replay-browser/test/hooks/scroll.test.ts`, [[2]](diffhunk://#diff-f447b26cebb602e4aaa13c839052771501e48afc2cebd9dc60ba62ce642b3502L99-R143)

These changes collectively enable UGC filtering for improved privacy and customization in session replay data.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/config/local-config.ts
+++ b/packages/session-replay-browser/src/config/local-config.ts
@@ -7,6 +7,7 @@ import {
   PrivacyConfig,
   SessionReplayPerformanceConfig,
   SessionReplayVersion,
+  UGCFilterRule,
 } from './types';
 import { SafeLoggerProvider } from '../logger';
 
@@ -30,6 +31,7 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
   storeType: StoreType;
   performanceConfig?: SessionReplayPerformanceConfig;
   experimental?: { useWebWorker: boolean };
+  ugcFilterRules?: UGCFilterRule[];
 
   constructor(apiKey: string, options: SessionReplayOptions) {
     const defaultConfig = getDefaultConfig();
@@ -53,6 +55,7 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
     this.version = options.version;
     this.performanceConfig = options.performanceConfig;
     this.storeType = options.storeType ?? 'idb';
+    this.ugcFilterRules = options.ugcFilterRules;
 
     if (options.privacyConfig) {
       this.privacyConfig = options.privacyConfig;
--- a/packages/session-replay-browser/src/helpers.ts
+++ b/packages/session-replay-browser/src/helpers.ts
@@ -1,6 +1,6 @@
-import { getGlobalScope, ServerZone } from '@amplitude/analytics-core';
+import { getGlobalScope, ILogger, ServerZone } from '@amplitude/analytics-core';
 import { getInputType } from '@amplitude/rrweb-snapshot';
-import { DEFAULT_MASK_LEVEL, MaskLevel, PrivacyConfig, SessionReplayJoinedConfig } from './config/types';
+import { DEFAULT_MASK_LEVEL, MaskLevel, PrivacyConfig, SessionReplayJoinedConfig, UGCFilterRule } from './config/types';
 import {
   KB_SIZE,
   MASK_TEXT_CLASS,
@@ -141,6 +141,54 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone, trackServerUr
   }
 
   return SESSION_REPLAY_SERVER_URL;
+};
+
+const globToRegex = (glob: string): RegExp => {
+  // Escape special regex characters, then convert globs
+  const escaped = glob
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&') // Escape regex specials
+    .replace(/\*/g, '.*') // Convert * to .*
+    .replace(/\?/g, '.'); // Convert ? to .
+
+  return new RegExp(`^${escaped}$`);
+};
+
+export const getPageUrl = (logger: ILogger, pageUrl: string, ugcFilterRules?: UGCFilterRule[]) => {
+  if (!ugcFilterRules) {
+    return pageUrl;
+  }
+
+  // validate ugcFilterRules
+  if (!ugcFilterRules.every((rule) => typeof rule.selector === 'string' && typeof rule.replacement === 'string')) {
+    logger.error('ugcFilterRules must be an array of objects with selector and replacement properties');
+    return pageUrl;
+  }
+
+  // validate ugcFilterRules are valid globs
+  if (
+    !ugcFilterRules.every((rule) => {
+      try {
+        new RegExp(rule.selector);
+        return true;
+      } catch (err) {
+        return false;
+      }
+    })
+  ) {
+    logger.error('ugcFilterRules must be an array of objects with valid globs');
+    return pageUrl;
+  }
+
+  // apply ugcFilterRules, order is important, first rule wins
+  for (const rule of ugcFilterRules) {
+    const regex = globToRegex(rule.selector);
+
+    if (regex.test(pageUrl)) {
+      return pageUrl.replace(regex, rule.replacement);
+    }
+  }
+
+  return pageUrl;
 };
 
 export const getStorageSize = async (): Promise<StorageData> => {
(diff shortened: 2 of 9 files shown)


packages/session-replay-browser/src/config/local-config.ts
```diff
@@ -30,6 +31,7 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
   storeType: StoreType;
   performanceConfig?: SessionReplayPerformanceConfig;
   experimental?: { useWebWorker: boolean };
+  ugcFilterRules?: UGCFilterRule[];
```
lewgordon-amplitude, line 34: I figured that this probably makes more sense in `interactionConfig` rather than a top level config since it really only applies to that.

lewgordon-amplitude: We should also probably add docs to the README.

packages/session-replay-browser/src/hooks/click.ts
```diff
@@ -102,14 +105,16 @@ export const clickHook: (logger: ILogger, options: Options) => mouseInteractionC
 
     const { left: scrollX, top: scrollY } = utils.getWindowScroll(globalScope as unknown as Window);
 
+    const maskedPageUrl = getPageUrl(logger, location.href, ugcFilterRules);
```
lewgordon-amplitude, line 108: Nit: Since this may or may not be masked.
```suggestion
    const pageUrl = getPageUrl(logger, location.href, ugcFilterRules);
```

packages/session-replay-browser/src/helpers.ts
```diff
@@ -143,6 +143,54 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone, trackServerUr
...
+  // Escape special regex characters, then convert globs
+  const escaped = glob
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&') // Escape regex specials
+    .replace(/\*/g, '.*') // Convert * to .*
+    .replace(/\?/g, '.'); // Convert ? to .
+
+  return new RegExp(`^${escaped}$`);
+};
+
+export const getPageUrl = (logger: ILogger, pageUrl: string, ugcFilterRules?: UGCFilterRule[]) => {
```
lewgordon-amplitude, line 156: Nit: It'd be nice to have ugcFilterRules be an empty array most of the time this way we don't always need to check to see if it's defined. Would it be hard to refactor so at the very beginning if it's undefined from local config it becomes an empty array?

packages/session-replay-browser/src/helpers.ts
lewgordon-amplitude, line 1: It'd be great to potentially move these functions out of this file. It's already a poorly named file and I put too much in here that I shouldn't have already.

packages/session-replay-browser/src/helpers.ts
```diff
@@ -143,6 +143,54 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone, trackServerUr
...
+  if (!ugcFilterRules) {
+    return pageUrl;
+  }
+
+  // validate ugcFilterRules
+  if (!ugcFilterRules.every((rule) => typeof rule.selector === 'string' && typeof rule.replacement === 'string')) {
+    logger.error('ugcFilterRules must be an array of objects with selector and replacement properties');
```
lewgordon-amplitude, line 163: I'm wondering if we should fail the call if we can't validate properly. This is intended to mask sensitive data, if there's an issue with the rules then we end up sending sensitive data rather than dropping it. I'd imagine in these cases people would want this data removed, so maybe it makes sense not to send them in the first place?

lewgordon-amplitude (commented):
Looks good overall and I think it's pretty close. Just wanted to discuss a few things before shipping this.

packages/session-replay-browser/src/helpers.ts
bravecod, line 163: validate this part at the beginning --> init

packages/session-replay-browser/src/helpers.ts
```diff
@@ -143,6 +143,54 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone, trackServerUr
...
 
+const globToRegex = (glob: string): RegExp => {
+  // Escape special regex characters, then convert globs
+  const escaped = glob
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&') // Escape regex specials
+    .replace(/\*/g, '.*') // Convert * to .*
+    .replace(/\?/g, '.'); // Convert ? to .
+
+  return new RegExp(`^${escaped}$`);
+};
```
bravecod, line 154: 1. (url) => url
2. filter rule set -> url ✔️

packages/session-replay-browser/src/helpers.ts
bravecod, line 163: throw an error to not send data to track this

packages/session-replay-browser/src/helpers.ts
```diff
@@ -143,6 +143,50 @@ export const getServerUrl = (serverZone?: keyof typeof ServerZone, trackServerUr
...
+
+export const validateUGCFilterRules = (ugcFilterRules: UGCFilterRule[]) => {
```
lewgordon-amplitude, line 156: Maybe this should return the sanitized filter rules? This way we don't need to build the regex every check.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -408,6 +409,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
             return;
           }
 
+          if (event.type === RRWebEventType.Meta) {
+            if (this.config?.interactionConfig?.enabled && this.config.interactionConfig.ugcFilterRules) {
+              event.data.href = getPageUrl(event.data.href, this.config.interactionConfig.ugcFilterRules);
```
lewgordon-amplitude: Can we determine if we should do this at the beginning of the function? That is something like:

```ts
const ugcFilterRules = this.config?.interactionConfig?.enabled ? this.config.interactionConfig.ugcFilterRules : []
...

if (event.type === RRWebEventType.Meta && ugcFilterRules) {
    event.data.href = getPageUrl(event.data.href, ugcFilterRules);
}
```

lewgordon-amplitude (commented):
One request then I think we're OK

PR #1071, opened 2025-05-09
fix(analytics-core): make network observer event callbacks handle exceptions

daniel-graham-amplitude:
### Summary

* If an event callback that's passed into "subscribe" throws an exception, window.fetch will fail to return the response. This adds a try-catch block to the event callback so that it still continues even in the event of an exception.
* also, make fix to start capturing Aborts too

### Checklist


packages/analytics-core/src/network-observer.ts
```diff
@@ -121,7 +121,11 @@ export class NetworkObserver {
 
   protected triggerEventCallbacks(event: NetworkRequestEvent) {
     this.eventCallbacks.forEach((callback) => {
-      callback.callback(event);
+      /* eslint-disable no-empty */
+      try {
+        callback.callback(event);
+      } catch (err) {}
+      /* eslint-enable no-empty */
```
Mercy811, line 128: Nit: can we log here? so that an exception won't be silence and causes confusion why callback is not called.

Mercy811 (approved):
Thanks Dan. LGTM

jxiwang (approved):

packages/analytics-core/src/network-observer.ts
daniel-graham-amplitude: Done!

PR #1074, opened 2025-05-09
refactor: make improvements to NetworkObserver

daniel-graham-amplitude:
### PROBLEMS
1. "networkObserver" does computations to calculate request body size and response body size. Although the computations are fairly basic, this could introduce a unnecessary overhead when an application does a very large amount of fetches **Performance issue**

2. our current "fetch" override implementation is too strongly typed and doesn't represent how it could be used in a browser (e.g.: we call input.toString() but input may be null or undefined when it's called in the wild) **bug**

3. our fetch override doesn't parse URL and HTTP Method when the first param is of type [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request), which has a ".url" and ".method" attribute. **bug**

4. restoring "fetch" to it's original fetch after unsubscribing is dangerous because if another library also overrided fetch after us, then it's implementation will get wiped out too. **bug**

5. the request "Headers" type was being cast as a Record<string, string> but it can also be a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance or an array of arrays (e.g.: `[[<header>, <value>],...]`)  **bug**

6. there's some redundancy in the "try" and "catch" blocks where 'originalFetch' is called (same code being called)

### SOLUTIONS
1. refactor "networkObserver" so that it doesn't calculate requestBodySize or responseBodySize, but rather the downstream subscribers can calculate those numbers, if they need them.

2. refactor the "fetch" override so that it accepts the 2 parameters as "optional" and then safely accesses everything (e.g.: replace `input.toString()` with `input?.toString?.()`

3. if the first parameter is of type Request, get URL from "input.url" and get method from "input.method". If not, fallback to url = "input?.toString?.()" and method = init.method (init is 2nd parameter) _If "input.method" and "init.method" are both set, default to "init.method" (as per the spec)_

4. don't restore fetch even when all callbacks are unsubscribed. Leave it in an overriden state so that we don't mess with others.

5. handle Arrays and Headers in the `get headers()` method of the request wrapper

6. move the NetworkRequestEvent creation after the original fetch is called, and then return/throw the original response/error that the original fetch returned/throws

## COMMENTS
* This implementation of the fetch override works on the assumption that everything could be undefined or throw an exception and we need to handle it.

e.g.) the fetch override looks like this (simplified)

```javascript
window.fetch = function (input, init) {
  try {
    beforeScript();
  } catch (e) {
    // suppress any error that beforeScript could possibly throw
  }
  
 let res, error;
  try {
    res = await originalFetch();
  } catch (e) {
    error = e;
  }
  
  try { 
    afterScript();
  } catch {}
    // suppress any error that afterScript could possibly throw
  }
  
  if (res) return res;
  else throw error;
```
* 

### Checklists


packages/analytics-core/src/network-request-event.ts
```diff
@@ -0,0 +1,198 @@
...
+        } else {
+          // encountered another unknown type
+          // we can't estimate the size of this entry
+          this._bodySize = undefined;
+          return;
+        }
+        // terminate if we reach the maximum number of entries
+        // to avoid performance issues in case of very large FormData
+        if (++count >= this.MAXIMUM_ENTRIES) {
+          this._bodySize = undefined;
```
jxiwang, line 103: This would essentially return undefined to the event property?

jxiwang (commented):
The performance concerns here are that the size calculation can bottleneck the request?

packages/analytics-core/src/network-request-event.ts
daniel-graham-amplitude: That's correct yes. Essentially it's "giving up" once it's reached the limit and not computing the size.

daniel-graham-amplitude:
> The performance concerns here are that the size calculation can bottleneck the request?

The size calculations and also the construction of the "requestHeaders".

`autocapture.networkTracking` filters out requests. This change restricts the calculation of the size to only the requests and responses that aren't being filtered. Which, in normal circumstances, would be >99% of requests (ie: requests with an HTTP status > 500)

packages/analytics-core/src/network-request-event.ts
```diff
@@ -0,0 +1,198 @@
...
+ */
+export class ResponseWrapper {
+  private _headers: Record<string, string> | undefined;
+  private _bodySize: number | undefined;
+  constructor(private response: Response) {}
+
+  get headers(): Record<string, string> {
+    if (this._headers) return this._headers;
+    const headers: Record<string, string> = {};
+    this.response.headers.forEach((value, key) => {
```
jxiwang, line 141: I think this still works, but do we want any additional header checks as well to be safe?

packages/analytics-core/src/network-observer.ts
```diff
@@ -109,91 +43,159 @@ export class NetworkObserver {
...
+      }
 
-        this.triggerEventCallbacks(requestEvent);
-        throw error;
+      // 4. return the original response or throw the original error
+      if (originalResponse) {
+        // if the response is not undefined, return it
+        return originalResponse;
+      } else {
+        throw originalError;
```
jxiwang, line 198: This is the original error thrown by the wrapped response?

packages/analytics-core/src/network-request-event.ts
daniel-graham-amplitude: According to [specification](https://developer.mozilla.org/en-US/docs/Web/API/Response/headers#value), a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object is the only type that is set in the `Response.headers` object so we shouldn't expect there to ever be a Record or Array type (like what request has).

packages/analytics-core/src/network-observer.ts
daniel-graham-amplitude: That is correct yes. This is where the original error gets caught

```typescript
// 2. make the call to the original fetch and preserve the response or error
let originalResponse, originalError;
try {
  originalResponse = await originalFetch(requestInfo as RequestInfo | URL, requestInit);
} catch (err) {
  // Capture error information
  originalError = err;
}
```

It's essential that `originalResponse` is the only object that is ever returned and that `originalError` is the only erro that is ever thrown.

jxiwang (approved):
Overall LGTM! Thanks for doing this! As long as we are minimizing the effect on the customer's network requests, then we should be good.

packages/analytics-core/src/network-request-event.ts
daniel-graham-amplitude: Yeah good thinking. I'll add a defensive check to it.

daniel-graham-amplitude:
> Overall LGTM! Thanks for doing this! As long as we are minimizing the effect on the customer's network requests, then we should be good.

Thanks! Yeah I'm mostly concerned about preserving fetch and making it future-proof so that nobody can break it (easily).

PR #1080, opened 2025-05-12
feat(analytics-browser): observe XHR along with "fetch"

daniel-graham-amplitude:
### Summary
* Adding support to `XMLHttpRequest` so we get full coverage of both "fetch" and "XMLHttpRequest"
* Allow setting `config.autocapture.networkTracking` as a NetworkTrackingOptions object. Deprecate `config.networkTrackingOptions` in favour of that
* Also, added some more safety checks and tests that ensure the code is safe now and safe from being meddled with in the future

## Comments
* Added tons of type safety. It may seem unreasonable, but considering the risks of overriding "fetch" and "XMLHttpRequest" it's warranted. Both to prevent defects/misuse now, and to prevent future maintainers from shooting themselves in the foot
* Added tests to ensure type safety
* The "fetch.html" and "xhr.html" files are there as tests to run in the browser

### Checklist


daniel-graham-amplitude:
> Where is `instrumentXhrAndFetchBeta`? I didn't a search on the PR and repo but didn't find it. Is that part of the public SDK config?

My mistake, I forgot to push it. 🤦 

It's there now.

packages/analytics-core/src/network-request-event.ts
```diff
@@ -65,58 +136,94 @@ export class RequestWrapper {
...
+  let body;
+  if (typeof bodyUnsafe === 'string') {
+    body = bodyUnsafe;
+    bodySize = new TextEncoder().encode(body).length;
+  } else if (bodyUnsafe instanceof Blob) {
+    body = bodyUnsafe as BlobSafe;
+    bodySize = bodyUnsafe.size;
+  } else if (bodyUnsafe instanceof URLSearchParams) {
+    body = bodyUnsafe as URLSearchParamsSafe;
+    bodySize = new TextEncoder().encode(bodyUnsafe.toString()).length;
```
jxiwang, line 175: Should this be
`new TextEncoder().encode(body.toString()).length;`?

packages/analytics-core/src/network-request-event.ts
```diff
@@ -65,58 +136,94 @@ export class RequestWrapper {
...
+  if (!TextEncoder) {
+    return;
+  }
+  let body;
+  if (typeof bodyUnsafe === 'string') {
+    body = bodyUnsafe;
+    bodySize = new TextEncoder().encode(body).length;
+  } else if (bodyUnsafe instanceof Blob) {
+    body = bodyUnsafe as BlobSafe;
+    bodySize = bodyUnsafe.size;
```
jxiwang, line 172: Should this be `body.size` ?

packages/analytics-core/src/network-request-event.ts
```diff
@@ -65,58 +136,94 @@ export class RequestWrapper {
...
+  } else if (bodyUnsafe instanceof URLSearchParams) {
+    body = bodyUnsafe as URLSearchParamsSafe;
+    bodySize = new TextEncoder().encode(bodyUnsafe.toString()).length;
+  } else if (ArrayBuffer.isView(bodyUnsafe)) {
+    body = bodyUnsafe as ArrayBufferViewSafe;
```
jxiwang, line 177: Maybe we do `bodySafe` here like below? Same with the rest of the typings.

jxiwang (approved):

packages/analytics-core/src/network-observer.ts
```diff
@@ -43,15 +80,25 @@ export class NetworkObserver {
     }
     this.eventCallbacks.set(eventCallback.id, eventCallback);
     if (!this.isObserving) {
+      /* istanbul ignore next */
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      const originalXhrOpen = this.globalScope?.XMLHttpRequest?.prototype?.open;
+      /* istanbul ignore next */
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      const originalXhrSend = this.globalScope?.XMLHttpRequest?.prototype?.send;
```
Mercy811, line 88: Looks like we never restore to origin. I understand that override won't persist when page reload (jump to another website). But just thinking about some edge cases here.

Mercy811 (approved):
Overall LGTM. Thanks @daniel-graham-amplitude.

test-server/browser-sdk/xhr.html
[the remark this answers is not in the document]
daniel-graham-amplitude: There's `network-observer.test.ts` which gives pretty good coverage. But it doesn't make actual network calls.

Do we have integration tests currently? I would love to have a place where I can make these calls in an automated way.

packages/analytics-core/src/network-observer.ts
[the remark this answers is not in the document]
daniel-graham-amplitude: Good thinking! Done.

PR #1081, opened 2025-05-13
chore(session replay): add metadata and config to rrweb events

jxiwang:
### Summary

Adding metadata event that propagates as an rrweb event. This will be processed to be used later as metadata information including versions and debugging. 

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -250,7 +250,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
   focusListener = () => {
     // Restart recording on focus to ensure that when user
     // switches tabs, we take a full snapshot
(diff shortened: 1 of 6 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -233,7 +247,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
   focusListener = () => {
     // Restart recording on focus to ensure that when user
     // switches tabs, we take a full snapshot
-    void this.recordEvents();
+    void this.recordEvents(false);
```
jxiwang, line 250: Metadata shouldn't change here, so reducing amount of logs.

lewgordon-amplitude (approved):
LGTM, this will be very useful to have. Thanks for this one!

PR #1083, opened 2025-05-14
 fix(plugin-page-view-tracking-browser): do not restore history.pushState on teardown

daniel-graham-amplitude:
### Summary

When we cal teardown on page view plugin, `window.history.pushState` is restored to whatever it was before we overwrote it. The problem with this is that if another plugin also overwrites `window.history.pushState` after we do, and then we restore it, their implementation will get overwritten to.

This PR stops the restoration of `pushState` and what it does instead is adds a flag `isTracking` that needs to be `true` for page tracking to work. So when "setup" is called, `isTracking=true`. When "teardown" is called", "isTracking=false" again and then the `history.pushState` callback becomes essentially a "no-op"

Also added a page to test-server that lets you manually test this.

### Checklist


Mercy811 (approved):
Thanks Dan. LGTM. Add the background here https://amplitude.slack.com/archives/C073N5DK70A/p1746782171920649.

PR #1088, opened 2025-05-16
chore: add tests for isNetworkTrackingEnabled

daniel-graham-amplitude:
### Checklist


Mercy811 (commented):
Can we also test when autocapture is undefined? False positive is more scary than false negative.

Mercy811 (approved):

PR #1099, opened 2025-05-22
fix(session-replay): map the type to the correct sdk plugin

jxiwang:
### Summary

Correct type mapping for SDK Plugins that are not Amplitude specific such as Segment.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -105,7 +105,15 @@ export class SessionReplay implements AmplitudeSessionReplay {
     );
     this.config = joinedConfig;
 
-    this.setMetadata(options.sessionId, joinedConfig, localConfig, remoteConfig, options.version?.version, VERSION);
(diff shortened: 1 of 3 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -524,13 +532,26 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
   }
 
+  private mapSDKType(sdkType: string | undefined) {
+    if (sdkType === 'plugin') {
+      return '@amplitude/plugin-session-replay-browser';
+    }
+
+    if (sdkType === 'segment') {
+      return '@amplitude/segment-session-replay-plugin';
+    }
```
junjie-amplitude, line 542: Can we return "unknown" when the type is an unknown valid string? Otherwise, return null.

junjie-amplitude (approved):
LGTM! Thanks for the fix!

PR #1100, opened 2025-05-23
fix(analytics-browser): prevent infinite Amplitude network requests

daniel-graham-amplitude:
### Summary

If network tracking is configured to include amplitude.com and allows status code 200, then it can cause infinite network request calls because it would trigger an Amplitude event, and then that would trigger a network request callback, which would trigger an Amplitude event...

This checks the events, and if it contains '[Amplitude] Network Request' as an event_type, then do not track it

### Checklist


Mercy811 (commented):
Should we ignore all requests to amplitude http v2 endpoint?

daniel-graham-amplitude:
> Should we ignore all requests to amplitude http v2 endpoint?

I think just requests that have `[Amplitude] Network Request` in their response. It's still useful for us, especially internally, to find errors from those endpoints.

Mercy811 (approved):
Thanks Dan. LGTM

PR #1106, opened 2025-05-27
chore: add plugin-stub-browser

daniel-graham-amplitude:
### Summary

Adding a new plugin called "plugin-stub-browser". It doesn't do anything and isn't meant to be used. It's here so that if you ever need to create a new plugin, you just copy and paste this, and replace "stub" with the name of your plugin, and you're up-and-running quickly.

### Checklist

### DIFF
--- a/packages/analytics-core/src/storage/cookie.ts
+++ b/packages/analytics-core/src/storage/cookie.ts
@@ -16,17 +16,17 @@ export class CookieStorage<T> implements Storage<T> {
     }
 
     CookieStorage.testValue = String(Date.now());
-    const testStrorage = new CookieStorage<string>(this.options);
+    const testStorage = new CookieStorage<string>(this.options);
     const testKey = 'AMP_TEST';
     try {
-      await testStrorage.set(testKey, CookieStorage.testValue);
-      const value = await testStrorage.get(testKey);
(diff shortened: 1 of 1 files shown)


packages/plugin-stub-browser/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "author": "Amplitude Inc",
+  "homepage": "https://github.com/amplitude/Amplitude-TypeScript",
+  "license": "MIT",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "lib/esm/index.d.ts",
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public",
+    "tag": "beta"
```
Mercy811, line 14: should it be alpha to align with its version?

packages/plugin-stub-browser/package.json
Mercy811, line 1: If it's not aimed to be used and is just a template to start with, we should not publish it to npm

Mercy811 (changes requested):

packages/plugin-stub-browser/package.json
daniel-graham-amplitude: Updated, thanks!

PR #1110, opened 2025-05-28
AMP-130401 create previous page page url plugin

cyhk:
### Checklist

Yes

### DIFF
--- a/packages/analytics-browser/src/plugins/page-url-previous-page.ts
+++ b/packages/analytics-browser/src/plugins/page-url-previous-page.ts
@@ -1,0 +1,143 @@
+import { BrowserClient, BrowserConfig, EnrichmentPlugin, Event, Logger } from '@amplitude/analytics-types';
+import { getGlobalScope } from '@amplitude/analytics-client-common';
+
+enum PreviousPageType {
+  Direct = 'direct', // for no prev page or referrer
+  Internal = 'internal', // for same domain - this excludes subdomains
+  External = 'external', // for different domains
+}
+
+type AdditionalEventProperties = {
+  '[Amplitude] Page Domain': string;
+  '[Amplitude] Page Location': string;
+  '[Amplitude] Page Path': string;
+  '[Amplitude] Page Title': string;
+  '[Amplitude] Page URL': string;
+  '[Amplitude] Previous Page Location': string;
+  '[Amplitude] Previous Page Type': PreviousPageType;
+};
+
+const defaultPageViewEvent = '[Amplitude] Page Viewed';
+const defaultElementClickedEvent = '[Amplitude] Element Clicked';
+
+type Options = {
+  restrictToAutocapture?: boolean;
+};
+
+export const pageUrlPreviousPagePlugin = ({ restrictToAutocapture = false }: Options): EnrichmentPlugin => {
+  const globalScope = getGlobalScope();
+  let loggerProvider: Logger | undefined = undefined;
+  let pushState: undefined | ((data: any, unused: string, url?: string | URL | null) => void);
+
+  const getDecodeURI = (locationStr: string): string => {
+    let decodedLocationStr = locationStr;
+    try {
+      decodedLocationStr = decodeURI(locationStr);
+    } catch (e) {
+      /* istanbul ignore next */
+      loggerProvider?.error('Malformed URI sequence: ', e);
+    }
+
+    return decodedLocationStr;
+  };
+
+  const getPrevPageType = (previousPage: string) => {
+    const currentDomain = (typeof location !== 'undefined' && location.hostname) || '';
+    const previousPageDomain = previousPage ? new URL(previousPage).hostname : undefined;
+
+    switch (previousPageDomain) {
+      case undefined:
+        return PreviousPageType.Direct;
+      case currentDomain:
+        return PreviousPageType.Internal;
+      default:
+        return PreviousPageType.External;
+    }
+  };
+
+  const getAdditionalEventProperties = (): AdditionalEventProperties => {
+    const locationHREF = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
+    const sessionStorage = getGlobalScope()?.sessionStorage;
+    const previousPage = sessionStorage?.getItem('previousPage') || document.referrer || '';
+
+    return {
+      '[Amplitude] Page Domain':
+        /* istanbul ignore next */ (typeof location !== 'undefined' && location.hostname) || '',
+      '[Amplitude] Page Location': locationHREF,
+      '[Amplitude] Page Path':
+        /* istanbul ignore next */ (typeof location !== 'undefined' && getDecodeURI(location.pathname)) || '',
+      '[Amplitude] Page Title': /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
+      '[Amplitude] Page URL': locationHREF.split('?')[0],
+      '[Amplitude] Previous Page Location': previousPage,
+      '[Amplitude] Previous Page Type': getPrevPageType(previousPage),
+    };
+  };
+
+  const trackURLChange = () => {
+    const sessionStorage = getGlobalScope()?.sessionStorage;
+    if (!sessionStorage) {
+      return;
+    }
+
+    const previousURL = sessionStorage.getItem('currentPage') || '';
+    const currentURL = (typeof location !== 'undefined' && location.href) || '';
+
+    sessionStorage.setItem('previousPage', previousURL);
+    sessionStorage.setItem('currentPage', currentURL);
+  };
+
+  const plugin: EnrichmentPlugin = {
+    name: '@amplitude/plugin-referral-page-url-browser',
+    type: 'enrichment',
+
+    setup: async (config: BrowserConfig, _: BrowserClient) => {
+      loggerProvider = config.loggerProvider;
+      loggerProvider.log('Installing @amplitude/plugin-previous-page-page-url-browser');
+
+      if (globalScope) {
+        globalScope.addEventListener('popstate', trackURLChange);
+
+        // Save reference to original push state, to be used in teardown
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        pushState = globalScope.history.pushState;
+
+        /* istanbul ignore next */
+        // There is no global browser listener for changes to history, so we have
+        // to modify pushState directly.
+        // https://stackoverflow.com/a/64927639
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        globalScope.history.pushState = new Proxy(globalScope.history.pushState, {
+          apply: (target, thisArg, [state, unused, url]) => {
+            target.apply(thisArg, [state, unused, url]);
+            void trackURLChange();
+          },
+        });
+      }
+    },
+    execute: async (event: Event) => {
+      if (
+        restrictToAutocapture &&
+        (event.event_type === defaultPageViewEvent || event.event_type === defaultElementClickedEvent)
+      ) {
+        return event;
+      }
+
+      event.event_properties = {
+        ...event.event_properties,
+        ...getAdditionalEventProperties(),
+      };
+
+      return event;
+    },
+    teardown: async () => {
+      if (globalScope) {
+        globalScope.removeEventListener('popstate', trackURLChange);
+        if (pushState) {
+          globalScope.history.pushState = pushState;
+        }
+      }
+    },
+  };
+
+  return plugin;
+};
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -27,6 +27,7 @@ import {
   getPageViewTrackingConfig,
   getElementInteractionsConfig,
   getNetworkTrackingConfig,
+  getPageUrlPreviousPageConfig,
   isAttributionTrackingEnabled,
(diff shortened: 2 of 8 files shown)


cyhk:
> ### Bug: Invalid URL Handling Causes Function Failure
> The `getPrevPageType` function lacks error handling for invalid URLs. The `new URL(previousPage)` constructor throws an error if `previousPage` (sourced from `document.referrer` or `sessionStorage`) is malformed or invalid, causing the function and potentially the plugin to crash.
> 
> `packages/analytics-browser/src/plugins/page-url-previous-page.ts#L27-L40`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7ImJ1ZyI6ImV5SnNiMk5oZEdsdmJuTWlPbHQ3SW1acGJHVWlPaUp3WVdOcllXZGxjeTloYm1Gc2VYUnBZM010WW5KdmQzTmxjaTl6Y21NdmNHeDFaMmx1Y3k5d1lXZGxMWFZ5YkMxd2NtVjJhVzkxY3kxd1lXZGxMblJ6SWl3aWMzUmhjblJNYVc1bElqb3lOeXdpWlc1a1RHbHVaU0k2TkRBc0ltTnZaR1ZNYVc1bGN5STZXMTE5WFN3aWFXUWlPaUl6T0RRNU5qQXdZeTFqTVRsaUxUUTJNRE10WVRjNE9TMW1ZemRqT1RRd05EZzBObVVpTENKa1pYTmpjbWx3ZEdsdmJpSTZJbFJvWlNCZ1oyVjBVSEpsZGxCaFoyVlVlWEJsWUNCbWRXNWpkR2x2YmlCc1lXTnJjeUJsY25KdmNpQm9ZVzVrYkdsdVp5Qm1iM0lnYVc1MllXeHBaQ0JWVWt4ekxpQlVhR1VnWUc1bGR5QlZVa3dvY0hKbGRtbHZkWE5RWVdkbEtXQWdZMjl1YzNSeWRXTjBiM0lnZEdoeWIzZHpJR0Z1SUdWeWNtOXlJR2xtSUdCd2NtVjJhVzkxYzFCaFoyVmdJQ2h6YjNWeVkyVmtJR1p5YjIwZ1lHUnZZM1Z0Wlc1MExuSmxabVZ5Y21WeVlDQnZjaUJnYzJWemMybHZibE4wYjNKaFoyVmdLU0JwY3lCdFlXeG1iM0p0WldRZ2IzSWdhVzUyWVd4cFpDd2dZMkYxYzJsdVp5QjBhR1VnWm5WdVkzUnBiMjRnWVc1a0lIQnZkR1Z1ZEdsaGJHeDVJSFJvWlNCd2JIVm5hVzRnZEc4Z1kzSmhjMmd1SWl3aVkyOXVabWxrWlc1alpTSTZNQzQ0TnpVeE9EQXhNemd6TXpFd01Ua3NJbU5oZEdWbmIzSjVJam9pVUU5VVJVNVVTVUZNWDBWRVIwVmZRMEZUUlNKOSIsImJyYW5jaCI6IkFNUC0xMzA0MDEtY3JlYXRlLXByZXZpb3VzLXBhZ2UtcGFnZS11cmwtcGx1Z2luIn0sImlhdCI6MTc0OTc5MDI4NSwiZXhwIjoxNzUwMzk1MDg1fQ.EqvtnKszQ8byHGQ8SK2NBrzW6raxjpemHSItN7F_kI9g6s_5mRAI8E65Bb3UbzMHIhC0F5E8ovIqnJuI5T6ekzCl05w1KDnosJr8Cj6UpJ8HqvNLUI6oRwDk9IcbfZ3NbRjtsuGPeWN0c_M5UFT6jnGZrdljLQqtQWoe2Why-NLJdHcUN6cX3j3HV_Lfn4RT_we66qAZtufiwJfsIPsYP9naXV4IXbCFTPp6PMvL8GekW9W6XjLx8neubZ4AxqYi3bSjIs1AVLjxkwRFk3xiQlrzk6kCjndlzZjp83gUWHEen-yDJX9MueRWfuYOFeZliImxgRMKxaonItUj6Vo1tA)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

this is fine since `currentURL` is never stored without first being run through `getDecodeURI` first

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,177 @@
...
+      /* istanbul ignore next */
+      loggerProvider?.error('Malformed URI sequence: ', e);
+    }
+
+    return decodedLocationStr;
+  };
+
+  const getPrevPageType = (previousPage: string) => {
+    const currentDomain = (typeof location !== 'undefined' && location.hostname) || '';
+    const previousPageDomain = previousPage ? new URL(previousPage).hostname : undefined;
```
daniel-graham-amplitude: `new URL` can throw an exception here if `previousPage` is malformed (or even if it's a relative URL). I would suggest catching the errors and logging a message.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,177 @@
...
+      case undefined:
+        return PreviousPageType.Direct;
+      case currentDomain:
+        return PreviousPageType.Internal;
+      default:
+        return PreviousPageType.External;
+    }
+  };
+
+  const trackURLChange = () => {
```
daniel-graham-amplitude: nit: the word "track" here made me think of "amplitude.track". Maybe something like "saveUrlChange" or "saveUrlInfo"?

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,177 @@
...
+
+          isProxied = true;
+        }
+      }
+    },
+    execute: async (event: Event) => {
+      const locationHREF = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
+
+      if (globalScope) {
+        const sessionStorage = globalScope.sessionStorage;
```
daniel-graham-amplitude: There's a wrapper class you can use here called [BrowserStorage](https://github.com/amplitude/Amplitude-TypeScript/blob/codex/update-globalscope.addeventlistener-error-handling/packages/analytics-browser/src/storage/session-storage.ts#L4) that encapsulates sessionStorage (and localStorage). It catches errors, has a check if it is enabled (so you can exit out if it isn't), and let's you set objects.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,177 @@
...
+  };
+
+  const trackURLChange = () => {
+    if (globalScope) {
+      const sessionStorage = globalScope.sessionStorage;
+      if (sessionStorage) {
+        const previousURL = getFromStorage(sessionStorage, 'currentPage', loggerProvider) || '';
+        const currentURL = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
+        setInStorage(sessionStorage, 'previousPage', previousURL, loggerProvider);
+        setInStorage(sessionStorage, 'currentPage', currentURL, loggerProvider);
```
daniel-graham-amplitude: There's a tiny risk that our customers websites could be using "previousPage" or "currentPage" and this would overwrite it. Perhaps a prefix would help here "AMP_".

I would also consider using [Browser Storage](https://github.com/amplitude/Amplitude-TypeScript/blob/codex/update-globalscope.addeventlistener-error-handling/packages/analytics-browser/src/storage/session-storage.ts#L4) and setting the entry as one single JSON like `{previousPage, currentPage}`.

daniel-graham-amplitude:
Thanks @cyhk ! Generally looks, had some comments.

General question though: was there a formal process for deciding the namings of the config and events?

For the config (`autocapture. pageUrlPreviousPage`) and for the event properties (`[Amplitude] Previous Page URL`, etc...). 

I'm asking just because API changes (in this case the config definition) and event properties can't be reverted once they've been released.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
cyhk: Nice, I'll use that instead!

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,175 @@
...
+    type: 'enrichment',
+
+    setup: async (config: BrowserConfig, _: BrowserClient) => {
+      loggerProvider = config.loggerProvider;
+      loggerProvider.log('Installing @amplitude/plugin-page-url-previous-page-browser');
+
+      isTracking = true;
+
+      if (globalScope) {
+        sessionStorage = new BrowserStorage<URLInfo>(globalScope.sessionStorage);
```
Mercy811: I'm thinking of cookie storage as a better option here as session storage is clean when a new tab opens, for example, the previous url of google.com is lost when i click on a search result.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        globalScope.addEventListener('popstate', saveUrlInfoWrapper);
+
+        if (!isProxied) {
+          /* istanbul ignore next */
+          // There is no global browser listener for changes to history, so we have
+          // to modify pushState and replaceState directly.
+          // https://stackoverflow.com/a/64927639
+          // eslint-disable-next-line @typescript-eslint/unbound-method
+          globalScope.history.pushState = new Proxy(globalScope.history.pushState, {
```
Mercy811: Note this only works for SPA. How do we track non-SPA?

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,175 @@
...
+        }
+      }
+    },
+    execute: async (event: Event) => {
+      const locationHREF = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
+
+      let previousPage = '';
+      if (sessionStorage && isStorageEnabled) {
+        const URLInfo = await sessionStorage.get(URL_INFO_STORAGE_KEY);
+        previousPage = URLInfo?.[PREVIOUS_PAGE_STORAGE_KEY] || document.referrer || '';
```
Mercy811: May result in different previous page location and referrer event properties and thus cause confusion.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
```diff
@@ -0,0 +1,175 @@
...
+          '[Amplitude] Page Location': locationHREF,
+          '[Amplitude] Page Path':
+            /* istanbul ignore next */ (typeof location !== 'undefined' && getDecodeURI(location.pathname)) || '',
+          '[Amplitude] Page Title':
+            /* istanbul ignore next */ (typeof document !== 'undefined' && document.title) || '',
+          '[Amplitude] Page URL': locationHREF.split('?')[0],
+          '[Amplitude] Previous Page Location': previousPage,
+          '[Amplitude] Previous Page Type': getPrevPageType(previousPage),
+        };
+      }
```
Mercy811: Any COGS concern @izaaz if we add extra event properties to all events? 

Any plan to add a filter to apply to specific events? I'm thinking about `config.autocapture.pageUrlPreviousPage.trackOn`, a function returning boolean.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
Mercy811, line 1: I know form interaction and file download plugins are in analytics-browser, but can we make page url previous page plugin as a separate plugin as plugin-page-view-tracking-browser as well? In general, we want to keep analytics-browser clean and modular.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
Dogfalo: For non-SPA we're using document.referrer.

packages/analytics-browser/src/plugins/page-url-previous-page.ts
Dogfalo: In this case, we utilize document.referrer. Also with cookie storage, it would get overwritten with multiple tabs

packages/analytics-browser/src/plugins/page-url-previous-page.ts
cyhk: Alan Okada spoke with Data Pipeline team and there were no particular problems raised with these new additions.

Dogfalo (approved):

daniel-graham-amplitude (approved):

packages/analytics-core/src/types/config/browser-config.ts
```diff
@@ -178,6 +178,11 @@ export interface AutocaptureOptions {
    * @experimental This feature is experimental and may not be stable
    */
   webVitals?: boolean;
+  /**
+   * Enables/disables adding additional referrer and page url event properties.
+   * @defaultValue `false`
+   */
+  pageUrlEnrichment?: boolean;
```
Mercy811, line 185: can we add `@experimental` tag here like `webVitals` above in case of api change?

PR #1127, opened 2025-06-03
chore: enhance publish workflow with from-package option

Mercy811:
### Summary

- Force to fail the job when there is no changed package to version 
- Adds `publishFrom` input parameter to support both `from-git` (default) and `from-package` publishing modes. The version creation step is now conditional and only runs for `from-git` mode.

### Checklist
* [x] Does your PR title have the correct title format?
* [ ] * Does your PR have a breaking change?: No

### DIFF
--- a/.github/workflows/publish-v2.yml
+++ b/.github/workflows/publish-v2.yml
@@ -85,7 +85,7 @@ jobs:
       # Only create release version when using from-git (default behavior)
       # from-package mode uses existing package.json versions and doesn't need git tags
       - name: Create release version
-        if: env.PUBLISH_FROM != 'from-package'
+        if: ${{ env.PUBLISH_FROM != 'from-package' }}
         run: |
           echo "Running lerna version..."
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-v2.yml
```diff
@@ -3,13 +3,10 @@ name: Publish v2.x
...
-      releaseType:
-        type: choice
-        description: Release Type
-        options:
-          - release
-          - prerelease
-          - graduate
+      publishFrom:
+        type: string
+        description: Publish source (leave empty for from-git, or enter "from-package")
```
jxiwang, line 8: Instead of `string` type, would `choice` be better since there is only 2? Default can be `from-git`.

Mercy811: I wanna raise the bar to use 'from-package' just like when you delete a github repo you have to input the repo name. But i'm open to discussion.

jxiwang: Ah gotcha, is the thought process to gate users from just selecting one and moving forward?

Mercy811: Yes!

jxiwang (approved):

daniel-graham-amplitude (approved):
Lgtm

PR #1133, opened 2025-06-04
chore: add element interactions performance test

daniel-graham-amplitude:
### Summary

add an element interactions test to compare performance difference when autocapture is enabled

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/e2e/autocapture.spec.ts
+++ b/packages/plugin-autocapture-browser/e2e/autocapture.spec.ts
@@ -1,0 +1,66 @@
+import { test, expect, Page, BrowserContext } from '@playwright/test';
+
+const TBT_DIFF_THRESHOLD = 200;
+
+async function runFormTest(page: Page, context: BrowserContext, browserName: string, skipAutocapture = false) {
+  // Enable CPU throttling through CDP only for Chromium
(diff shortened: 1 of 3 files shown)


packages/plugin-autocapture-browser/e2e/autocapture.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { test, expect, Page, BrowserContext } from '@playwright/test';
+
+const TBT_DIFF_THRESHOLD = 200;
```
Mercy811, line 3: how is this value decided? For performance I was thinking of giving a report of web vitals with autocapture on/off. But generally, I don't have any objection of adding this as an test. At least, it serves as an example of showing that the impact of autocapture is fairly small.

Mercy811 (approved):

packages/plugin-autocapture-browser/e2e/autocapture.spec.ts
daniel-graham-amplitude: I decided on this value because in my tests, the maximum difference I observed, on several retries, was about 150 ms.

I think the purpose of this test would be that we want to know if autocapture ever causes big performance degradation, we want this test to fail, so that we'd know.

PR #1146, opened 2025-06-10
DRAFT: feat(analytics-browser): add dead click tracking to autocapture

daniel-graham-amplitude:
### Summary

Add dead click tracking to element interactions. 

### How it works

When a user sets a "deadClickAllowlist" (currently defaults to `[]` until we decide on configuration) any elements that match that list, when clicked on, will fire a dead click event if a navigation event or a mutation event does not happen within 3 seconds.

(This works the opposite of "action" click which only triggers if a mutation/navigation event happens with 500ms)

### Checklist


daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude:
bugbot run

packages/analytics-core/src/types/frustration-interactions.ts
```diff
@@ -0,0 +1,132 @@
...
+  'input[type="reset"]',
+  'input[type="image"]',
+  'input[type="file"]',
+  ...CLICKABLE_ELEMENT_SELECTORS,
+];
+
+/**
+ * Default CSS selectors for rage clicks tracking
+ */
+export const DEFAULT_RAGE_CLICK_ALLOWLIST = [
```
jxiwang: I believe rage clicks are usually captured on any element, not just interactive elements.

packages/plugin-autocapture-browser/src/autocapture/track-dead-click.ts
```diff
@@ -0,0 +1,90 @@
...
+    throttleTime(DEAD_CLICK_TIMEOUT),
+  );
+
+  return actionClicks.subscribe((actionClick) => {
+    const deadClickEvent: EventDeadClick = {
+      X: (actionClick.event as MouseEvent).clientX,
+      Y: (actionClick.event as MouseEvent).clientY,
+    };
+    /* istanbul ignore next */
+    amplitude?.track(
```
jxiwang: How come this is optional?

packages/plugin-autocapture-browser/src/frustration-plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+import { AllWindowObservables, ObservablesEnum } from './autocapture-plugin';
+import { createClickObservable, createMutationObservable } from './observables';
+
+type BrowserEnrichmentPlugin = EnrichmentPlugin<BrowserClient, BrowserConfig>;
+
+export const frustrationPlugin = (options: FrustrationInteractionsOptions): BrowserEnrichmentPlugin => {
+  const name = constants.FRUSTRATION_PLUGIN_NAME;
+  const type = 'enrichment';
+
+  // TODO: add visualTagging stuff here too
```
jxiwang: Is this required to release?

packages/plugin-autocapture-browser/src/autocapture/track-dead-click.ts
daniel-graham-amplitude: I'm not sure, I just copied and pasted that from "track-click" which did that too.

packages/analytics-core/src/types/frustration-interactions.ts
daniel-graham-amplitude: It think different vendors do it differently. Sentry does it on [buttons](https://amplitude.com/docs/data/visual-labeling). FullStory does it on anything.

I'm thinking capture it on any element is the way to go. I'll just change this allowlist to `['*']` because it should probably still be configurable.

packages/plugin-autocapture-browser/src/frustration-plugin.ts
daniel-graham-amplitude: I'm going to confirm with the team, but I don't think so. Seems like an advanced feature. I'll take the comment out, but still raise the conversation.

packages/analytics-core/src/types/frustration-interactions.ts
```diff
@@ -0,0 +1,123 @@
...
+  '[role="option"]',
+  '[role="tab"]',
+  '[role="treeitem"]',
+  '[contenteditable="true" i]',
+];
+
+/**
+ * Default CSS selectors for dead clicks tracking
+ */
+export const DEFAULT_DEAD_CLICK_ALLOWLIST = [
```
Dogfalo, line 88: Should we have the default dead click allowlist be a copy of the CSS_SELECTOR_ALLOWLIST?

packages/analytics-core/src/types/frustration-interactions.ts
```diff
@@ -0,0 +1,123 @@
...
+  ...CLICKABLE_ELEMENT_SELECTORS,
+];
+
+/**
+ * Default CSS selectors for rage clicks tracking
+ */
+export const DEFAULT_RAGE_CLICK_ALLOWLIST = ['*'];
+
+/**
+ * Default time window for dead clicks (1 second)
```
Dogfalo: 3 second*

packages/plugin-autocapture-browser/src/autocapture/track-dead-click.ts
```diff
@@ -0,0 +1,90 @@
...
+  /* istanbul ignore next */
+  if (navigateObservable) {
+    changeObservables.push(navigateObservable);
+  }
+  const mutationOrNavigate = merge(...changeObservables);
+
+  const actionClicks = filteredClickObservable.pipe(
+    mergeMap((click) => {
+      // Create a timer that emits after 500ms
+      let timeoutId: NodeJS.Timeout;
```
Dogfalo: ReturnType<typeof setTimeout> probably better since this isn't a NodeJS environment

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
```diff
@@ -0,0 +1,90 @@
+import { AllWindowObservables } from 'src/autocapture-plugin';
+import { filter, map, bufferTime } from 'rxjs';
+import { BrowserClient } from '@amplitude/analytics-core';
+import { filterOutNonTrackableEvents, shouldTrackEvent } from '../helpers';
+import { AMPLITUDE_ELEMENT_RAGE_CLICKED_EVENT } from '../constants';
+
+let RAGE_CLICK_THRESHOLD = 5;
+let RAGE_CLICK_WINDOW_MS = 3000; // 3 seconds
```
Dogfalo: Seems like we should be able to use the defaults we defined above?
nit: 3_000 for readability.

packages/plugin-autocapture-browser/src/frustration-plugin.ts
Dogfalo: What is meant by visual Tagging stuff here?

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -327,6 +341,18 @@ export const addAdditionalEventProperties = <T>(
   };
 
   if (isElementBasedEvent(baseEvent) && baseEvent.event.target !== null) {
+    if (isCapturingCursorPointer) {
```
Dogfalo, line 348: Can you help add comments to explain this a bit more? I'm not entirely understanding what we're trying to accomplish

packages/plugin-autocapture-browser/src/frustration-plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const subscriptions: Subscription[] = [];
+
+  const rageCssSelectors = options.rageClicks?.cssSelectorAllowlist ?? DEFAULT_RAGE_CLICK_ALLOWLIST;
+  const deadCssSelectors = options.deadClicks?.cssSelectorAllowlist ?? DEFAULT_DEAD_CLICK_ALLOWLIST;
+
+  /* istanbul ignore next */
```
Dogfalo: use of istanbul ignore next seems a bit liberal. Are these all non-testable or is this temporary?

packages/plugin-autocapture-browser/src/frustration-plugin.ts
daniel-graham-amplitude: Element Interactions supports `visual tagging`: https://github.com/amplitude/Amplitude-TypeScript/blob/AMP-125613/add-dead-clicks-to-element-interactions/packages/analytics-core/src/types/element-interactions.ts#L72

I left this TODO here in-case we want to add `visual tagging` on a future iteration.

packages/analytics-core/src/types/frustration-interactions.ts
daniel-graham-amplitude: They're not the exact same list. `CSS_SELECTOR_ALLOWLIST` has `input`. `DEFAULT_DEAD_CLICK_ALLOWLIST` has only clickable "input" types (button, submit, reset, image, file).

What I could do is make a common default called something like:

```
export const CLICKABLE_ELEMENT_SELECTORS = [
  'a',
  'button',
  'select',
  'textarea',
  'label',
  'video',
  'audio',
  '[contenteditable="true" i]',
  '[data-amp-default-track]',
  '.amp-default-track',
];
```

And then

`CSS_SELECTOR_ALLOWLIST = ['input', ...CLICKABLE_ELEMENT_SELECTORS]`

```
DEFAULT_DEAD_CLICK_ALLOWLIST = [
  ...CLICKABLE_ELEMENT_SELECTORS,
   '[role="button"]',
  '[role="link"]',
  '[role="menuitem"]',
  '[role="menuitemcheckbox"]',
  '[role="menuitemradio"]',
  '[role="option"]',
  '[role="tab"]',
  '[role="treeitem"]',
  'input[type="button"]',
  'input[type="submit"]',
  'input[type="reset"]',
  'input[type="image"]',
  'input[type="file"]',
].
```

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
daniel-graham-amplitude: Which defaults are those?

packages/plugin-autocapture-browser/src/frustration-plugin.ts
daniel-graham-amplitude: I just got rid of a bunch. Some of them were unnecessary. Some them I added tests to cover them.

The only ones remaining now are istanbul ignores that were copy-pasted over from "autocapture-plugin.ts".

packages/plugin-autocapture-browser/src/helpers.ts
daniel-graham-amplitude: Added a comment, but not here, it's in the function parameter definition.

I'll paste it here for convenience:

```
// capture the event if the cursor is a "pointer" when this element was clicked on
// reason: a "pointer" cursor indicates that an element should be interactable
//         regardless of the element's tag name
```

packages/plugin-autocapture-browser/src/autocapture/track-dead-click.ts
daniel-graham-amplitude: I'm going to change this to just `amplitude.track`. There's no reason why `amplitude` should not be there.

daniel-graham-amplitude:
I just added an E2E test fix that's unrelated to this. We were testing a page that makes calls to https://httpstat.us which is a 3rd party endpoint and it was down. This fix mocks calls to that endpoint so we don't need to rely on it being available (and makes it more performant because no HTTP roundtrips!)

packages/plugin-autocapture-browser/src/frustration-plugin.ts
jxiwang: For this iteration, we can leave this out for now.

packages/analytics-core/src/types/frustration-interactions.ts
Dogfalo: Right, that's what I'm saying. I think they should be the same list in almost all cases.
If the CSS_SELECTOR_ALLOWLIST is saying, this is the list of elements that are interactable and we want to track clicks on it. Then it would make sense that the dead click tracks those same elements when it was clicked and there was no other recorded change afterwards.

I was thinking of this:
```js
if (options.rageClicks && options.rageClicks.cssSelectorAllowlist === undefined) {
  options.rageClicks.cssSelectorAllowlist = options.cssSelectorAllowlist
}
```

What are you thoughts?

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
Dogfalo: https://github.com/amplitude/Amplitude-TypeScript/pull/1146/files#diff-a7e1313731c1686ec0074f9a85b61664cb26fe4b04838503d30bf432fdf79adbR115
`export const DEFAULT_RAGE_CLICK_THRESHOLD = 5;`

packages/analytics-core/src/types/frustration-interactions.ts
daniel-graham-amplitude: The issue is that the `cssSelectorAllowlist` is part of the `autocaptureInteractions` scope which is separate from `frustrationInteractions`.

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
daniel-graham-amplitude: Aha I understand now thanks.

packages/analytics-core/src/types/frustration-interactions.ts
Dogfalo: What do you mean by scope here?

daniel-graham-amplitude: The config (which is still open for revision) looks like this:

```
{
  elementInteractions: {
    cssSelectorAllowlist: [...],
  },
  frustrationInteractions: {
    deadClicks: {
      cssSelectorAllowlist: [...]
    },
  },
}
```

It would mean that the config for `frustrationInteractions` would reach into the `elementInteractions` config.

jxiwang (approved):

packages/analytics-core/src/types/frustration-interactions.ts
Dogfalo: I don't see that as a problem since they're all under autocapture. It could be argued that deadClicks and rage clicks could live inside elementInteractions too

daniel-graham-amplitude: We had considered putting them under elementInteractions but decided not too because we wanted a low friction way of opting into rage/dead clicks (by setting `autocapture.frustrationInteractions = true`) without having to also opt into `elementInteractions`.

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
daniel-graham-amplitude: Fix pushed.

Dogfalo (approved):

PR #1152, opened 2025-06-12
fix: add more attributes to web-vitals

daniel-graham-amplitude:
### Summary
* Add "page" information to web-vitals events
* Add "navigationStart" to metrics (along with "timestamp" which was already there). And switched to Epoch

Extra: fixed "watch" script. Make it use Nx instead to fix problem.

### Checklist


Mercy811 (approved):
LGTM

PR #1156, opened 2025-06-13
feat(session replay): dynamically import rrweb

jxiwang:
### Summary

Only import rrweb package when recording.

- Removes any instance of the main `rrweb` package and imports the `rrweb-record` package.
- Move certain utils that are not exposed in `rrweb` into Amplitude-Typescript (getWindowHeight, etc).
- Import `rrweb-utils` instead of `rrweb` for util functions.

### Checklist

### DIFF
--- a/packages/session-replay-browser/README.md
+++ b/packages/session-replay-browser/README.md
@@ -95,6 +95,17 @@ sessionReplay.shutdown()
 |`performanceConfig.timeout`|`number`|No|`undefined`|Optional timeout in milliseconds for the `requestIdleCallback` API. If specified, this value will be used to set a maximum time for the browser to wait before executing the deferred compression task, even if the browser is not idle.|
 |`experimental.useWebWorker`|`boolean`|No|`false`|If the SDK should compress the replay events using a webworker.|
 
+## Bundle Size Optimization
+The Session Replay SDK uses dynamic imports to optimize bundle size and improve initial page load performance. Key modules are loaded on-demand rather than being included in the initial bundle:
+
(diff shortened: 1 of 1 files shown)


packages/session-replay-browser/src/utils/rrweb.ts
```diff
@@ -0,0 +1,76 @@
+import { getGlobalScope } from '@amplitude/analytics-core';
+import dom from '@amplitude/rrweb-utils';
+import type { eventWithTime, scrollCallback } from '@amplitude/rrweb-types';
+
+// These functions are not exposed in rrweb package, so we will define it here to use
+// Ignoring this function since this is copied from rrweb
+/* istanbul ignore next */
```
jxiwang, line 7: Pulling this from `rrweb`, so ignoring from tests

lewgordon-amplitude: I think I get the reason why, but feel free to correct me. Is the reason we need to separate these out because we need to have these functions before we dynamically pull the rrweb-record?

lewgordon-amplitude (commented):
I think overall the changes look good and make sense. I just did have a question before we merge this. Does this impact customers pulling from npm/yarn? I know we recommend customers to add our CDN to their CSP, but would this break existing customers not doing this? Say if they bundle this and host it from their own CDN.

jxiwang:
@lewgordon-amplitude This wouldn't break any existing customers. For the use case of uploading to their CDN, are you saying that they would bundle this package themselves and upload to their own CDN provider?

packages/session-replay-browser/src/utils/rrweb.ts
jxiwang: We're importing the entire `rrweb` package just for this function. I think this is primarily used for the scroll events.

lewgordon-amplitude (approved):
LGTM!

PR #1159, opened 2025-06-17
refactor: make autocapture functions re-usable

daniel-graham-amplitude:
### Summary
* make "addAdditionalEventProperties" and "getEventProperties" re-usable functions
* make the document.click and mutation observers re-usable global observers

### Why
This is to set up a subsequent PR to add the frustration plugin. Without this refactor, the subsequent PR would be very large and unruly.

### Extra
Also added a test page to the test server for testing element interactions

### Checklist


daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude:
bugbot rn

packages/plugin-autocapture-browser/src/observables.ts
```diff
@@ -0,0 +1,38 @@
...
+import { share } from 'rxjs/operators';
+
+let mutationObservableInstance: Observable<MutationRecord[]> | null = null;
+let clickObservableInstance: Observable<MouseEvent> | null = null;
+
+/**
+ * Creates an observable that tracks DOM mutations on the document body.
+ * This observable is created lazily and shared across different plugins to avoid creating multiple mutation observers.
+ */
+export const getGlobalMutationObservable = (): Observable<MutationRecord[]> => {
```
Dogfalo: I would question whether this is actually necessary to try and share it since most of the performance impact is due to what happens inside the event callback function rather than the observer of the mutations itself. Also we wouldn't want to create cross plugin dependencies where some other plugin depends on autocapture being available.

daniel-graham-amplitude: Sounds good, yeah I can change it so that creates a new one each time. I wasn't sure if we should be limiting mutation observers or not.

I'll also update the click observer.

Dogfalo (approved):

daniel-graham-amplitude:
@Dogfalo I just merged in main and it changed quite a bit. Probably needs another review. Sorry for the inconvenience.

Dogfalo:
@daniel-graham-amplitude looks good to me still 👍

daniel-graham-amplitude:
bugbot run

PR #1160, opened 2025-06-18
test: add base tests e2e playwright tests for session-replay-browser

cely404:
### Summary
Add base tests e2e playwright tests for session-replay-browser

### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -11,7 +11,9 @@ on:
 jobs:
   e2e:
     name: E2E Tests
-    runs-on: ops-amd64-general
+    runs-on:
+      group: ops-amd64-general
+      labels: large
 
     steps:
       - name: Check out git repository


packages/session-replay-browser/e2e/integration.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    await page.waitForTimeout(2000); // 2s pause
+    await expect(eventLog).toContainText('form_submitted');
+    await page.waitForTimeout(2000); // 2s pause
+
+    // 5. Properties verification with pacing
+    const propertiesDisplay = page.locator('#properties-display');
+    await page.click('#get-current-properties');
+    await page.waitForTimeout(2000); // 2s pause
+    await expect(propertiesDisplay).toContainText('[Amplitude] Session Replay ID');
+    await page.waitForTimeout(2000); // 2s pause
```
lewgordon-amplitude, line 105: It would be nice to output the replay ID or even the URL so we can go right to the replay. The URL format is `https://apps.stag2.amplitude.com/analytics/<org_url>/session-replay/project/<project_id>/search/replay?sessionReplayId=<sessionReplayId>`.

cely404:
> LGTM this will definitely be useful to have. Any thoughts on moving the JS for the test server to typescript or even using a simple React/NextJS app here?

I went back and forth on this a lot. I ultimately did not go with a simple react app because I didn't think we gained much from it. The goal here is to test either the sr browser sdk alone or integrate it with the analytics browser sdk and Im not quite sure how much more we benefit from having a full react app. 
It would make sense if we need to have a more complicated app to test on but as of now a basic html seemed like it would take us pretty far. WDYT? I can also bring this up to the mobile team to get their general thoughts on it.

test-server/session-replay/js/sdk-init.js
```diff
@@ -0,0 +1,91 @@
...
+        updateSessionReplayStatus(StatusType.COMPLETE);
+    } catch (error) {
+        console.error('Session Replay SDK initialization error:', error);
+        updateSessionReplayStatus(StatusType.ERROR);
+        throw error; // Re-throw to prevent Analytics SDK initialization
+    }
+}
+
+// Save the original fetch
+const originalFetch = window.fetch;
```
cely404, line 64: This is potentially overkill but we are basically testing that EVERY single request succeeds including config, sr tracking and event tracking

lewgordon-amplitude:
> > LGTM this will definitely be useful to have. Any thoughts on moving the JS for the test server to typescript or even using a simple React/NextJS app here?
> 
> I went back and forth on this a lot. I ultimately did not go with a simple react app because I didn't think we gained much from it. The goal here is to test either the sr browser sdk alone or integrate it with the analytics browser sdk and Im not quite sure how much more we benefit from having a full react app. It would make sense if we need to have a more complicated app to test on but as of now a basic html seemed like it would take us pretty far. WDYT? I can also bring this up to the mobile team to get their general thoughts on it.

I just figured that it'd be easier for us to maintain and it's a slightly more realistic app that customers would use.

PR #1162, opened 2025-06-23
feat(analytics-browser): add dead + rage clicks to analytics-browser

daniel-graham-amplitude:
### Summary

You can now capture dead clicks and rage clicks via a new configuration called "autocapture.frustrationInteractions".

### Checklist


packages/analytics-core/src/types/browser-config.ts
```diff
@@ -160,6 +161,11 @@ export interface AutocaptureOptions {
    * @defaultValue `false`
    */
   elementInteractions?: boolean | ElementInteractionsOptions;
+  /**
+   * Enables/disables frustration interactions tracking.
+   * @defaultValue `false`
```
Mercy811, line 166: based on the code it looks like the default value is true?

Mercy811 (approved):
LGTM! See nit.

packages/analytics-core/src/types/browser-config.ts
daniel-graham-amplitude: It should be "false" by default. Just like elementInteractions. Is it not?

daniel-graham-amplitude: I implemented "isFrustrationInteractionsEnabled" the same as "isElementInteractionsEnabled" so it should be the same

PR #1163, opened 2025-06-24
ci: add single package publish workflow

Mercy811:
This workflow allows publishing individual packages with the following features:
- Package name input (e.g., unified, analytics-browser)
- Release tag selection (beta, alpha)
- Dry run mode for testing without actual publishing
- Automatic version bumping using semantic versioning
- Changelog updates and npm publishing
- Package privacy management (temporary public, restore to private)

The workflow includes proper authorization checks, builds, tests, and cleanup steps to ensure safe and reliable package publishing.

### Checklist

### DIFF
--- a/.github/workflows/publish-single-package.yml
+++ b/.github/workflows/publish-single-package.yml
@@ -105,12 +105,11 @@ jobs:
           CURRENT_VERSION=$(npm pkg get version | tr -d '"')
           echo "Current version: $CURRENT_VERSION"
           
-          # Determine version bump type based on release tag
+          # Set version bump type and determine preid based on release tag
+          BUMP_TYPE="prerelease"
           if [ "$RELEASE_TAG" = "alpha" ]; then
-            BUMP_TYPE="prerelease"
             PREID="alpha"
           else
-            BUMP_TYPE="prerelease" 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-single-package.yml
```diff
@@ -0,0 +1,244 @@
...
+      - name: Calculate new version
+        run: |
+          cd $PACKAGE_PATH
+          # Get current version
+          CURRENT_VERSION=$(npm pkg get version | tr -d '"')
+          echo "Current version: $CURRENT_VERSION"
+          
+          # Determine version bump type based on release tag
+          if [ "$RELEASE_TAG" = "alpha" ]; then
+            BUMP_TYPE="prerelease"
```
daniel-graham-amplitude: nit: seems like you can just set `BUMP_TYPE="prerelease"` above the "if" tag because it's always "prerelease" regardless of condition.

daniel-graham-amplitude (approved):
Lgtm. Let's test this right away though.

.github/workflows/publish-single-package.yml
Mercy811: Good catch. Updated.

PR #1168, opened 2025-06-25
feat(autocapture): fetch page actions from remote config

Dogfalo:
### Summary

- Remote config for key `analyticsSDK.pageActions` is fetched now after plugin initialization, this requires `fetchRemoteConfig` to be enabled.
- It is designed to be non-blocking and doesn't slow down initialization of the plugin.
- Fixed an error where the wrong operator name was used for checking Element Text in the Labeled Event definition evaluation

### Checklist


packages/plugin-autocapture-browser/src/autocapture-plugin.ts
```diff
@@ -180,12 +182,53 @@ export const autocapturePlugin = (options: ElementInteractionsOptions = {}): Bro
...
+      })
+        .then(async (remoteConfigFetch) => {
+          try {
+            const remotePageActions = await remoteConfigFetch.getRemoteConfig('analyticsSDK', 'pageActions');
+            recomputePageActionsData(remotePageActions as ElementInteractionsOptions['pageActions']);
+          } catch (error) {
+            // Log error but don't fail the setup
+            /* istanbul ignore next */
+            config?.loggerProvider?.error(`Failed to fetch remote config: ${String(error)}`);
+          }
```
daniel-graham-amplitude, line 198: Nitpick: this try -> catch block isn't necessary. If the error gets thrown from inside of 'then' it will be caught by 'catch'.

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
```diff
@@ -180,12 +182,53 @@ export const autocapturePlugin = (options: ElementInteractionsOptions = {}): Bro
...
 
+    // Fetch remote config for pageActions in a non-blocking manner
+    if (config.fetchRemoteConfig) {
+      createRemoteConfigFetch({
+        localConfig: config,
+        configKeys: ['analyticsSDK.pageActions'],
+      })
+        .then(async (remoteConfigFetch) => {
+          try {
+            const remotePageActions = await remoteConfigFetch.getRemoteConfig('analyticsSDK', 'pageActions');
```
daniel-graham-amplitude, line 192: Does the remote config endpoint allow us to return `analyticsSDK: { browserSDK, pageActions}` instead of calling these 2 separately?

packages/plugin-autocapture-browser/src/pageActions/triggers.ts
```diff
@@ -74,8 +92,37 @@ export const matchLabeledEventsToTriggers = (labeledEvents: LabeledEvent[], leTo
...
+      event,
+      Array.from(groupedLabeledEvents[event.type]).map((id) => pageActions.labeledEvents[id]),
+    );
+    // Find matching conditions
+    const matchingTriggers = matchLabeledEventsToTriggers(matchingLabeledEvents, labeledEventToTriggerMap);
+    for (const trigger of matchingTriggers) {
+      executeActions(trigger.actions, event);
+    }
+
+    return event;
```
daniel-graham-amplitude, line 128: Could this function be moved into a helper function? It looks like it does the same thing as `evaluateTriggers` from `autocapture-plugin.ts`

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
```diff
@@ -180,12 +182,53 @@ export const autocapturePlugin = (options: ElementInteractionsOptions = {}): Bro
...
+            recomputePageActionsData(remotePageActions as ElementInteractionsOptions['pageActions']);
```
daniel-graham-amplitude, line 193: Just to confirm, this means that:
1. if the fetch fails, it will fallback to the hardcoded configuration?
2. up until the fetch is complete, it's going to use the hardcoded config?

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
Dogfalo: Currently in ananytics-browser if it is fetched at that point it causes initialization to be delayed which we want to avoid.

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
Dogfalo: Yes. Although for more context for this feature, I would say that hardcoded local configuration is very unlikely to be used at all due to complexity of configuration. It can be helpful for debugging/development though since remote config requires upto 10 mins to take effect.

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
Dogfalo: It is actually catching errors at different points, the .catch() is only handling promise rejects for the `createRemoteConfigFetch` function, whereas the try..catch is preventing any failed fetches and malformed responses from causing unexpected behavior

Dogfalo: Looking at createRemoteConfigFetch a bit closer, I'm not entirely sure if that needs to be a promise, but there may have been some reason in the past. Since that remote config client will be replaced in the future I think it's better to not make changes to it though.

packages/plugin-autocapture-browser/src/pageActions/triggers.ts
Dogfalo: Yes, thought I actually removed the other `evaluateTriggers` function already. Refactored it so its behavior actually a bit more clear using a class now

daniel-graham-amplitude (approved):

Dogfalo:
bugbot run

PR #1172, opened 2025-06-26
feat(session-replay-react-native) add @amplitude/session-replay-react-native package

aliaksandr-kazarez:
### Summary
Adds beta-1 version of `@amplitude/session-replay-react-native` package

### Checklist

* [x] Does your PR have a breaking change?: no  
* [x] Updated README.md with new usage
* [x] Improve Unit tests


aliaksandr-kazarez:
bugbot run

aliaksandr-kazarez:
bugbot run

aliaksandr-kazarez:
bugbot review

packages/session-replay-react-native/ios/RCTAmpMaskViewManager.m
```diff
@@ -0,0 +1,29 @@
...
+  return [[RCTView alloc] init];
+}
+
+RCT_CUSTOM_VIEW_PROPERTY(mask, NSString, RCTView)
+{
+  NSString* mask = [RCTConvert NSString:json];
+  if ([mask isEqualToString:@"amp-mask"]) {
+    view.amp_isBlocked = true;
+  } else if ([mask isEqualToString:@"amp-block"]) {
+    view.amp_isBlocked = true;
```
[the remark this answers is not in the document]
aliaksandr-kazarez: this logic were there for quite some time, don't need to change it

aliaksandr-kazarez:
bugbot run

packages/session-replay-react-native/android/src/main/AndroidManifestNew.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
crleona, line 1: needed?

packages/session-replay-react-native/ios/ConsoleLogger.swift
```diff
@@ -0,0 +1,49 @@
...
+@objc(AMPLogLevel)
+public enum LogLevelEnum: Int {
+    case OFF
+    case ERROR
+    case WARN
+    case LOG
+    case DEBUG
+}
+
+public class ConsoleLogger: CoreLogger {
```
crleona, line 15: pull this in from `AmplitudeCore`, no need to repeat it here

packages/session-replay-react-native/ios/NativeSessionReplay.swift
```diff
@@ -0,0 +1,149 @@
+import Foundation
+import AmplitudeSessionReplay
+
+@objc(NativeSessionReplay)
```
crleona: I realize we haven't don't this everywhere, but let's try to prefix objc classes with `AMP`

crleona (approved):

packages/session-replay-react-native/ios/NativeSessionReplay.swift
aliaksandr-kazarez: updated

packages/session-replay-react-native/ios/ConsoleLogger.swift
aliaksandr-kazarez: What might be the impact of pulling in logger from AmplitudeCore over cocoapods?

packages/session-replay-react-native/android/src/main/AndroidManifestNew.xml
aliaksandr-kazarez: That's needed for project to work with Gradle 7.3.0, which is required to support ReactNative <=0.72.
I believe at this point we're supporting 0.69+

packages/session-replay-react-native/ios/ConsoleLogger.swift
crleona: It's already a dependency, though indirectly via SR, so I wouldn't expect any changes.

Thinking about this a bit more - longer term we'll probably move to some shared logger across all amplitude products on a particular platform.  does it make sense to bridge logging back to the RN side so we can support that?

aliaksandr-kazarez: If we want to do something with the logs, then yes, otherwise I'll probably try to avoid unnecessary data travels between native and JS, as this adds performance overhead, which could be quite significant for Old Architecture, less noticeable for the new architecture, but still.

crleona: makes sense - can tackle when we figure out the use case a bit more.

PR #1175, opened 2025-06-30
fix(session-replay-browser): reduces network calls to remote config from many to 1

lewgordon-amplitude:
### Summary

We were previously making a remote call for each key of the config. This changes it to only look up the namespace and fetch the keys in JS.

Tested using tampermonkey:

image

### Checklist

### DIFF
--- a/packages/analytics-remote-config/src/remote-config.ts
+++ b/packages/analytics-remote-config/src/remote-config.ts
@@ -34,11 +34,10 @@ export class RemoteConfigFetch<RemoteConfig extends { [key: string]: object }>
     this.configKeys = configKeys;
   }
 
-  getRemoteConfig = async <K extends keyof RemoteConfig>(
+  getRemoteNamespaceConfig = async (
     configNamespace: string,
-    key: K,
     sessionId?: number | string,
-  ): Promise<RemoteConfig[K] | undefined> => {
+  ): Promise<RemoteConfig | undefined> => {
     const fetchStartTime = Date.now();
     // Finally fetch via API
     const configAPIResponse = await this.fetchWithTimeout(sessionId);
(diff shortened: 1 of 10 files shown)


packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -56,43 +56,29 @@ export class SessionReplayJoinedConfigGenerator {
...
-        'sessionReplay',
-        'sr_logging_config',
-        sessionId,
-      );
-
-      if (samplingConfig || privacyConfig) {
-        remoteConfig = {};
-        if (samplingConfig) {
-          remoteConfig.sr_sampling_config = samplingConfig;
+      const namespaceConfig = await this.remoteConfigFetch.getRemoteNamespaceConfig('sessionReplay', sessionId);
```
jpollock-ampl, line 59: :100:

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/config/joined-config.ts
jpollock-ampl: Non-blocking, but wanted to flag - if this call fails, we lose _all_ remote config.  I do not think this is significantly different from the previous implementation as all of the separate fetches were in the same try/catch.

lewgordon-amplitude: Yeah, exactly.

packages/analytics-remote-config/src/remote-config.ts
```diff
@@ -34,25 +34,33 @@ export class RemoteConfigFetch<RemoteConfig extends { [key: string]: object }>
...
         this.metrics.fetchTimeAPISuccess = Date.now() - fetchStartTime;
-        return remoteConfig[key];
+        return remoteConfig;
       }
     }
     this.metrics.fetchTimeAPIFail = Date.now() - fetchStartTime;
     return undefined;
   };
 
+  getRemoteConfig = async <K extends keyof RemoteConfig>(
```
jpollock-ampl, line 55: Non-blocking - Given `getRemoteNamespaceConfig()`, when would we use this?  Perhaps we should try to deprecate this to avoid the issue we are fixing in this PR.  

Alternatively, we could accept a list of keys, (`keys: K[]`) to allow this method to extract several keys from a single call.

lewgordon-amplitude: I wanted to keep it since analytics-browser-sdk currently still uses this I believe.

lewgordon-amplitude: I do like the idea to extract several keys though. I would say that this is a quicker fix until we move to the new remote client.

jpollock-ampl (approved):

PR #1176, opened 2025-06-30
perf(session replay): use text length instead of blob for size calc

jxiwang:
### Summary

Switching from Blob size calculation to string length. Tested manually with some performance traces to determine if any changes can be seen and also with a test file (ty Cursor).

[compression-comparison.txt](https://github.com/user-attachments/files/21007431/compression-comparison.txt)

```
Original event: 2203 chars
Compressed event: 54 chars
Compression ratio: 40.80x

Original Event:
  String Length: 0.15ms
  TextEncoder:   4.00ms
  Blob:          28.73ms
  String vs TextEncoder: 27.4x slower
  String vs Blob:        196.6x slower

Compressed Event:
  String Length: 0.04ms
  TextEncoder:   1.42ms
  Blob:          23.39ms
  String vs TextEncoder: 37.7x slower
  String vs Blob:        621.6x slower

=== ACCURACY COMPARISON: When String Length vs Byte Size Matters ===

asciiOnly: 1100 chars, 1100 bytes (0 diff, 0.0%)
unicodeLight: 400 chars, 500 bytes (+100 diff, 25.0%)
unicodeHeavy: 600 chars, 1200 bytes (+600 diff, 100.0%)
mixedContent: 1150 chars, 1300 bytes (+150 diff, 13.0%)
```

### Length 
shouldSplitEvents_encoder

### Blob 
shouldSplitEvents_blob

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/events/base-events-store.ts
+++ b/packages/session-replay-browser/src/events/base-events-store.ts
@@ -55,7 +55,7 @@ export abstract class BaseEventsStore<KeyType> implements EventsStore<KeyType> {
       totalSize += this.getStringSize(event);
     }
 
-    // Account for JSON array serialization overhead:
+    // Additional overhead from using length instead of byte size
     // - Array brackets: [] = 2 characters
     // - Commas between events: events.length - 1
     // - Double quotes around each event: events.length * 2


packages/session-replay-browser/src/events/base-events-store.ts
```diff
@@ -37,23 +37,51 @@ export abstract class BaseEventsStore<KeyType> implements EventsStore<KeyType> {
...
+
+  /**
+   * Calculates the total character length of events array
+   */
+  private getEventsArraySize(events: Events): number {
+    let totalSize = 0;
+    for (const event of events) {
+      totalSize += this.getStringSize(event);
+    }
+    return totalSize;
```
lewgordon-amplitude: You may also want to assume additional bytes for `,` and `[`, `]`. The commas in particular should just add `totalSize += events.length`.

jxiwang: Thanks, good catch!

PR #1179, opened 2025-07-01
feat(session-replay-browser): allows users to omit scripts and comments from being captured

lewgordon-amplitude:
### Summary

Currently in shadow root contexts script tags may show "SCRIPT_PLACEHOLDER" which is not ideal. This is due to `<script>` tags being captured and replaced with this text. On replay rrweb injects `display: none !important`, but is not added to shadow roots sometimes causing the text to appear.

This can be fixed in the player too, but it's good to have this option if we need it later on. I changed the name of the option on our side since I thought it was clearer, but happy to discuss that one!

[SR-1374](https://amplitude.atlassian.net/browse/SR-1374)

### Checklist

[SR-1374]: https://amplitude.atlassian.net/browse/SR-1374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/plugin-session-replay-browser/src/session-replay.ts
+++ b/packages/plugin-session-replay-browser/src/session-replay.ts
@@ -77,6 +77,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin<BrowserClient, Brow
         performanceConfig: this.options.performanceConfig,
         storeType: this.options.storeType,
         experimental: this.options.experimental,
+        omitElementTags: this.options.omitElementTags,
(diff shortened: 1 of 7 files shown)


jpollock-ampl (approved):
>  I changed the name of the option on our side since I thought it was clearer, but happy to discuss that one!

I like the name you chose.

PR #1194, opened 2025-07-04
fix: reduce redundant array creation in getHierarchy

daniel-graham-amplitude:
### Checklist


packages/plugin-autocapture-browser/test/hierarchy.test.ts
```diff
@@ -1,6 +1,64 @@
...
+  //   //   return {
+  //   //     port1: {
+  //   //       onmessage: () => {},
+  //   //     },
+  //   //     port2: {
+  //   //       postMessage: () => {},
+  //   //     },
+  //   //   };
+  //   // };
+  // });
```
Mercy811, line 61: Can we just delete it? 
```suggestion
```

PR #1203, opened 2025-07-09
feat: targeting replay capture

bravecod:
### Summary

This pull request introduces several enhancements to the `plugin-session-replay-browser` package, focusing on improving the handling of user properties, integrating new session replay functionality, and adding comprehensive unit tests. The most important changes include the addition of the `parseUserProperties` helper function, updates to session replay integration, and expanded test coverage.

### Enhancements to user property handling:
- **`packages/plugin-session-replay-browser/src/helpers.ts`:** Added the `parseUserProperties` function to extract and process user properties based on specific `IdentifyOperation` types. This function ensures only supported operations are included in the resulting object.
- **`packages/plugin-session-replay-browser/src/constants.ts`:** Defined `PROPERTY_ADD_OPERATIONS` to list supported `IdentifyOperation` types for user property handling.

### Updates to session replay integration:
- **`packages/plugin-session-replay-browser/src/session-replay.ts`:** Enhanced the `SessionReplayPlugin` class to include new methods like `evaluateTargetingAndCapture` and updated session replay initialization options to include `userProperties`. [[1]](diffhunk://#diff-3fe57928f5097a7852ccabdf51be26876203c64fb472c02ce246578cf1a6faa7L1-L5) [[2]](diffhunk://#diff-3fe57928f5097a7852ccabdf51be26876203c64fb472c02ce246578cf1a6faa7L16-R35) [[3]](diffhunk://#diff-3fe57928f5097a7852ccabdf51be26876203c64fb472c02ce246578cf1a6faa7R67-R68) [[4]](diffhunk://#diff-3fe57928f5097a7852ccabdf51be26876203c64fb472c02ce246578cf1a6faa7R94) [[5]](diffhunk://#diff-3fe57928f5097a7852ccabdf51be26876203c64fb472c02ce246578cf1a6faa7R161-R165)
- **`packages/plugin-session-replay-browser/package.json`:** Added new dependencies, including `@amplitude/analytics-client-common` and `@amplitude/analytics-types`, to support updated session replay functionality.

### Expanded test coverage:
- **`packages/plugin-session-replay-browser/test/plugin-helpers.test.ts`:** Added unit tests for the `parseUserProperties` function to validate its behavior across different `IdentifyOperation` types and edge cases.
- **`packages/plugin-session-replay-browser/test/session-replay.test.ts`:** Expanded test coverage for the `SessionReplayPlugin` class, including tests for session replay initialization, error handling, and the new `evaluateTargetingAndCapture` method for `IDENTIFY` events. [[1]](diffhunk://#diff-fc0192921acc8a0a584c787015829e8db027d267282a7dc3911ba8e2e1388c87L1-R9) [[2]](diffhunk://#diff-fc0192921acc8a0a584c787015829e8db027d267282a7dc3911ba8e2e1388c87R113-R137) [[3]](diffhunk://#diff-fc0192921acc8a0a584c787015829e8db027d267282a7dc3911ba8e2e1388c87R412-R432)
### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/src/constants.ts
+++ b/packages/plugin-session-replay-browser/src/constants.ts
@@ -1,0 +1,11 @@
+import { IdentifyOperation } from '@amplitude/analytics-types';
+
+export const PROPERTY_ADD_OPERATIONS = [
+  IdentifyOperation.SET,
+  IdentifyOperation.SET_ONCE,
+  IdentifyOperation.ADD,
+  IdentifyOperation.APPEND,
+  IdentifyOperation.PREPEND,
+  IdentifyOperation.POSTINSERT,
+  IdentifyOperation.POSTINSERT,
+];
--- a/packages/plugin-session-replay-browser/src/constants.ts
+++ b/packages/plugin-session-replay-browser/src/constants.ts
@@ -7,5 +7,4 @@ export const PROPERTY_ADD_OPERATIONS = [
   IdentifyOperation.APPEND,
   IdentifyOperation.PREPEND,
   IdentifyOperation.POSTINSERT,
-  IdentifyOperation.POSTINSERT,
 ];
--- a/packages/session-replay-browser/README.md
+++ b/packages/session-replay-browser/README.md
@@ -47,7 +47,17 @@ sessionReplay.init(API_KEY, {
 });
 ```
 
-### 3. Get session replay event properties
+### 3. Evaluate targeting (optional)
+Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. This should be done *before* step 4, getting the event properties. If you are not using the targeting condition logic provided via the Amplitude UI, this step is not required.
+```typescript
+const sessionTargetingMatch = sessionReplay.evaluateTargetingAndCapture({ event: {
+  event_type: EVENT_NAME,
+  time: EVENT_TIMESTAMP,
+  event_properties: eventProperties
+} });
+```
+
+### 4. Get session replay event properties
 Any event that occurs within the span of a session replay must be tagged with properties that signal to Amplitude to include it in the scope of the replay. The following shows an example of how to use the properties
 ```typescript
 const sessionReplayProperties = sessionReplay.getSessionReplayProperties();
@@ -57,7 +67,7 @@ track(EVENT_NAME, {
 })
 ```
 
-### 4. Update session id
+### 5. Update session id
 Any time that the session id for the user changes, the session replay SDK must be notified of that change. Update the session id via the following method:
 ```typescript
 sessionReplay.setSessionId(UNIX_TIMESTAMP)
@@ -67,7 +77,7 @@ You can optionally pass a new device id as a second argument as well:
 sessionReplay.setSessionId(UNIX_TIMESTAMP, deviceId)
 ```
 
-### 5. Shutdown (optional)
+### 6. Shutdown (optional)
 If at any point you would like to discontinue collection of session replays, for example in a part of your application where you would not like sessions to be collected, you can use the following method to stop collection and remove collection event listeners.
 ```typescript
 sessionReplay.shutdown()
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -285,7 +285,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
   };
 
   evaluateTargetingAndCapture = async (
-    targetingParams?: Pick<TargetingParameters, 'event' | 'userProperties'>,
+    targetingParams: Pick<TargetingParameters, 'event' | 'userProperties'>,
     isInit = false,
   ) => {
     if (!this.identifiers || !this.identifiers.sessionId || !this.config) {
@@ -298,7 +298,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
     }
 
     if (this.config.targetingConfig && !this.sessionTargetingMatch) {
-      let eventForTargeting = targetingParams?.event;
+      let eventForTargeting = targetingParams.event;
       if (
         eventForTargeting &&
         Object.values(SpecialEventType).includes(eventForTargeting.event_type as SpecialEventType)
@@ -314,7 +314,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
         targetingConfig: this.config.targetingConfig,
         loggerProvider: this.loggerProvider,
         apiKey: this.config.apiKey,
-        targetingParams: { userProperties: targetingParams?.userProperties, event: eventForTargeting },
+        targetingParams: { userProperties: targetingParams.userProperties, event: eventForTargeting },
       });
     }
 
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -68,6 +68,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
   eventCompressor: EventCompressor | undefined;
   sessionTargetingMatch = false;
   private lastTargetingParams?: Pick<TargetingParameters, 'event' | 'userProperties'>;
+  private lastShouldRecordDecision?: boolean;
 
   // Visible for testing only
   pageLeaveFns: PageLeaveFn[] = [];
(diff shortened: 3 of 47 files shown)


packages/plugin-session-replay-browser/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { IdentifyOperation } from '@amplitude/analytics-types';
+
+export const PROPERTY_ADD_OPERATIONS = [
+  IdentifyOperation.SET,
+  IdentifyOperation.SET_ONCE,
+  IdentifyOperation.ADD,
+  IdentifyOperation.APPEND,
+  IdentifyOperation.PREPEND,
+  IdentifyOperation.POSTINSERT,
+  IdentifyOperation.POSTINSERT,
```
lewgordon-amplitude: Why is this listed twice?

packages/session-replay-browser/README.md
```diff
@@ -47,7 +47,17 @@ sessionReplay.init(API_KEY, {
...
-### 3. Get session replay event properties
+### 3. Evaluate targeting (optional)
+Any event that occurs within the span of a session replay must be passed to the SDK to evaluate against targeting conditions. This should be done *before* step 4, getting the event properties. If you are not using the targeting condition logic provided via the Amplitude UI, this step is not required.
+```typescript
+const sessionTargetingMatch = sessionReplay.evaluateTargetingAndCapture({ event: {
+  event_type: EVENT_NAME,
+  time: EVENT_TIMESTAMP,
+  event_properties: eventProperties
+} });
+```
```
lewgordon-amplitude, line 58: Nit: I feel like we should discourage using the properties in the config or even prevent them from being used. I'd rather people use the UI since we don't have documentation on our targeting objects?

packages/session-replay-browser/src/targeting/targeting-idb-store.ts
lewgordon-amplitude, line 1: Given the issues we've seen with idb it's probably worth while to have a memory store. I also think that this should respect the store type given in the session replay browser options. It would seem weird to have multiple store type configs.

packages/targeting/src/targeting-idb-store.ts
lewgordon-amplitude, line 1: What's the difference between this and the other targeting idb store?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -325,11 +376,27 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        return false;
+      }
+      this.loggerProvider.log(
+        `Capturing replays for session ${this.identifiers.sessionId} due to matching targeting conditions.`,
+      );
+    } else {
+      const isInSample = isSessionInSample(this.identifiers.sessionId, this.config.sampleRate);
+      if (!isInSample) {
+        this.loggerProvider.log(`Opting session ${this.identifiers.sessionId} out of recording due to sample rate.`);
+      }
```
lewgordon-amplitude: Some of this may be covered by the console log work, but it may be nice to put additional details in the debugging event to capture why targeting succeeded (possibly why it didn't, too?)

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -273,6 +283,47 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        sessionId: Number(this.identifiers.sessionId),
+        targetingConfig: this.config.targetingConfig,
+        loggerProvider: this.loggerProvider,
+        apiKey: this.config.apiKey,
+        targetingParams: { userProperties: targetingParams.userProperties, event: eventForTargeting },
+      });
+    }
+
+    if (isInit) {
+      void this.initialize(true);
```
lewgordon-amplitude, line 328: Isn't this async? Why not await this?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -273,6 +283,47 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        Object.values(SpecialEventType).includes(eventForTargeting.event_type as SpecialEventType)
+      ) {
+        eventForTargeting = undefined;
+      }
+
+      // We're setting this on this class because fetching the value from idb
+      // is async, we need to access this value synchronously (for record
+      // and for getSessionReplayProperties - both synchronous fns)
+      this.sessionTargetingMatch = await evaluateTargetingAndStore({
+        sessionId: Number(this.identifiers.sessionId),
```
lewgordon-amplitude: We need to change the typing here since we want to support string session ids, too. Change we change this before merging?

lewgordon-amplitude (commented):
Looks good overall, just a few things to discuss before merging 💪

packages/session-replay-browser/src/targeting/targeting-idb-store.ts
jxiwang: Would there be an issue with conditions not properly matched since it's not persisted? Case I'm thinking is if the user closes their client but an event has occured. Although, I'm not sure if the `EvalEngine` would already handle it as well.

lewgordon-amplitude: Yeah makes sense. It seems like we probably need some type of client side storage. I think maybe we could just support event triggering rather than event counting (not sure if that's already in scope). All we really need to get to though perhaps is whether or not we should capture for the session which could be a different (more lightweight storage).

bravecod: I'm not entirely familiar with the targeting changes, and since this is based on a stale branch, it will likely require some refactoring.
Let me schedule some time next week so we can walk through it together. @jxiwang @lewgordon-amplitude

jxiwang: Since we are getting this out to select customers, this could be something we follow up on as well.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -273,6 +283,47 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        targetingConfig: this.config.targetingConfig,
+        loggerProvider: this.loggerProvider,
+        apiKey: this.config.apiKey,
+        targetingParams: { userProperties: targetingParams.userProperties, event: eventForTargeting },
```
bravecod, line 323: custom rrweb event contains this and capture

packages/targeting/src/targeting.ts
```diff
@@ -0,0 +1,46 @@
+import { EvaluationEngine } from '@amplitude/experiment-core';
```
bravecod, line 1: if this supports string sessionId

packages/session-replay-browser/src/session-replay.ts
jxiwang: Would make sure we aren't doing `Number(...)` or else if will convert to a `NaN` value.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -326,11 +381,46 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
     }
-    return isInSample;
+
+    // Send custom rrweb event for targeting decision
+    void this.addCustomRRWebEvent(CustomRRwebEvent.TARGETING_DECISION, {
+      message,
+      sessionId: this.identifiers.sessionId,
+      matched,
+      targetingParams: this.lastTargetingParams,
+    });
```
lewgordon-amplitude: Doesn't this fire every time they check the properties? Can we only fire this if shouldRecord changes?

bravecod: yeah, I think we can do that!

bravecod: https://github.com/amplitude/Amplitude-TypeScript/pull/1203/commits/17006e5b8b56c419baa40507e23841a87fb25011#diff-5701ec875ae41b5ed40495ead788b110bc03e50345eb0313f36c2efe3097830c

lewgordon-amplitude (approved):
LGTM, we'll probably need to follow up with some thorough testing here though since it's such a large PR.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -326,11 +383,49 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        shouldRecord = false;
+        matched = false;
+      } else {
+        shouldRecord = true;
+        matched = true;
+      }
+    }
+
+    // Only send custom rrweb event for targeting decision when the decision changes
+    if (this.lastShouldRecordDecision !== shouldRecord) {
```
jxiwang: Should this be in the `this.config.targetingConfig` if statement ? We don't care about this metadata if targeting is not set up.

PR #1209, opened 2025-07-15
feat(analytics-browser): add experimental frustrationInteractions

daniel-graham-amplitude:
### Summary

Make frustration interactions available as an experimental feature for beta users.

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/frustration-plugin.ts
+++ b/packages/plugin-autocapture-browser/src/frustration-plugin.ts
@@ -24,16 +24,7 @@ import { createClickObservable, createMutationObservable } from './observables';
 
 type BrowserEnrichmentPlugin = EnrichmentPlugin<BrowserClient, BrowserConfig>;
 
-export const frustrationPlugin = (
-  options: FrustrationInteractionsOptions = {
-    deadClicks: {
-      cssSelectorAllowlist: DEFAULT_DEAD_CLICK_ALLOWLIST,
-    },
-    rageClicks: {
-      cssSelectorAllowlist: DEFAULT_RAGE_CLICK_ALLOWLIST,
-    },
-  },
-): BrowserEnrichmentPlugin => {
+export const frustrationPlugin = (options: FrustrationInteractionsOptions = {}): BrowserEnrichmentPlugin => {
   const name = constants.FRUSTRATION_PLUGIN_NAME;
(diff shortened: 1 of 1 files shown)


packages/analytics-browser/test/default-tracking.test.ts
```diff
@@ -10,8 +11,54 @@ import {
...
   isNetworkTrackingEnabled,
+  isFrustrationInteractionsEnabled,
 } from '../src/default-tracking';
 
+describe('isFrustrationInteractionsEnabled', () => {
+  test('should return true with true parameter', () => {
+    expect(isFrustrationInteractionsEnabled(true)).toBe(true);
+  });
+
+  test('should return true with undefined parameter', () => {
```
Mercy811, line 22:
```suggestion
  test('should return false with undefined parameter', () => {
```

packages/plugin-autocapture-browser/src/frustration-plugin.ts
```diff
@@ -24,7 +24,16 @@ import { createClickObservable, createMutationObservable } from './observables';
...
+export const frustrationPlugin = (
+  options: FrustrationInteractionsOptions = {
+    deadClicks: {
+      cssSelectorAllowlist: DEFAULT_DEAD_CLICK_ALLOWLIST,
+    },
+    rageClicks: {
+      cssSelectorAllowlist: DEFAULT_RAGE_CLICK_ALLOWLIST,
+    },
+  },
+): BrowserEnrichmentPlugin => {
```
Mercy811: What's the point of adding a partial default value here if we do the following a few lines below? 
```
const rageCssSelectors = options.rageClicks?.cssSelectorAllowlist ?? DEFAULT_RAGE_CLICK_ALLOWLIST;
  const deadCssSelectors = options.deadClicks?.cssSelectorAllowlist ?? DEFAULT_DEAD_CLICK_ALLOWLIST;
  const dataAttributePrefix = options.dataAttributePrefix ?? DEFAULT_DATA_ATTRIBUTE_PREFIX;
```

This newly added blocks brings in confusion, for example, if i pass custom a css selector allow list for dead clicks, do i loose the default css selector allow list for rage click?

Mercy811 (commented):
Thanks Dan. Overall looks good to me. Please see comments.

PR #1214, opened 2025-07-17
fix(analytics-browser): disable frustrationInteractions and webVitals when autocapture=true

daniel-graham-amplitude:
### Summary

When "autocapture=true", disable web-vitals and frustration-interactions because they're still in beta.

### Checklist


Mercy811 (approved):
LGTM! I missed it too when reviewing the last PR. I was thinking autocapture true should turn everything on but forgot that they are beta. Thanks Dan for correcting it!

PR #1220, opened 2025-07-28
chore: exclusively use Nx for CI workflows

daniel-graham-amplitude:
### Summary
* Replace ci.yml with ci-nx.yml
* Remove ci-nx.yml

### Checklist


Mercy811:
Can we leave ci.yml there and set the trigger as manual? Just in case of edge case when ci nx fails or we wanna manually run non-nx ci. wdyt?

PR #1221, opened 2025-07-29
feat(analytics-browser): allow capture of request/response header/body information

daniel-graham-amplitude:
### Checklist

### DIFF
--- a/packages/analytics-core/src/network-request-event.ts
+++ b/packages/analytics-core/src/network-request-event.ts
@@ -50,7 +50,14 @@ type HeadersInitSafe = HeadersRequestSafe | Record<string, string> | string[][];
 type ResponseSafe = {
   status: number;
   headers: HeadersResponseSafe | undefined;
+  clone(): ResponseCloneSafe;
 };
+
+type ResponseCloneSafe = {
+  text(): Promise<string>;
+};
+
+const TEXT_READ_TIMEOUT = 500;
 
 export type RequestInitSafe = {
   method?: string;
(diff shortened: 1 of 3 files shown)


daniel-graham-amplitude:
bugbot run

packages/analytics-core/src/network-request-event.ts
```diff
@@ -145,7 +152,7 @@ export class RequestWrapperFetch implements IRequestWrapper {
     return this.request.method;
   }
 
-  get body(): FetchRequestBody | XMLHttpRequestBodyInitSafe | null {
+  get body(): string | null {
```
jxiwang, line 155: How come this is a string now?

daniel-graham-amplitude: I had noticed that it only returns a "string" in it's current implementation, so I just narrowed down the return type to "string" to simplify the implementation.

I don't expect we'll want to capture any [type in the Request Init body](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#body) aside from just string, because that's how JSON is represented.

The other types are:

[ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)
[TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
[DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView)
[Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)
[File](https://developer.mozilla.org/en-US/docs/Web/API/File)
[URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
[FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)
[ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)

If there's demand, we could capture them in a future version.

Mercy811 (approved):
LGTM! Thanks @daniel-graham-amplitude

jxiwang (approved):

PR #1227, opened 2025-07-30
fix(analytics-browser): support evaluation window for cross domain t…

Mercy811:
### Summary

AMP-134981

This is feature request to provide a evaluation window for cross-domain tracking. Other analytics platforms do enforce a time-validation between the link being generated and cross domain tracking being supported. More details of what others do on slack thread linked in the jira ticket.

- Support ampTimestamp url param, the expiration timestamp
- The SDK would not take device id, or session id in url param if ampTimestamp passes

Note that ampTimestamp is a future timestamp and should be generated and appended to the url by customers. The advantage is that time window is more flexible. GA is 2 minutes while Adobe is 5 minutes. We can just let customers choose how long they want.

### Checklist

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -120,7 +120,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
 
     // Check if ampTimestamp is present and valid
     const ampTimestamp = queryParams.ampTimestamp ? Number(queryParams.ampTimestamp) : undefined;
-    const isTimestampValid = ampTimestamp !== undefined && !Number.isNaN(ampTimestamp) && Date.now() < ampTimestamp;
+    const isTimestampValid = ampTimestamp ? Date.now() < ampTimestamp : true;
 
     const querySessionId =
       isTimestampValid && !Number.isNaN(Number(queryParams.ampSessionId))
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -112,14 +112,20 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
 
     // Step 3: Set session ID
     // Priority 1: `options.sessionId`
-    // Priority 2: sessionId from url if it's Number
+    // Priority 2: sessionId from url if it's Number and ampTimestamp is valid
     // Priority 3: last known sessionId from user identity storage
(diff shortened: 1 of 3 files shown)


packages/analytics-browser/src/config.ts
```diff
@@ -242,11 +242,14 @@ export const useBrowserConfig = async (
   const previousCookies = await cookieStorage.get(getCookieName(apiKey));
   const queryParams = getQueryParams();
 
+  // Check if ampTimestamp is present and valid
+  const ampTimestamp = queryParams.ampTimestamp ? Number(queryParams.ampTimestamp) : undefined;
+  const isTimestampValid = ampTimestamp ? Date.now() < ampTimestamp : true;
```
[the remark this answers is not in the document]
Mercy811: If ampTimestamp doesn't exist or invalid, for backwards compatibility, we still take session id and device id from url

packages/analytics-browser/src/browser-client.ts
```diff
@@ -112,14 +112,20 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
...
     // Default: `Date.now()`
     // Session ID is handled differently than device ID and user ID due to session events
     const queryParams = getQueryParams();
-    const querySessionId = Number.isNaN(Number(queryParams.ampSessionId))
-      ? undefined
-      : Number(queryParams.ampSessionId);
+
+    // Check if ampTimestamp is present and valid
+    const ampTimestamp = queryParams.ampTimestamp ? Number(queryParams.ampTimestamp) : undefined;
+    const isTimestampValid = ampTimestamp ? Date.now() < ampTimestamp : true;
```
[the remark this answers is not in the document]
Mercy811: If ampTimestamp doesn't exist or invalid, for backwards compatibility, we still take session id and device id from url

daniel-graham-amplitude: nitpick: `isTimestampValid` makes me think that this line of code is checking to see if the timestamp was formatted correctly. I think a better name would be something like `isTimeLimitExceeded`.

daniel-graham-amplitude: Or I suppose `isWithinTimeLimit`

Mercy811: Rename at https://github.com/amplitude/Amplitude-TypeScript/pull/1227/commits/960d0f94688e9603a9890bf03c29b79d5bb2f832

PR #1232, opened 2025-07-31
fix: add function to translate remoteConfig to localConfig

daniel-graham-amplitude:
### Summary

To keep Remote Config consistent across platforms, we want remote config to not use union types (e.g.: `frustrationInteractions = Object | boolean`), because they don't work well in Swift or Kotlin.

Instead of the `Object | boolean` union type, we settled on a singular type that looks like this:

```
{ enabled: true|false, ...otherParams }
```

Objects with this shape can then be translated into local union types

`{ prop: { enabled: true, hello: 'world' }}` becomes `{ prop: { hello: 'world' }}`
`{ prop: { enabled: true }}` becomes `{ prop: true }`
`{ prop: { enabled: false, ... }}` becomes `{ prop: false}`

This PR adds runtime translations to the remote config object so that when it encounters objects of the shape `{ enabled: true|false, ...otherParams}`  it translates them at runtime to a format that's compatible with local config

For example

```
{ autocapture: { networkTracking: { enabled: true }}
```

Will be translated to

```
{ autocapture: { networkTrackings: true }}
```

Which is compatible with the type `NetworkTrackingOptions | boolean`

### Checklist


daniel-graham-amplitude:
bugbot run

packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -20,6 +20,56 @@ type RemoteConfigBrowserSDK = {
...
+          // If enabled is false, set the value to false
+          (config as any)[propertyName] = false;
+        }
+      }
+
+      // recursively translate properties of the value
+      translateRemoteConfigToLocal(value as Record<string, any>);
+    } catch (e) {
+      // a failure here means that an accessor threw an error
+      // so don't translate it
```
Mercy811, line 68:
```suggestion
      // a failure here means that an accessor threw an error
      // so don't translate it
      // TODO(diagnostics)
```

packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -20,6 +20,56 @@ type RemoteConfigBrowserSDK = {
   autocapture?: AutocaptureOptionsRemoteConfig | boolean;
 };
 
+/**
+ * Performs deep translation of remote config so that
+ * the schema of the remote config is compatible with
+ * the schema of the local config
+ * @param config Config from remote config
+ */
```
Mercy811, line 28: nit:
```suggestion
/**
 * Performs a deep transformation of a remote config object so that
 * it matches the expected schema of the local config.
 * 
 * Specifically, it normalizes nested `enabled` flags into concise union types.
 * 
 * ### Transformation Rules:
 * - If an object has `enabled: true`, it is replaced by the same object without the `enabled` field.
 * - If it has only `enabled: true`, it is replaced with `true`.
 * - If it has `enabled: false`, it is replaced with `false` regardless of other fields.
 * 
 * ### Examples:
 * Input:  { prop: { enabled: true, hello: 'world' }}  
 * Output: { prop: { hello: 'world' } }
 * 
 * Input:  { prop: { enabled: true }}  
 * Output: { prop: true }
 * 
 * Input:  { prop: { enabled: false, hello: 'world' }}  
 * Output: { prop: false }
 * 
 * Input:  { prop: { hello: 'world' }}  
 * Output: { prop: { hello: 'world' } } // No change
 * 
 * @param config Remote config object to be transformed
 * @returns Transformed config object compatible with local schema
 */
```

Mercy811 (approved):
Thanks @daniel-graham-amplitude. LGTM! See one nit.

We should be careful when using the new "enabled" schema on the backend as older versions of the SDK won't support translation, specifically element interactions.
For element interactions, we should keep using the old schema without "enabled", for example, on an older version of the SDK, the following won't disable elementInteractions.

```
{
  autocapture: {
     elementInteractions: {
      enabled: false
    }
  }
}
```
So far, it's the only field we support union type.

Mercy811 (changes requested):
Actually, after re-thinking, this change won't take effect on older SDK versions (2.11.1 and before), so we we cannot set "enabled: false" to disable all existing fields with union types in autocapture https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8yNDU0M2M3M2M1ZGZhNTY2M2Y1ZTkzNzU1OTMzZjU5NTQzZjBmZDIxL2YvcGFja2FnZXMvYW5hbHl0aWNzLWNvcmUvc3JjL3R5cGVzL2NvbmZpZy9icm93c2VyLWNvbmZpZy50cz91cmw9Z2l0JTQwZ2l0aHViLmNvbSUzQWFtcGxpdHVkZSUyRkFtcGxpdHVkZS1UeXBlU2NyaXB0LmdpdA%3D%3D?origin=gitlens

They are:
- attribution
- pageViews
- elementInteractions
- frustrationInteractions (experimental)
- networkTracking
- webVitals (experimental)

Running this test on main branch
```
    test('should keep backward compatibility with older SDK versions', () => {
      localConfig = createConfigurationMock(
        createConfigurationMock({
          autocapture: true,
        }),
      );

      const remoteConfig = {
        autocapture: {
          elementInteractions: {
            enabled: false,
          },
        },
      };

      const expectedAutocapture = {
        sessions: true,
        fileDownloads: true,
        formInteractions: true,
        pageViews: true,
        attribution: true,
        elementInteractions: false,
        frustrationInteractions: true,
        webVitals: true,
      };

      updateBrowserConfigWithRemoteConfig(remoteConfig, localConfig);
      console.log(localConfig.autocapture);
      expect(localConfig.autocapture).toStrictEqual(expectedAutocapture);
      expect(localConfig.defaultTracking).toStrictEqual(expectedAutocapture);
    });

```
image

Mercy811 (approved):
LGTM. Schema validation: https://amplitude.slack.com/archives/C0981RU8QQG/p1754440876284279?thread_ts=1754416163.871069&cid=C0981RU8QQG

PR #1243, opened 2025-08-09
refactor(analytics-browser): add JSON query helper methods

daniel-graham-amplitude:
### Summary
* Add a helper module called "json-query" which exposes a method called `pruneJson(jsObj: Json, allowlist: string[], excludelist: string[])` that:
  * Takes a plain Javascript object, allowlist and excludelist as it's parameters
  * Deletes any attributes in the object that aren't part of the `allowlist`
  * Deletes any attributes in the object that are defined in the `excludelist`
  * Doesn't return anything, this is a mutative function. Prevents creating duplicate objects.
* Stubbed-in + commented out configuration changes
  * Leaving out config changes from the scope of this MR because it's not ready
  * Commented out the config changes that I made already so I can easily enable them later
  
### Comments
* Since this new module is unused, it's behaviour isn't set in stone. We can still easily iterate on this and make changes if we decide that it's behaviour isn't what we desire.

### Checklist

### DIFF
--- a/packages/analytics-core/src/types/network-tracking.ts
+++ b/packages/analytics-core/src/types/network-tracking.ts
@@ -18,15 +18,24 @@ export interface NetworkTrackingOptions {
 }
 
 // export interface BodyCaptureRule {
-//   /**
-//    * Keys to allow in the request/response body.
-//    */
 //   // TODO: Change this to whatever is the settled configuration name before merging
+//   /**
+//    * List of JSON pointers to capture from a request or response body (JSON objects only)
+//    *
+//    * If this is empty or undefined, no attributes are captured
+//    *
+//    * Follows a syntax similar to JSON Pointer, except:
+//    * - The leading / is optional
+//    * - A wildcard * can be used to match any key
+//    * - A wildcard ** can be used to match any number of keys (or no keys)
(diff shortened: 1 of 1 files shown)


daniel-graham-amplitude:
bugbot run

packages/analytics-core/src/types/network-tracking.ts
```diff
@@ -17,6 +17,19 @@ export interface NetworkTrackingOptions {
...
+// export interface BodyCaptureRule {
+//   /**
+//    * Keys to allow in the request/response body.
+//    */
+//   // TODO: Change this to whatever is the settled configuration name before merging
+//   allowlist?: string[];
+//   /**
+//    * Keys to block in the request/response body.
+//    */
+//   // TODO: Change this to whatever is the settled configuration name before merging
```
Mercy811: nit: explain how * and ** work and the format of flat json as JSDoc comments are visible in IDE when hovering

Mercy811 (approved):
Thanks @daniel-graham-amplitude. LGTM!

packages/analytics-core/src/types/network-tracking.ts
daniel-graham-amplitude: Changes submitted thanks.

PR #1246, opened 2025-08-12
chore: refactor httpstat.us to use a local service

daniel-graham-amplitude:
### Summary
* httpstat.us is down so refactoring XHR and Fetch tests to use a local API server (I should have done that from the start, using an external service like that isn't reliable, and also kind of inconsiderate)
* Added a new API service that does the same thing as httpstat.us
* Removed redundant fetch.html and xhr.html files from test-server

### Checklist


daniel-graham-amplitude:
bugbot run

Mercy811 (approved):
LGTM!

PR #1252, opened 2025-08-14
feat(analytics-browser): add urls matching attribute to network capture rules (experimental)

daniel-graham-amplitude:
### Summary
Add `urls` attribute to `NetworkCaptureRule`, which, when set, means the URL has to match a url pattern in `urls` in order for it to be captured. `url` can be `string|regex`

### Commentary
This is experimental for now. Leaving it as _draft_ until design decision is ready.

### Checklist

### DIFF
--- a/packages/analytics-core/src/types/network-tracking.ts
+++ b/packages/analytics-core/src/types/network-tracking.ts
@@ -47,7 +47,7 @@ export interface NetworkCaptureRule {
   hosts?: string[];
   /**
    * URL patterns to allow for network capture. Supports wildcard.
-   * 
+   *
    * This takes precedence over `hosts`
    * @experimental This feature is experimental and may not be stable
    * @defaultValue `["*"]` all URLs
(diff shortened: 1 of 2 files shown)


packages/analytics-core/src/types/network-tracking.ts
```diff
@@ -45,6 +45,17 @@ export interface NetworkCaptureRule {
    * @defaultValue `["*"]` all hosts (except amplitude)
    */
   hosts?: string[];
+  /**
+   * URL patterns to allow for network capture. Supports wildcard.
+   * @experimental This feature is experimental and may not be stable
+   * @defaultValue `["*"]` all URLs
+   */
+  urls?: (string | RegExp)[];
```
Mercy811: Can we mention that urls takes precedence over hosts and if urls is set, hosts is ignored?

Mercy811: question: are we deprecating hosts in the future? or the plan is to keep both of them.

Mercy811 (approved):
Overall LGTM. See comments.

packages/analytics-core/src/types/network-tracking.ts
daniel-graham-amplitude: The plan is to deprecate hosts.

daniel-graham-amplitude: Done!

PR #1253, opened 2025-08-15
feat(plugin-network-capture-browser): add ability to capture headers (experimental)

daniel-graham-amplitude:
### Summary

* Add ability for users to capture Response Headers and Request Headers from network request events.
  * Added `responseHeaders` and `requestHeaders` as config rules that have an `allowlist` and `captureSafeHeaders` feature
  * When rule is set, it captures headers and sends them as part of the network request event

### Checklist


packages/plugin-network-capture-browser/src/track-network-event.ts
```diff
@@ -156,6 +201,33 @@ export function shouldTrackNetworkEvent(networkEvent: NetworkRequestEvent, optio
...
+        if (networkEvent.requestWrapper && requestHeadersRule) {
+          const requestHeaders = networkEvent.requestWrapper.headers(
+            requestHeadersRule.allowlist,
+            requestHeadersRule.captureSafeHeaders,
+          );
+          if (requestHeaders) {
+            networkEvent.requestHeaders = requestHeaders;
+          }
+        }
+      }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: The mutation of `networkEvent` is an "enrichment" of the event, that makes it include headers (response/request). And the consumption of `headers()` is intentional; this is the only place that should call `headers`.

packages/analytics-core/src/types/network-tracking.ts
```diff
@@ -39,6 +39,21 @@ export interface NetworkTrackingOptions {
...
+   * @defaultValue `[]` no headers
+   */
+  allowlist?: string[];
+  /**
+   * Capture all [Safe Headers](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/constants.ts)
+   *
+   * If true, these safe headers will all be captured automatically.
+   * @defaultValue `true`
+   */
+  captureSafeHeaders?: boolean;
```
Mercy811: For simplicity and consistence, can we have just allowlist and set its default value to safe headers? We can export safe headers and customers can import and reuse it if they want to extend it. Example: [elementInteractionsOptions.cssAllowlist](https://github.com/amplitude/Amplitude-TypeScript/blob/da213cc33c4986bcebff2b4264b2c17314f5f310/packages/analytics-core/src/types/element-interactions.ts#L37-L44)

daniel-graham-amplitude:
promptless run

packages/analytics-core/src/types/network-tracking.ts
```diff
@@ -63,12 +63,24 @@ export interface NetworkCaptureRule {
...
+  /**
+   * Determines what to capture from the response headers.
+   * @experimental This feature is experimental and may not be stable
+   */
+  responseHeaders?: string[] | boolean;
+  /**
+   * Determines what to capture from the request headers.
+   * @experimental This feature is experimental and may not be stable
+   */
+  requestHeaders?: string[] | boolean;
```
Mercy811: It seems that responseHeaders can be just `string[] | undefined`. 
- if undefined, then default to safe headers, equals to current true
- if not 
  - [] empty array, equals to current false

packages/analytics-core/src/types/network-tracking.ts
Mercy811: It's updated in following PRs.

packages/analytics-core/src/types/network-tracking.ts
daniel-graham-amplitude: Once it gets out of "experimental", the plan is to capture headers by default. So allowing boolean means customers can set `requestHeaders=false` to explicitly opt out.

This follows a pattern similar to `networkTracking` too.

Mercy811: Can we add some comments to explain what are requestHeaders values when true & false?
- true: default safe headers. Not all headers (this might cause confusion) 
- false: no headers will be tracked

daniel-graham-amplitude: Added a comment, thanks!

Mercy811 (approved):
Thanks @daniel-graham-amplitude LGTM!

PR #1255, opened 2025-08-15
refactor(analytics-core): change pruneHeaders to be non-mutative

daniel-graham-amplitude:
### Summary

Upon careful consideration, I think it's unwise to make the `pruneHeaders` function mutate the headers object. There's a risk that somebody could unwittingly pass the original headers object into `pruneHeaders` by mistake and mutate it. Not worth the performance boost.

This PR changes it so that `pruneHeaders` returns a new headers object that has the key/value pairings pruned from the original. The original object is untouched.

### Checklist


Mercy811 (approved):
I agree it won't lower performance even we duplicate headers and then prune. LGTM. Thanks Dan!

PR #1257, opened 2025-08-18
feat(autocapture): add filtering for email in isNonSensitiveString

achhunna:
### Summary

Add email address regex match filtering along with credit card and SSN.

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/helpers.ts
+++ b/packages/plugin-autocapture-browser/src/helpers.ts
@@ -95,6 +95,10 @@ export const isNonSensitiveString = (text: string | null) => {
     if (ssnRegex.test(text)) {
       return false;
     }
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+    if (emailRegex.test(text)) {
+      return false;
+    }
   }
   return true;
 };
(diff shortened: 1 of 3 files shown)


packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -95,6 +95,10 @@ export const isNonSensitiveString = (text: string | null) => {
     if (ssnRegex.test(text)) {
       return false;
     }
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
cyhk: email regex is unfortunately notoriously complicated:  would prefer to used some library or tried and true regex, e.g. https://emailregex.com/ if the one you have isn't one already.

Dogfalo: We also need to have it match emails between other text

packages/plugin-autocapture-browser/test/helpers.test.ts
```diff
@@ -50,6 +50,12 @@ describe('autocapture-plugin helpers', () => {
       expect(result).toEqual(false);
     });
 
+    test('should return false when text is email address format', () => {
```
Dogfalo, line 53: add more use cases `This is my email user@example.com`

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: Done

packages/plugin-autocapture-browser/test/helpers.test.ts
achhunna: Done

Dogfalo (approved):

PR #1259, opened 2025-08-19
feat(autocapture): add redactTextRegex option to autocapture

Dogfalo:
### Summary
- add option `redactTextRegex`
  - This option is a RegExp[], or it can be a list objects passed in primarily from remote config JSON.
- Refactor data extraction from elements into DataExtractor class
  - This class helps to keep text filtering functions portable without having to pass around the options

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/data-extractor.ts
+++ b/packages/plugin-autocapture-browser/src/data-extractor.ts
@@ -17,13 +17,16 @@ import type { JSONValue } from './helpers';
 export class DataExtractor {
   private readonly additionalRedactTextPatterns: RegExp[];
 
-  constructor(options: ElementInteractionsOptions = {}) {
+  constructor(options: ElementInteractionsOptions) {
     const rawPatterns = options.redactTextRegex ?? [];
     const compiled: RegExp[] = [];
     for (const entry of rawPatterns) {
+      if (compiled.length >= constants.MAX_REDACT_TEXT_PATTERNS) {
+        break;
+      }
       if (entry instanceof RegExp) {
         compiled.push(entry);
-      } else if ('pattern' in entry && typeof (entry.pattern === 'string')) {
+      } else if ('pattern' in entry && typeof entry.pattern === 'string') {
         try {
           compiled.push(new RegExp(entry.pattern));
         } catch {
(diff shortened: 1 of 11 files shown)


daniel-graham-amplitude:
bugbot run

Dogfalo:
@cursor review

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -0,0 +1,242 @@
...
+    if (text == null) {
+      return false;
+    }
+
+    if (typeof text !== 'string') {
+      return true;
+    }
+
+    // Check for credit card number
+    const ccRegex =
```
daniel-graham-amplitude: _Optional_ the `ccRegex` and `ssnRegex` could be moved into constants outside the class so that you don't have to declare it on every invocation of `isNonSensitiveString`

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -0,0 +1,242 @@
...
+        }
+      }
+    }
+    this.additionalRedactTextPatterns = compiled;
+  }
+
+  isNonSensitiveString = (text: string | null): boolean => {
+    // if text is null, it can't be sensitive
+    if (text == null) {
+      return false;
```
daniel-graham-amplitude: I think you meant to return `true` here? If it's null, it is non sensitive.

daniel-graham-amplitude: Actually I think you can just take this conditional out completely because it would be covered by `if (typeof test !== 'string'` below

daniel-graham-amplitude:
Mostly looks good, one comment that needs to be addressed.

packages/plugin-autocapture-browser/src/data-extractor.ts
Dogfalo: Yea good callout. Another thing I'm thinking of doing in the next pass is to combine them which I believe will also improve the performance

daniel-graham-amplitude:
Lgtm. You may want to sync with @achhunna on this one. This PR and his [PR](https://github.com/amplitude/Amplitude-TypeScript/pull/1263) change the same parts of the code and could cause conflicts.

Probably best @achhunna merges his first.

PR #1263, opened 2025-08-20
feat(autocapture): attribute redaction

achhunna:
### Summary

Exclude attribute capture from specific elements on the page, allowing Autocapture usage on the page without capturing attributes in the specified element.
```
For example:
<div data-amp-mask-attributes=”name”>
<span name=”John D”>Account</span>
</div>
```
The above example would exclude the “name” attribute from Autocapture data at the SDK level.

### Checklist

### DIFF
--- a/packages/plugin-autocapture-browser/src/helpers.ts
+++ b/packages/plugin-autocapture-browser/src/helpers.ts
@@ -141,10 +141,51 @@ export const getText = (element: Element): string => {
   return text;
 };
 
+/**
+ * Collects redacted attribute names from ancestor elements with data-amp-redact-attributes
+ * Redaction rules only apply to children, not the element that defines the rule
+ * The 'id' and 'class' attributes cannot be redacted as they're critical for element identification
+ * @param element - The target element to check ancestors for
+ * @returns Set of attribute names that should be redacted
+ */
+export const getRedactedAttributeNames = (element: Element): Set<string> => {
+  const redactedAttributes = new Set<string>();
+  let currentElement: Element | null = element.parentElement; // Start from parent, not element itself
+
+  // Walk up the DOM tree to find any data-amp-redact-attributes
+  while (currentElement) {
+    const redactValue = currentElement.getAttribute('data-amp-redact-attributes');
+    if (redactValue) {
+      // Parse comma-separated attribute names and add to set
+      const attributesToRedact = redactValue
+        .split(',')
+        .map((attr) => attr.trim())
+        .filter((attr) => attr.length > 0);
+      attributesToRedact.forEach((attr) => {
+        // Prevent 'id' and 'class' from being redacted as they're critical for element identification
+        if (attr !== 'id' && attr !== 'class') {
+          redactedAttributes.add(attr);
+        }
+      });
+    }
+    currentElement = currentElement.parentElement;
+  }
+
+  return redactedAttributes;
+};
+
 export const getAttributesWithPrefix = (element: Element, prefix: string): { [key: string]: string } => {
+  const redactedAttributes = getRedactedAttributeNames(element);
+
   return element.getAttributeNames().reduce((attributes: { [key: string]: string }, attributeName) => {
     if (attributeName.startsWith(prefix)) {
       const attributeKey = attributeName.replace(prefix, '');
+
+      // Skip redacted attributes
+      if (redactedAttributes.has(attributeKey)) {
+        return attributes;
+      }
+
       const attributeValue = element.getAttribute(attributeName);
       if (attributeKey) {
         attributes[attributeKey] = attributeValue || '';
@@ -288,19 +329,18 @@ export const getEventProperties = (actionType: ActionType, element: Element, dat
   /* istanbul ignore next */
   const rect =
     typeof element.getBoundingClientRect === 'function' ? element.getBoundingClientRect() : { left: null, top: null };
-  const ariaLabel = element.getAttribute('aria-label');
+
+  const redactedAttributes = getRedactedAttributeNames(element);
   const attributes = getAttributesWithPrefix(element, dataAttributePrefix);
   const nearestLabel = getNearestLabel(element);
+
   /* istanbul ignore next */
   const properties: Record<string, any> = {
-    [constants.AMPLITUDE_EVENT_PROP_ELEMENT_ID]: element.getAttribute('id') || '',
-    [constants.AMPLITUDE_EVENT_PROP_ELEMENT_CLASS]: element.getAttribute('class'),
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_HIERARCHY]: getHierarchy(element),
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_TAG]: tag,
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_TEXT]: getText(element),
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_POSITION_LEFT]: rect.left == null ? null : Math.round(rect.left),
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_POSITION_TOP]: rect.top == null ? null : Math.round(rect.top),
-    [constants.AMPLITUDE_EVENT_PROP_ELEMENT_ARIA_LABEL]: ariaLabel,
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_ATTRIBUTES]: attributes,
     [constants.AMPLITUDE_EVENT_PROP_ELEMENT_PARENT_LABEL]: nearestLabel,
     [constants.AMPLITUDE_EVENT_PROP_PAGE_URL]: window.location.href.split('?')[0],
@@ -308,9 +348,27 @@ export const getEventProperties = (actionType: ActionType, element: Element, dat
     [constants.AMPLITUDE_EVENT_PROP_VIEWPORT_HEIGHT]: window.innerHeight,
     [constants.AMPLITUDE_EVENT_PROP_VIEWPORT_WIDTH]: window.innerWidth,
   };
-  if (tag === 'a' && actionType === 'click' && element instanceof HTMLAnchorElement) {
+
+  // Add non-redacted attributes conditionally
+  // id is never redacted, so always include it
+  properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_ID] = element.getAttribute('id') || '';
+
+  // class is never redacted, so always include it
+  properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_CLASS] = element.getAttribute('class');
+
+  if (!redactedAttributes.has('aria-label')) {
+    properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_ARIA_LABEL] = element.getAttribute('aria-label');
+  }
+
+  if (
+    tag === 'a' &&
+    actionType === 'click' &&
+    element instanceof HTMLAnchorElement &&
+    !redactedAttributes.has('href')
+  ) {
     properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_HREF] = element.href;
   }
+
   return removeEmptyProperties(properties);
 };
 
--- a/packages/plugin-autocapture-browser/src/hierarchy.ts
+++ b/packages/plugin-autocapture-browser/src/hierarchy.ts
@@ -1,4 +1,4 @@
-import { isNonSensitiveElement, type JSONValue } from './helpers';
+import { isNonSensitiveElement, getRedactedAttributeNames, type JSONValue } from './helpers';
 import type { Hierarchy, HierarchyNode } from './typings/autocapture';
 
 const BLOCKED_ATTRIBUTES = [
@@ -41,6 +41,7 @@ export function getElementProperties(element: Element | null): HierarchyNode | n
   }
 
   const tagName = String(element.tagName).toLowerCase();
+  const redactedAttributes = getRedactedAttributeNames(element);
   const properties: HierarchyNode = {
     tag: tagName,
   };
@@ -74,6 +75,11 @@ export function getElementProperties(element: Element | null): HierarchyNode | n
   // if input is hidden or password or for SVGs, skip attribute collection entirely
   if (!HIGHLY_SENSITIVE_INPUT_TYPES.includes(String(element.getAttribute('type'))) && !SVG_TAGS.includes(tagName)) {
     for (const attr of filteredAttributes) {
+      // Skip redacted attributes
+      if (redactedAttributes.has(attr.name)) {
+        continue;
+      }
+
       // If sensitive element, only allow certain attributes
       if (isSensitiveElement && !SENSITIVE_ELEMENT_ATTRIBUTE_ALLOWLIST.includes(attr.name)) {
         continue;
--- a/packages/plugin-autocapture-browser/test/helpers.test.ts
+++ b/packages/plugin-autocapture-browser/test/helpers.test.ts
@@ -4,6 +4,8 @@ import {
   isNonSensitiveElement,
   getText,
   getAttributesWithPrefix,
+  getRedactedAttributeNames,
+  getEventProperties,
   isEmpty,
   removeEmptyProperties,
   getNearestLabel,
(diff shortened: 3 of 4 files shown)


packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -141,10 +141,51 @@ export const getText = (element: Element): string => {
...
+ * @param element - The target element to check ancestors for
+ * @returns Set of attribute names that should be redacted
+ */
+export const getRedactedAttributeNames = (element: Element): Set<string> => {
+  const redactedAttributes = new Set<string>();
+  let currentElement: Element | null = element.parentElement; // Start from parent, not element itself
+
+  // Walk up the DOM tree to find any data-amp-mask-attributes
+  while (currentElement) {
+    const redactValue = currentElement.getAttribute(constants.DATA_AMP_MASK_ATTRIBUTES);
```
daniel-graham-amplitude: I think a more performant way of doing this would be to make use of the [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) selector.

Instead of doing
```
let currentElement: Element | null = element.parentElement;
```
You can do
```
let currentElement: Element | null = element.closest(`[${constants.DATA_AMP_MASK_ATTRIBUTES}]`);
```
And always call `closest` instead of `.parentElement`.

The rationale for doing it this way is that `closest` invokes native libraries and is more performant than using JS to visit every ancestor.

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -41,6 +41,7 @@ export function getElementProperties(element: Element | null): HierarchyNode | n
   }
 
   const tagName = String(element.tagName).toLowerCase();
+  const redactedAttributes = getRedactedAttributeNames(element);
```
daniel-graham-amplitude: I'm concerned about the performance of this. This gets called on every ancestor of an element (via `getHierarchy`), meaning that for every visit to an ancestor (n), we're also visiting the ancestor's ancestors (n * n).

Would this approach work?
* don't do the redaction in `getElementProperties`
* when calling `getElementProperties`, make sure that `DATA_AMP_MASK_ATTRIBUTES` is always one of the `attrs` that we capture
* after calling `ancestors.map((el) => getElementProperties(el))` (in `getHierarchy`), iterate through them from left to right (leftmost being the root node)
  * whenever we find an attr `DATA_AMP_MASK_ATTRIBUTES`, add it's value as a list of attributes to a list (or possibly a Set) called something like `redactedAttrs`
  * if one of the `attrs` in the `HierarchyNode` that we're visiting is in `redactedAttrs`, delete it
  
This could reduce the operation from O(n*n) back to being O(n).

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: Done

packages/plugin-autocapture-browser/src/hierarchy.ts
achhunna: Moved `DATA_AMP_MASK_ATTRIBUTES` attr check into `getHierarchy`

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -122,6 +143,29 @@ export const getHierarchy = (element: Element | null): Hierarchy => {
...
+    }
+
+    // Apply redaction to current node
+    if (node?.attrs) {
+      for (const attrName of redactedAttrs) {
+        if (node.attrs[attrName]) {
+          delete node.attrs[attrName];
+        }
+      }
+    }
```
daniel-graham-amplitude: I think you'll want to swap the two if statements, because IIRC you don't want to apply the redaction rules on the element that has the rules set, but rather on that element's descendants.

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -122,6 +143,29 @@ export const getHierarchy = (element: Element | null): Hierarchy => {
     MAX_HIERARCHY_LENGTH,
   ) as Hierarchy;
 
+  // Collect redacted attributes and apply redaction in a single pass
+  const redactedAttrs = new Set<string>();
+  const reversedHierarchy = [...hierarchy].reverse(); // root to target order
```
daniel-graham-amplitude: Because `hierarchy` gets passed through `ensureListUnderLimit`, it might trim some of the hierarchy, meaning we might miss some of the redaction rules. You'll want to  apply the redactions to the array returned by `ancestors.map((el) => getElementProperties(el))` (line 142 above)

packages/plugin-autocapture-browser/src/hierarchy.ts
daniel-graham-amplitude: Note: I don't actually know what the desired behaviour is. Do we want this attribute to only redact attributes on it's descendants, or should it do it on itself too?

achhunna: We are redacting contents of the attribute, so will be its descendants not on the attribute itself so this suggestion is correct. We are matching what Heap/Pendo offer https://developers.heap.io/docs/ignoring-sensitive-data-and-pii#precise-data-redaction-via-heap-redact `The content of these attributes will be redacted before it leaves the browser.`

packages/plugin-autocapture-browser/src/hierarchy.ts
achhunna: Done

daniel-graham-amplitude: I just refreshed and I'm not seeing it yet. Let me refresh again...

daniel-graham-amplitude: Okay nvm I see what you did.

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -141,10 +141,57 @@ export const getText = (element: Element): string => {
...
+    if (redactValue) {
+      // Parse comma-separated attribute names and add to set
+      const attributesToRedact = parseAttributesToRedact(redactValue);
+      attributesToRedact.forEach((attr) => {
+        redactedAttributes.add(attr);
+      });
+    }
+    currentElement = currentElement.parentElement;
```
daniel-graham-amplitude: This can also be `currentElement = currentElement.closest(`[${constants.DATA_AMP_MASK_ATTRIBUTES}]`);`

daniel-graham-amplitude (approved):
Lgtm!

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: Done

packages/plugin-autocapture-browser/src/helpers.ts
Dogfalo: I wouldn't say this is always the case. accessing parentElement isn't an expensive operation and getAttribute has to be called anyways. So I would say it's actually extra work that needs to be done on the Selector parser

daniel-graham-amplitude: Yeah looking at it more, it probably doesn't make a huge difference. If you want to revert back to the parentElement way I'll accept it.

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: This was causing an infinite loop so changed to `currentElement = currentElement.parentElement?.closest(`[${constants.DATA_AMP_MASK_ATTRIBUTES}]`) || null;` to move up and find closest with `data-amp-mask-attributes` attribute.

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -122,6 +131,45 @@ export const getHierarchy = (element: Element | null): Hierarchy => {
...
+        node.attrs[DATA_AMP_MASK_ATTRIBUTES] = String(maskAttr).substring(0, MAX_ATTRIBUTE_LENGTH);
+      }
+    }
+  }
+
+  // Collect redacted attributes and apply redaction in a single pass
+  const redactedAttrs = new Set<string>();
+  const reversedHierarchy = [...hierarchy].reverse(); // root to target order
+
+  for (const node of reversedHierarchy) {
```
Dogfalo: This looks like we may run into issues if there are duplicate amp-mask for the same attribute at different levels.

update: nvm, was thinking this was building the masked attr array

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -66,9 +70,14 @@ export function getElementProperties(element: Element | null): HierarchyNode | n
...
   }
 
+  // Collect redacted attributes from element and ancestors first
+  const redactedAttrs = new Set<string>();
+
   const attributes: Record<string, string> = {};
   const attributesArray = Array.from(element.attributes);
-  const filteredAttributes = attributesArray.filter((attr) => !BLOCKED_ATTRIBUTES.includes(attr.name));
+  const filteredAttributes = attributesArray.filter(
+    (attr) => !BLOCKED_ATTRIBUTES.includes(attr.name) && !redactedAttrs.has(attr.name),
```
Dogfalo: redactedAttrs is always empty, so this is a noop

achhunna: rm

packages/plugin-autocapture-browser/src/hierarchy.ts
achhunna: refactored

packages/plugin-autocapture-browser/test/helpers.test.ts
```diff
@@ -10,6 +10,57 @@ import {
...
+        redactedAttributeNames.add(attr);
+      });
+    }
+    currentElement = currentElement.parentElement?.closest(`[${DATA_AMP_MASK_ATTRIBUTES}]`) || null;
+  }
+
+  return redactedAttributeNames;
+};
+
+const getAttributesWithPrefixAndAttributesWithPrefix = (
```
Dogfalo: Delete?

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+        elementToRedactedAttributesMap.set(node, allRedactedAttributes);
+      }
+    }
+
+    hierarchy = ancestors.map((el) =>
+      getElementProperties(el, elementToRedactedAttributesMap.get(el) ?? new Set<string>()),
+    );
+
+    // Mask value in attributes
+    for (const hieraryNode of hierarchy) {
```
daniel-graham-amplitude: Typo `hieraryNode` -> `hierarchyNode`

Dogfalo: spelling: hierarchyNode

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+  getHierarchy = (element: Element | null): Hierarchy => {
+    let hierarchy: Hierarchy = [];
+    if (!element) {
+      return [];
+    }
+
+    // Get list of ancestors including itself and get properties at each level in the hierarchy
+    const ancestors = getAncestors(element);
+
+    const elementToRedactedAttributesMap = new Map<Element, Set<string>>();
```
Dogfalo: elementToAttributesToMaskMap

Dogfalo: Add comment:

```
// Build attributes to mask map

File: packages/plugin-autocapture-browser/src/data-extractor.ts [hunk 11]
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+    const reversedAncestors = [...ancestors].reverse(); // root to target order
```
Dogfalo: We don't need to reverse the ancestors, we can just for..loop starting from the end.

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+
+    for (let i = 0; i < reversedAncestors.length; i++) {
+      const node = reversedAncestors[i];
+      if (node) {
+        const redactedAttributes = parseAttributesToRedact(node.getAttribute(constants.DATA_AMP_MASK_ATTRIBUTES));
```
Dogfalo: redactedAttributes -> attributesToMask

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+        const ancestorRedactedAttributes =
```
Dogfalo: ancestorRedactedAttributes -> ancestorAttributesToMask

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+          i === 0 ? [] : elementToRedactedAttributesMap.get(reversedAncestors[i - 1]) ?? new Set<string>();
+        const allRedactedAttributes = new Set([...ancestorRedactedAttributes, ...redactedAttributes]);
```
Dogfalo: allRedactedAttributes -> combinedAttributesToMask

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -83,6 +72,48 @@ export class DataExtractor {
...
+        elementToRedactedAttributesMap.set(node, allRedactedAttributes);
+      }
+    }
+
+    hierarchy = ancestors.map((el) =>
+      getElementProperties(el, elementToRedactedAttributesMap.get(el) ?? new Set<string>()),
+    );
+
+    // Mask value in attributes
```
Dogfalo: `// Search for and mask any sensitive attribute values`

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: fixed, good catch

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -109,29 +140,41 @@ export class DataExtractor {
...
+    // Add non-redacted attributes conditionally
+    // id is never redacted, so always include it
+    properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_ID] = element.getAttribute('id') || '';
+
+    // class is never redacted, so always include it
+    properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_CLASS] = element.getAttribute('class');
+
+    properties[constants.AMPLITUDE_EVENT_PROP_ELEMENT_ARIA_LABEL] = currentElementAttributes?.['aria-label'];
+
     if (tag === 'a' && actionType === 'click' && element instanceof HTMLAnchorElement) {
```
Dogfalo, line 173: Add comment about why we do not use the attribute value from hierarchy

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -92,11 +92,35 @@ export const isNonSensitiveElement = (element: Element) => {
...
+ * The 'id' and 'class' attributes cannot be redacted as they're critical for element identification
+ * @param element - The target element to check for redaction attributes
+ * @returns Set of attribute names that should be redacted
+ */
+/**
+ * Parses a comma-separated string of attribute names and filters out protected attributes
+ * @param attributeString - Comma-separated string of attribute names
+ * @returns Array of valid attribute names, excluding 'id' and 'class'
+ */
+export const parseAttributesToRedact = (attributeString: string | null): string[] => {
```
Dogfalo: parseAttributesToMask

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -92,11 +92,35 @@ export const isNonSensitiveElement = (element: Element) => {
...
+  return attributeString
+    ? attributeString
+        .split(',')
+        .map((attr) => attr.trim())
+        .filter((attr) => attr.length > 0)
```
Dogfalo: combine filters

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -92,11 +92,35 @@ export const isNonSensitiveElement = (element: Element) => {
...
+        .filter((attr) => attr !== 'id' && attr !== 'class') // Prevent 'id' and 'class' from being redacted as they're critical for element identification
+    : [];
+};
+
+export const getAttributesWithPrefix = (
```
Dogfalo: Name is a bit confusing now since the behavior has changed.

getAttributesWithPrefix -> extractPrefixedAttributes

packages/plugin-autocapture-browser/src/hierarchy.ts
```diff
@@ -27,15 +28,21 @@ const BLOCKED_ATTRIBUTES = [
   'data-reactid',
   'data-react-checksum',
   'data-reactroot',
-];
+
+  // Amplitude specific - used for redaction but should not be included in getElementProperties
+  DATA_AMP_MASK_ATTRIBUTES,
```
Dogfalo, line 33: add 
`TEXT_MASK_ATTRIBUTE`

Dogfalo (changes requested):

packages/plugin-autocapture-browser/test/helpers.test.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done, good call

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/data-extractor.ts
achhunna: done

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: done

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: done

packages/plugin-autocapture-browser/src/helpers.ts
achhunna: done

packages/plugin-autocapture-browser/src/hierarchy.ts
achhunna: done

PR #1265, opened 2025-08-21
fix(plugin-autocapture-browser): make rage click less noisy

daniel-graham-amplitude:
### Summary

Changes definition of rage click.

A rage click _starts_ when all conditions are true:
1. the same element has been clicked on 4 times 
2. the clicks were all within 50px
3. the first click and last click were within 1000ms

A rage click _ends_ when one condition is true:
1. a different element is clicked on
2. a click's coordinates are outside of 50px
3. 1000ms passes since the last click event
4. the most recent click event and the 4th most recent click event aren't within 1000ms

When a rage click ends, an `[Amplitude] Rage Click` event is fired, and all of the clicks that were buffered between the start and end of the rage click are used to make up the `clicks` and `click count` attributes.

### Checklist


packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
```diff
@@ -50,56 +127,51 @@ export function trackRageClicks({
...
-          return null;
+        // if we have enough clicks to be a rage click, set a timout to trigger the rage
+        // click event after the time threshold is reached.
+        // This will be cancelled if a new click is tracked.
+        if (clickWindow.length >= RAGE_CLICK_THRESHOLD) {
+          triggerRageClickTimeout = setTimeout(() => {
+            const { rageClickEvent, time } = getRageClickAnalyticsEvent(clickWindow);
+            amplitude.track(AMPLITUDE_ELEMENT_RAGE_CLICKED_EVENT, rageClickEvent, { time });
+            resetClickWindow();
+          }, RAGE_CLICK_WINDOW_MS);
```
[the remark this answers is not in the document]
daniel-graham-amplitude: It's not the first click we're interested in, it's that we want to wait the whole 1000ms after the end of the rage click window, so that we know for sure that the rage click window is closed. And it would only trigger once after unsubscribe, so not a huge problem there.

packages/analytics-core/src/types/frustration-interactions.ts
```diff
@@ -107,6 +107,11 @@ export const DEFAULT_RAGE_CLICK_WINDOW_MS = 1_000;
  */
 export const DEFAULT_RAGE_CLICK_THRESHOLD = 4;
 
+/**
+ * Default threshold for rage clicks to be considered out of bounds (15 pixels)
+ */
+export const DEFAULT_RAGE_CLICK_OUT_OF_BOUNDS_THRESHOLD = 15; // pixels
```
jxiwang: Is this consistent with the other platforms?

daniel-graham-amplitude: Shoot, that should be 50px. Let me fix that.

50px is what iOS uses.

packages/plugin-autocapture-browser/src/autocapture/track-rage-click.ts
```diff
@@ -29,6 +34,66 @@ type ClickEvent = {
...
+  yMax?: number;
+  xMin?: number;
+  xMax?: number;
+  isOutOfBounds?: boolean;
+};
+
+function addCoordinates(regionBox: ClickRegionBoundingBox, click: ClickEvent) {
+  const { clientX, clientY } = click.event as MouseEvent;
+  if (regionBox.yMin === undefined || clientY < regionBox.yMin) {
+    regionBox.yMin = clientY;
```
Dogfalo: minor: think this logic is more readable as 

```js
regionBox.yMin = Math.min(regionBox.yMin ?? clientY, clientY);
regionBox.yMax = Math.max(regionBox.yMax ?? clientY, clientY);
regionBox.xMin = Math.min(regionBox.xMin ?? clientX, clientX);
regionBox.xMax = Math.max(regionBox.xMax ?? clientX, clientX);
```

daniel-graham-amplitude: Updated to that, thanks. That is better.

jxiwang (approved):

PR #1273, opened 2025-08-26
chore: test default amplitude is BrowserClient

daniel-graham-amplitude:
### Summary

Add test that verifies that the export of "analytics-browser" index is instance of BrowserClient (structural typing test).

Methodology: Create a "FakeBrowserClient" that implements "BrowserClient" and then check that the methods in "analytics-browser" default export match methods in `BrowserClient.prototype`

### Checklist


Mercy811 (approved):
Thanks Dan LGTM!

PR #1276, opened 2025-08-26
fix(analytics-core): handle XHR responseType JSON

daniel-graham-amplitude:
#### Problem
The [responseText](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText) attribute in XHR is a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) that throws an [Invalid State Error](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText#exceptions) if the [XMLHttpRequest.prototype.responseType](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType) is not one of `'string'` or `''`. When the `responseType` is a type other than `string` or `''` (like `json`), then the body is accessed via the [response](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response) property.

_Note: I never noticed this because I tested on Axios which only uses strings; and the error is caught, it just prevents body capturing when responseType is JSON_

#### Solution
* Instead of passing [XMLHttpRequest.prototype.responseText](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText) property to the downstream `ResponseXhrWrapper`, pass a function that returns a JSON object (if the body is JSON). 
  * If `responseType=="json"` returns a clone of the [response](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response). 
   * If `responseType==='text || ''` then returns JSON.parse of the text. If not valid JSON, return `null`.
  * Otherwise return `null`

### Checklist


daniel-graham-amplitude:
bugbot run

packages/analytics-core/src/network-observer.ts
```diff
@@ -254,6 +255,46 @@ export class NetworkObserver {
...
+          // if response is a JS object, clone it so that subscribers can't mutate it
+          if (context.globalScope?.structuredClone) {
+            /* eslint-disable-next-line @typescript-eslint/no-unsafe-return */
+            return context.globalScope.structuredClone(xhrUnsafe.response);
+          }
+        } else if (['text', ''].includes(xhrUnsafe.responseType)) {
+          // if response is a string, parse it as JSON
+          /* eslint-disable-next-line @typescript-eslint/no-unsafe-return */
+          return JSON.parse(xhrUnsafe.responseText);
+        }
```
Mercy811, line 281: Not a blocker, mainly a question. response type can be "document", "blob" too https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType#value. Should we add an else block for those values and add a debug log?

Mercy811 (approved):
LGTM!

packages/analytics-core/src/network-observer.ts
daniel-graham-amplitude: Yeah that was on purpose. We only capture JSON response body, and text and JSON are the only 2 response types that can be JSON Serializable. The other types we would just ignore.

If I have a capture rule that looks like this, for example:

```
{
  "urls": "https://api.amplitude.com/api/endpoint/*",
  "responseBody": { "allowlist": ["query"] }
}
```

What I'm telling it to do is capture URLs that match the pattern `https://api.amplitude.com/api/endpoint/*` and if they have a JSON body, capture the `query` attribute (if it exists).

If the endpoint returns a non JSON parseable type, then I would say that it falls under the category of being a response that doesn't have the `query` attribute, and doesn't warrant having an error/debug message.

PR #1277, opened 2025-08-29
chore: playwright remote config smoke tests

daniel-graham-amplitude:
### Summary

Add a bash script that runs a round of playwright tests against the local server. These tests test that the remote config endpoint is being called and that after calling "init" the page is still interactable. They run across a matrix of different API keys and server zones (staging + us)

### Checklist


packages/e2e-remote-config/remote-config-test.sh
Mercy811, line 1: could you add some comments about why those versions are chosen or save them as a table in readme?

playwright.config.ts
```diff
@@ -36,10 +36,10 @@ export default defineConfig({
...
     //   },
     // },
-    {
-      name: 'webkit',
-      use: { ...devices['Desktop Safari'] },
-    },
+    // {
+    //   name: 'webkit',
+    //   use: { ...devices['Desktop Safari'] },
+    // },
```
Mercy811: this is the root level playwright config. If we comment this, the e2e GHA won't run on safari. Can we keep this uncomment?

package.json
```diff
@@ -41,7 +42,8 @@
   "devDependencies": {
     "@commitlint/cli": "^17.3.0",
     "@commitlint/config-conventional": "^17.3.0",
-    "@playwright/test": "^1.52.0",
+    "@nx/workspace": "^21.2.1",
+    "@playwright/test": "1.55.0",
```
Mercy811, line 46: nit: why are we using a pined version instead of version range?

packages/e2e-remote-config/test/e2e/fetch-remote-config.spec.ts
Mercy811, line 1: question: It looks like we only test remote config is fetched. Do we have plan to test the merged config of local and remote and the SDK behaves as expected?

packages/e2e-remote-config/test/e2e/fetch-remote-config.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { test, expect } from '@playwright/test';
+
+/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return, no-restricted-globals */
+
+// TODO: make this settable via env variable
+const remoteConfigUrl = 'https://sr-client-cfg.amplitude.com/config*';
```
Mercy811: question: the other one should just be the staging url right?

package.json
daniel-graham-amplitude: The reason for that is because the Docker container where the tests run from is pinned to a specific version so this ensures that they're on the same version.

packages/e2e-remote-config/test/e2e/fetch-remote-config.spec.ts
daniel-graham-amplitude: Yeah, I totally did this wrong it's fixed now.

But yeah the idea is that the page will always request from `https://sr-client-cfg.amplitude.com/` and if you want to run it elsewhere, set an environment variable.

packages/e2e-remote-config/test/e2e/fetch-remote-config.spec.ts
daniel-graham-amplitude: I didn't have any plans, I get concerned about adding too much to an E2E test suite because it can cause brittleness and flakiness. My original thinking for this task was just that we check that 1. the remote fetch was called; 2. the remote fetch doesn't cause the application to blow up.

Mercy811 (approved):
LGTM!

PR #1281, opened 2025-09-08
feat(analytics-core): add diagnostics client

Mercy811:
### Summary

Add a diagnostics client with APIs: 
- setTag() for tag
- increment() for counter
- recordHistogram for histogram
- recordEvent for event. This will be added later after the diagnostic server supports events in the request payload

The client has two storages
- memory storage as cache
  - auto save to persistent storage every 1 second if there is any data
- indexedDB as persistent storage
  - auto flush every 5 minutes if there is any data
 
The client also flushes on initialization when last flush time stored in indexedDB has passed 5 minutes.

The code is based on the diagnostics client design [here](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3270639769/Diagnostics+client+design).

I'm heavily using `istanbul ignore next` in diagnostics-storage for test coverage as it's mainly just indexddb APIs call. I will set up a e2e test to further test them

### Checklist

### DIFF
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -137,7 +137,8 @@ export interface IDiagnosticsClient {
    * });
    * ```
    */
-  recordEvent(name: string, properties: EventProperties): void;
+  // TODO(AMP-139569)
+  // recordEvent(name: string, properties: EventProperties): void;
 
   // Flush storage
   _flush(): void;
@@ -293,6 +294,17 @@ export class DiagnosticsClient implements IDiagnosticsClient {
     //   time: record.time,
     //   event_properties: record.event_properties,
     // }));
+
+    // Early return if all data collections are empty
+    if (
+      Object.keys(tags).length === 0 &&
+      Object.keys(counters).length === 0 &&
+      Object.keys(histogram).length === 0
+      // TODO(AMP-139569)
+      // && Object.keys(events).length === 0
+    ) {
+      return;
+    }
 
     // Create flush payload
     const payload: FlushPayload = {
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -1,9 +1,7 @@
 /// <reference lib="dom" />
-import { DiagnosticsStorage } from './diagnostics-storage';
+import { DiagnosticsStorage, IDiagnosticsStorage } from './diagnostics-storage';
 
-// Flush interval: 1 hour in milliseconds
-const FLUSH_INTERVAL_MS = 60 * 60 * 1000; // 1 hour
-
+const FLUSH_INTERVAL_MS = 5 * 1000; // 5 minutes
 
 // === Core Data Types ===
 
@@ -30,8 +28,6 @@ interface DiagnosticsEvent {
   readonly time: number;
   readonly event_properties: EventProperties;
 }
-
-// === Histogram Types ===
 
 /**
  * Computed histogram statistics for final payload
@@ -84,7 +80,7 @@ interface FlushPayload {
  *
  * Key Features:
  * - IndexedDB storage
- * - Time-based persistent storage flush interval (1 hour since last flush)
+ * - Time-based persistent storage flush interval (5 minutes since last flush)
  * - 1 second time-based memory storage flush interval
  * - Histogram statistics calculation (min, max, avg)
  * - Thread-safe operations through async/await
@@ -109,7 +105,6 @@ interface FlushPayload {
  * // Record diagnostic events
  * await diagnostics.recordEvent('error', {
  *   stack_trace: '...',
- *   api_key: 'hidden'
  * });
  *
  * // Manual flush (auto-flushes every 1 hour)
@@ -136,8 +131,17 @@ export interface IDiagnosticsClient {
 }
 
 export class DiagnosticsClient implements IDiagnosticsClient {
-  private storage: DiagnosticsStorage;
+  private storage: IDiagnosticsStorage;
   private flushTimer: ReturnType<typeof setTimeout> | null = null;
+
+  // In-memory storages
+  private inMemoryTags: DiagnosticsTags = {};
+  private inMemoryCounters: DiagnosticsCounters = {};
+  private inMemoryHistograms: DiagnosticsHistogramStats = {};
+  private inMemoryEvents: DiagnosticsEvent[] = [];
+
+  // Global timer for 1-second persistence
+  private globalSaveTimer: ReturnType<typeof setTimeout> | null = null;
 
   constructor(apiKey: string) {
     this.storage = new DiagnosticsStorage(apiKey);
@@ -145,39 +149,77 @@ export class DiagnosticsClient implements IDiagnosticsClient {
   }
 
   async setTag(name: string, value: string): Promise<void> {
-    try {
-      await this.storage.setTag(name, value);
-    } catch (error) {
-      console.error('[Amplitude] DiagnosticsClient: Failed to set tag', error);
-    }
+    this.inMemoryTags[name] = value;
+    this.startSaveTimerIfNeeded();
   }
 
   async increment(name: string, size = 1): Promise<void> {
-    try {
-      // Get existing counter or create new one
-      const existingCounter = await this.storage.getCounter(name);
-      const currentValue = existingCounter ? existingCounter.value : 0;
-
-      await this.storage.setCounter(name, currentValue + size);
-    } catch (error) {
-      console.error('[Amplitude] DiagnosticsClient: Failed to increment counter', error);
-    }
+    this.inMemoryCounters[name] = (this.inMemoryCounters[name] || 0) + size;
+    this.startSaveTimerIfNeeded();
   }
 
   async recordHistogram(name: string, value: number): Promise<void> {
-    try {
-      // Add the new record to IndexedDB
-      await this.storage.addHistogramRecord(name, value);
-    } catch (error) {
-      console.error('[Amplitude] DiagnosticsClient: Failed to record histogram', error);
+    const existing = this.inMemoryHistograms[name];
+    if (existing) {
+      // Update existing stats incrementally
+      existing.count += 1;
+      existing.min = Math.min(existing.min, value);
+      existing.max = Math.max(existing.max, value);
+      existing.sum += value;
+    } else {
+      // Create new stats
+      this.inMemoryHistograms[name] = {
+        count: 1,
+        min: value,
+        max: value,
+        sum: value,
+      };
     }
+    this.startSaveTimerIfNeeded();
   }
 
   async recordEvent(name: string, properties: EventProperties): Promise<void> {
+    this.inMemoryEvents.push({
+      event_name: name,
+      time: Date.now(),
+      event_properties: properties,
+    });
+    this.startSaveTimerIfNeeded();
+  }
+
+  private startSaveTimerIfNeeded(): void {
+    if (!this.globalSaveTimer) {
+      this.globalSaveTimer = setTimeout(() => {
+        void this.saveAllDataToStorage();
+      }, 1000); // Save every 1 second
+    }
+  }
+
+  private async saveAllDataToStorage(): Promise<void> {
     try {
-      await this.storage.addEventRecord(name, properties);
+      // Make copies and clear immediately to avoid race conditions
+      const tagsToSave = { ...this.inMemoryTags };
+      const countersToSave = { ...this.inMemoryCounters };
+      const histogramsToSave = { ...this.inMemoryHistograms };
+      const eventsToSave = [...this.inMemoryEvents];
+
+      // Clear in-memory data immediately
+      this.inMemoryEvents = [];
+      this.inMemoryTags = {};
+      this.inMemoryCounters = {};
+      this.inMemoryHistograms = {};
+
+      // Clear the timer since we're processing the data
+      this.globalSaveTimer = null;
+
+      await Promise.all([
+        this.storage.setTags(tagsToSave),
+        this.storage.setCounters(countersToSave),
+        this.storage.setHistogramStats(histogramsToSave),
+        this.storage.addEventRecords(eventsToSave),
+      ]);
     } catch (error) {
-      console.error('[Amplitude] DiagnosticsClient: Failed to record event', error);
+      console.error('[Amplitude] DiagnosticsClient: Failed to save data to storage', error);
     }
   }
 
@@ -189,7 +231,7 @@ export class DiagnosticsClient implements IDiagnosticsClient {
       // Get all stored data from IndexedDB
       const tagRecords = await this.storage.getAllTags();
       const counterRecords = await this.storage.getAllCounters();
-      const histogramRecords = await this.storage.getAllHistogramRecords();
+      const histogramStatsRecords = await this.storage.getAllHistogramStats();
       const eventRecords = await this.storage.getAllEventRecords();
 
       // Convert records to the expected format
@@ -203,8 +245,16 @@ export class DiagnosticsClient implements IDiagnosticsClient {
         counters[record.key] = record.value;
       });
 
-      // Calculate histogram results from records
-      const histogram = this.calculateHistogramResults(histogramRecords);
+      // Convert histogram stats to histogram results
+      const histogram: DiagnosticsHistograms = {};
+      histogramStatsRecords.forEach((stats) => {
+        histogram[stats.key] = {
+          count: stats.count,
+          min: stats.min,
+          max: stats.max,
+          avg: stats.count > 0 ? Math.round((stats.sum / stats.count) * 100) / 100 : 0,
+        };
+      });
 
       // Convert event records to the expected format
       const events: DiagnosticsEvent[] = eventRecords.map((record) => ({
@@ -223,6 +273,9 @@ export class DiagnosticsClient implements IDiagnosticsClient {
 
       // Clear all data after successful flush preparation
       await this.clearAllData();
+
+      // Clear in-memory data
+      this.clearAllInMemoryData();
 
       // Update the last flush timestamp
       await this.storage.setLastFlushTimestamp(Date.now());
@@ -243,47 +296,17 @@ export class DiagnosticsClient implements IDiagnosticsClient {
     }
   }
 
-  private calculateHistogramResults(records: Array<{key: string, value: number}>): DiagnosticsHistograms {
-    const histogramMap: { [name: string]: number[] } = {};
+  private clearAllInMemoryData(): void {
+    this.inMemoryTags = {};
+    this.inMemoryCounters = {};
+    this.inMemoryHistograms = {};
+    this.inMemoryEvents = [];
 
-    // Group values by histogram name
-    records.forEach((record) => {
-      if (!histogramMap[record.key]) {
-        histogramMap[record.key] = [];
-      }
-      histogramMap[record.key].push(record.value);
-    });
-
-    // Calculate statistics for each histogram
-    const results: DiagnosticsHistograms = {};
-    Object.keys(histogramMap).forEach((name) => {
-      const values = histogramMap[name];
-      if (values.length > 0) {
-        results[name] = this.calculateStatistics(values);
-      }
-    });
-
-    return results;
-  }
-
-  private calculateStatistics(values: number[]): HistogramResult {
-    if (values.length === 0) {
-      return { count: 0, min: 0, max: 0, avg: 0 };
+    // Clear the global save timer
+    if (this.globalSaveTimer) {
+      clearTimeout(this.globalSaveTimer);
+      this.globalSaveTimer = null;
     }
-
-    const sortedValues = [...values].sort((a, b) => a - b);
-    const count = values.length;
-    const min = sortedValues[0];
-    const max = sortedValues[count - 1];
-    const sum = values.reduce((acc, val) => acc + val, 0);
-    const avg = Math.round((sum / count) * 100) / 100; // Round to 2 decimal places
-
-    return {
-      count,
-      min,
-      max,
-      avg,
-    };
   }
 
   private async clearAllData(): Promise<void> {
@@ -296,24 +319,17 @@ export class DiagnosticsClient implements IDiagnosticsClient {
 
   /**
    * Initialize flush interval logic.
-   * Check if 1 hour has passed since last flush, if so flush immediately.
+   * Check if 5 minutes has passed since last flush, if so flush immediately.
    * Otherwise set a timer to flush when the interval is reached.
    */
   private async initializeFlushInterval(): Promise<void> {
     try {
-      const lastFlushTimestamp = await this.storage.getLastFlushTimestamp();
       const now = Date.now();
-
-      if (!lastFlushTimestamp) {
-        // First time - flush immediately to establish initial timestamp
-        await this.flushAndUpdateTimestamp();
-        return;
-      }
-
+      const lastFlushTimestamp = (await this.storage.getLastFlushTimestamp()) || now;
       const timeSinceLastFlush = now - lastFlushTimestamp;
 
       if (timeSinceLastFlush >= FLUSH_INTERVAL_MS) {
-        // More than 1 hour has passed, flush immediately
+        // More than 5 minutes has passed, flush immediately
         await this.flushAndUpdateTimestamp();
       } else {
         // Set timer for remaining time
@@ -354,30 +370,28 @@ export class DiagnosticsClient implements IDiagnosticsClient {
       await this._flush();
     } catch (error) {
       console.error('[Amplitude] DiagnosticsClient: Failed to flush and update timestamp', error);
-      // Set timer for retry on error
-      this.setFlushTimer(FLUSH_INTERVAL_MS);
     }
   }
 
   // Utility method to get current data for debugging (not part of interface)
   async _getCurrentData() {
+    // Ensure latest data is saved to storage first
+    await this.saveAllDataToStorage();
+
     return {
       tags: await this.storage.getAllTags(),
       counters: await this.storage.getAllCounters(),
-      histogramEntries: await this.storage.getAllHistogramRecords(),
+      histogramStats: await this.storage.getAllHistogramStats(),
       events: await this.storage.getAllEventRecords(),
+      inMemoryTags: this.inMemoryTags,
+      inMemoryCounters: this.inMemoryCounters,
+      inMemoryHistograms: this.inMemoryHistograms,
+      inMemoryEvents: this.inMemoryEvents,
     };
   }
 
-  // Utility method to get histogram values for a specific key, sorted by value
-  // This leverages the compound index [key, value] for efficient querying
-  async _getHistogramValuesSorted(key: string): Promise<number[]> {
-    try {
-      const records = await this.storage.getHistogramRecordsSortedByValue(key);
-      return records.map((record) => record.value);
-    } catch (error) {
-      console.error('[Amplitude] DiagnosticsClient: Failed to get sorted histogram values', error);
-      return [];
-    }
+  // Utility method to get current histogram stats for a specific key
+  _getHistogramStats(key: string): HistogramStats | undefined {
+    return this.inMemoryHistograms[key];
   }
 }
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -4,7 +4,7 @@ import { DiagnosticsStorage, IDiagnosticsStorage } from './diagnostics-storage';
 import { ServerZoneType } from 'src/types/server-zone';
 
 const SAVE_INTERVAL_MS = 1000; // 1 second
-const FLUSH_INTERVAL_MS = 5 * 1000; // 5 minutes
+const FLUSH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
 const US_SERVER_URL = 'https://diagnostics.prod.us-west-2.amplitude.com/v1/capture';
 const EU_SERVER_URL = 'https://diagnostics.prod.eu-central-1.amplitude.com/v1/capture';
 
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -1,7 +1,7 @@
 /// <reference lib="dom" />
-import { ILogger } from 'src/logger';
+import { ILogger } from '../logger';
 import { DiagnosticsStorage, IDiagnosticsStorage } from './diagnostics-storage';
-import { ServerZoneType } from 'src/types/server-zone';
+import { ServerZoneType } from '../types/server-zone';
 
 const SAVE_INTERVAL_MS = 1000; // 1 second
 const FLUSH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -1,4 +1,3 @@
-/// <reference lib="dom" />
 import { ILogger } from '../logger';
 import { DiagnosticsStorage, IDiagnosticsStorage } from './diagnostics-storage';
 import { ServerZoneType } from '../types/server-zone';
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -306,6 +306,11 @@ export class DiagnosticsClient implements IDiagnosticsClient {
     if (!this.storage) {
       return;
     }
+
+    await this.saveAllDataToStorage();
+    this.saveTimer = null;
+    this.flushTimer = null;
+
     // Get all data from storage and clear it
     const {
       tags: tagRecords,
--- a/packages/analytics-core/src/diagnostics/diagnostics-client.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-client.ts
@@ -3,6 +3,7 @@ import { ILogger } from 'src/logger';
 import { DiagnosticsStorage, IDiagnosticsStorage } from './diagnostics-storage';
 import { ServerZoneType } from 'src/types/server-zone';
 
+const SAVE_INTERVAL_MS = 1000; // 1 second
 const FLUSH_INTERVAL_MS = 5 * 1000; // 5 minutes
 const US_SERVER_URL = 'https://diagnostics.prod.us-west-2.amplitude.com/v1/capture';
 const EU_SERVER_URL = 'https://diagnostics.prod.eu-central-1.amplitude.com/v1/capture';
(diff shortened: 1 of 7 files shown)


packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,435 @@
...
+              resolve();
+            }
+          };
+
+          request.onerror = () => reject(new Error('Failed to set counters'));
+        });
+      });
+    } catch (error) {
+      this.logger.debug('DiagnosticsStorage: Failed to set counters', error);
+    }
```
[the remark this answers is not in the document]
Mercy811: metric data type formats are slight different, aim for clarity and take duplication as trade offs

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,392 @@
...
+      this.logger?.error('DiagnosticsClient: Failed to save data to storage', error);
+    }
+  }
+
+  _flush() {
+    // Clear the current timer since we're flushing now
+    this.clearFlushTimer();
+
+    // Perform flush operations asynchronously without blocking
+    void this.performFlushOperations();
```
daniel-graham-amplitude: I don't think this will prevent blocking. This will enqueue the`peformFlushOperations` function onto the microtask queue (similar to how the [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide) function works). And then when the current **macro**task is complete (ie: the callstack is cleared), it will immediately start dequeuing all the tasks on the microtask queue, and run them one-by-one, on the main thread. The only way to perform this operation in a "non-blocking" environment is to run this from a Web Worker.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,392 @@
...
+        tags: tagRecords,
+        counters: counterRecords,
+        histogramStats: histogramStatsRecords,
+        events: eventRecords,
+      } = await this.storage.getAllAndClear();
+
+      // Update the last flush timestamp
+      void this.storage.setLastFlushTimestamp(Date.now());
+
+      // Convert metrics to the expected format
```
daniel-graham-amplitude: Is there a reason that we have to "convert" the metrics? Could we not just track the metrics in the same form that we want them to be when we send them?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude: Another option is [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) which will wait for the main thread to be idle before calling an operation. According to docs: _ This enables developers to perform background and low priority work on the main thread, without impacting latency-critical events such as animation and input response_

It's imperfect because even if you call it while the thread is "idle" it's still going to cause blocking of the main thread, so if the user tries interacting with the document while it's running, they'll see performs blips. But still better than just calling it right away.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,435 @@
...
+  async getDB(): Promise<IDBDatabase> {
+    if (!this.dbPromise) {
+      this.dbPromise = this.openDB();
+    }
+    return this.dbPromise;
+  }
+
+  openDB(): Promise<IDBDatabase> {
+    return new Promise((resolve, reject) => {
+      if (typeof indexedDB === 'undefined') {
```
daniel-graham-amplitude: I think we should get into the habit of always using `globalScope` when referencing variables like this one. `globalScope.indexedDB === 'undefined'`.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: It simplifies calculation as we have to update avg = (avg * count + newValue) / (count + 1) and do the same when saving to storage.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: synced offline, when running an non-await async function, it will execute synchronously until the await part which is queued up in the microtask queue

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: updated to use getGlobalScope()

test-server/diagnostics-test.html
```diff
@@ -0,0 +1,346 @@
...
+            margin-left: 10px;
+        }
+    </style>
+</head>
+<body>
+    <div class="container">
+        <h1>🔬 Diagnostics Client Test</h1>
+        
+        <div class="network-info">
+            <strong>API Key:</strong> 5b9a9510e261f9ead90865bbc5a7ad1d<br>
```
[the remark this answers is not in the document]
Mercy811: fixed

test-server/diagnostics-test.html
```diff
@@ -0,0 +1,346 @@
...
+            }
+        }
+
+        // Initialize diagnostics client
+        window.initializeDiagnostics = function() {
+            try {
+                logger = new UILogger();
+                logger.enable(LogLevel.Debug); // Enable debug level logging
+                
+                diagnosticsClient = new DiagnosticsClient('5b9a9510e261f9ead90865bbc5a7ad1d', logger);
```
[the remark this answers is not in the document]
Mercy811: fixed

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,377 @@
...
+    const payload: FlushPayload = {
+      tags,
+      histogram,
+      counters,
+      // TODO(AMP-139569)
+      // events,
+    };
+
+    // Send payload to diagnostics server
+    void this.fetch(payload);
```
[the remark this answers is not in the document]
Mercy811: all errors are caught by this.fetch()

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,377 @@
...
+  constructor(apiKey: string, logger: ILogger, serverZone: ServerZoneType = 'US') {
+    this.apiKey = apiKey;
+    this.logger = logger;
+    this.serverUrl = serverZone === 'US' ? DIAGNOSTICS_US_SERVER_URL : DIAGNOSTICS_EU_SERVER_URL;
+    this.storage = new DiagnosticsStorage(apiKey, logger);
+    void this.initializeFlushInterval();
+  }
+
+  setTag(name: string, value: string) {
+    this.inMemoryTags[name] = value;
```
daniel-graham-amplitude, line 192: Is there a risk that these `inMemoryTags` could just keep growing and cause a memory leak?

If there is, a simple solution would just be to limit how big this array can get (like a Circular Buffer/Ring Buffer)

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,377 @@
...
+  inMemoryCounters: DiagnosticsCounters = {};
+  inMemoryHistograms: DiagnosticsHistogramStats = {};
+  inMemoryEvents: DiagnosticsEvent[] = [];
+
+  // Timer for 1-second persistence
+  saveTimer: ReturnType<typeof setTimeout> | null = null;
+  // Timer for flush interval
+  flushTimer: ReturnType<typeof setTimeout> | null = null;
+
+  constructor(apiKey: string, logger: ILogger, serverZone: ServerZoneType = 'US') {
```
daniel-graham-amplitude, line 168: Will this class be supported in every browser context? Do we need to do an `isSupported` check, maybe for an environment that doesn't support indexedDb?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,377 @@
...
+    // Send payload to diagnostics server
+    void this.fetch(payload);
+  }
+
+  /**
+   * Send diagnostics data to the server
+   */
+  async fetch(payload: FlushPayload) {
+    try {
+      const response = await fetch(this.serverUrl, {
```
daniel-graham-amplitude, line 388: 1. Let's add `globalScope.fetch` here
2. We should add some retry logic if the fetch fails. I would say we should retry a certain number of times (using exponential backoff), and if it keeps failing, then we just shut down the diagnostics client.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude, line 376: I don't think we need "fetch" to be it's own method, I think inlining the contents of `async fetch(payload){ ... }` here would have the same effect.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,377 @@
...
+        method: 'POST',
+        headers: {
+          'X-ApiKey': this.apiKey,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(payload),
+      });
+
+      if (!response.ok) {
```
daniel-graham-amplitude, line 397: Are there any response status codes that would indicate that the diagnostics service is down? If we run into any of those, than we ought to stop sending requests to it.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,566 @@
...
+          resolve();
+          return;
+        }
+
+        const handleCompletion = () => {
+          completed++;
+          if (completed === entries.length) {
+            resolve();
+          }
+        };
```
daniel-graham-amplitude, line 207: Since you're making use of [transactions](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/transaction), I think you can replace `handleCompletion` with

```
transaction.oncomplete = () => resolve();
```

`oncomplete` I believe will only be triggered once all of the transactions are complete

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,566 @@
...
+
+        entries.forEach(([key, value]) => {
+          const request = store.put({ key, value });
+
+          request.onsuccess = handleCompletion;
+
+          /* istanbul ignore next */
+          request.onerror = () => {
+            this.logger.debug('DiagnosticsStorage: Failed to set tag', key, value);
+            handleCompletion();
```
daniel-graham-amplitude: If we want the transaction to keep going, even on failure, we'll need to do this:

```
request.onerror = (evt) => {
   ...
   evt.preventDefault();
};
```

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,566 @@
...
+      const db = await this.getDB();
+      const transaction = db.transaction([TABLE_NAMES.TAGS], 'readwrite');
+      const store = transaction.objectStore(TABLE_NAMES.TAGS);
+
+      return new Promise((resolve) => {
+        let completed = 0;
+        const entries = Object.entries(tags);
+
+        if (entries.length === 0) {
+          resolve();
```
daniel-graham-amplitude, line 206: We should move this conditional up to before you call `this.getDB()`. If there's no entries to set, then let's skip dealing with the indexeddb altogether.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: To elaborate, I mean that if one of the puts fails, it will rollback the whole transaction.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,566 @@
...
+
+        // First, read all existing values
+        entries.forEach(([key, incrementValue]) => {
+          const getRequest = store.get(key);
+
+          getRequest.onsuccess = () => {
+            const existingRecord = getRequest.result as CounterRecord | undefined;
+            /* istanbul ignore next */
+            const existingValue = existingRecord ? existingRecord.value : 0;
+            updatedCounters[key] = existingValue + incrementValue;
```
daniel-graham-amplitude: Are you certain that there needs to be one read phase and one write phase?

Couldn't you just call `store.put({ key, value: existingValue + incrementValue })` instead of setting `updatedCounters`?

All these 'gets' and 'sets' will then be batched together as a "transaction" which can resolve the promise in `transaction.onsuccess`.

daniel-graham-amplitude:
```typescript
transaction.onsuccess = () => resolve();

// ...

const getRequest = store.get(key);
getRequest.onsuccess = () => {
    const existingRecord = getRequest.result as CounterRecord | undefined;
    /* istanbul ignore next */
    const existingValue = existingRecord ? existingRecord.value : 0;
    store.put(key, existingValue + incrementValue);
}
```

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,566 @@
...
+          pendingWrites = entries.length;
+
+          entries.forEach(([key]) => {
+            const stats = updatedHistograms[key];
+            const putRequest = store.put(stats);
+
+            putRequest.onsuccess = handleWriteCompletion;
+
+            /* istanbul ignore next */
+            putRequest.onerror = () => {
```
daniel-graham-amplitude: I'd like to give more thought about what to do with an error. It seems like if there's an error, the whole db is probably unusable and we need to just abort completely?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: memory storage will be flushed to persistent storage in 1 sec and cleaned.

daniel-graham-amplitude: Okay then, is there a limit to how much will be stored in persistent storage?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: it's available for 95+% users https://caniuse.com/indexeddb.`DiagnosticsStorage` already handles the case where indexedDB is not available. In this case, less than 5%, no persistent storage and no diagnostics events will be tracked. I'm less worried about this as remote config client v1 had used indexedDB for a while too.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: 1. will use `getGlobalScope()`
2. we've agreed on client design doc that diagnostics client doesn't retry on network failure, it simply drops

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: IMHO, it's cleaner to bring fetch out of flush

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: So far no, since there is no retry, it should send at most 1 request every 5 minutes.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude: I think it's fine to miss 5% of users as this is a feature that doesn't effect users experience. But for those 5% we should check for support earlier so that our script doesn't do any unnecessary work.

Right now, what would happen in browsers where `IndexedDB` is not available:
1. `getDB()` will be rejected and log a debug message every time it's rejected (which is every time data is synced to IndexedDB`)
2. tags, histograms, etc... will still keep being saved in memory even though there's no possibility of them being synced to disk

A pattern I've seen for this is to just add a static method to the class

```
DiagnosticsClient.isSupported = () => { ...};
```

And then don't instantiate this class unless it's supported:

```
if (!DiagnosticsClient.iSupported) {
  return;
}
const diagnosticsClient = new DiagnosticsClient();
```

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: there is a cap on 10 events but no limit on other metrics as they are all small.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: Also, I just noticed that `tags` is of type `Record<string, string>` but we're checking for `entries.length`. Did you perhaps mean `Object.keys(tags).length === 0`?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: Sounds good. I agree we should check `DiagnosticsStorage.isSupported()` in diagnostics client constructor.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: Synced on zoom. It makes sense to set a memory limit on every metric as setTimer put the task into a marco queue and a better way to interpret setTimer is to read it as perform a task in at least timeout.

Example where memory can grow larger than we expected:
```
for loop running for 2 seconds
	diagnosticClient.setTags()
```

Limit for each metric is 100 except for events

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: Dan's main concern is that if the diagnostics server is down, customers with monitoring system like Sentry will notice a high volume of diagnostics request failure. It's safe to tell them to mute as it won't affect event tracking. For http v2 api, the SDK is sending a request every 1 second with linear backoff and max retry 5, which would cause even higher volume if the http v2 server is down. 

Rate limit will be covered in tomorrow's weekly sync.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: I misread this, I agree weather fetch is a function or not will not make a difference on execution time.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: > Also, I just noticed that tags is of type Record<string, string> but we're checking for entries.length. Did you perhaps mean Object.keys(tags).length === 0?

there should be no difference on checking entries or keys. 
image

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest/error. Request errors are two types 1. db level, like quota exceeded error and version error 2. request level, like unknown error, not readable error, and abort error (actually transaction level).

We can abort on 1) and all other errors and continue on 2)

UPD:
I'm re-thinking this. It's safer to just abort the whole transaction regardless of error types. Even if it's a request level error, it highly likely that the next request will hit unknown error again when memory is low

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: listen for abort instead as it's suggested by https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/error_event

And no preventDefault as we want to abort the whole transaction when a request fails because it's likely the next request will also fail

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: Okay that makes sense!

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: Updated to use transaction oncomplete to resolve the promise

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: Updated to batch request into a single transaction and use transaction's oncomplete to resolve the promise

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,494 @@
...
+
+    // Create events table
+    if (!db.objectStoreNames.contains(TABLE_NAMES.EVENTS)) {
+      const eventsStore = db.createObjectStore(TABLE_NAMES.EVENTS, {
+        keyPath: 'id',
+        autoIncrement: true,
+      });
+
+      // Create index on time for chronological queries
+      eventsStore.createIndex('time_idx', 'time', { unique: false });
```
daniel-graham-amplitude, line 183: Are we doing anything with this "time_idx"?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,494 @@
...
+          resolve();
+          return;
+        }
+
+        const handleCompletion = () => {
+          completed++;
+          if (completed === events.length) {
+            resolve();
+          }
+        };
```
daniel-graham-amplitude, line 364: Can we replace this with `transaction.onComplete` and `transaction.onAbort`?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: it's created because a table needs a unique key but there can be multiple events with the same event name

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,494 @@
...
+      const transaction = db.transaction([TABLE_NAMES.INTERNAL], 'readonly');
+      const store = transaction.objectStore(TABLE_NAMES.INTERNAL);
+
+      return new Promise((resolve, reject) => {
+        const request = store.get(key);
+
+        request.onsuccess = () => resolve(request.result as InternalRecord | undefined);
+
+        /* istanbul ignore next */
+        request.onerror = () => reject(new Error('Failed to get internal value'));
```
daniel-graham-amplitude, line 421: I think it's preferable to use the `transaction.onsuccess` and `transaction.onabort` to resolve/reject. If the db is closed midway through this operation, `onerror` will fail to fire and this whole promise will get stuck.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,494 @@
...
+    }
+  }
+
+  async setLastFlushTimestamp(timestamp: number): Promise<void> {
+    try {
+      await this.setInternal(INTERNAL_KEYS.LAST_FLUSH_TIMESTAMP, timestamp.toString());
+    } catch (error) {
+      /* istanbul ignore next */
+      this.logger.debug('DiagnosticsStorage: Failed to set last flush timestamp', error);
+    }
```
daniel-graham-amplitude, line 447: I think the try->catch block here is unnecessary since `setInternal` already has a try-catch block that logs out a debug statement.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: updated

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,490 @@
...
+      const db = await this.getDB();
+      const transaction = db.transaction([TABLE_NAMES.COUNTERS], 'readwrite');
+      const store = transaction.objectStore(TABLE_NAMES.COUNTERS);
+
+      return new Promise((resolve) => {
+        const entries = Object.entries(counters);
+
+        transaction.oncomplete = () => {
+          resolve();
+        };
```
daniel-graham-amplitude, line 242: nit: this can be shortened to transaction.oncomplete = resolve (optional)

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,490 @@
...
+   * Check if IndexedDB is supported in the current environment
+   * @returns true if IndexedDB is available, false otherwise
+   */
+  static isSupported(): boolean {
+    return getGlobalScope()?.indexedDB !== undefined;
+  }
+
+  async getDB(): Promise<IDBDatabase> {
+    if (!this.dbPromise) {
+      this.dbPromise = this.openDB();
```
crleona, line 120: are there any error conditions which could terminate the db connection while this is running, and should we try to re-open if there are?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,437 @@
...
+      this.logger.debug('DiagnosticsClient: Early return setTags as reaching memory limit');
+      return;
+    }
+
+    this.inMemoryTags[name] = value;
+    this.startTimersIfNeeded();
+  }
+
+  increment(name: string, size = 1) {
+    if (!this.storage) {
```
crleona, line 197: how many environments do we expect to not have indexdb support? would it be worth keeping in-memory events to send in that case?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,437 @@
...
+      return;
+    }
+
+    if (Object.keys(this.inMemoryCounters).length >= MAX_MEMORY_STORAGE_COUNT) {
```
crleona, line 201: why do we need a limit for the number of counters/histograms?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: Added `transaction.onabort` but since this is a single operation transaction, it should be enough to solve on `request.onsuccess`.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: but `setInternal` returns a promise which can reject

packages/analytics-core/src/diagnostics/diagnostics-client.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    await Promise.all([
+      this.storage.setTags(tagsToSave),
+      this.storage.incrementCounters(countersToSave),
+      this.storage.setHistogramStats(histogramsToSave),
+      this.storage.addEventRecords(eventsToSave),
+    ]);
+  }
+
+  async _flush() {
```
crleona, line 305: should flush also save any in-memory data?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: I've tried. The issue is:
resolve expects: (value: void | PromiseLike<void>) => void
transaction.oncomplete expects: (this: IDBTransaction, ev: Event) => any

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: it's because of how javascript handle timeout callback. See an example and more details at https://github.com/amplitude/Amplitude-TypeScript/pull/1281#discussion_r2350240375

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: No because flush's job is purely just to pull data from storage and send them. `saveAllDataToStorage` clears in memory data.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: it supports around 95% users. Similar discussion https://github.com/amplitude/Amplitude-TypeScript/pull/1281#discussion_r2350165890.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
crleona: We should not be creating a timer to flush each metric, just one timer to flush them all.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
crleona: Let me rephrase this: if I am calling flush, would I expect that to send every bit of data I have or just the data that's been synced to disk?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
crleona: How about react native or other environments using the shared TS core?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
crleona: how about react native, browser extensions, or any other context that TS core supports?

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude: It's a rate limit so that that the client can't set an unlimited number of counters or histograms. It's a limit that's very unlikely to ever be hit, I advocated for it because it could prevent crazy edge cases where an array/object grows indefinitely.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: From [W3C spec](https://www.w3.org/TR/IndexedDB/?utm_source=chatgpt.com#database-connection)

_A [connection](https://www.w3.org/TR/IndexedDB/) may be closed by a user agent in exceptional circumstances, for example due to loss of access to the file system, a permission change, or clearing of the [storage key](https://storage.spec.whatwg.org/#storage-key)’s storage. If this occurs the user agent must run [close a database connection](https://www.w3.org/TR/IndexedDB/) with the [connection](https://www.w3.org/TR/IndexedDB/) and with the forced flag set to true_


I don't think any of those conditions are conditions where we'd want to attempt re-opening if the database is closed.

daniel-graham-amplitude: That does raise the question of how are we handling database closures?

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: Ah yes, Typescript.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
crleona: We're going to cause more issues by randomly dropping data. This is small enough as not to be a concern.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
crleona: fail current attempt, next one try to reconnect? 😄

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude: Just data that's been synced to disk.

crleona: If we're uploading we might as well upload all the data we have.

packages/analytics-core/src/diagnostics/diagnostics-client.ts
daniel-graham-amplitude: Browser Extensions will work ✅ . React Native does not support IndexedDB ❌ 

Possible RN implementations:
1. Cut out the "sync to disk" step and just push data from memory to diagnostics service
2. Add an adapter setter to ClientDiagnostics that let's us replace the the indexed db storage with another storage that has the same interface, but uses different persistence mechanism (like [Async Storage](https://reactnative.dev/docs/asyncstorage))
3. Have a totally separate RN implementation of client diagnostics

I don't know about #1, that would create a lot of network request noise (memory is held for just 1 second, disk is 5 minutes).

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -0,0 +1,497 @@
...
+      request.onerror = () => {
+        reject(new Error('Failed to open IndexedDB'));
+      };
+
+      request.onsuccess = () => {
+        resolve(request.result);
+      };
+
+      request.onupgradeneeded = (event) => {
+        const db = (event.target as IDBOpenDBRequest).result;
```
daniel-graham-amplitude, line 151: I'm thinking we should add a handler here for `db.onerror` that aborts everything (closes diagnostics storage).

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: synced offline, will increase the limit to 10k

packages/analytics-core/src/diagnostics/diagnostics-client.ts
Mercy811: Synced offline, keep the current storage implementation and call save to db in flush. Kill the save to db timer

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: Thanks @crleona, good catch. Fix at https://github.com/amplitude/Amplitude-TypeScript/pull/1281/commits/190d4b94e5619e318195e9ac74a9a93a1f5bf5f8. It sets `dbPromise` to null when 1. the promise rejects 2. the promise resolves, but later the connection breaks

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: This is a good point. I added it at https://github.com/amplitude/Amplitude-TypeScript/pull/1281/commits/a49e4e2184eccbb3681a4c5d11c3c6c441566984. It will also trigger `db.onclose` to set `dbPromise` to null and later if `setTags()` is called, it will try to re-connect

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
daniel-graham-amplitude: Now that we set `db = null` that means that the calls to `getDB()` can be null too.

Example:
```
const db = await this.getDB();
const transaction = db.transaction([TABLE_NAMES.TAGS], 'readwrite');
```

So we'll need to return when db is null:
```
const db = await this.getDB();
if (!db) return;
const transaction = db.transaction([TABLE_NAMES.TAGS], 'readwrite');
```

daniel-graham-amplitude (approved):
Alright, good to go. Next steps let's start thinking about using this internally.

PR #1282, opened 2025-09-09
fix(autocapture): make the default for patterns in mask-text-regex to be case insensitive

Dogfalo:
### Summary

For text mask regex patterns, change regex into case insensitive by default.
This allows for easier RegEx patterns to be made. In the future, if necessary, an option can be added in Remote Config to allow case-sensitive to be applied.

### Checklist


achhunna (approved):
👍

PR #1287, opened 2025-09-12
feat(autocapture): set page url enrichment plugin to default on and add/fix tests

cyhk:
### Summary
Fix tests and release page-url-enrichment-plugin to default **on** for customers

### Checklist

### DIFF
--- a/packages/analytics-browser/test/default-tracking.test.ts
+++ b/packages/analytics-browser/test/default-tracking.test.ts
@@ -1,19 +1,19 @@
 import {
   getAttributionTrackingConfig,
-  getPageViewTrackingConfig,
   getElementInteractionsConfig,
   getFrustrationInteractionsConfig,
+  getNetworkTrackingConfig,
+  getPageViewTrackingConfig,
   isAttributionTrackingEnabled,
+  isElementInteractionsEnabled,
   isFileDownloadTrackingEnabled,
   isFormInteractionTrackingEnabled,
+  isFrustrationInteractionsEnabled,
+  isNetworkTrackingEnabled,
+  isPageUrlEnrichmentEnabled,
   isPageViewTrackingEnabled,
   isSessionTrackingEnabled,
-  isElementInteractionsEnabled,
-  getNetworkTrackingConfig,
-  isNetworkTrackingEnabled,
-  isFrustrationInteractionsEnabled,
   isWebVitalsEnabled,
-  isPageUrlEnrichmentEnabled,
 } from '../src/default-tracking';
 
 describe('isWebVitalsEnabled', () => {


packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -644,8 +665,8 @@ describe('Web attribution', () => {
               events: [
                 generateAttributionEvent(++eventId, url, initReferrer),
                 generateSessionStartEvent(++eventId),
-                generatePageViewEvent(++eventId, 1, url, initReferrer),
-                generatePageViewEvent(++eventId, 2, newURL),
+                generatePageViewEvent(++eventId, 1, url, initReferrer, { previousPageUrl: '' }),
```
cyhk: The one lingering concern is this test, where I expected the previousPageUrl to be the initial url (`https://www.example.com?utm_source=test_utm_source`), but sessionStorage (which the plugin uses to store previous pages) was empty.

I am unsure whether it is the test or the plugin that is the problem here. The plugin should already be able to handle SPA routing.

cyhk: This is resolved. It was a async/timing issue.

packages/analytics-browser-test/test/web-attribution.test.ts
```diff
@@ -357,12 +371,12 @@ describe('Web attribution', () => {
               client_upload_time: event_upload_time,
               events: [
                 generateAttributionEvent(++eventId, url),
-                generateSessionStartEvent(++eventId),
-                generatePageViewEvent(++eventId, 1, url),
-                generateSessionEndEvent(++eventId),
+                addPageUrlEnrichmentProperties(generateSessionStartEvent(++eventId), url),
+                generatePageViewEvent(++eventId, 1, url, undefined, { previousPageUrl: '' }),
+                addPageUrlEnrichmentProperties(generateSessionEndEvent(++eventId), directUrl, url),
```
cyhk: Tests are passing and make sense to me, except this (and other similar cases with session_end events have what is expected to be the next page's data.

When do session end events get fired?

test-server/browser-sdk/page-url-enrichment.html
```diff
@@ -63,20 +63,19 @@ <h1>Page 1</h1>
...
-        import.meta.env.VITE_AMPLITUDE_USER_ID || 'amplitude-typescript test user',
-        {
-          autocapture: {
-            pageUrlPreviousPage: true,
-          },
-        },
+        import.meta.env.VITE_AMPLITUDE_USER_ID || 'amplitude-typescript test user'
       );
       amplitude.add(pageViewPlugin);
+      amplitude.add(pageUrlPlugin);
```
Mercy811: why don't we just turn them on in init config?

cyhk: updated!

packages/analytics-browser/src/default-tracking.ts
```diff
@@ -237,3 +237,21 @@ export const getAttributionTrackingConfig = (config: BrowserOptions): Attributio
...
+  if (typeof autocapture === 'boolean') {
+    return autocapture;
+  }
+
+  if (typeof autocapture === 'object' && autocapture.pageUrlEnrichment === false) {
+    return false;
+  }
+
+  return true;
+};
```
daniel-graham-amplitude: It looks like this is the same criteria as attribution, file download, form interaction, page view and sessions.

I think you can just do it like they do

```
export const isPageUrlEnrichmentEnabled = (autocapture: AutocaptureOptions | boolean | undefined) =>
  isTrackingEnabled(autocapture, "pageUrlEnrichment")
```

+ would be good to group it together with the rest of those.

packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
```diff
@@ -146,19 +145,21 @@ export const pageUrlEnrichmentPlugin = (): EnrichmentPlugin => {
...
-            [PREVIOUS_PAGE_STORAGE_KEY]: previousPage,
+            [PREVIOUS_PAGE_STORAGE_KEY]: document.referrer || '',
           });
         }
 
+        const updatedURLInfo = await sessionStorage.get(URL_INFO_STORAGE_KEY);
+        let previousPage = '';
+        if (updatedURLInfo) {
+          previousPage = updatedURLInfo[PREVIOUS_PAGE_STORAGE_KEY] || '';
+        }
```
daniel-graham-amplitude, line 170: Optional nitpick: 158-161 can be shortened to one line

```javascript
let previousPage = updateURLInfo?.[PREVIOUS_PAGE_STORAGE_KEY] || '';
```

cyhk: Ah I remember I ended up with what I have in the PR to get 100% coverage on tests 😅 - I did have what you are suggesting before. I'll keep it as is in this case.

packages/analytics-browser/src/default-tracking.ts
cyhk: good call, updated!

packages/analytics-browser-test/test/web-attribution.test.ts
daniel-graham-amplitude: I just noticed this now. `session_end` events get tracked when a user visits a website that has Amplitude tracking on it and the "lastEventTime" (which gets updated to the latest time every event) is >30 minutes in the past. When that happens, the client creates a "session_end" event and fires it with "lastEventTime + 1" as it's timestamp.

So the session end events are uploaded at least 30 minutes after the last event in the session takes place. But usually it's way after that.

cyhk: I see - that makes sense for why the events are showing up the way they are in tests then.

PR #1294, opened 2025-09-17
fix(session-replay-browser): prevent multiple instances from starting

jpollock-ampl:
### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -529,6 +529,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
     if (!shouldRecord || !sessionId || !config) {
       return;
     }
+
+    // NOTE: If there is already an active instance, do not start a new one
+    if (this.recordCancelCallback) {
+      return;
+    }
+
     this.stopRecordingEvents();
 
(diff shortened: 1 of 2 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -529,6 +529,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
     if (!shouldRecord || !sessionId || !config) {
       return;
     }
+
+    // NOTE: If there is already an active instance, do not start a new one
+    if (this.recordCancelCallback) {
+      return;
+    }
```
lewgordon-amplitude, line 536: I'll be curious to see how testing goes, but I'm wondering if this will break some assumptions about this function. Specifically, it's more of a start or restart recording events at the moment.

jpollock-ampl: +1 - I will report back from testing. What are the scenarios in which we would want to restart?  Focus changes?

lewgordon-amplitude: https://github.com/amplitude/Amplitude-TypeScript/blob/5d8cbbd3293cd827de9752f970015fbe6076b526/packages/session-replay-browser/src/session-replay.ts#L274

PR #1295, opened 2025-09-17
perf(session-replay): reduce record initializations for targeting 

jxiwang:
### Summary

The evaluateTargetingAndCapture function is currently being used to call `recordEvent`. We should limit this two only 2 scenarios: 

- Initialization
- TRC evaluation

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -305,15 +305,14 @@ export class SessionReplay implements AmplitudeSessionReplay {
       return;
     }
 
-    // If there's no targeting config and this is not initialization, skip initialize/recordEvents
-    if (!this.config.targetingConfig && !isInit) {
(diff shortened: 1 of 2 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -300,6 +305,17 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    // If there's no targeting config and this is not initialization, skip initialize/recordEvents
+    if (!this.config.targetingConfig && !isInit) {
+      this.loggerProvider.log('No targeting config set, skipping initialization/recording for event.');
+      return;
+    }
+
+    // If there's no targeting config but this is initialization, log and continue
+    if (!this.config.targetingConfig) {
+      this.loggerProvider.log('Targeting config has not been set yet, cannot evaluate targeting.');
+    }
```
lewgordon-amplitude: nit: Wonder if it would be cleaner to put this in the other if statement.

PR #1297, opened 2025-09-18
fix(analytics-browser): add URI decoding to Element Clicked event attribute

daniel-graham-amplitude:
### Summary
* Fix attribute in "Element Clicked" so that Page URL looks like https://www.topps.com/products/2025-bowman-chrome®-baseball-mega-box instead of https://www.topps.com/products/2025-bowman-chrome%C2%AE-baseball-mega-box to be consistent with pageViewed
* Moved `getDecodeURI` into a helper function so that it can be re-used. It was copied/hardcoded in a few places.

### Checklist

### DIFF
--- a/packages/analytics-core/test/utils/url-utils.test.ts
+++ b/packages/analytics-core/test/utils/url-utils.test.ts
@@ -41,8 +41,8 @@ describe('isUrlMatchAllowlist', () => {
 });
 
 describe('getDecodeURI', () => {
-  test('should return the original string if decoding fails', () => {
-    const result = getDecodeURI('https://amplitude.com/blog');
-    expect(result).toEqual('https://amplitude.com/blog');
+  test('should decode URI', () => {
+    const result = getDecodeURI('https://www.topps.com/products/2025-bowman-chrome%C2%AE-baseball-mega-box');
+    expect(result).toEqual('https://www.topps.com/products/2025-bowman-chrome®-baseball-mega-box');
   });
 });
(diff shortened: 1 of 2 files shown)


cyhk (approved):
LGTM. Thanks for the fix and the refactor!

packages/plugin-autocapture-browser/test/data-extractor.test.ts
```diff
@@ -1161,4 +1161,21 @@ describe('data extractor', () => {
       document.body.removeChild(element);
     });
   });
+
+  describe('getEventProperties with page URL', () => {
+    test('should escape special characters in page URL', () => {
+      mockWindowLocationFromURL(new URL('https://www.topps.com/products/2025-bowman-chrome®-baseball-mega-box'));
```
Mercy811: Can we add a test where original url is `https://www.topps.com/products/2025-bowman-chrome%C2%AE-baseball-mega-box ` and expect `https://www.topps.com/products/2025-bowman-chrome®-baseball-mega-box` like what's in your description?

packages/analytics-core/test/utils/url-utils.test.ts
```diff
@@ -39,3 +39,10 @@ describe('isUrlMatchAllowlist', () => {
     expect(result).toEqual(true);
   });
 });
+
+describe('getDecodeURI', () => {
+  test('should return the original string if decoding fails', () => {
+    const result = getDecodeURI('https://amplitude.com/blog');
```
Mercy811: I don't think in this example decoding will fail

daniel-graham-amplitude: Fixed the comment and made it a better test.

packages/plugin-autocapture-browser/test/data-extractor.test.ts
daniel-graham-amplitude: Fixed.

PR #1299, opened 2025-09-19
fix(plugin-ga-events-forwarder-browser): do not log errors on relative URLs

daniel-graham-amplitude:
### Summary

Problem: relative URL's cause `new URL` to throw an exception in the `intercept` function, which logs a console error in the catch block, which is creating a lot of noise for one of our customers.

Solution: move the `new URL` function before the try catch block into it's own try catch block. In this block, the catch doesn't log an error. No need to log an error, because if it reaches that block it means that it's a relative URL.

### Checklist

### DIFF
--- a/packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
+++ b/packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
@@ -130,8 +130,15 @@ export const gaEventsForwarderPlugin = ({ measurementIds = [] }: Options = {}): 
    * 3b. Sends events to Amplitude after Amplitude SDK is initialized
    */
   const intercept = (requestUrl: string | URL, data: BodyInit | null | undefined): boolean => {
+    let url: URL;
     try {
-      const url = new URL(requestUrl);
+      url = new URL(requestUrl);
+    } catch (e) {
+      // it's a relative URL, therefore not a GA4 event
+      return false;
+    }
+
+    try {
       if (
         GA_SERVICE_ROOT_DOMAIN_VALUES.some((rootDomain) => url.hostname.endsWith(rootDomain)) &&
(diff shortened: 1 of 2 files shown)


packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
```diff
@@ -152,6 +159,7 @@ export const gaEventsForwarderPlugin = ({ measurementIds = [] }: Options = {}):
     } catch (e) {
       /* istanbul ignore next */
       logger?.error(e);
```
Mercy811, line 161: Can we just change it to logger.debug() instead so that we don't silently catch an error when url is invalid

UPD:
> because if it reaches that block it means that it's a relative URL.

Does it mean that if it reaches there, the url is always valid?

dyao-amp (approved):

packages/plugin-ga-events-forwarder-browser/src/ga-events-forwarder.ts
daniel-graham-amplitude: > we don't silently catch an error when url is invalid

The thing that's different here is that this callback is invoked via monkeypatching of fetch and sendBeacon. It intercepts whatever URL was sent to that

```
globalScope.navigator.sendBeacon = new Proxy(globalScope.navigator.sendBeacon, {
  apply: (target: SendBeaconFn, thisArg: any, argArray: Parameters<SendBeaconFn>) => {
    // Intercepts request and attempt to send to Amplitude
    interceptBeacon.apply(thisArg, argArray);
    // Execute sendBeacon
    return target.apply(thisArg, argArray);
  },
});
```

So even if the URL is invalid, we don't really have any reason to be logging errors or debug, because it's not our error.

Mercy811 (approved):

PR #1300, opened 2025-09-22
fix(session-replay-browser): Fix webworker packaging for ESM

lewgordon-amplitude:
#### 🐛 Problem
The webworker string replacement performed by Rollup was being reverted by subsequent tsc compilation, causing the webworker functionality to fail for npm consumers using the library files (lib/cjs/ and lib/esm/). The failure mode is that it just wouldn't run with the webworker at all and it fell back to the main thread processing.
#### ✅ Solution
- Fixed build order: Changed from yarn bundle && yarn build:es5 && yarn build:esm to yarn build:es5 && yarn build:esm && yarn bundle to ensure TypeScript compilation happens before Rollup bundling
- Added webworker replacement for library files: Created a postprocess script to apply webworker string replacement to the compiled library files after bundling
- Fixed "use strict" directive: Configured Rollup TypeScript plugin with noImplicitUseStrict: false to ensure proper strict mode emission
#### 🔬 Testing
Tested locally using a basic HTML app and a Next JS app.

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/events/event-compressor.ts
+++ b/packages/session-replay-browser/src/events/event-compressor.ts
@@ -124,7 +124,7 @@ export class EventCompressor {
           // fallback: serialize
           this.worker.postMessage(JSON.stringify({ event, sessionId }));
         } else {
-          throw err; // rethrow unexpected
+          this.config.loggerProvider.warn('Unexpected error while posting message to worker:', err);
         }
       }
     } else {
--- a/packages/session-replay-browser/test/event-compressor.test.ts
+++ b/packages/session-replay-browser/test/event-compressor.test.ts
@@ -287,7 +287,7 @@ describe('EventCompressor', () => {
     expect(terminateMock).toHaveBeenCalled();
   });
 
-  test('should rethrow unexpected errors in webworker', async () => {
+  test('should log warning for unexpected errors in webworker', async () => {
     let postMessageMock = jest.fn();
     let onMessageMock = jest.fn();
     let onErrorMock = jest.fn();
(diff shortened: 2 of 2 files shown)


lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/events/event-compressor.ts
```diff
@@ -121,7 +116,17 @@ export class EventCompressor {
...
     if (this.worker) {
       // This indirectly compresses the event.
-      this.worker.postMessage({ event, sessionId });
+      try {
+        this.worker.postMessage({ event, sessionId });
+      } catch (err: any) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        if (err.name === 'DataCloneError') {
+          // fallback: serialize
+          this.worker.postMessage(JSON.stringify({ event, sessionId }));
```
jxiwang, line 125: Is this the fall back that you mentioned where it would go back to normal compression?

lewgordon-amplitude: This is a weird case that I noticed where I think our custom events can't be passed to the webworker so we need to serialize them. Really all the fallback logic is in the constructor (if the worker script is undefined).

lewgordon-amplitude: Also there's fallback logic here: https://github.com/amplitude/Amplitude-TypeScript/pull/1300/files#diff-5701ec875ae41b5ed40495ead788b110bc03e50345eb0313f36c2efe3097830cR191

jxiwang (approved):

packages/session-replay-browser/src/events/event-compressor.ts
jxiwang: Gotcha that makes sense, does that mean on error throw the event could get dropped?

lewgordon-amplitude: Yeah though maybe there's a better way to handle it? I also wonder if I should just silently skip instead of throwing.

jxiwang: Hmm, maybe warning? I think we set anything that the customer can't fix as a warning to not trigger their diagnostic tools. As far handling, we could add the event to the non-worker queue instead of throwing but not sure if there are any implications with that. I'm okay with failing silently since we are still testing this.

lewgordon-amplitude: Oh yeah that's a good idea

lewgordon-amplitude: I'm thinking maybe we re-evaluate the fallback on error per event. I'm worried that it may happen more often than not defeating the purpose of the webworker. Let's see how this goes and I think we can definitely revisit in a later PR.

jxiwang: Yeah agreed! :ship:

PR #1301, opened 2025-09-22
fix(analytics-core): add events support for diagnostics client

Mercy811:
### Checklist

### DIFF
--- a/packages/analytics-core/src/diagnostics/diagnostics-storage.ts
+++ b/packages/analytics-core/src/diagnostics/diagnostics-storage.ts
@@ -2,6 +2,8 @@ import { getGlobalScope } from '../global-scope';
 import { ILogger } from '../logger';
 import { HistogramStats } from './diagnostics-client';
 
+const MAX_PERSISTENT_STORAGE_EVENTS_COUNT = 10;
+
 // Database configuration
 const DB_VERSION = 1;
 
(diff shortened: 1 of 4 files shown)


packages/analytics-core/src/diagnostics/diagnostics-storage.ts
```diff
@@ -363,17 +364,38 @@ export class DiagnosticsStorage implements IDiagnosticsStorage {
...
+            );
+          }
+
+          // Only add events up to the available slots (take the least recent ones)
+          events.slice(0, availableSlots).forEach((event) => {
+            const request = store.add({
+              event_name: event.event_name,
+              event_properties: event.event_properties,
+              time: event.time,
+            });
```
daniel-graham-amplitude: Nit: this seems redundant to me. It's making an identical copy of the object. Couldn't you just do `store.add(event);`?

store.add docs:  "(in a)  separate thread, creates a [structured clone](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#structured-clone) of the value, and stores the cloned value in the object store". So need to worry about it performing a destructive action on the original object.

daniel-graham-amplitude (approved):
Approved, pending my one comment.

packages/analytics-core/src/diagnostics/diagnostics-storage.ts
Mercy811: Thanks Dan! I mis-read it as what I should do https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/add. Fixed at https://github.com/amplitude/Amplitude-TypeScript/pull/1301/commits/0c3dc0a026aec706248c3889a21fdf92a205766b

PR #1302, opened 2025-09-22
fix(session-replay): fix builds for iife

jxiwang:
### Summary

Users that have the IIFE script installed on their site could not access the correct `window.sessionReplay` object. Fixing this issue by pointing the build to `index.ts`.

Also including an HTML script that helps with manual testing of our SR builds. Can be leveraged for E2E testing down then line. 

### Checklist


lewgordon-amplitude (approved):
LGTM!

PR #1306, opened 2025-09-25
chore: migrate GTM snippet to amplitude-typescript

daniel-graham-amplitude:
### Summary

The goal of this PR is to deprecate https://github.com/amplitude/amplitude-js-gtm/ in favour of having the GTM code snippet being deployed in amplitude-typescript, alongside `analytics-browser`

#### In this PR
* Added an EJS template for amplitude-wrapper that takes the contents of the analytics-browser and plugin-session-replay-browser scripts and puts them into "amplitude-wrapper.js" (similar how we did in amplitude-js-gtm)
* Added a rollup script that takes the js file that's generated above, and then gzips it
* Added a test page to the test-server that runs a test on the above gzipped page
* Set the version of the new script to be the same as "analytics-browser" so that they are always released concurrently

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
         node-version: [18.17.x, 20.x, 22.x]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
-    retries: 2
 
     steps:
       - name: Check out git repository


packages/gtm-snippet/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@amplitude/gtm-snippet",
+  "version": "3.20.0",
```
Mercy811: If my [understanding](https://github.com/amplitude/Amplitude-TypeScript/pull/1306#issuecomment-3352923227) is correct, just make sure that this version matches analytics-bowser latest version when merging this PR

packages/gtm-snippet/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "scripts": {
+    "build": "yarn build:template && yarn build:bundle",
+    "build:bundle": "rollup --config rollup.config.js",
+    "build:template": "node ./scripts/build-snippet.js",
+    "prepublishOnly": "yarn build",
+    "deploy": "node ./scripts/upload-to-s3.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/amplitude/amplitude-js-gtm.git"
```
Mercy811: nit: update url

test-server/gtm-snippet/gtm-snippet.html
```diff
@@ -0,0 +1,21 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/amplitude.png" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <script src="/gtm-snippet/lib/scripts/analytics-browser-gtm-wrapper.min.js.gz"></script>
```
Mercy811, line 7: is this path valid? I'm not sure if the the html file could find the path outside of the test server

packages/gtm-snippet/amplitude-wrapper.js.ejs
Mercy811, line 1: I know this is just a draft PR. Just wanna double check

> Added an EJS template for amplitude-wrapper that takes the contents of the analytics-browser and plugin-session-replay-browser scripts and puts them into "amplitude-wrapper.js" (similar how we did in amplitude-js-gtm

We're going to replace SDK scripts with placeholders right?

packages/gtm-snippet/scripts/build-snippet.js
Mercy811, line 1: What's the plan to update plugin-user-agent-enrichment-browser? It's on v1.x branch. I'm thinking: 
1. upload plugin-user-agent-enrichment-browser script to cdn when it publishes 
2. checkout v1.x and build and then read from its lib
3. move it to main branch so that we can directly read from lib. (i'm more into this one)

For G&S, I guess we will need to go either 1 or 2

test-server/gtm-snippet/gtm-snippet.html
daniel-graham-amplitude: It can now. I updated the Vite config so that it finds static files in the root path.

packages/gtm-snippet/scripts/build-snippet.js
daniel-graham-amplitude: Plan: defer as another task

packages/gtm-snippet/amplitude-wrapper.js.ejs
daniel-graham-amplitude: Could you elaborate on that? I don't think I understand.

Mercy811: `packages/gtm-snippet/amplitude-wrapper.js.ejs` is the template and are we going to have placeholders like these https://github.com/amplitude/amplitude-browser-sdk-gtm-template/blob/324beb0d170a54b0d67f72197ec35349a776f60e/template.tpl.ejs#L11-L20?

daniel-graham-amplitude: Yeah the next iteration we can add the SDK scripts like G&S, page enrichment, etc...

PR #1307, opened 2025-09-26
feat(engagement): Add Guides and Surveys support for unified SDK

tkainrad:
Jira Ticket: https://amplitude.atlassian.net/browse/FOO-2348

I tested via a hello world project and made sure that all the bundles initialize correctly.

**Technical Description:**
Added engagement plugin support to the unified package by:
- Adding `@amplitude/engagement-browser` as a dependency
- Importing and initializing the engagement plugin in the unified client
- Extending the UnifiedOptions interface to include engagement configuration options

Note that the Engagement SDK npm package differs from other web SDKs as it _does not_ bundle all the code. Instead, it contains only initialization code that will then download the full bundle from our CDN. With the unified package, we want to use remote config to skip that bundle download unless the Engagement SDK has been enabled explicitly in the dashboard.

Related PRs:
- https://github.com/amplitude/javascript/pull/83781
- https://github.com/amplitude/assistance-browser/pull/731


Mercy811 (approved):
LGTM!

packages/unified/src/unified.ts
```diff
@@ -80,6 +82,9 @@ export class AmplitudeUnified extends AmplitudeBrowser implements UnifiedClient
     } else {
       this.sessionReplay = (srPlugin as SessionReplayPlugin).sessionReplay;
     }
+
+    await super.add(EngagementPlugin({ ...{ enabled: false }, ...unifiedOptions?.engagement, ...sharedOptions }))
+      .promise;
```
Mercy811: Engagement plugin is a thin loader of the actual G&S SDK. It's disabled by default so existing customers won't load the G&S SDK. They can opt in by remote config.

PR #1310, opened 2025-09-29
fix(session-replay-browser): Optimize function restart logic

jpollock-ampl:
### Summary

This PR optimizes some of the logic around starting/restarting replay capture.  

Previously, the following three actions would cause the replay to start or restart:
- Focus Change
  - This is desired, we want to restart when this happens
- Session ID Changes (via `setSessionId()`)
  - This is desired, we want to restart when this happens
- An event being emitted from the SDK
  - Undesired, a restart is unnecessary in this scenario

This PR stops the replay capture from being restarted when an event is emitted from the SDK.  

How does this fix work?

The primary change is in `recordEvents()`:
```
if (this.recordCancelCallback && !forceRestart) { 
  // Exit Early
}
```

We will check for an existing callback and if the user does **not** pass forceRestart (default is `forceRestart=false`), we will exit early and not re-create the SDK. 

In the case of "focus" or "session ID changes", we pass `forceRestart=true`, to ensure that a new capture is created.

### Verification

Beyond unit tests, I created a test app and installed this version of the SDK.  I was able to observe the behavior of the SDK on a focus event and then when sending 5 events via a button click:

Screenshot 2025-09-29 at 3 41 46 PM

The first 3 logs represent the "focus" event.  The fourth log represents the button click.  

--- 

Notes:

Given a call to `track()` in your application, it goes through the Amplitude Analytics SDK's pipeline:

For us, that means the plugin execution pipeline.  So it goes through [timeline.ts, which executes the various types of plugins](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/timeline.ts#L83):
- before plugins
- enrichment plugins (this is what our plugin is)
- destination plugins

Our plugin (the SR plugin) is an enrichment plugin ([source](https://github.com/amplitude/Amplitude-TypeScript/blob/a1fc5b7/packages/plugin-session-replay-browser/src/session-replay.ts#L21)), so its execute() [method](https://github.com/amplitude/Amplitude-TypeScript/blob/a1fc5b7/packages/plugin-session-replay-browser/src/session-replay.ts#L132) is called for every single event ([source](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/timeline.ts#L131))

The execute() [method](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-session-replay-browser/src/session-replay.ts#L132) calls:

- [evaluateTargetingAndCapture](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/plugin-session-replay-browser/src/session-replay.ts#L165), which will evaluate targeting conditions ([source code for evaluateTargetingAndCapture](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L290))
- It then calls await [this.recordEvents()](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L344) (if not during initialization)
- OR calls void [this.initialize(true)](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L342) during initialization, which [also calls recordEvents()](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L370)

The recordEvents() method:
- Calls [this.getShouldRecord()](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L527) to determine if recording should happen
- If yes, starts the rrweb capture process (including tearing down the existing instance via [stopRecordingEvents()](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/session-replay-browser/src/session-replay.ts#L532))

### Checklist


jxiwang (approved):
Nice! Thanks for adding this!

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -546,13 +555,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
     const config = this.config;
     const shouldRecord = this.getShouldRecord();
     const sessionId = this.identifiers?.sessionId;
     if (!shouldRecord || !sessionId || !config) {
       return;
     }
+
+    // NOTE: If there is already an existing active recording, exit early unless forceRestart is true
+    if (this.recordCancelCallback && !forceRestart) {
+      this.loggerProvider.debug(`A capture is already in progress and forceRestart is false. Exiting.`);
```
lewgordon-amplitude: Nit: Maybe instead of "Exiting" we could say "not restarting". I wouldn't want this to leak to users who made get the wrong impression.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -546,13 +555,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
     }
   }
 
-  async recordEvents(shouldLogMetadata = true) {
+  async recordEvents(recordEventsConfig?: { shouldLogMetadata?: boolean; forceRestart?: boolean }) {
+    const { shouldLogMetadata = true, forceRestart = true } = recordEventsConfig || {};
```
lewgordon-amplitude: I was thinking these could be non null since their defaults may not be obvious and maybe it's better to be more explicit here. It also saves a null check.
```suggestion
  async recordEvents({shouldLogMetadata, forceRestart}: { shouldLogMetadata: boolean; forceRestart: boolean }) {
```

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -546,13 +555,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
     if (!shouldRecord || !sessionId || !config) {
       return;
     }
+
+    // NOTE: If there is already an existing active recording, exit early unless forceRestart is true
+    if (this.recordCancelCallback && !forceRestart) {
+      this.loggerProvider.debug(`A capture is already in progress and forceRestart is false. Exiting.`);
+      return;
+    }
+    this.loggerProvider.debug(`Starting new capture for session.`);
```
lewgordon-amplitude: Minor nit: Maybe we should include the session replay ID or session ID? I think it wouldn't hurt and would potentially help with debugging.

lewgordon-amplitude (approved):
LGTM!

packages/session-replay-browser/src/session-replay.ts
jpollock-ampl: Thank you for the feedback - https://github.com/amplitude/Amplitude-TypeScript/pull/1310/commits/c81dab9a6990efbe52cd7fa4ddbe63a125547345

packages/session-replay-browser/src/session-replay.ts
jpollock-ampl: Great call - https://github.com/amplitude/Amplitude-TypeScript/pull/1310/commits/f75a51292a5cb894504d4f4a65f59b10d0e1ebbb

packages/session-replay-browser/src/session-replay.ts
jpollock-ampl: Done! https://github.com/amplitude/Amplitude-TypeScript/pull/1310/commits/c68312f19611c7f5811d1fc1ee82146bc205a89b

lewgordon-amplitude (approved):
LGTM!

PR #1314, opened 2025-10-02
ci: publish new package with latest tag

Mercy811:
### Summary

Publish a new package which doesn't have a non beta/alpha version yet with the latest npm tag but with a beta/alpha version. 

Beta 3,4, and 5 were published with this GAH but with beta tag. Customers won't be able to install them without explicitly specify the version. This also ensure that `yarn add @amplitude/unified` automatically installs the latest version. 

image

### Checklist

### DIFF
--- a/.github/workflows/publish-single-package.yml
+++ b/.github/workflows/publish-single-package.yml
@@ -3,6 +3,9 @@
 # Use this workflow for new packages that will only have beta/alpha versions
 # The 'latest' tag will always point to the most recent version (even if beta/alpha)
 # since these packages don't have stable releases yet
+# 
+# IMPORTANT: Once a stable (non-beta/alpha) version is published to latest,
+# this workflow will prevent further beta/alpha publishing to maintain stability
 name: Publish New Package (Beta/Alpha with Latest Tag)
 
 on:
@@ -14,7 +17,7 @@ on:
         required: true
       releaseTag:
         type: choice
-        description: 'Release type - choose beta for testing releases or alpha for experimental releases (package will always be published with latest tag)'
+        description: 'Release type - choose beta for testing releases or alpha for experimental releases (package will be published with latest tag, but only if no stable version exists)'
         required: true
         options:
           - beta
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-single-package.yml
```diff
@@ -1,26 +1,32 @@
...
     inputs:
       packageName:
         type: string
-        description: 'Package name (e.g., unified, analytics-browser)'
+        description: 'Package name (e.g., unified, analytics-browser) - must be a valid package in packages/ directory'
         required: true
       releaseTag:
         type: choice
-        description: 'Release tag'
+        description: 'Release type - choose beta for testing releases or alpha for experimental releases (package will always be published with latest tag)'
```
daniel-graham-amplitude: `(package will always be published with latest tag)` <-- shouldn't this be `(package will be published with latest tag by default)`

daniel-graham-amplitude: Nvm I understand now. The version tag says "beta" or "alpha" but it's published to the "latest" channel so that users don't have to explicitly install from another channel.

One area of concern for this approach is that once you take out the "beta" or "alpha" tag from a package, and publish a semantically versioned, prod tag to latest, we can't publish any more "beta" or "alpha" tags for that package. Which is fine, but do we have a way of blocking us from doing that?

Example:
1. publish `my-package@1.0.0-beta.1 --latest` ✅ 
2. publish `my-package@1.0.0 --latest` ✅ 
3. publish `my-package@1.0.0-beta.2 --latest` ❌ _this should be forbidden_

Mercy811: The idea is that we only go forward, once graduate, we should not use this GHA to publish.

Mercy811: Let me see if we can auto check if current latest is non beta/alpha version, if so, it will abort

daniel-graham-amplitude: Right, I'm just concerned about the possibility that somebody may mistakenly run this script on a package that's already been graduated to GA. Maybe it's okay though.

Mercy811: I totally agree. We should do our best to prevent it. I added a step to check latest npm version

daniel-graham-amplitude: Nice!

PR #1317, opened 2025-10-02
fix: Allow timestamp override for experiment events

tyiuhc:
### Summary

This fix will allow experiment events to override the default timestamp provided by the Analytics SDK. This will allow the timestamp of experiment events to reflect the time of exposure, rather than when the event is processed by the Analytics SDK.

### Checklist


Mercy811 (approved):
LGTM! `ExposureTrackingProvider.track(exposure: Exposure)` takes `Exposure` which is not amplitude event but just event properties. So it makes sense to track time as an event property and convert it to event option.

PR #1318, opened 2025-10-03
fix: use file path for validate version publishing rules require statement

tkainrad:
## Problem

The "Validate version publishing rules" step in the publish-single-package workflow was failing with:
```
Error: Cannot find module 'packages/unified/package.json'
```

## Root Cause

On line 90, the require statement was using a relative path without the `./` prefix:
```bash
PACKAGE_JSON_NAME=$(node -p "require('$PACKAGE_PATH/package.json').name")
```

When `$PACKAGE_PATH` expands to `packages/unified`, Node.js's `require()` function doesn't recognize it as a file path - it needs `./` or `../` prefix to be treated as a relative file path.

## Solution

Added `./` prefix to the require path:
```bash
PACKAGE_JSON_NAME=$(node -p "require('./$PACKAGE_PATH/package.json').name")
```

Now the path correctly resolves to `./packages/unified/package.json` which Node.js can properly load.

## Testing

Tested during dry-run of publishing the unified package.

### DIFF
--- a/.github/workflows/publish-single-package.yml
+++ b/.github/workflows/publish-single-package.yml
@@ -3,7 +3,7 @@
 # Use this workflow for new packages that will only have beta/alpha versions
 # The 'latest' tag will always point to the most recent version (even if beta/alpha)
 # since these packages don't have stable releases yet
-# 
+#
 # IMPORTANT: Once a stable (non-beta/alpha) version is published to latest,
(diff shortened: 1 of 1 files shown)


Mercy811 (approved):
Thanks for fixing it @tkainrad. LGTM!

PR #1320, opened 2025-10-03
fix: resolve pluralization issue frustrationInteractions

daniel-graham-amplitude:
### Summary

If remote config supplies "rageClick" or "deadClick" rename it to "rageClicks" and "deadClicks" (respectively).

### Checklist


daniel-graham-amplitude:
@Mercy811 that's what I had suggested but the problem is iOS already implemented it with singular, and the team is insisting that they be consistent across SDK's.

PR #1330, opened 2025-10-07
feat(autocapture): add excluded events list and remove deprecated pac…

cyhk:
…kage dependencies

### Summary
This cleans up some things for the plugin before the PR for general release.

It does two things: 1) Adds an excluded events list to exclude special events `i$dentify`, `$groupidentify` and `revenue_amount` and 2) removes dependency from deprecated packages `analytics-types` and `analytics-client-common`

### Checklist

### DIFF
--- a/packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
+++ b/packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
@@ -28,10 +28,10 @@ enum PreviousPageType {
   External = 'external', // for different domains
 }
 
-export const EXCLUDED_DEFAULT_EVENT_TYPES = new Set([
-  SpecialEventType.IDENTIFY as string,
-  SpecialEventType.GROUP_IDENTIFY as string,
-  SpecialEventType.REVENUE as string,
+export const EXCLUDED_DEFAULT_EVENT_TYPES = new Set<string>([
+  SpecialEventType.IDENTIFY,
+  SpecialEventType.GROUP_IDENTIFY,
+  SpecialEventType.REVENUE,
 ]);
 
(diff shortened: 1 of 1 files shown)


packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
```diff
@@ -17,6 +25,8 @@ enum PreviousPageType {
   External = 'external', // for different domains
 }
 
+export const EXCLUDED_DEFAULT_EVENT_TYPES = new Set(['$identify', '$groupidentify', 'revenue_amount']);
```
daniel-graham-amplitude: Instead of hardcoding these can you import them from [analytics-core](https://github.com/amplitude/Amplitude-TypeScript/blob/no-ticket-add-import-checks-to-eslint/packages/analytics-core/src/types/event/event.ts#L98-L102)

cyhk: sure, changed!

packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
```diff
@@ -17,6 +26,12 @@ enum PreviousPageType {
   External = 'external', // for different domains
 }
 
+export const EXCLUDED_DEFAULT_EVENT_TYPES = new Set([
```
Dogfalo: avoid `as`
instead use:
```
new Set<string>
```

cyhk: oh nice, didn't know that would resolve the typing too.

PR #1328, opened 2025-10-07
Revert "feat(analytics-client-common): refactor getPageTitle from aut…

Mercy811:
…ocapture to make it reusable (#1298)"

This reverts commit 3c931eeb7bb4d2482523c48cf796113187d9b078.

### Summary

analytics-client-common is deprecated and is not dependency of plugin autocpature. It will break customer website 
image

CI didn't catch it because this is a monorepo and it's resolved by the top level build. I will have a follow-up PR to fix CI.

### Checklist

* [ ] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* Does your PR have a breaking change?:

PR #1332, opened 2025-10-08
chore: add import checks to eslint

daniel-graham-amplitude:
### Summary
* Adds the "plugin:import/recommended" which contains rules about import statements
  * Most importantly it contains "import/no-extraneous-dependencies" which will help prevent [this](https://github.com/amplitude/Amplitude-TypeScript/pull/1328)
  * Turned off most of the rules for tests and specs because it's not production and if it works, no need to fix it
* Added missing dependencies that were exposed by the "no-extraneous-dependencies" check (unified, session-replay-browser, ...)
* Added some ignores to certain files that had failing lint tests (didn't fix the lint failures because that's out of the scope of this PR)

Here's a screenshot showing it failing when it tries importing "analytics-client-common" from "plugin-autocapture-browser"

Screenshot 2025-10-07 at 4 58 29 PM

### Checklist


packages/segment-session-replay-plugin/src/helpers.ts
Mercy811, line 1: I'm a little bit scared to disable eslint on the whole file especially this is a new and active plugin. Once we disable it, the chance is low that we will enable it again. I'm fine to disable it for `packages/plugin-web-attribution-browser/src/web-attribution.ts` as it's deprecated.

daniel-graham-amplitude: Removed.

packages/session-replay-browser/package.json
```diff
@@ -48,6 +48,7 @@
     "@amplitude/rrweb-plugin-console-record": "2.0.0-alpha.32",
     "@amplitude/rrweb-record": "2.0.0-alpha.32",
     "@amplitude/rrweb-utils": "2.0.0-alpha.32",
+    "@amplitude/rrweb-types": "2.0.0-alpha.32",
```
Mercy811, line 51: Seems we don't need this for eslint? 

UPD:
Oh I see, we added it here because this package is imported in session-replay-browser. cc: @lewgordon-amplitude and @jxiwang

daniel-graham-amplitude: Not having this was causing a `import/no-extraneous-dependencies` error

PR #1351, opened 2025-10-18
feat(session-replay-browser): migrate to new remote config client

Mercy811:
### Summary

Replace the old remote config with the new one for session replay browser. 

If initialize SR as a plugin with analytics SDK, 3 remote config requests are made. This will be fixed in the remote config client at [here](https://github.com/amplitude/Amplitude-TypeScript/blob/394f18b52c383f30fbed85cb0dcf4fa80df527b5/packages/analytics-core/src/remote-config/remote-config.ts#L34-L35). 

After: 3 remote config requests to the new endpoint
Screenshot 2025-10-20 at 2 29 42 PM

Before (analytics-browser 2.27.0 & SR plugin 1.22.25): 2 remote config requests (one from analytics and one from autocapture) by the new client and one by the old client
image

### Checklist

### DIFF
--- a/test-server/session-replay-browser/sr-plugin-with-analytics-sdk.html
+++ b/test-server/session-replay-browser/sr-plugin-with-analytics-sdk.html
@@ -1,0 +1,151 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/amplitude.png" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Session Replay Plugin Test</title>
+    <style>
+      body {
+        font-family: Arial, sans-serif;
+        max-width: 600px;
+        margin: 50px auto;
+        padding: 20px;
+      }
+      h1 {
+        color: #333;
+      }
+      button {
+        background: #0066cc;
+        color: white;
+        border: none;
+        padding: 12px 24px;
+        margin: 10px 5px;
+        border-radius: 4px;
+        cursor: pointer;
+        font-size: 16px;
+      }
+      button:hover {
+        background: #0052a3;
+      }
+      .status {
+        padding: 10px;
+        margin: 10px 0;
+        border-radius: 4px;
+        background: #f0f0f0;
+        font-family: monospace;
+      }
+      input {
+        padding: 8px;
+        margin: 5px;
+        font-size: 14px;
+        width: 300px;
+      }
+      .section {
+        margin: 20px 0;
+        padding: 20px;
+        background: #f9f9f9;
+        border-radius: 8px;
+      }
+    </style>
+  </head>
+  <body>
+    <h1>Session Replay Plugin Test</h1>
+    
+    <div class="status" id="status">Status: Not initialized</div>
+    
+    <input type="text" id="apiKey" placeholder="API Key (optional)" />
+    <br />
+    <button id="initBtn">Initialize Analytics + Session Replay Plugin</button>
+
+    <div class="section">
+      <h2>Test Events</h2>
+      <button id="trackBtn">Track Event</button>
+      <button id="getInfoBtn">Get Session Info</button>
+    </div>
+
+    <script type="module">
+      import * as amplitude from '@amplitude/analytics-browser';
+      import { sessionReplayPlugin } from '@amplitude/plugin-session-replay-browser';
+      
+      window.amplitude = amplitude;
+      
+      const statusEl = document.getElementById('status');
+      const apiKeyEl = document.getElementById('apiKey');
+      const initBtn = document.getElementById('initBtn');
+      const trackBtn = document.getElementById('trackBtn');
+      const getInfoBtn = document.getElementById('getInfoBtn');
+
+      function updateStatus(message) {
+        statusEl.textContent = `Status: ${message}`;
+        console.log('[Status]', message);
+      }
+
+      // Auto-fill API key from environment
+      if (import.meta.env.VITE_AMPLITUDE_API_KEY) {
+        apiKeyEl.value = import.meta.env.VITE_AMPLITUDE_API_KEY;
+      }
+
+      initBtn.addEventListener('click', async () => {
+        const apiKey = apiKeyEl.value || import.meta.env.VITE_AMPLITUDE_API_KEY || 'test-api-key';
+        
+        try {
+          updateStatus('Initializing Analytics Browser SDK...');
+          
+          console.log('[Init] Starting with API key:', apiKey);
+          
+          // Create and add Session Replay plugin
+          console.log('[Plugin] Creating Session Replay plugin...');
+          const sessionReplayTracking = sessionReplayPlugin({
+            sampleRate: 1.0,
+            logLevel: 'debug',
+          });
+          amplitude.add(sessionReplayTracking);
+
+          // Initialize Analytics Browser SDK
+          amplitude.init(apiKey, 'test-user@amplitude.com', {
+            logLevel: amplitude.Types.LogLevel.Debug,
+          });
+          
+          const sessionId = amplitude.getSessionId();
+          const deviceId = amplitude.getDeviceId();
+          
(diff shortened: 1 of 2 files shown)


Mercy811:
bugbot run

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
-      this.localConfig.loggerProvider.warn(knownError.message);
-      config.captureEnabled = false;
-    }
 
-    if (!remoteConfig) {
+          // Resolve on first callback
+          resolve();
+        },
+      );
+    });
```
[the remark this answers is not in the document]
Mercy811: This is handled by the remote config client

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
+    // Subscribe to remote config client to get the config (uses cache if available)
+    await new Promise<void>((resolve) => {
+      this.remoteConfigClient.subscribe(
+        'configs.sessionReplay',
+        'all',
+        (remoteConfig: RemoteConfig | null, source: Source) => {
+          this.localConfig.loggerProvider.debug(
+            `Session Replay remote configuration received from ${source}:`,
+            JSON.stringify(remoteConfig, null, 2),
+          );
```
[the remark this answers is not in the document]
Mercy811: duplicate of https://github.com/amplitude/Amplitude-TypeScript/pull/1351#discussion_r2441477877

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
-          if (samplingConfig) {
-            remoteConfig.sr_sampling_config = samplingConfig;
+    let sessionReplayRemoteConfig: SessionReplayRemoteConfig | undefined;
+
+    // Subscribe to remote config client to get the config (uses cache if available)
+    await new Promise<void>((resolve) => {
+      this.remoteConfigClient.subscribe(
+        'configs.sessionReplay',
+        'all',
+        (remoteConfig: RemoteConfig | null, source: Source) => {
```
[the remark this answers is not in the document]
Mercy811: cancel subscription won't prevent this. This is actually fine and it's by design to return the first available result from either remote or from cache.

packages/session-replay-browser/src/config/joined-config.ts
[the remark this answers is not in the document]
Mercy811: duplicate of the previous one

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
+              if (samplingConfig) {
+                sessionReplayRemoteConfig.sr_sampling_config = samplingConfig;
+              }
+              if (privacyConfig) {
+                sessionReplayRemoteConfig.sr_privacy_config = privacyConfig;
+              }
+              if (targetingConfig) {
+                sessionReplayRemoteConfig.sr_targeting_config = targetingConfig;
+              }
+            }
```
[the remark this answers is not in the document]
Mercy811: The merging logic remains unchanged and is designed by SR team.

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
-    }
 
-    if (!remoteConfig) {
+          // Resolve on first callback
+          resolve();
+        },
+      );
+    });
+
+    if (!sessionReplayRemoteConfig) {
```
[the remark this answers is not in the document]
Mercy811: it's handled by the remote config client. it will never throw an error.

packages/session-replay-browser/test/integration/sampling.test.ts
```diff
@@ -113,29 +147,10 @@ describe('module level integration', () => {
...
-        return initPromise.then(() => {
-          const sampleRate = sessionReplay.config?.sampleRate;
-          // Ensure that sample rate matches what's passed in the options
-          expect(sampleRate).toBe(1);
-          const sessionRecordingProperties = sessionReplay.getSessionReplayProperties();
-          expect(sessionRecordingProperties).toMatchObject({});
-          expect(mockRecordFunction).not.toHaveBeenCalled();
-        });
-      });
-    });
```
Mercy811, line 135: this test is deleted because all errors are handled by the new remote config client.

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -40,70 +39,83 @@ export const removeInvalidSelectorsFromPrivacyConfig = (privacyConfig: PrivacyCo
...
-        config.loggingConfig = namespaceConfig.sr_logging_config;
-
-        if (samplingConfig || privacyConfig || targetingConfig) {
-          remoteConfig = {};
-          if (samplingConfig) {
-            remoteConfig.sr_sampling_config = samplingConfig;
+    let sessionReplayRemoteConfig: SessionReplayRemoteConfig | undefined;
+
+    // Subscribe to remote config client to get the config (uses cache if available)
+    await new Promise<void>((resolve) => {
```
lewgordon-amplitude, line 59: Could we make this return a config or at least some object? I'm thinking it's slightly harder to reason when the config variable is being modified in the promise.

test-server/session-replay-browser/sr-plugin-with-analytics-sdk.html
lewgordon-amplitude, line 1: Nice! This is awesome. We have an old app that we used for testing that may be useful here.

Mercy811: where is the old app? I can move it here

lewgordon-amplitude: It's a bit messy and it's a full react app :/ https://github.com/amplitude/tasks-app

packages/session-replay-browser/src/config/joined-config.ts
Mercy811: i don't think it's possible because the callback in `remoteConfigClient.subscribe()` is supposed to be called when remote config is fetched. The promise and resolve here is to make it synchronous and awaitable. 

> I'm thinking it's slightly harder to reason when the config variable is being modified in the promise.

The execution order will be 
1. Define a variable `sessionReplayRemoteConfig`
2. Create and await a promise
  2.1. Fetch remote config
  2.2. Update `sessionReplayRemoteConfig`
  2.3 Resolve the promise

lewgordon-amplitude: Yeah that makes sense. I guess this changes the semantics slightly then if I'm reading this right as we will still continue if the remote config fetch fails (returns null). Before the idea was to force it here and make sure that capture was disabled in that case.

Mercy811: Oh I see. We won't be able to tell whether remote config fetch fails because the new remote config client won't throw an error. I think it's fine to make this change here as it would only affect the case where remote config fetch fails. Before it forced capture disabled and after it will just fall back to local config. wdyt?

lewgordon-amplitude: Yeah I think we want to force capture disabled still. The rationale was that someone could remotely disable session replay as a kill switch if need be.

test-server/session-replay-browser/sr-plugin-with-analytics-sdk.html
Mercy811: Good to know that! I will try to do it in a separate pr so that we can test against local changes without linking.

PR #1353, opened 2025-10-20
perf(session-replay): reduce recording initialization when trc enabled

jxiwang:
### Summary

We do not re-record when a recording is already initialized for TRC. This prevents record initialization when new events are fired. 

Callstack with TRC ON - 

Screenshot 2025-10-21 at 11 20 48 AM
Screenshot 2025-10-21 at 11 21 28 AM

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -6,6 +6,8 @@ import {
   LogLevel,
   returnWrapper,
   SpecialEventType,
+  generateHashCode,
+  isTimestampInSample,
 } from '@amplitude/analytics-core';
 
 // Import only specific types to avoid pulling in the entire rrweb-types package
@@ -32,7 +34,6 @@ import {
 import { EventCompressor } from './events/event-compressor';
 import { createEventsManager } from './events/events-manager';
 import { MultiEventManager } from './events/multi-manager';
-import { generateHashCode, isTimestampInSample } from '@amplitude/analytics-core';
 import { getDebugConfig, getPageUrl, getStorageSize, maskFn } from './helpers';
 import { clickBatcher, clickHook, clickNonBatcher } from './hooks/click';
 import { ScrollWatcher } from './hooks/scroll';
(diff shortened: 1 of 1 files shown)


packages/plugin-session-replay-browser/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.22.26-perfrecord-init-prelease.0](https://github.com/amplitude/Amplitude-TypeScript/compare/@amplitude/plugin-session-replay-browser@1.22.25...@amplitude/plugin-session-replay-browser@1.22.26-perfrecord-init-prelease.0) (2025-10-20)
+
+**Note:** Version bump only for package @amplitude/plugin-session-replay-browser
```
lewgordon-amplitude: I feel like this shouldn't go in the changelog?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -366,7 +367,11 @@ export class SessionReplay implements AmplitudeSessionReplay {
     if (isInit) {
       void this.initialize(true);
     } else {
-      await this.recordEvents();
+      // Only call recordEvents if we're not already recording, unless forceRestart is true
+      if (forceRestart || !this.recordCancelCallback) {
```
lewgordon-amplitude: Nit: Why not just else if?

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -237,7 +237,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
     }
 
     if (this.config?.targetingConfig) {
-      await this.evaluateTargetingAndCapture({ userProperties: options?.userProperties });
+      await this.evaluateTargetingAndCapture({ userProperties: options?.userProperties }, false, true);
```
lewgordon-amplitude, line 241: So does this forceRestart if the session ID is set?

packages/plugin-session-replay-browser/CHANGELOG.md
lewgordon-amplitude: I guess you're in the middle of testing ATM?

stevenchien-amplitude (approved):

packages/session-replay-browser/src/session-replay.ts
jxiwang: Ah good catch, let me change this.

packages/session-replay-browser/src/session-replay.ts
jxiwang: Yeah I kept the original record initializations the same.

packages/plugin-session-replay-browser/CHANGELOG.md
jxiwang: Hmm, I'm not sure why this is in here.. maybe cause I deployed a pre-release. Will remove.

PR #1360, opened 2025-10-22
fix: Previous Page Location set to current page on first load in SPAs

### LINKED ISSUE #1359: [plugin-page-url-enrichment-browser] Previous Page incorrectly set to current page on direct visit in Next.js/SPAs
## Expected Behavior

When a user directly visits a page (no referrer) for the first time in a session, the `@amplitude/plugin-page-url-enrichment-browser` plugin should set:
- `[Amplitude] Previous Page Location`: Empty string or the actual `document.referrer` value
- `[Amplitude] Previous Page Type`: `"direct"` (when no referrer exists)

The previous page should NOT be the same as the current page on a direct visit.

## Current Behavior

When a user directly visits a page in a Next.js application, the plugin incorrectly sets:
- `[Amplitude] Previous Page Location`: Same as current page URL
- `[Amplitude] Previous Page Type`: `"internal"`

This happens because the plugin's `saveURLInfo()` function is triggered during Next.js hydration/initialization (via `history.replaceState` proxy) before the first event is tracked. The storage gets initialized incorrectly:

1. Storage is empty initially
2. Next.js calls `history.replaceState()` during hydration (normal SPA behavior)
3. Plugin's proxy triggers `saveURLInfo()`
4. `saveURLInfo()` moves `AMP_CURRENT_PAGE` → `AMP_PREVIOUS_PAGE` (but current page is the only page!)
5. Session storage ends up with both current and previous set to the same URL

**Session Storage State After Bug:**
```json
{
  "AMP_CURRENT_PAGE": "https://www.example.com/",
  "AMP_PREVIOUS_PAGE": "https://www.example.com/"
}
```

## Possible Solution

The issue is in the [`saveURLInfo()` function](https://github.com/amplitude/Amplitude-TypeScript/blob/6cea10aa528d37483ca79759ab83d7a5c4178332/packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts#L84-L95) in `page-url-enrichment.ts`:

```typescript
previousURL = (URLInfo?.[CURRENT_PAGE_STORAGE_KEY]) || '';
```

This blindly copies the current page to previous page on every navigation, even when it's the very first initialization.

**Suggested Fix:**

The plugin should distinguish between:
1. **First initialization** (no session storage exists yet) - Should use `document.referrer`
2. **Actual navigation** (storage exists, user is navigating) - Should move current → previous

Modified logic:
```typescript
const saveURLInfo = async () => {
  if (sessionStorage && isStorageEnabled) {
    const URLInfo = await sessionStorage.get(URL_INFO_STORAGE_KEY);
    const currentURL = getDecodeURI(location.href || '');

    // On first initialization, use document.referrer instead of current page
    const previousURL = URLInfo?.[CURRENT_PAGE_STORAGE_KEY]
      ? URLInfo[CURRENT_PAGE_STORAGE_KEY]  // Navigation: use previous current
      : (document.referrer || '');          // First load: use referrer

    await sessionStorage.set(URL_INFO_STORAGE_KEY, {
      [CURRENT_PAGE_STORAGE_KEY]: currentURL,
      [PREVIOUS_PAGE_STORAGE_KEY]: previousURL,
    });
  }
};
```

Alternatively, the `execute()` method already has logic to handle first initialization (lines 144-147), but `saveURLInfo()` gets called by the navigation proxies before `execute()` runs, causing the corruption.

## Steps to Reproduce

1. Install `@amplitude/plugin-page-url-enrichment-browser` in a Next.js application
2. Initialize Amplitude with the plugin:
```javascript
import { pageUrlEnrichmentPlugin } from '@amplitude/plugin-page-url-enrichment-browser';

const pageUrlPlugin = pageUrlEnrichmentPlugin();
amplitude.add(pageUrlPlugin);
amplitude.init('YOUR_API_KEY');
```
3. Clear session storage (or open an incognito window)
4. Navigate directly to your homepage (e.g., type URL in address bar or bookmark)
5. Track a page view event after page load:
```javascript
amplitude.track('Page View');
```
6. Inspect the event properties

**Expected:**
- `[Amplitude] Previous Page Location`: `""` (empty)
- `[Amplitude] Previous Page Type`: `"direct"`

**Actual:**
- `[Amplitude] Previous Page Location`: `"https://www.example.com/"`
- `[Amplitude] Previous Page Type`: `"internal"`

**Session Storage Contents:**
```javascript
sessionStorage.getItem('AMP_URL_INFO')
// {"AMP_CURRENT_PAGE":"https://www.example.com/","AMP_PREVIOUS_PAGE":"https://www.example.com/"}
```

## Environment

- JS SDK Version: `@amplitude/analytics-browser@^2.16.1`, `@amplitude/plugin-page-url-enrichment-browser@^0.4.3`
- Installation Method: NPM
- Browser and Version: Chrome 131 (reproduced across browsers)
- Framework: Next.js 15.1.6

## Workaround

Pre-initialize the session storage before calling `amplitude.init()`:

```javascript
const urlInfoKey = 'AMP_URL_INFO';
const existingUrlInfo = window.sessionStorage.getItem(urlInfoKey);
if (!existingUrlInfo) {
  const initialUrlInfo = {
    AMP_CURRENT_PAGE: window.location.href,
    AMP_PREVIOUS_PAGE: document.referrer || '',
  };
  window.sessionStorage.setItem(urlInfoKey, JSON.stringify(initialUrlInfo));
}

amplitude.add(pageUrlPlugin);
amplitude.init('YOUR_API_KEY');
```

This prevents the plugin from incorrectly initializing the storage during Next.js hydration.

tmchow:
### Summary
When a user directly visits a page in an SPA (e.g., Next.js), the plugin's saveURLInfo() function was being triggered by SPA hydration events (history.replaceState) before the first event was tracked. This caused the storage to be initialized with both current and previous page set to the same URL.

The fix updates saveURLInfo() to check if storage already exists:
- If storage exists: move current page to previous (normal navigation)
- If storage is empty: use document.referrer as previous page (first load)

This ensures that on direct visits with no referrer, Previous Page Type is
set to "direct" and Previous Page Location is empty, rather than being
incorrectly set to the current page URL.

Fixes #1359

### Checklist

### DIFF
--- a/packages/analytics-browser/e2e/cookie-consent.spec.ts
+++ b/packages/analytics-browser/e2e/cookie-consent.spec.ts
@@ -29,7 +29,7 @@ test.describe('Cookie Consent Page', () => {
     await page.click('#track-before-consent-btn');
     
     // Wait a moment to ensure any potential requests would have been made
-    await page.waitForTimeout(1000);
+    await page.waitForTimeout(1500);
     
     // Assert: no network request made by Amplitude (since SDK is not initialized)
     expect(requests.length).toBe(0);
--- a/packages/analytics-browser/e2e/events.spec.ts
+++ b/packages/analytics-browser/e2e/events.spec.ts
@@ -64,7 +64,7 @@ test.describe('Events Page', () => {
 
     // Wait for all network requests to complete
     await page.waitForLoadState('networkidle');
-    await page.waitForTimeout(1000);
+    await page.waitForTimeout(1500);
 
     const events = requests[0].events;
     // Verify the network requests


packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
```diff
@@ -84,8 +84,11 @@ export const pageUrlEnrichmentPlugin = (): EnrichmentPlugin => {
   const saveURLInfo = async () => {
     if (sessionStorage && isStorageEnabled) {
       const URLInfo = await sessionStorage.get(URL_INFO_STORAGE_KEY);
-      const previousURL = URLInfo?.[CURRENT_PAGE_STORAGE_KEY] || '';
-      const currentURL = getDecodeURI((typeof location !== 'undefined' && location.href) || '');
+      const currentURL = getDecodeURI(location.href || '');
```
cyhk: any reason why the definition of currentURL is changed to exclude `typeof location !== 'undefined'`?

tmchow: Good catch! That was an unintentional simplification. I'll restore the `typeof location !== 'undefined'` check to maintain consistency with the rest of the file.

tmchow:
> Looks good to me. Thanks for the fix again, @tmchow.

I see a failure with cookie consent which should be totally unrelated. How to proceed?

cyhk:
> > Looks good to me. Thanks for the fix again, @tmchow.
> 
> I see test playwright failures which should be totally unrelated. How to proceed?

Not quite sure what is happening - for what it's worth I ran the tests locally and they passed, then again on a separate branch and they passed as well. I just reran the e2e tests for this PR again to see what happens, though it is likely that they will still fail. 

Looking at the failures, I wonder if this could be a timeout issue? Could you try increasing the timeouts to some of the failed tests?

PR #1361, opened 2025-10-22
ci: npm publish token check

Mercy811:
### Summary

Add a token expiration check at the beginning of feature branch prerelease. One package can only have one trusted GHA. The spot is taken by publish-v2. Feature branch prerelease will have to use granular access token with 90 days max life limits.

image

### Checklist

### DIFF
--- a/.github/workflows/feature-branch-prerelease.yml
+++ b/.github/workflows/feature-branch-prerelease.yml
@@ -20,16 +20,18 @@ jobs:
       - name: Check NPM token validity
         run: |
           RESPONSE=$(curl -s -H "Authorization: Bearer ${{ secrets.NPM_PUBLISH_TOKEN }}" https://registry.npmjs.org/-/whoami)
-          if [ "$RESPONSE" != '{"username":"sdk.dev"}' ]; then
+          USERNAME=$(echo "$RESPONSE" | jq -r '.username')
+          if [ "$USERNAME" != "sdk.dev" ]; then
             echo "❌ NPM token validation failed!"
-            echo "Expected: {\"username\":\"sdk.dev\"}"
-            echo "Got: $RESPONSE"
+            echo "Expected username: sdk.dev"
+            echo "Got username: $USERNAME"
+            echo "Full response: $RESPONSE"
             echo ""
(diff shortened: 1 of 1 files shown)


.github/workflows/feature-branch-prerelease.yml
```diff
@@ -13,9 +13,28 @@ on:
...
 
 jobs:
+  check-npm-token:
+    name: Check NPM Token
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check NPM token validity
+        run: |
+          RESPONSE=$(curl -s -H "Authorization: Bearer ${{ secrets.NPM_PUBLISH_TOKEN }}" https://registry.npmjs.org/-/whoami)
+          if [ "$RESPONSE" != '{"username":"sdk.dev"}' ]; then
```
daniel-graham-amplitude: Is there any chance that the response cold look something like `{"username": "sdk.dev", "other": "somethingElse"}` ?

I think a slightly more reliable check would be to us jq. 

```
RESPONSE=$(curl -s -H "Authorization: Bearer ${{ secrets.NPM_PUBLISH_TOKEN }}" https://registry.npmjs.org/-/whoami)
USERNAME=$(echo "$RESPONSE" | jq -r '.username')

if [ "$USERNAME" != "sdk.dev" ]; then
```

.github/workflows/feature-branch-prerelease.yml
```diff
@@ -13,9 +13,28 @@ on:
...
+            echo "❌ NPM token validation failed!"
+            echo "Expected: {\"username\":\"sdk.dev\"}"
+            echo "Got: $RESPONSE"
+            echo ""
+            echo "⚠️  Please update the NPM_PUBLISH_TOKEN secret in GitHub repository settings."
+            echo "The token may have expired or been revoked."
```
daniel-graham-amplitude, line 30: I would also add that they need to rotate the token in NPM first before updating the NPM_PUBLISH_TOKEN.

Mercy811:
@daniel-graham-amplitude Both comments make sense to me. I applied the first one and for the second one I added the guide. 
image

daniel-graham-amplitude (approved):

PR #1371, opened 2025-10-27
feat: more diagnostics metrics

Mercy811:
### Summary

- Track a counter `sdk.diagnostics.sampled.in.and.enabled`. This is a diagnostics internal metric. It will be cross compared with page view event. The total number of this counter should be approximately the multiple of sample rate times the total number of page view during a fixed period of time
- Track a counter `analytics.dropped.events`
- Track an event `analytics.dropped.events` which will be forwarded to [datadog as logs](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3454174130/Diagnostics+-+runbook#Event). Only event type is captured.
- Track `plugin.autocapture.version` tag with version 

### Checklist

---

> [!NOTE]
> Integrates DiagnosticsClient across browser and plugins, records diagnostics for dropped events, and tags the autocapture plugin version, with supporting exports, tests, and a simplified diagnostics test page.
> 
> - **Core**:
>   - **Destination**: accepts `diagnosticsClient`; on non-200 responses increments `analytics.dropped.events` and records a diagnostics event with codes/messages/stack; constructor updated and tests added.
>   - **DiagnosticsClient**: logs sampling init by incrementing `sdk.diagnostics.sampled.in.and.enabled`; maintains sampling via `_setSampleRate`; minor initialization/flush behavior covered by tests.
>   - **Exports**: re-export `LabeledEvent`, `Trigger`, `DataSource`.
> - **Browser**:
>   - Initialize `DiagnosticsClient`, set `library` tag, and pass to plugins; install `Destination({ diagnosticsClient })` and `autocapturePlugin(..., { diagnosticsClient })`.
> - **Autocapture Plugin**:
>   - Set diagnostics tag `plugin.autocapture.version` on setup; imports updated; tests added for diagnostics tag and type exports.
> - **Test Server**:
>   - Replace `diagnostics-test.html` with simplified `test-server/diagnostics.html` demonstrating init, plugin injection, and diagnostics sampling.
> - **Tests**: expand unit tests for diagnostics sampling, destination diagnostics recording, and plugin imports/behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f7a45174dc2d17a1c03c45fc043bd0354e53727. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -362,15 +362,15 @@ export class Destination implements DestinationPlugin {
   fulfillRequest(list: Context[], code: number, message: string) {
     // Record diagnostics for dropped events (non-success status codes)
     if (!isSuccessStatusCode(code)) {
-      this.diagnosticsClient?.increment('analytics.dropped.events', list.length);
-      this.diagnosticsClient?.recordEvent('analytics.dropped.events', {
+      this.diagnosticsClient?.increment('analytics.events.dropped', list.length);
+      this.diagnosticsClient?.recordEvent('analytics.events.dropped', {
         events: list.map((context) => context.event.event_type),
         code: code,
         message: message,
         stack_trace: getStacktrace(),
       });
     } else {
-      this.diagnosticsClient?.increment('analytics.sent.events', list.length);
+      this.diagnosticsClient?.increment('analytics.events.sent', list.length);
     }
     this.removeEvents(list);
     list.forEach((context) => context.callback(buildResult(context.event, code, message)));
--- a/packages/analytics-core/test/plugins/destination.test.ts
+++ b/packages/analytics-core/test/plugins/destination.test.ts
@@ -759,10 +759,10 @@ describe('destination', () => {
       destination.fulfillRequest(contexts, 400, 'Bad Request');
 
       // Verify diagnostics increment was called with the correct count
-      expect(incrementSpy).toHaveBeenCalledWith('analytics.dropped.events', 2);
+      expect(incrementSpy).toHaveBeenCalledWith('analytics.events.dropped', 2);
 
       // Verify diagnostics recordEvent was called with the correct data
-      expect(recordEventSpy).toHaveBeenCalledWith('analytics.dropped.events', {
+      expect(recordEventSpy).toHaveBeenCalledWith('analytics.events.dropped', {
         events: ['event1', 'event2'],
         code: 400,
         message: 'Bad Request',
@@ -804,7 +804,7 @@ describe('destination', () => {
         destination.fulfillRequest([context], 500, 'Internal Server Error');
 
         // Verify stack_trace defaults to empty array when stack is undefined
-        expect(recordEventSpy).toHaveBeenCalledWith('analytics.dropped.events', {
+        expect(recordEventSpy).toHaveBeenCalledWith('analytics.events.dropped', {
           events: ['event1'],
           code: 500,
           message: 'Internal Server Error',
@@ -837,7 +837,7 @@ describe('destination', () => {
       destination.fulfillRequest(contexts, 200, 'OK');
 
       // Verify diagnostics increment was called with the correct count for sent events
-      expect(incrementSpy).toHaveBeenCalledWith('analytics.sent.events', 2);
+      expect(incrementSpy).toHaveBeenCalledWith('analytics.events.sent', 2);
 
       // Verify callbacks were called
       expect(callback1).toHaveBeenCalledTimes(1);


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -351,6 +358,16 @@ export class Destination implements DestinationPlugin {
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
+    // Record diagnostics for dropped events (non-success status codes)
+    if (code !== 200) {
```
daniel-graham-amplitude: I think we should probably cover 200 to 299. I just had a look at the destination plugin, and it [looks like](https://github.com/amplitude/Amplitude-TypeScript/blob/main/packages/analytics-core/src/transports/base.ts#L85) we deem a "successful" request to be anything >=200 and <300

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -351,6 +358,16 @@ export class Destination implements DestinationPlugin {
...
+      this.diagnosticsClient?.increment('analytics.droppedEvents', list.length);
```
daniel-graham-amplitude: We should also consider tracking the opposite: `analytics.capturedEvents`. I feel like `droppedEvents` isn't as useful of a metric without knowing what percentage of events are being dropped (`droppedEvents / capturedEvents`)

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -351,6 +358,16 @@ export class Destination implements DestinationPlugin {
...
+      this.diagnosticsClient?.recordEvent('analytics.droppedEvents', {
+        events: list.map((context) => context.event),
+        code: code,
+        message: message,
+        stack_trace: new Error().stack?.split('\n').filter((line) => line.trim()) || [], // convert to array of strings
```
daniel-graham-amplitude: `new Error().stack?.split('\n').filter((line) => line.trim())` <-- I don't think this would give us useful information. I just plugged it into my console and it returns this: `(2) ['Error', '    at <anonymous>:1:1']`

I'm guessing we need to pass an array. Could we just pass an empty array `[]` since there's no actual error object or exception being caught here?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -351,6 +358,16 @@ export class Destination implements DestinationPlugin {
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
+    // Record diagnostics for dropped events (non-success status codes)
+    if (code !== 200) {
+      this.diagnosticsClient?.increment('analytics.droppedEvents', list.length);
+      this.diagnosticsClient?.recordEvent('analytics.droppedEvents', {
```
daniel-graham-amplitude: Wouldn't these 2 having the same name (both: `analytics.droppedEvents`) cause them to overwrite?

packages/analytics-core/src/plugins/destination.ts
Mercy811: Here is a real example. Direct console log might not be a good example
image

packages/analytics-core/src/plugins/destination.ts
Mercy811: No, one is metric and the other is log. they are different types

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: I understand now. [Error.prototype.stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack) will give us the stacktrace of wherever the error was instantiated.

daniel-graham-amplitude: Optional: maybe a helper method called something like `getStacktrace()` could make this more idiomatic?

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: Okay 👍

packages/analytics-core/src/plugins/destination.ts
Mercy811: nice catch! Updated.

packages/analytics-core/src/plugins/destination.ts
Mercy811: new metric added

packages/analytics-core/src/plugins/destination.ts
Mercy811: There is actually an existing `getStacktrace()`. Reuse that

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -351,6 +360,18 @@ export class Destination implements DestinationPlugin {
   }
 
   fulfillRequest(list: Context[], code: number, message: string) {
+    // Record diagnostics for dropped events (non-success status codes)
+    if (!isSuccessStatusCode(code)) {
+      this.diagnosticsClient?.increment('analytics.dropped.events', list.length);
```
daniel-graham-amplitude: Just had a thought. Would it make more sense for this to be called `analytics.events.dropped` (and also `analytics.events.sent`). Seems like it would make sense for these two to have the same prefix.

Mercy811: make sense! updated

daniel-graham-amplitude: 👍

PR #1378, opened 2025-10-31
refactor: track action click RxJS -> Zen

daniel-graham-amplitude:
### Summary
* refactor "track-action-click" to use Zen Observable over RxJS

### Checklist

---

> [!NOTE]
> Refactors action-click tracking to ZenObservable with new `clickObservableZen`, `mutationObservableZen`, and `navigateObservableZen`, updating plugin wiring, tests, and demo page.
> 
> - **Autocapture core (`src/autocapture-plugin.ts`)**:
>   - Add Zen streams: `clickObservableZen`, `mutationObservableZen`, `navigateObservableZen`; enrich via `DataExtractor` and `multicast`.
>   - Retain RxJS `clickObservable`; remove RxJS `mutationObservable` usage.
>   - Create Zen navigate observable when `window.navigation` exists.
>   - Expose new observables in `AllWindowObservables`; guard `trackActionClick` subscription.
> - **Action-click tracking (`src/autocapture/track-action-click.ts`)**:
>   - Rewrite to use Zen observables and `merge`/`asyncMap` from `@amplitude/analytics-core`.
>   - Filter, map, and debounce clicks; detect action clicks from click followed by mutation/navigate within 500ms; track `[Amplitude] Element Clicked`.
> - **Observables (`src/observables.ts`)**:
>   - Implement `createMutationObservableZen` with `document.body` guard.
>   - Implement `createClickObservableZen` with `{ capture: true }` listeners.
> - **Tests (`test/.../track-action-clicks.test.ts`)**:
>   - Update tests to Zen-based flow; add mocked `window.navigation` and assert navigation-triggered action clicks.
> - **Demo page (`test-server/autocapture/element-interactions.html`)**:
>   - Add `#content-changer-div` and handler to trigger DOM changes; adjust copy.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 73ab0b1aa1f150b00956fe1b9150a8afc7a64676. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
+++ b/packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
@@ -56,7 +56,7 @@ export function trackActionClick({
   const mutationOrNavigate = navigateObservableZen
     ? merge(mutationObservableZen, navigateObservableZen)
     : mutationObservableZen;
-  
+
   const clickMutationNavigateObservable = merge(filteredClickObservable, mutationOrNavigate);
 
   let actionClickTimer: ReturnType<typeof setTimeout> | null = null;


daniel-graham-amplitude:
bugbot run

packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
```diff
@@ -41,36 +46,63 @@ export function trackActionClick({
...
+      actionClickTimer = setTimeout(() => {
+        actionClickTimer = null;
+        lastClickEvent = null;
+        return null;
+      }, 500);
+      return Promise.resolve(null);
+    } else {
+      // if mutation/navigation + last click event, then it's an action click
+      if (lastClickEvent) {
+        try {
```
Dogfalo: I think try..catch is not necessary here, this seems more straightforward:

```js
} else {
  // if mutation/navigation + last click event, then it's an action click
  if (lastClickEvent) {
    const event = lastClickEvent;
    lastClickEvent = null;
    return Promise.resolve(event);
  }
}
```

packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
```diff
@@ -41,36 +46,63 @@ export function trackActionClick({
...
-  if (navigateObservable) {
-    changeObservables.push(navigateObservable);
-  }
-  const mutationOrNavigate = merge(...changeObservables);
+  let mutationOrNavigate = navigateObservableZen
+    ? merge(mutationObservableZen, navigateObservableZen)
+    : mutationObservableZen;
+  mutationOrNavigate = asyncMap(mutationOrNavigate, (event) => {
+    // defer mutation/navigation events to new task to prevent
+    // from being emitted before the click event
```
Dogfalo: When does this actually happen?

daniel-graham-amplitude: In my observations, I've found that there's cases where the "mutation" events will actually happen before the click handler.

```
# Sequence of events
click happens
dom mutation happens
dom dispatches mutation event
dom dispatches click event
```

The mutation needs to happen after the click event to track an "action click".

My solution was to just put the mutation on a `setTimeout(() => {}, 0)` so that it's invoked on a separate task from the click handler.

I think I should look into why the mutation happens first too, that is surprising, I would have expected the ordering to look like this:

```
# Sequence of events
click happens
dom dispatches click event
dom mutation happens
dom dispatches mutation event
```

packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
daniel-graham-amplitude: Yeah I'll do that. That was just a hacky way I came up with to avoid the need to copy the event.

packages/plugin-autocapture-browser/src/autocapture/track-action-click.ts
Dogfalo: I would only expect this to happen if the page has handlers on the mousedown/pointerdown event, but this is fairly rare. Mutation events are already batched so it seems even stranger that it's happening before the click event is received.

daniel-graham-amplitude: I'm going to take this setTimeout out. The problem was that my unit tests got the ordering wrong. But in a live browser environment it's always going to be `click -> click event -> mutation event`

PR #1385, opened 2025-11-05
fix(session-replay-browser): fix heatmap click x/y

lewgordon-amplitude:
### Summary

Use current scroll depth from the scroll watcher to correctly track click x/y. Before we just relying on the window scrolling element which is incorrect in cases where the window is not the scrolling element. The scroll events are correctly tracking depth so we're just reusing that to calculate it for the clicks.

SR-1975

### Checklist

### DIFF
--- a/packages/session-replay-browser/src/hooks/click.ts
+++ b/packages/session-replay-browser/src/hooks/click.ts
@@ -1,12 +1,13 @@
 import type { mouseInteractionCallBack } from '@amplitude/rrweb-types';
 import { MouseInteractions } from '@amplitude/rrweb-types';
-import { getWindowScroll, Mirror } from '../utils/rrweb';
+import { Mirror } from '../utils/rrweb';
 import { SessionReplayEventsManager as AmplitudeSessionReplayEventsManager } from '../typings/session-replay';
-import { PayloadBatcher } from 'src/track-destination';
+import { PayloadBatcher } from '../track-destination';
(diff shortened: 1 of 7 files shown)


jxiwang:
> incorrect in cases where the window is not the scrolling element

@lewgordon-amplitude Is this for cases like a carousel or a scrollable component?

lewgordon-amplitude:
@jxiwang here's an example HTML

```html
<!DOCTYPE html>
<html>
  <head>
    <style>
      html, body {
        height: 100%;
        margin: 0;
        overflow: auto;
      }
      body {
        background: #fafafa;
      }
      .content {
        height: 500vh; /* taller than viewport */
        padding: 20px;
        background: linear-gradient(white, lightblue);
      }
    </style>
    <script src="http://localhost:55031/lib/scripts/session-replay-browser-min.js"></script>
  </head>
  <body>
    <script>
      window.sessionReplay.init('', {
        deviceId: '...', // uuid v4
        sessionId: Date.now(), // uuid v4
        sampleRate: 1,
        interactionConfig: {
          enabled: true,
          batch: false,
        }
      });
    </script>
    <div class="content">
      <h1>Scroll inside the body</h1>
      <p>Keep scrolling down...</p>
    </div>
  </body>
</html>
```

If you do window.scrollableElement.scrollTop it'll always return 0 in the old way. This new way should fix that.

PR #1396, opened 2025-11-12
fix: handle null remote config cases for session capture

christiancho:
### Summary

When merging remote and local configs, the current implementation ignored null remote configs and still allowed for session captures. This PR adds a rejection to the promise so we can handle cases where there is no remote config to disable captures.

## Test assertion changes:

- Updated unit test to check for `enableCapture` to be false when remote config is missing
- Changed mocks for some tests to explicitly return a remote config to allow tests to pass
- Mocked `console.error` for api tests to make sure tests don't fail

### Checklist

### DIFF
--- a/packages/session-replay-browser/test/config/joined-config.test.ts
+++ b/packages/session-replay-browser/test/config/joined-config.test.ts
@@ -214,7 +214,7 @@ describe('SessionReplayJoinedConfigGenerator', () => {
       });
       test('should set return an undefined remote config', async () => {
         mockRemoteConfig = null;
-        const { localConfig } = await joinedConfigGenerator.generateJoinedConfig(); 
+        const { localConfig } = await joinedConfigGenerator.generateJoinedConfig();
         expect(localConfig).toEqual(mockLocalConfig);
       });
     });
--- a/packages/session-replay-browser/test/helpers.test.ts
+++ b/packages/session-replay-browser/test/helpers.test.ts
@@ -9,7 +9,7 @@ import {
   UNMASK_TEXT_CLASS,
 } from '../src/constants';
 import { ServerZone } from '@amplitude/analytics-core';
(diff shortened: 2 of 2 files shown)


packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -55,53 +55,68 @@ export class SessionReplayJoinedConfigGenerator {
...
-            config.loggingConfig = namespaceConfig.sr_logging_config;
-
-            if (samplingConfig || privacyConfig || targetingConfig) {
-              sessionReplayRemoteConfig = {};
-              if (samplingConfig) {
-                sessionReplayRemoteConfig.sr_sampling_config = samplingConfig;
-              }
-              if (privacyConfig) {
-                sessionReplayRemoteConfig.sr_privacy_config = privacyConfig;
+            if (remoteConfig) {
```
jpollock-ampl: Are we able to remove this if we have the `if (!remoteConfig)` above?

packages/session-replay-browser/src/config/joined-config.ts
jpollock-ampl, line 1: Mind adding a unit test(s) for the new functionality?

christiancho: Yep, I'll add that as well.

christiancho:
@Mercy811 I updated a lot of tests related to remote config in the SDK - do you mind taking a look at the tests especially to make sure I didn't change the assertions to break?

Mercy811:
Hi SR team (@lewgordon-amplitude @jpollock-ampl), could you confirm if this is the desired behavior? The original logic from v1.x branch is to https://github.com/amplitude/Amplitude-TypeScript/blob/568554a4849ff9069ae6948783458bc3c2997523/packages/session-replay-browser/src/config/joined-config.ts#L53-L55.

lewgordon-amplitude (approved):
LGTM!

PR #1408, opened 2025-11-19
ci: apply tags

Mercy811:
### Summary

Follow up of https://github.com/amplitude/Amplitude-TypeScript/pull/1407.
"Update test package dependencies" step will let publish step think there is no package changes to publish because when we amended the commit, the tags still pointed to the old commit SHA. This PR update tags to point to the new commit
image

### Checklist


sojingle (approved):
LGTM

PR #1411, opened 2025-11-20
fix(analytics-core): record unsuccessful response from catch error

Mercy811:
### Summary

- A follow up of https://github.com/amplitude/Amplitude-TypeScript/pull/1405. 
- All unsuccessful events have 0 error code. So track the only case where we can get 0 with the key `analytics.events.unsuccessful.from.catch.error`
image

### Checklist


sojingle (approved):
LGTM

PR #1415, opened 2025-11-20
fix: update analytics/kotlin sdk dependency to include Dexguard fix

polbins:
### Summary

Update Analytics/Kotlin SDK to include the fix for Dexguard compilation issue.

### Checklist


polbins:
I made it include a range instead so we don't have to do this often

igor-amp (approved):

PR #1416, opened 2025-11-21
feat(session-replay-browser): move web worker out of experimental

lewgordon-amplitude:
### Summary

This has been in production for awhile now and seems stable. We will still look at `experimental.useWebWorker` to maintain backwards compatibility.

### Checklist

### DIFF
--- a/packages/plugin-session-replay-browser/test/session-replay.test.ts
+++ b/packages/plugin-session-replay-browser/test/session-replay.test.ts
@@ -335,6 +335,20 @@ describe('SessionReplayPlugin', () => {
       );
     });
 
+    test('should not set useWebWorker when not provided (default)', async () => {
+      const sessionReplay = new SessionReplayPlugin({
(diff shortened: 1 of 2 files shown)


packages/plugin-session-replay-browser/test/session-replay.test.ts
```diff
@@ -334,6 +334,36 @@ describe('SessionReplayPlugin', () => {
...
+      });
+      await sessionReplay.setup?.(mockConfig, mockAmplitude);
+
+      expect(init).toHaveBeenCalledTimes(1);
+      expect(init.mock.calls[0][1]).toEqual(
+        expect.objectContaining({
+          useWebWorker: true,
+        }),
+      );
+    });
```
jpollock-ampl, line 380: WDYT about a test that checks for `useWebWorker` default value (i.e. if it isn't provided)?

Issue #1400, opened 2025-11-14
`Previous Page Type` incorrectly stated as `external` when `Previous Page Location` is a subdomain in the same domain


tmchow:
## Expected Behavior

`Previous Page Type` event property should be `internal` when the previous page location is still in the same domain.

Example:
```
[Amplitude] Page URL = https://my.bigcartel.biz/login
[Amplitude] Previous Page Location = https://app.bigcartel.biz/
[Amplitude] Previous Page Type =  internal
```

## Current Behavior

`Previous Page Type` event property is incorrectly stating `external` when the previous page is still in our domain.

Example just now:
```
[Amplitude] Page URL = https://my.bigcartel.biz/login
[Amplitude] Previous Page Location = https://app.bigcartel.biz/
[Amplitude] Previous Page Type =  external
```

## Environment

- JS SDK Version: 
  - `@amplitude/analytics-browser@2.29.0`
  - `@amplitude/plugin-web-attribution-browser@2.1.92`
  - `@amplitude/analytics-client-common@2.4.11` (contains the bug)
- Installation Method: NPM
- Browser and Version: All browsers
- Next.js Version: 15.1.6

PR #1433, opened 2025-12-11
fix(plugin-page-url-enrichment-browser): add internalDomains config for subdomain matching

cyhk:
### Summary

Current logic for `Previous Page URL Type` explicitly ignores subdomains and classifies them as `external`.
This adds a config for the user so that they can specify a list of domains that will be matched to domains and subdomains of the previous page url to decide if it is internal (when there is a match) or an external (when there is no match).

This PR is in response to the issue raised [here](https://github.com/amplitude/Amplitude-TypeScript/issues/1400).

### Checklist

---

> [!NOTE]
> Adds an `internalDomains` option to `pageUrlEnrichmentPlugin` to classify previous pages from specified domains/subdomains as internal, with updated types and tests.
> 
> - **Plugin (`packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts`)**
>   - `pageUrlEnrichmentPlugin` now accepts `{ internalDomains?: string[] }`.
>   - Update `getPrevPageType` to treat previous pages as `internal` when current and/or previous hostnames match any `internalDomains` (supports subdomains); otherwise `external`; `direct` when missing.
>   - Minor typing tweak: `previousURL` can be `string | undefined`.
> - **Types (`packages/analytics-core/src/types/page-url-enrichment.ts`)**
>   - Introduce `PageUrlEnrichmentOptions` with optional `internalDomains: string[]`.
> - **Tests (`packages/plugin-page-url-enrichment-browser/test/page-url-enrichment.test.ts`)**
>   - Add cases validating internal/external classification across subdomains and multiple domains via `internalDomains`.
>   - Adjust existing tests for titles/expectations and direct type handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5a369311414c331800dfbcdc8f2a51ac143b41c. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
+++ b/packages/plugin-page-url-enrichment-browser/src/page-url-enrichment.ts
@@ -24,7 +24,7 @@ export type URLInfo = {
 
 enum PreviousPageType {
   Direct = 'direct', // for no prev page or referrer
-  Internal = 'internal', // for same domain - this excludes subdomains
+  Internal = 'internal', // for internal domains - exact domain matches or when current and previous page are both internal domains
   External = 'external', // for different domains
 }
 


packages/analytics-core/src/types/page-url-enrichment.ts
```diff
@@ -2,4 +2,6 @@
  * @experimental this feature is experimental and may not be stable
  * Using an empty object for now until we have actual options
  */
-export type PageUrlEnrichmentOptions = Record<string, never>;
+export type PageUrlEnrichmentOptions = {
+  internalDomains?: string[];
```
Mercy811, line 8: Can we support regex? Exact match seems not to be enough for pages like `example.com/blog/{blog-id}`. Or they just want to add the whole subdomain as internal

cyhk: Hmm, since we are matching with domains and not full URLs in the plugin logic, this should be fine. Do let me know if I am missing something.

Mercy811: Oh i see!

PR #1434, opened 2025-12-11
fix(analytics-browser): remoteConfig.fetchRemoteConfig was broken

daniel-graham-amplitude:
### Summary

"remoteConfig.fetchRemoteConfig === false" wasn't causing remote config to not be fetched.

This patches it so the logic is as follows:
* if fetchRemoteConfig is false then set remoteConfig.fetchRemoteConfig to false
* if remoteConfig.fetchRemoteConfig is false then set remoteConfig.fetchRemoteConfig to false
* any other case (default) set "remoteConfig.fetchRemoteConfig=true"

### Checklist

---

> [!NOTE]
> Makes remote config fetching opt-out correctly (false if either flag is false) and only initializes the remote config client when enabled.
> 
> - **Config** (`packages/analytics-browser/src/config.ts`):
>   - Normalize `remoteConfig.fetchRemoteConfig` to `false` if either `options.remoteConfig?.fetchRemoteConfig === false` or `options.fetchRemoteConfig === false`; default to `true` otherwise.
>   - Ensure `this.remoteConfig` is initialized to `{}` before setting flags.
> - **Browser client** (`packages/analytics-browser/src/browser-client.ts`):
>   - Read `browserOptions.remoteConfig?.fetchRemoteConfig` into `fetchRemoteConfig` and only create/subscribe the `RemoteConfigClient` when it is truthy.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9afd48e44707049730bf581b37810059d3bc2a79. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -121,12 +121,16 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     this.diagnosticsClient = diagnosticsClient;
 
     // Backward compatibility for fetchRemoteConfig
-    let _fetchRemoteConfig = true;
-    if (
-      remoteConfig?.fetchRemoteConfig !== true && 
-      (remoteConfig?.fetchRemoteConfig === false || fetchRemoteConfig === false)
-    ) {
+    let _fetchRemoteConfig;
+    if (remoteConfig?.fetchRemoteConfig === true) {
+      // set to true if remoteConfig explicitly set to true
+      _fetchRemoteConfig = true;
+    } else if (remoteConfig?.fetchRemoteConfig === false || fetchRemoteConfig === false) {
+      // set to false if either are set to false explicitly
       _fetchRemoteConfig = false;
+    } else {
+      // default to true if both undefined
+      _fetchRemoteConfig = true;
     }
     this.remoteConfig = this.remoteConfig || {};
     this.remoteConfig.fetchRemoteConfig = _fetchRemoteConfig;


packages/analytics-browser/src/config.ts
```diff
@@ -121,15 +121,11 @@ export class BrowserConfig extends Config implements IBrowserConfig {
...
-        this.remoteConfig = {
-          fetchRemoteConfig: fetchRemoteConfig,
-        };
-      }
-      if (!this.remoteConfig.fetchRemoteConfig) {
-        this.remoteConfig.fetchRemoteConfig = fetchRemoteConfig;
-      }
+    this.remoteConfig = this.remoteConfig || {};
+    if (remoteConfig?.fetchRemoteConfig === false || fetchRemoteConfig === false) {
+      this.remoteConfig.fetchRemoteConfig = false;
```
Mercy811: This doesn't seem right to me. IMO, remoteConfig.fetchRemoteConfig should always takes precedence once it's set. 

An edge example is that fetchRemoteConfig = false while remoteConfig.fetchRemoteConfig = true. The current code change will disable remote config, but it should be on

daniel-graham-amplitude: Change made.

packages/analytics-browser/src/browser-client.ts
```diff
@@ -112,7 +112,9 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
 
     let remoteConfigClient: IRemoteConfigClient | undefined;
     // Create remote config client and subscribe to analytics configs
-    if (browserOptions.remoteConfig?.fetchRemoteConfig) {
+    /* istanbul ignore next */
+    const fetchRemoteConfig = browserOptions.remoteConfig && browserOptions.remoteConfig.fetchRemoteConfig;
+    if (fetchRemoteConfig) {
```
Mercy811, line 117: If we are using both of them, we have to keep them synced. I just added one line to the newly added test and it will fail.
Image

We can just keep the original code and make sure we update the deprecated one based on the new one

packages/analytics-browser/src/config.ts
```diff
@@ -121,15 +121,13 @@ export class BrowserConfig extends Config implements IBrowserConfig {
...
-    if (fetchRemoteConfig) {
-      if (!this.remoteConfig) {
-        this.remoteConfig = {
-          fetchRemoteConfig: fetchRemoteConfig,
-        };
-      }
-      if (!this.remoteConfig.fetchRemoteConfig) {
-        this.remoteConfig.fetchRemoteConfig = fetchRemoteConfig;
-      }
+    this.remoteConfig = this.remoteConfig || {};
```
Mercy811: maybe you want this? otherwise you lose remoteConfig.serverUrl
```suggestion
    this.remoteConfig = remoteConfig || {};
```

packages/analytics-browser/src/config.ts
```diff
@@ -121,15 +121,13 @@ export class BrowserConfig extends Config implements IBrowserConfig {
...
     // Backward compatibility for fetchRemoteConfig
-    if (fetchRemoteConfig) {
-      if (!this.remoteConfig) {
-        this.remoteConfig = {
-          fetchRemoteConfig: fetchRemoteConfig,
-        };
-      }
-      if (!this.remoteConfig.fetchRemoteConfig) {
-        this.remoteConfig.fetchRemoteConfig = fetchRemoteConfig;
-      }
```
Mercy811, line 132: this is the fix i'm thinking

```
    if (fetchRemoteConfig) {
      if (this.remoteConfig?.fetchRemoteConfig === undefined) {
        this.remoteConfig = {
          fetchRemoteConfig: fetchRemoteConfig,
        };
      }
```

Mercy811: if the deprecated one presents, then check if the new one presents, and update the one new value accordingly

daniel-graham-amplitude: If the deprecated `fetchRemoteConfig` is undefined though and so is `this.remoteConfig.fetchRemoteConfig` it should be set to `true`.

Mercy811: oh that's true

daniel-graham-amplitude: I just changed it again to make it more intuitive + added some comments.

Mercy811: I noticed that. I was about to comment an edge case where the final result should be false if i only set remoteConfig.fetchRemoteConfig = false

daniel-graham-amplitude: Yeah that case is covered (it was covered before, but it was harder to understand).

PR #1446, opened 2025-12-24
chore: use workspaces

daniel-graham-amplitude:
### Summary

This PR updates versions in package.json to use "workspace:*" when it's an internal package. This is so that it uses the internal package instead of fetching it from the NPM repository.

### Checklist

---

> [!NOTE]
> Adopts pnpm workspaces and links internal packages via `workspace:*`, updating tooling and minor code to align with the new setup.
> 
> - Switches repo to pnpm: set `npmClient` to `pnpm` in `lerna.json`, add pnpm setup/caching to `.github/actions/build-and-test`, run `pnpm install` in `.husky/pre-commit`, tweak root build script (`--skip-nx-cache`), and refresh `pnpm-lock.yaml`
> - Replace pinned internal versions with `workspace:*` across packages (e.g., `analytics-*`, `plugin-*`, `session-replay-*`, `gtm-snippet`, `targeting`, `unified`)
> - Add `@aws-sdk/client-s3` as a dev dependency in the root and multiple packages used by publish scripts
> - Unified SDK: re-export `setTransport` directly from `client` and drop explicit type import; minor import tidy in `analytics-core` `debug.ts`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6fc0511cce0f6ec14c4119d3075c35a59007789d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/utils/debug.ts
+++ b/packages/analytics-core/src/utils/debug.ts
@@ -2,15 +2,9 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { LogConfig, DebugContext, ILogger } from '../logger';
+import { AmplitudeCore } from '../core-client';
+import { LogConfig, DebugContext } from '../logger';
 import { LogLevel } from '../types/loglevel';
-
-interface DebugClient {
-  config: {
-    loggerProvider: ILogger;
-    logLevel: LogLevel;
-  };
-}
 
 export const getStacktrace = (ignoreDepth = 0): string[] => {
   const trace = new Error().stack || '';
@@ -21,7 +15,7 @@ export const getStacktrace = (ignoreDepth = 0): string[] => {
 };
 
 // This hook makes sure we always get the latest logger and logLevel.
-export const getClientLogConfig = (client: DebugClient) => (): LogConfig => {
+export const getClientLogConfig = (client: AmplitudeCore) => (): LogConfig => {
   const { loggerProvider: logger, logLevel } = { ...client.config };
   return {
     logger,
@@ -43,7 +37,7 @@ export const getValueByStringPath = (obj: any, path: string): any => {
   return obj;
 };
 
-export const getClientStates = (client: DebugClient, paths: Array<string>) => (): { [key: string]: any } => {
+export const getClientStates = (client: AmplitudeCore, paths: Array<string>) => (): { [key: string]: any } => {
   const res: { [key: string]: any } = {};
   for (const path of paths) {
     res[path] = getValueByStringPath(client, path);


package.json
```diff
@@ -4,7 +4,7 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "lerna run build --stream --include-dependencies",
+    "build": "lerna run build --stream --include-dependencies --skip-nx-cache",
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is intentional. We want the "build" script to run on every package to ensure a completely clean build.

packages/analytics-core/src/utils/debug.ts
Mercy811, line 1: why do we need to replace AmplitudeCore with DebugClient? Can we separate it to another pr and keep only changes related to use workspaces

packages/plugin-network-capture-browser/package.json
```diff
@@ -44,6 +44,7 @@
     "tslib": "^2.4.1"
   },
   "devDependencies": {
+    "@aws-sdk/client-s3": "^3.229.0",
```
Mercy811, line 47: where is this used? I did a global search but didn't find anywhere it's used

packages/unified/src/index.ts
Mercy811, line 1: it seems to be the same as https://github.com/amplitude/Amplitude-TypeScript/pull/1447.

packages/unified/src/unified-client-factory.ts
Mercy811, line 1: the same as above

Mercy811 (changes requested):

packages/plugin-network-capture-browser/package.json
daniel-graham-amplitude: These are all used in `upload-to-s3.js`. I had to add these to all of the individual packages because PNPM doesn't allow hoisting (ie: accessing dependencies in the top level node_modules). 

(note: hoisting was the cause of the analytics-client-common problem we had)

packages/analytics-core/src/utils/debug.ts
daniel-graham-amplitude: I'm going to try reverting it. I put that there because there was a build error, but I think this may be another case of `pnpm build` not working (similar to yesterday).

packages/unified/src/index.ts
daniel-graham-amplitude: Yeah, the change was made in 2 pr's. This undoes the change.

packages/analytics-core/src/utils/debug.ts
daniel-graham-amplitude: Yeah, I'll leave it out. I think it was another outdated build issue.

Mercy811:
Synced offline. We merged this PR despite of the failed "[Continuous Integration (Nx) / Check Deprecated Packages (pull_request)](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/20607750791/job/59186769787?pr=1451)" as it's intended to update client-common

PR #1478, opened 2026-01-10
feat(analytics-core): add console observer helper

daniel-graham-amplitude:
### Summary

Add new utility module `consoleObserver` that has 2 methods: `addListener` and `removeListener` that can listen for calls to `console` (`log`, `warn`, `error`, etc...)

It works by overriding [Console](https://developer.mozilla.org/en-US/docs/Web/API/console) so that calls to `console` are intercepted and trigger callbacks.

### Checklist

---

> [!NOTE]
> Adds a utility to observe console calls across log levels.
> 
> - New `consoleObserver` with `addListener`/`removeListener` to intercept `console` methods; preserves originals and includes a re-entrancy guard to avoid recursion
> - Supports multiple callbacks per level and a testing-only `_restoreConsole`
> - Unit tests validate callback invocation, multiple listeners, non-function properties, recursion safety, exception handling, and listener removal
> - Demo page `test-server/observers/console.html` wires `consoleObserver` to echo errors and then disconnects after a timeout
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7e8134577d97ab55db72c0e8f067c63750e26ab7. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/observers/console.ts
+++ b/packages/analytics-core/src/observers/console.ts
@@ -5,7 +5,7 @@ type Callback = (logLevel: ConsoleLogLevel, args: any[]) => void;
 
 const globalScope = getGlobalScope();
 /* istanbul ignore next */
-const originalConsole: Console | undefined = globalScope?.console;
+const originalConsole: any = globalScope?.console;
 
 let handlers: { [key in ConsoleLogLevel]?: Array<Callback> } = {};
 
@@ -21,6 +21,7 @@ function overrideConsole(logLevel: ConsoleLogLevel): boolean {
   }
 
   // should not override if original console property is not a function
+  /* eslint-disable-next-line @typescript-eslint/no-unsafe-member-access */
   if (typeof originalConsole[logLevel] !== 'function') {
     return false;
   }
@@ -53,9 +54,10 @@ function overrideConsole(logLevel: ConsoleLogLevel): boolean {
     inConsoleOverride = false;
     return originalFn[logLevel]!.apply(originalConsole, args);
   };
+  /* eslint-disable-next-line @typescript-eslint/no-unsafe-member-access */
   originalFn[logLevel] = originalConsole[logLevel] as (...args: any[]) => void;
   /* eslint-disable-next-line @typescript-eslint/no-unsafe-member-access */
-  (originalConsole as any)[logLevel] = handler;
+  originalConsole[logLevel] = handler;
   return true;
 }
 
@@ -102,7 +104,7 @@ function _restoreConsole() {
   for (const [key, originalHandler] of Object.entries(originalFn)) {
     if (originalHandler) {
       /* eslint-disable-next-line @typescript-eslint/no-unsafe-member-access */
-      (originalConsole as any)[key] = originalHandler;
+      originalConsole[key] = originalHandler;
     }
   }
   originalFn = {};


packages/analytics-core/src/index.ts
```diff
@@ -40,6 +40,7 @@ export { RemoteConfigClient, IRemoteConfigClient, RemoteConfig, Source } from '.
 export { LogLevel } from './types/loglevel';
 export { AMPLITUDE_PREFIX, STORAGE_PREFIX } from './types/constants';
 export { Storage, IdentityStorageType } from './types/storage';
+export { consoleObserver } from './observers/console';
```
Mercy811: Can we add expose `consoleObserver` when it's actually used by another package in case if there is any API change?

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,118 @@
...
+  handlers.clear();
+}
+
+const consoleObserver = {
+  addListener,
+  removeListener,
+  _restoreConsole,
+};
+
+export { consoleObserver, ConsoleLogLevel };
```
Mercy811, line 120: if only export `consoleObserver` as a single object, and there are multiple analytics browser instances, all instances will import the same object. I'd export it as a class and each consumer should initialize their own console observer

daniel-graham-amplitude: > all instances will import the same object

I meant for it to work that way. By being a global singleton object, it ensures that the console override (`global.console = new Proxy(...)`) will only happen once (lazily), and we don't have a scenario where several observers are created and we get a long chain of overrides. Since the only 2 methods are to `addListener` and `removeListener` (which requires you to pass in a reference to the callback used in `addListener`), the functions that call this singleton won't be able to step on eachother's toes.

Network Observer works similarly. Network Observer is defined as a "class" (`class NetworkObserver { ... }`) but then exported as a singleton object (`export const networkObserver = new NetworkObserver();`). The same rationale being that it prevents the native `fetch` and `XHR` from being repeatedly overridden.

I chose to skip making a `ConsoleObserver` class because singleton objects add unnecessary boilerplate. An ESM module is, in a way, already a "singleton" object (ie: you can define public functions and maintain state within them).

packages/analytics-core/src/index.ts
daniel-graham-amplitude: Sure!

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Disconnect a callback function from a console log method
+ * @param callback - The callback function to disconnect
+ */
+function removeListener(callback: Callback) {
+  handlers.forEach((callbacks) => {
+    if (callbacks.includes(callback)) {
+      callbacks.splice(callbacks.indexOf(callback), 1);
+    }
+  });
```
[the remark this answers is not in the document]
daniel-graham-amplitude: Fixed by iterating through the callback list backwards and then deleting them all.

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,117 @@
...
+            inConsoleOverride = true;
+            const callbacks = handlers.get(prop);
+            if (callbacks) {
+              callbacks.forEach((callback) => {
+                try {
+                  callback(prop, args);
+                } catch {
+                  // do nothing
+                }
+              });
```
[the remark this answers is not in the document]
daniel-graham-amplitude: Fixed by doing reverse iteration.

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,118 @@
...
+            }
+            inConsoleOverride = false;
+          }
+        } catch {
+          // do nothing
+        }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This catch was added as a extra safety bit to protect against breaking console log. It's very, very unlikely to ever be reached.

jxiwang (approved):

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,122 @@
+import { getGlobalScope } from '../global-scope';
+
+type ConsoleLogLevel = keyof Console;
+
+const globalScope = getGlobalScope();
+/* istanbul ignore next */
+const originalConsole: Console | undefined = globalScope?.console;
+const isProxySupported = typeof Proxy === 'function';
+const isMapSupported = typeof Map === 'function';
```
jxiwang: qq: What is `Map` and `Proxy`?

daniel-graham-amplitude: [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) is a JS implementation of a key-value store (similar to Java hash map).

[Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) let's you "proxy" an existing object and let's you intercept and redefine behavours of the original object. In this case it's re-defining the behaviour of `window.console`.

packages/analytics-core/src/observers/console.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+// this should only be used for testing
+// restoring console can break console overrides
+function _restoreConsole() {
+  for (const [key, originalHandler] of Object.entries(originalFn)) {
+    if (originalHandler) {
+      /* eslint-disable-next-line @typescript-eslint/no-unsafe-member-access */
+      (originalConsole as any)[key] = originalHandler;
```
jxiwang: nit: Is it possible to not cast as any here? Not a blocker

daniel-graham-amplitude: Done thanks. I just set `originalConsole` as an "any" object.

Mercy811:
bugbot run

PR #1527, opened 2026-02-07
fix(analytics-browser): prevent Form from being bound to 2x

daniel-graham-amplitude:
### Summary

Prevent the same Form from being bound to 2x. This can happen when the same form shows up more than once in the same MutationRecord. A case where this can happen is if a form is appended to the body, and then moved to inside a container

### Checklist

---

> [!NOTE]
> **Low Risk**
> Small, localized change to DOM listener registration with added test coverage; low likelihood of unintended side effects beyond form event tracking.
> 
> **Overview**
> Prevents duplicate form interaction tracking by deduplicating listener attachment per `HTMLFormElement` using a `WeakSet`, avoiding double-bound `change`/`submit` handlers when a form is moved/observed multiple times.
> 
> Adds a regression test that reproduces the duplicate-mutation scenario (append then move into a container) and asserts `form_start`/`form_submit` are only tracked once.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2db56b9e0385edc0aaec7012d4cacfef207ffa51. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-browser/test/plugins/form-interaction-tracking.test.ts
+++ b/packages/analytics-browser/test/plugins/form-interaction-tracking.test.ts
@@ -484,43 +484,60 @@ describe('formInteractionTracking', () => {
       });
     });
 
-    test('should not track form more than once', async () => {
-      const config = createConfigurationMock({
-        defaultTracking: {
-          formInteractions: {
-            shouldTrackSubmit: () => true,
+    describe('duplicate form tracking', () => {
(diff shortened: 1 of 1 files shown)


daniel-graham-amplitude:
bugbot run

PR #1541, opened 2026-02-16
fix(analytics-browser): handle XHR open vs send race condition

daniel-graham-amplitude:
### Summary

There's a race condition where, if `XHR.prototype.open` is monkey-patched after a call to `open` than the monkey-patched `send` and `setRequestHeader` will not have access to the analytics object that is created by `open`. What this PR does is just skips over the analytics capturing logic for `send` and `setRequestHeader` if there was no corresponding call to `open` that was previously captured.

This also changes the log level from `error` to `debug` so that we don't spam customers console. Plus it wraps calls to `debug` in a method called `safeInvoke` that swallows any exceptions, so that we don't worry about the case where loggerProvider could have been mistakenly overridden.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches browser network instrumentation by changing how XHR `send`/`setRequestHeader` are wrapped; mistakes could drop or alter network event capture, but changes are guarded and backed by targeted tests.
> 
> **Overview**
> Fixes an XHR monkey-patching race by **skipping analytics capture in `XMLHttpRequest.prototype.send` and `setRequestHeader` when `$$AmplitudeAnalyticsEvent` was never set by a patched `open`**, while still calling the original XHR methods.
> 
> Reduces customer console noise and hardens error handling by switching several internal logs from `error` to `debug` and wrapping all debug logging in a new `safeInvoke()` to prevent logger failures from impacting network flows. Adds tests covering the `open`-before-patch vs `open`-after-patch behavior and ensures XHR prototype patches are reset between tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a32eae34ecc0002b4cafdb6f9bc697044a12c858. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/observers/network.ts
+++ b/packages/analytics-core/src/observers/network.ts
@@ -47,7 +47,7 @@ type RequestUrlAndMethod = {
 // Use great care when modifying this type, make sure you only use read-only properties and only add
 // what you need to access, nothing more.
 type AmplitudeXMLHttpRequestSafe = {
-  $$AmplitudeAnalyticsEvent?: AmplitudeAnalyticsEvent;
+  $$AmplitudeAnalyticsEvent: AmplitudeAnalyticsEvent;
   status: number;
   responseText: string;
   responseType: XMLHttpRequestResponseType;


packages/analytics-core/src/observers/network.ts
```diff
@@ -368,35 +382,42 @@ export class NetworkObserver {
...
-          const responseHeaders = xhrSafe.getAllResponseHeaders();
-          const responseBodySize = xhrSafe.getResponseHeader('content-length');
-
-          const responseWrapper = new ResponseWrapperXhr(
-            xhrSafe.status,
-            responseHeaders,
+      // if xhrSafe.$$AmplitudeAnalyticsEvent is not set, it means that
+      // the xhr.open method was called before we monkey-patched XHR and
+      // the event is missed
+      if (xhrSafe.$$AmplitudeAnalyticsEvent) {
```
daniel-graham-amplitude, line 388: This diff overcomplicates the actual logic. All it does is wrap the method in this `if` statement. The inner contents are the same.

jxiwang (approved):
Is it throwing an error because `networkObserverContext` isn't available yet?

packages/analytics-core/src/observers/network.ts
Mercy811: Hide whitespace will give a better diff 
image

packages/analytics-core/src/observers/network.ts
```diff
@@ -223,7 +233,7 @@ export class NetworkObserver {
         timestamps = this.getTimestamps();
       } catch (error) {
         /* istanbul ignore next */
-        this.logger?.debug('an unexpected error occurred while retrieving timestamps', error);
+        safeInvoke(() => this.logger?.debug('an unexpected error occurred while retrieving timestamps', error));
```
Mercy811, line 236: Super nit: I just saw others have `?.` so i'm thinking maybe this needs it too. Or if it's the other way around
```suggestion
        safeInvoke(() => this.logger?.debug?.('an unexpected error occurred while retrieving timestamps', error));
```

packages/analytics-core/src/observers/network.ts
```diff
@@ -56,6 +56,14 @@ type AmplitudeXMLHttpRequestSafe = {
   addEventListener: (type: 'loadend', listener: () => void) => void;
 };
 
+function safeInvoke(fn: () => void) {
```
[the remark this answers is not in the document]
daniel-graham-amplitude: I'm going to close this one. Even if the typeguard is removed mistakenly, any error that's thrown as a result of the object not being present will be swallowed by try-catch.

packages/analytics-core/src/observers/network.ts
daniel-graham-amplitude: Fixed. I removed the `?.` on them all.

PR #1542, opened 2026-02-17
feat(analytics-browser): support gzip request body compression

Mercy811:
### Summary

- Gzip request if server url is default. 
- Add a new configuration `enableRequestBodyCompression` to disable it if server url is set to a proxy url
- Same logic as https://github.com/amplitude/Amplitude-Swift/commit/12bbd67a4d8c08b58525e6cdddd40075e7ca3b1a
- Only payload string length larger than 2k will be compressed

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches the core event upload pipeline and transport interface; mistakes could break ingestion or proxy setups, though changes are gated by endpoint checks, size thresholds, and extensive tests.
> 
> **Overview**
> Adds **optional gzip compression for event upload bodies** by extending the core `Transport.send()` API and plumbing a new `enableRequestBodyCompression` config through browser/core config and the destination upload path.
> 
> Compression is **forced for Amplitude’s default ingestion endpoints**, while custom `serverUrl`s only compress when explicitly enabled; browser transports (`XHRTransport` and a browser-local `FetchTransport`) gzip payloads �≥2MB when `CompressionStream` is available, and `SendBeaconTransport` explicitly opts out since it can’t set headers. E2E and unit tests are updated/added to handle gzipped bodies (including request parsing helpers) and to validate compression behavior across transports and endpoint types.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df2cac1b7996effa6836912390407fa0c9368c7d. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/test/plugins/destination.test.ts
+++ b/packages/analytics-core/test/plugins/destination.test.ts
@@ -636,7 +636,11 @@ describe('destination', () => {
 
         await destination.send([context]);
 
-        expect(transportProvider.send).toHaveBeenCalledWith(customServerUrl, expect.any(Object), expectedShouldCompress);
+        expect(transportProvider.send).toHaveBeenCalledWith(
+          customServerUrl,
+          expect.any(Object),
+          expectedShouldCompress,
+        );
       },
     );
 


packages/analytics-core/src/transports/gzip.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Compress a string to gzip and return the result as an ArrayBuffer.
+ * Payload is small so buffering is fine. Used by Fetch and XHR transports.
+ */
+export async function compressToGzipArrayBuffer(data: string): Promise<ArrayBuffer> {
+  const CompressionStreamImpl = CompressionStream;
+  if (typeof CompressionStreamImpl === 'undefined') {
+    throw new Error('CompressionStream is not available');
+  }
+  const stream = new Blob([data]).stream().pipeThrough(new CompressionStreamImpl('gzip'));
+  return new Response(stream).arrayBuffer();
```
daniel-graham-amplitude: I think returning an Array Buffer is an unnecessary "middle-person". If we just return `stream`, then the result of `stream` I believe can be passed directly into the fetch body?

Mercy811: the steam is [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) type which is better used for larger payload (> 1MB) and to send ReadableStream, we also have to set an additional header [duplex: half](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#duplex) which might not be supported by the backend

Mercy811:
bugbot run

packages/analytics-core/src/transports/fetch.ts
```diff
@@ -2,27 +2,37 @@ import { BaseTransport } from './base';
...
+    let body: string | ArrayBuffer = bodyString;
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      Accept: '*/*',
+      ...this.customHeaders,
+    };
+    if (this.shouldCompressUploadBody && isCompressionStreamAvailable()) {
+      body = await compressToGzipArrayBuffer(bodyString);
+      headers['Content-Encoding'] = 'gzip';
+    }
```
daniel-graham-amplitude: Adding compression has some trade-offs

_Benefits_: the payload is compressed to a smaller size and less bandwidth is consumed
_Drawbacks_: the compression algorithm runs on the user's machine (not blocking but still consuming CPU); an ArrayBuffer is created which 2x's the memory usage; the creation of the ArrayBuffer uses some CPU cycles (this is blocking)

In sum, we're trading memory and cpu cycles for reduced bandwidth.

I just did some benchmarking from my browser console to see how much compression gives us 
1. `hello world` before compression is 11 bytes before compression and ~30 bytes after (increases the size after compression)
2. a sample payload (taken from our website) that has 1 event is 2kb before compression and 0.9kb after compression (>50% compression)
3. a sample payload that has 28 events is 58kb before compression and  5.3kb after compression (>90% compression!)

Based on this data, I think we ought to set a minimum threshold at which we apply compression. @Mercy811 do you have any thoughts on what that should be? I think 1.5kb (uncompressed) would be a good minimum.

packages/analytics-browser/src/transports/xhr.ts
```diff
@@ -37,10 +46,25 @@ export class XHRTransport extends BaseTransport implements Transport {
...
+        xhr.send(body);
+      };
+
+      if (this.shouldCompressUploadBody && isCompressionStreamAvailable()) {
+        compressToGzipArrayBuffer(bodyString)
+          .then((body) => sendBody(body, 'gzip'))
+          .catch(reject);
+      } else {
+        sendBody(bodyString);
       }
```
daniel-graham-amplitude: Do we have any sense of how many of our customers actually use `xhr` instead of the default `fetch` as their `transport` type?

I'm just wondering, because if the number is very low and XHR is essentially "legacy" technology, then I'm not sure if it's worthwhile for us to be adding features to it (adding more bytes to the bundle).

packages/analytics-core/src/transports/fetch.ts
Mercy811: isn't 1 less realistic, 2 and 3 reflect the real situations as we would send at least one event. I'd keep the logic simple and same as other platforms.

packages/analytics-browser/src/transports/xhr.ts
Mercy811: > Do we have any sense of how many of our customers actually use xhr instead of the default fetch as their transport type?

We don't know that, but we cannot assume that no one uses it. Network tracking supports xhr too. To support xhr here we just need to add less than 30 lines of code, I don't think that will affect bundle size much.

packages/analytics-browser/src/config.ts
```diff
@@ -454,18 +457,32 @@ export const shouldFetchRemoteConfig = (options: BrowserOptions = {}): boolean =
...
-export const createTransport = (transport?: TransportTypeOrOptions) => {
+/**
+ * Custom servers may not support gzip compression—only enable it if opted in.
+ * When using the SDK's default endpoints, request bodies are always compressed.
+ */
+export const getShouldCompressUploadBody = (
+  serverUrl: string | undefined,
+  enableRequestBodyCompression = false,
+): boolean => {
+  if (serverUrl !== undefined && serverUrl !== '') {
```
Mercy811: Note that `getShouldCompressUploadBody()` returns `enableRequestBodyCompression` when server url is an empty string too because that's the default value set by browser config at https://github.com/amplitude/Amplitude-TypeScript/blob/e082ac7bbc50d1e8fc2d3e77ae7f6607924dfc9b/packages/analytics-browser/src/config.ts#L87 Looks like we can just update it to `serverUrl?: string,` not sure why we had it like this initially. 

server url is later set to the real default value in core config at https://github.com/amplitude/Amplitude-TypeScript/blob/e082ac7bbc50d1e8fc2d3e77ae7f6607924dfc9b/packages/analytics-core/src/config.ts#L81-L83

packages/analytics-core/src/transports/fetch.ts
daniel-graham-amplitude: 1 isn't realistic in our use-case, but considering that this class is implemented as a general purpose "fetch" class, it's not impossible that another package could use this class and that package may have a need for lower payloads.

I don't believe #2 is a "compression-worthy" payload size. 2kb is a tiny amount which would have no effect on the round trip time (RTT) because 2kb is well below the TCP ingestion window (14kb). We would be missing the benefits of compression, but get the drawbacks (increased ram usage, increased CPU usage, increased main thread blocking time from CompressionStream* and .arrayBuffer() invocation)

> I'd keep the logic simple and same as other platforms

We need to consider the different requirements of Web vs Mobile. Because of the single-threadedness of JS, it's quite a bit more "performance-sensitive" than iOS or Android.

* _I was incorrect previously. CompressionStream does block the main thread, but does so in small pieces_

packages/analytics-core/src/transports/gzip.ts
daniel-graham-amplitude: Okay, yeah sounds like SR had this same issue. We can keep it as-is, but this does increase the "compression tax" because it adds another variable and computation to this method.

packages/analytics-core/src/transports/gzip.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Compress a string to gzip and return the result as an ArrayBuffer.
+ * Payload is small so buffering is fine. Used by Fetch and XHR transports.
+ */
+export async function compressToGzipArrayBuffer(data: string): Promise<ArrayBuffer> {
+  const CompressionStreamImpl = CompressionStream;
+  if (typeof CompressionStreamImpl === 'undefined') {
+    throw new Error('CompressionStream is not available');
```
daniel-graham-amplitude: Idea: what if instead of throwing exceptions, we refactor `compressToGzipArrayBuffer` to be a "best-effort" function and have it just return `undefined` if it doesn't work

```
if (typeof CompressionStreamImpl === 'undefined') {
  return undefined;
}
try {
  const stream //...
} catch (e) {
  return undefined;
}
```

My concern is that if any other packages use this method, they'll all need to remember to check for `window.CompressionStream` before calling this, or risk an uncaught exception.

packages/analytics-core/src/transports/fetch.ts
Mercy811: added 2kb threshold

packages/analytics-browser/src/config.ts
Mercy811: move the shouldCompressUploadBody calculation logic to destination plugin to avoid undecided serverUrl here

packages/analytics-core/src/transports/gzip.ts
Mercy811: Good idea! Updated

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+      const compressed = await compressToGzipArrayBuffer(bodyString);
+      if (compressed) {
+        headers['Content-Encoding'] = 'gzip';
+        body = compressed;
+      }
+    }
+    headers = {
+      ...headers,
+      ...this.customHeaders,
+    };
```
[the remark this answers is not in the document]
Mercy811: this is intentional

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+    /* istanbul ignore if */
+    if (typeof fetch === 'undefined') {
+      throw new Error('FetchTransport is not supported');
+    }
+
+    const bodyString = JSON.stringify(payload);
+    const shouldCompressBody =
+      shouldCompressUploadBody &&
+      getStringSizeInBytes(bodyString) >= MIN_GZIP_UPLOAD_BODY_SIZE_BYTES &&
+      isCompressionStreamAvailable();
```
daniel-graham-amplitude, line 32: Looking at the `getStringSizeInBytes` helper functions, it looks like it runs `new TextEncoder(value)`, which creates a copy of the payload and then returns the size, meaning there's an additional copy of the payload in memory (in byte format) that is generated in O(n) time (because it has to iterate through each character in the string and copy it over).

I'm thinking because this threshold is just a heuristic anyways, maybe could just define the minimum upload size as a minimum string length rather than a minimum byte size, then we'd just need to check `bodyString.length`. 

I understand that this goes against our agreement to make it 2kb, but string length should be "good enough". Most characters in a string are 1 byte each, and at most they're 3 bytes each, which gives an upper bound on how much we can under-exaggerate the "size" of a string.

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+      'Content-Type': 'application/json',
+      Accept: '*/*',
+    };
+
+    if (shouldCompressBody) {
+      const compressed = await compressToGzipArrayBuffer(bodyString);
+      if (compressed) {
+        headers['Content-Encoding'] = 'gzip';
+        body = compressed;
+      }
```
daniel-graham-amplitude, line 45: Awesome! This nice.

packages/analytics-browser/src/transports/send-beacon.ts
```diff
@@ -10,7 +18,8 @@ export class SendBeaconTransport extends BaseTransport implements Transport {
       }
       try {
         const data = JSON.stringify(payload);
-        const success = globalScope.navigator.sendBeacon(serverUrl, JSON.stringify(payload));
+        // SendBeacon cannot set Content-Encoding, so we always send uncompressed
+        const success = globalScope.navigator.sendBeacon(serverUrl, data);
```
daniel-graham-amplitude, line 22: Nice catch on fixing the payload being "stringified" twice.

packages/analytics-browser/src/transports/xhr.ts
```diff
@@ -31,16 +40,45 @@ export class XHRTransport extends BaseTransport implements Transport {
...
+            headers['Content-Encoding'] = 'gzip';
+            sendBody(compressed);
+          } else {
+            sendBody(bodyString);
+          }
+        } else {
+          sendBody(bodyString);
+        }
+      };
+      void doSend();
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This seems legitimate. Maybe we could just do it all from inside the promise constructor without inline functions, to avoid this issue?

```
let body: string | ArrayBuffer = bodyString;
if (shouldCompressBody) {
  const compressed = await compressToGzipArrayBuffer(bodyString);
  if (compressed) {
    headers['Content-Encoding'] = 'gzip';
    body = compressed; 
  }
}

// Merge headers: custom headers override defaults (consistent with FetchTransport)
headers = {
  ...headers,
  ...this.customHeaders,
};

for (const [key, value] of Object.entries(headers)) {
  xhr.setRequestHeader(key, value);
}
xhr.send(body);
```

packages/analytics-browser/src/transports/send-beacon.ts
```diff
@@ -1,7 +1,15 @@
...
+ * SendBeacon does not support custom headers (e.g. Content-Encoding: gzip),
+ * so request body compression is not applied even when enableRequestBodyCompression is true.
+ */
 export class SendBeaconTransport extends BaseTransport implements Transport {
-  async send(serverUrl: string, payload: Payload): Promise<Response | null> {
+  constructor() {
+    super();
+  }
+
+  async send(serverUrl: string, payload: Payload, _enableRequestBodyCompression = false): Promise<Response | null> {
```
daniel-graham-amplitude, line 12: Since `_enableRequestBodyCompression` is an optional parameter in the BaseTransport method, you should just be able to leave it out of the `send` type signature, since it's not being used.

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+  MIN_GZIP_UPLOAD_BODY_SIZE_BYTES,
+  Payload,
+  Response,
+  Transport,
+} from '@amplitude/analytics-core';
+
+// Temporary browser-specific fetch transport with gzip support.
+// TODO: Merge this implementation back into @amplitude/analytics-core FetchTransport
+// once React Native SDK supports request body gzip.
+export class FetchTransport extends BaseTransport implements Transport {
```
daniel-graham-amplitude, line 14: What was the reason for making a new fetch.ts in analytics-browser? Is it because the core fetch.ts fails to compile in other environments (like React Native), when you add the compression logic?

packages/analytics-browser/src/transports/xhr.ts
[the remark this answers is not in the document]
Mercy811: Thanks for taking a look. I updated to `doSend().catch(reject);` to solve this. 

I didn't have the same code with fetch because I wanna avoid using `await` inside the new Promise constructor  which then need to be 
```
return new Promise(async (resolve, reject) => {
```
This double promise is hard to read and easily cause unhandled rejection.

packages/analytics-browser/src/transports/fetch.ts
Mercy811: This is a good point! English letter, number, and common punctuation are all 1 byte each. So I'm keeping `2 * 1024 * 1024` and compare it with string length.

packages/analytics-browser/src/transports/fetch.ts
Mercy811: Actually I re-thought about this and felt it makes more sense to not allow custom headers override default headers (`Content-Type`, `Accept`, `Content-Encoding`). If it's useless and even harmful to override default headers if transport providers keeps their behavior to set json or gzip payload. Custom headers are support to support proxy authorization in real world for example.

packages/analytics-browser/src/transports/fetch.ts
Mercy811: React Native import core fetch. I don't wanna let it use the gzip feature now because by default we compress

packages/analytics-browser/src/transports/send-beacon.ts
Mercy811: Oh I just wanna make it look aligned in browser platform. 

>_enableRequestBodyCompression is an optional parameter in the BaseTransport method,

Yes! So we can keep the http transport in node sdk remains unchanged for now

packages/analytics-browser/src/transports/xhr.ts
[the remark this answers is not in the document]
Mercy811: @cursoragent I just pushed commits to fix this. Review again

packages/analytics-browser/src/transports/fetch.ts
daniel-graham-amplitude: Does React Native fail at runtime or compilation time though? Couldn't we just do a check on `!!globalScope.CompressionStream` and if it returns `false` don't run gzip for that platform? I'm just concerned about having 2 separate implementations of fetch that would need to be maintained.

packages/analytics-core/src/transports/base.ts
```diff
@@ -8,7 +8,7 @@ import { Response } from '../types/response';
 import { isSuccessStatusCode } from '../utils/status-code';
 
 export class BaseTransport implements Transport {
-  send(_serverUrl: string, _payload: Payload): Promise<Response | null> {
+  send(_serverUrl: string, _payload: Payload, _enableRequestBodyCompression?: boolean): Promise<Response | null> {
```
daniel-graham-amplitude, line 11: Thinking about the signature of this some more. I feel like there's a chance, in the future, that there could be more options that we need to add to `send` and the method signature would be pretty long.

Would it be much trouble if we did it like this:
```
type SendOptions = { _enableRequestBodyCompression?: boolean };
send(_serverUrl: string, _payload: Payload, _sendOptions?: SendOptions)
```

I'll leave this one as "optional".

packages/analytics-browser/src/transports/fetch.ts
Mercy811: The browser fetch is for short term. After dogfooding, we can have another pr for react native and merge it to the gzip feature branch and then merge gzip feature branch into main together. I tend to avoid making this giant pr bigger.

packages/analytics-core/src/transports/base.ts
Mercy811: I thought about this too but realized we probably need a refactor on the whole class. It doesn't make sense to me to set custom headers as a class property but pass server url as a parameter to send. They should be treated the same and as well as the should compress flag.

Mercy811: I’ll keep it as it is because it’s an internal interface, unlike storage provider.

packages/analytics-core/src/types/config/core-config.ts
```diff
@@ -62,6 +62,12 @@ export interface IConfig {
    * Amplitude internal use.
    */
   ingestionMetadata?: IngestionMetadata;
+  /**
+   * Controls request body compression **only** when a custom `serverUrl` is configured.
+   * When using the SDK's default endpoints, request bodies are always compressed regardless of this setting.
+   * @defaultValue `false`
```
daniel-graham-amplitude, line 72: Can we re-word this JSDoc to a format that says what it does briefly in the first line; and then in the next line(s) gives more details

Something like
```
Compress network request body payloads with gzip

(only applies compression when it is supported by the platform; the instance hasn't been configured to use a proxy server via `serverUrl`; the payload size is below the minimum size threshold; the transport method supports setting request headers)
```

Mercy811: updated

crleona (approved):

PR #1548, opened 2026-02-21
feat(analytics-browser): add config attribution.excludeInternalReferrers

daniel-graham-amplitude:
### Summary

Adds a new configuration to `attribution` called `excludeInternalReferrers` that
* when `true` or `excludeInternalReferrers.condition === 'always'`, do not track campaign information if document.referrer and location.hostname are on the same domain (ie: block all internal referrers)
* when `excludeInternalReferrers.condition === 'ifEmptyCampaign'`, do not track campaign information if document.referrer and location.hostname are on the same domain and the campaign is empty (no UTM parameters or click ID's)

Why: this is so that customers can easily safeguard against unwanted internal referrers. 

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes campaign attribution decision logic to optionally ignore same-site referrers, which can alter when `$identify` attribution updates fire and relies on new domain-parsing heuristics.
> 
> **Overview**
> Adds a new attribution config, `excludeInternalReferrers`, to prevent campaign updates from being recorded when the referrer is on the same effective domain as the current page (either **always** or only when the campaign has no UTM/click-id params via `condition: 'ifEmptyCampaign'`).
> 
> Threads a computed `topLevelDomain` through `BrowserConfig` into `WebAttribution`/`isNewCampaign` and introduces domain/subdomain helpers (`getDomain`, `isSubdomainOf`, known 2LD list) to classify internal referrers; exports `ExcludeInternalReferrersOptions` and `EXCLUDE_INTERNAL_REFERRERS_CONDITIONS` from `analytics-core` and updates unit tests and test-server examples accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 012529dccc260dc1fdaf0475fc8ce19271ef5dd1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-browser/src/config.ts
+++ b/packages/analytics-browser/src/config.ts
@@ -296,6 +296,9 @@ export const useBrowserConfig = async (
   // Step 1: Create identity storage instance
   const identityStorage = options.identityStorage || DEFAULT_IDENTITY_STORAGE;
   let topLevelDomain = '';
+
+  // use the getTopLevelDomain function to find the TLD only if identity storage
+  // is cookie (because getTopLevelDomain() uses cookies)
   if (identityStorage === DEFAULT_IDENTITY_STORAGE) {
     topLevelDomain = await getTopLevelDomain();
   }


packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -32,6 +70,34 @@ export const isNewCampaign = (
...
+    ) {
+      logger.debug(
+        `This is not a new campaign because ` +
+          `referring_domain=${current.referring_domain} is on the same domain as the current page ` +
+          `and it is configured to exclude internal referrers with empty campaign parameters.`,
+      );
+      return false;
+    } else {
+      logger.error('Invalid configuration provided for attribution.excludeInternalReferrers.');
+    }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: Fixed

daniel-graham-amplitude:
bugbot run

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -22,6 +60,42 @@ const isDirectTraffic = (current: Campaign) => {
...
+  if (
+    typeof excludeInternalReferrers === 'boolean' ||
+    (typeof excludeInternalReferrers === 'object' &&
+      typeof excludeInternalReferrers.condition === 'string' &&
+      ['always', 'ifEmptyCampaign'].includes(excludeInternalReferrers.condition))
+  ) {
+    return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
daniel-graham-amplitude: @cursoragent in the next conditional it checks if `typeof excludeInternalReferrers.condition === 'undefined'` which covers the case of `excludeInternalReferrers = {}`

packages/analytics-core/src/types/constants.ts
```diff
@@ -109,3 +109,8 @@ export const SAFE_HEADERS = [
 
 // list of Network headers to never capture
 export const FORBIDDEN_HEADERS = ['authorization', 'cookie', 'set-cookie'];
+
+export const EXCLUDE_INTERNAL_REFERRERS_CONDITIONS = {
+  always: 'always',
+  ifEmptyCampaign: 'ifEmptyCampaign',
+};
```
Mercy811: It seems that this is not used anywhere and can be deleted

daniel-graham-amplitude: Fixed, good catch thx.

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -93,3 +129,102 @@ export const getDefaultExcludedReferrers = (cookieDomain: string | undefined) =>
...
+
+  return `${name}.${tld}`;
+};
+
+const isInternalReferrer = (referringDomain: string) => {
+  const globalScope = getGlobalScope();
+  /* istanbul ignore if */
+  if (!globalScope) return false;
+  return isSameDomain(referringDomain, globalScope.location.hostname);
+};
```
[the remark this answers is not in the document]
daniel-graham-amplitude: Will leave this as a known issue for now. Ports has always been a problem.

test-server/autocapture/element-interactions.html
```diff
@@ -274,7 +274,18 @@ <h2>Content Changing Button</h2>
...
+            attribution: {
+              excludeReferrers: [
+                'google.com',
+              ],
+              excludeInternalReferrers: {
+                condition: 'ifEmptyCampaign',
+                //condition: 'always',
+              },
+              //excludeInternalReferrers: false,
+            },
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is okay.

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -22,16 +32,41 @@ const isDirectTraffic = (current: Campaign) => {
...
+  return Object.values(campaignWithoutReferrer).every((value) => !value);
+};
+
 export const isNewCampaign = (
   current: Campaign,
   previous: Campaign | undefined,
   options: Options,
   logger: ILogger,
   isNewSession = true,
+  cookieDomain?: string,
```
Mercy811: Why do we add cookieDomain here? IMO we should never mix two features (cookieOptions and attribution) together. If a customer sets cookieDomain to an empty string, regardless of referring domain, all traffic will be considered as external referrer in current implementation of `isInternalReferrer()`

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -93,3 +135,104 @@ export const getDefaultExcludedReferrers = (cookieDomain: string | undefined) =>
...
+  'github.io',
+  'gitlab.io',
+  'cloudfront.net',
+  'herokuapp.com',
+  'appspot.com',
+  'azurewebsites.net',
+  'firebaseapp.com',
+];
+
+export const getDomain = (hostname: string) => {
```
Mercy811, line 217: I just realized that we probably need to apply the same logic to `getReferrer()` in https://github.com/amplitude/Amplitude-TypeScript/blob/d1fc50dbc4a8f7f3bb1989e1cc5b5ec4b48b8814/packages/analytics-core/src/campaign/campaign-parser.ts#L54


Before 2ld fix, let's say we have 

- corp1.co.uk
- corp2.co.uk

corp1 and corp2 are owned by different corps and are external traffic to each other. If a user first browsed corp1.co.uk internally and then went from corp1 to corp2. The SDK would set both previous referring domain and current referring domain = co.uk and `hasNewDomain`. It would falsely consider this is not a new campaign.

packages/analytics-browser/src/attribution/helpers.ts
daniel-graham-amplitude: My rationale is that the only use-case for setting `cookieDomain` is because you're operating from a multi-tenanted web domain and need to isolate your subdomain from other subdomains of that same domain. And thus "cookieDomain" indicates that you consider only a domain that's a subdomain of this domain to be internal.

For example, if you have a site running from mysite.shopify.com you'd want to set your cookie domain to mysite.shopify.com so that other subdomains of shopify.com can't access those cookies. I'd make the logical conclusion from this that if you set the cookie domain to "mysite.shopify.com" you consider any other domain under "shopify.com" that isn't a subdomain of "mysite.shopify.com" to be external.

daniel-graham-amplitude: If it's an "empty string" it will fall back to the `getDomain` function:

```
const internalDomain = cookieDomain || getDomain(globalScope.location.hostname);
```

packages/analytics-browser/src/attribution/helpers.ts
daniel-graham-amplitude: Yeah let's make note of this

packages/analytics-browser/src/attribution/helpers.ts
daniel-graham-amplitude: @Mercy811 I have a better idea coming up shortly.

daniel-graham-amplitude: @Mercy811 I refactored it. Now it passes the top level domain (as computed in `getTopLevelDomain`) as an argument in Attribution instead of using `cookieOptions.domain`.

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -93,3 +135,104 @@ export const getDefaultExcludedReferrers = (cookieDomain: string | undefined) =>
...
+  'org.cn',
+  'com.mx',
+  'github.io',
+  'gitlab.io',
+  'cloudfront.net',
+  'herokuapp.com',
+  'appspot.com',
+  'azurewebsites.net',
+  'firebaseapp.com',
+];
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is just a fallback in case `getTopLevelDomain` fails.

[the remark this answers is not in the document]
daniel-graham-amplitude: This is only a fallback in-case we don't have top level domain

packages/analytics-browser/src/config.ts
```diff
@@ -293,9 +295,12 @@ export const useBrowserConfig = async (
 ): Promise<IBrowserConfig> => {
   // Step 1: Create identity storage instance
   const identityStorage = options.identityStorage || DEFAULT_IDENTITY_STORAGE;
+  let topLevelDomain = '';
+  if (identityStorage === DEFAULT_IDENTITY_STORAGE) {
+    topLevelDomain = await getTopLevelDomain();
+  }
```
Mercy811, line 304: why is topLevelDomain decided by identityStorage? Would it make more sense to make treat top level domain as a value solely based on location.hostname?

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -22,16 +32,41 @@ const isDirectTraffic = (current: Campaign) => {
...
+  return Object.values(campaignWithoutReferrer).every((value) => !value);
+};
+
 export const isNewCampaign = (
   current: Campaign,
   previous: Campaign | undefined,
   options: Options,
   logger: ILogger,
   isNewSession = true,
+  topLevelDomain?: string,
```
Mercy811, line 46: Thanks! topLevelDomain makes much more sense to me!

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -93,3 +135,104 @@ export const getDefaultExcludedReferrers = (cookieDomain: string | undefined) =>
...
+
+  return `${name}.${tld}`;
+};
+
+const isInternalReferrer = (referringDomain: string, topLevelDomain?: string) => {
+  const globalScope = getGlobalScope();
+  /* istanbul ignore if */
+  if (!globalScope) return false;
+  // if referring domain is subdomain of config.cookieDomain, return true
+  const internalDomain = (topLevelDomain || '').trim() || getDomain(globalScope.location.hostname);
```
Mercy811, line 236: nice! topLevelDomain should always be a non empty string in a browser environment but getDomain with 2LDs is a safe net when topLevelDomain is an empty string.

packages/analytics-core/src/types/config/browser-config.ts
```diff
@@ -226,6 +227,12 @@ export interface AttributionOptions {
    * @defaultValue `[/your-domain\.com$/]`
    */
   excludeReferrers?: (string | RegExp)[];
+  /**
+   * Exclude internal referrers from campaign attribution.
+   * (a referrer is 'internal' if it is on the same domain as the current page)
+   * @experimental this feature is experimental and may not be stable
+   */
+  excludeInternalReferrers?: true | false | ExcludeInternalReferrersOptions;
```
Mercy811, line 236: can we add the default value here too? I know we have default value inside ExcludeInternalReferrersOptions, but IDE doesn't display nested options. Here is what I got by hovering on attribution for example

Image

packages/analytics-browser/src/config.ts
daniel-graham-amplitude: > why is topLevelDomain decided by identityStorage

"getTopLevelDomain" only works if cookies is supported, because it works by trying to set cookies and if it sticks, that's the TLD. I believe that was probably the original reason we only retrieve that value when identityStorage is `cookies`.

> Would it make more sense to make treat top level domain as a value solely based on location.hostname

Exclude Internal Referrers falls back to getting domain from `location.hostname`. But the problem is it's not as accurate as `getTopLevelDomain` because we don't have an exhaustive list of 2LD's to check the domain against.

daniel-graham-amplitude: Added comment

Mercy811: Customers sometimes set identity storage for something else for privacy too. That's why I don't think we should ever tie them together. 

Sync offline that we would keep it as it is now as it's there since the beginning but add a comment to explain for future reference

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -22,16 +32,41 @@ const isDirectTraffic = (current: Campaign) => {
...
+    ) {
+      if (condition === 'always') {
+        debugLogInternalReferrerExclude(condition, current.referring_domain, logger);
+        return false;
+      } else if (condition === 'ifEmptyCampaign' && isEmptyCampaign(current)) {
+        debugLogInternalReferrerExclude(condition, current.referring_domain, logger);
+        return false;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is expected. This feature is more meant for people who explicitly set excludeReferrers to begin with

packages/analytics-browser/src/attribution/helpers.ts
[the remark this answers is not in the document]
daniel-graham-amplitude: This is just a fallback in case the more robust 'getTopLevelDomain' doesn't work.

packages/analytics-browser/src/attribution/helpers.ts
```diff
@@ -93,3 +135,104 @@ export const getDefaultExcludedReferrers = (cookieDomain: string | undefined) =>
...
+  return `${name}.${tld}`;
+};
+
+const isInternalReferrer = (referringDomain: string, topLevelDomain?: string) => {
+  const globalScope = getGlobalScope();
+  /* istanbul ignore if */
+  if (!globalScope) return false;
+  // if referring domain is subdomain of config.cookieDomain, return true
+  const internalDomain = (topLevelDomain || '').trim() || getDomain(globalScope.location.hostname);
+  return isSubdomainOf(referringDomain, internalDomain);
```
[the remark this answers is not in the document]
daniel-graham-amplitude: Known issue. Small edge case

PR #1559, opened 2026-02-27
fix: make default excluded referrers work better

daniel-graham-amplitude:
### Summary

Our [documentation for exclude referrers](https://amplitude.com/docs/sdks/analytics/browser/browser-sdk-2#advanced-configuration-for-marketing-attribution-tracking) says "When this option isn't set, the SDK excludes the current domain (and its subdomains)". Technically this isn't true right now. It's only true if `identityStorage` is not equal to `cookie` and if `getTopLevelDomain` does not fail.

What this PR does is, if it fails to retrieve "topLevelDomain" via `getTopLevelDomain`, than fall back to the "getDomain" (non-cookie parsing method). And then when creating "default referrers", use topLevelDomain if cookie domain is not present.

(extra: beefed up the KNOWN_2LD's list to include some more commonly used domains)

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches domain-detection used for marketing attribution and default referrer exclusion, which can subtly change campaign attribution behavior across sites/domains. Risk is mitigated by added/updated unit tests, but edge-case hostname parsing regressions are still possible.
> 
> **Overview**
> Improves default *excluded referrers* behavior by deriving a usable domain even when cookie-based `topLevelDomain` detection isn’t available: `BrowserConfig` now falls back to `getDomain()` (from `window.location.hostname`) when `topLevelDomain` is unset, and `WebAttribution` builds default excluded referrers from `cookieOptions.domain` **or** `topLevelDomain`.
> 
> Extends `getDomain` to accept an optional hostname (or use the global hostname), returns `''` when hostname is missing, and expands the `KNOWN_2LDS` list to better handle ccTLD / multi-tenant domain parsing. Updates tests to cover the new `getDomain()` behavior and the new `topLevelDomain` defaulting in config.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b9b09fe656559c18fffa15c942c0f330bb78620. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/analytics-browser/src/config.ts
```diff
@@ -135,7 +136,8 @@ export class BrowserConfig extends Config implements IBrowserConfig {
     this.remoteConfig = this.remoteConfig || {};
     this.remoteConfig.fetchRemoteConfig = _fetchRemoteConfig;
     this.fetchRemoteConfig = _fetchRemoteConfig;
-    this.topLevelDomain = topLevelDomain;
+
+    this.topLevelDomain = topLevelDomain || getDomain();
```
Mercy811, line 140: >  It's only true if identityStorage is not equal to cookie and if getTopLevelDomain does not fail.

If the goal of the pr is to fix this. Should we fix it here in useBrowserConfig?  

https://github.com/amplitude/Amplitude-TypeScript/blob/bb3dc622fc1de7675600abe7802b49c10f73a82a/packages/analytics-browser/src/config.ts#L298-L304

I'm thinking about 
```
 const topLevelDomain = await getTopLevelDomain(undefined, diagnosticsClient) || getDomain()
```

daniel-graham-amplitude: I thought about that, but we can't. The reason is because `getDomain` doesn't always return the top level domain accurately (ie: missing 2LD's) and it would cause setting a cookie to fail.

Mercy811: The current implementation doesn't fallback to other identity storage when cookie is unavailable

with fallback to getDomain:
- if cookie is available, getTopLevelDomain returns non empty value
- if cookie is not available, getTopLevelDomain returns empty string and fall back to getDomain, and this is fine because setting cookie is supposed to fail

daniel-graham-amplitude: > because setting cookie is supposed to fail

I don't think we should be letting it fail. If it returns "empty", it will use the `hostOnly` domain (the source of the two cookie problem). If the cookie tracking fails, it means we won't track the page at all

This PR was meant just for default excluded referrers to have a better default.

packages/analytics-browser/src/attribution/web-attribution.ts
```diff
@@ -26,7 +26,7 @@ export class WebAttribution {
     this.options = {
       initialEmptyValue: 'EMPTY',
       resetSessionOnNewCampaign: false,
-      excludeReferrers: getDefaultExcludedReferrers(config.cookieOptions?.domain),
+      excludeReferrers: getDefaultExcludedReferrers(config.cookieOptions?.domain || config.topLevelDomain),
```
Mercy811, line 29: when `getDefaultExcludedReferrers(config.cookieOptions?.domain)` returns `[]`, `||` returns `[]`

packages/analytics-browser/src/config.ts
Mercy811: If it returns "empty", it means cookie is not available, then it doesn't matter what value we set. 

> it will use the hostOnly domain (the source of the two cookie problem). 

setting hostOnly cookie is not intentional but more like a bug that we set it accidentally knowing that cookie is not available. 

> This PR was meant just for default excluded referrers to have a better default

I understand this but I am concerned about having a unclear definition of `config.topLevelDomain` which could cause more confusion and bugs.

packages/analytics-browser/src/attribution/web-attribution.ts
Mercy811: nvm, it thought it's `getDefaultExcludedReferrers(config.cookieOptions?.domain) || config.topLevelDomain`

packages/analytics-browser/src/config.ts
Mercy811: Sync over zoom, this is more like a naming issue. Will rename to solve this

PR #1576, opened 2026-03-06
fix(session-replay): guarantee in-order batch delivery including retries

lewgordon-amplitude:
### Summary

Session replay event batches were not guaranteed to arrive at the server in the order they were recorded.

**Root cause 1 — concurrent sends:** `flush()` used `Promise.all` to send all ready batches simultaneously. Since each batch is an independent `fetch` request, network conditions could cause them to arrive out of order.

**Root cause 2 — retries breaking ordering:** When a batch failed, `handleOtherResponse` re-enqueued it with a backoff timeout and immediately moved on to the next batch. This meant a failed batch could be overtaken by a later batch before its retry completed.

**Fix:**
- Replace `Promise.all` with a sequential `for` loop in `flush()` so batches are sent one at a time
- Move retry logic inline into `handleOtherResponse` — it now awaits the backoff delay and retries the same batch before returning, so the loop naturally blocks until each batch fully completes (including any retries)
- Remove the now-dead deferred-timeout queue logic from `addToQueue` and `flush`

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core session replay delivery/retry behavior by switching from concurrent to sequential sending with in-loop backoff retries, which could affect throughput and timing. Risk is mitigated by updated tests covering ordering and retry paths.
> 
> **Overview**
> Ensures **in-order delivery** of session replay batches by making `SessionReplayTrackDestination.flush()` drain the queue and `await` each `send()` sequentially (replacing concurrent `Promise.all`).
> 
> Retry handling is moved inline: `handleReponse()`/`handleOtherResponse()` are now async and *block* on backoff + retry of the same batch rather than re-queuing with a timeout, and the now-dead timeout-based queue logic is removed.
> 
> Updates/expands tests in `track-destination.test.ts` to assert sequential send order and timer-based retry behavior, and bumps versions/changelogs across related packages (including `session-replay-browser`, `plugin-session-replay-browser`, `segment-session-replay-plugin`, and `unified`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6c6ba01b83c785f0b793528d78b0fbfe9eac572a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/session-replay-browser/src/track-destination.ts
```diff
@@ -89,17 +81,17 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+    const list = this.queue;
+    this.queue = [];
 
     if (this.scheduled) {
       clearTimeout(this.scheduled);
       this.scheduled = null;
     }
 
-    await Promise.all(list.map((context) => this.send(context, useRetry)));
+    for (const context of list) {
```
stevenchien-amplitude, line 92: was curious about how the latency impact would be, but if this fully guarantees order, we should roll with this.

stevenchien-amplitude (approved):

packages/session-replay-browser/src/track-destination.ts
lewgordon-amplitude: Yeah not sure. I was chatting with @ericyang-amplitude and @weibrian and I want to see if this helps with replay quality.

ericyang-amplitude: FWIW this is what the event collection SDK does

that said, my typescript is very rusty, but looks like this synchronously `await` the send until it completes and tries the next. But the behavior when you exceed max # tries you complete the request, does that drop the replay event? (I noticed it completes the request)

I'm not sure what the event collection SDK in this case, but it could also be good to check in with them!

lewgordon-amplitude: Yeah good call out. I think they do drop the events after too many retries just as a safeguard.

[removeEventsExceedFlushMaxRetries](https://github.com/amplitude/Amplitude-TypeScript/blob/aa0703c8b5da571cb32485f870ee3532d9069a02/packages/analytics-core/src/plugins/destination.ts#L134) (L134) — drops events exceeding flushMaxRetries, calls fulfillRequest with 500
[fulfillRequest](https://github.com/amplitude/Amplitude-TypeScript/blob/aa0703c8b5da571cb32485f870ee3532d9069a02/packages/analytics-core/src/plugins/destination.ts#L407) (L407) — fires the failure callback, then calls removeEvents
[removeEvents](https://github.com/amplitude/Amplitude-TypeScript/blob/aa0703c8b5da571cb32485f870ee3532d9069a02/packages/analytics-core/src/plugins/destination.ts#L443) (L443) — removes from this.queue
[saveEvents](https://github.com/amplitude/Amplitude-TypeScript/blob/aa0703c8b5da571cb32485f870ee3532d9069a02/packages/analytics-core/src/plugins/destination.ts#L431) (L431) — persists the updated queue to storage, permanently removing the dropped event

@Mercy811 @daniel-graham-amplitude feel free to correct me though!

PR #1582, opened 2026-03-09
feat(session-replay-browser): add opt-in request and response body capture to network events

lewgordon-amplitude:
### Summary

Extends NetworkRequestEvent with requestBody, responseBody, and responseBodyStatus fields. Body capture is opt-in via loggingConfig.network.body and supports truncation, binary skipping, and error handling.

Tested with playwright tests and locally with tampermonkey. Truncated example:

```json
  {
    "typeName": "Custom",
    "type": 5,
    "timestamp": 1773669315874,
    "data": {
      "tag": "fetch-request",
      "payload": {
        "timestamp": 1773669315841,
        "type": "fetch",
        "method": "GET",
        "url": "https://jsonplaceholder.typicode.com/todos/1",
        "status": 200,
        "duration": 30,
        "responseHeaders": {
          "access-control-allow-credentials": "true",
          "age": "18870",
          "alt-svc": "h3=\":443\"; ma=86400",
          "cache-control": "max-age=43200",
          "cf-cache-status": "HIT",
          "cf-ray": "9dd4412829410688-IAD",
          "content-encoding": "zstd",
          "content-type": "application/json; charset=utf-8",
          "date": "Mon, 16 Mar 2026 13:55:15 GMT",
          "etag": "W/\"53-hfEnumeNh6YirfjyjaujcOPPT+s\"",
          "expires": "-1",
          "nel": "{\"report_to\":\"heroku-nel\",\"response_headers\":[\"Via\"],\"max_age\":3600,\"success_fraction\":0.01,\"failure_fraction\":0.1}",
          "pragma": "no-cache",
          "priority": "u=1,i",
          "report-to": "{\"group\":\"heroku-nel\",\"endpoints\":[{\"url\":\"https://nel.heroku.com/reports?s=ugBqa2N%2FQep2Mpv9K2PhqV7tx8ao5vuJS3n%2FbNJbf4E%3D\\u0026sid=e11707d5-02a7-43ef-b45e-2cf4d2036f7d\\u0026ts=1764098794\"}],\"max_age\":3600}",
          "reporting-endpoints": "heroku-nel=\"https://nel.heroku.com/reports?s=ugBqa2N%2FQep2Mpv9K2PhqV7tx8ao5vuJS3n%2FbNJbf4E%3D&sid=e11707d5-02a7-43ef-b45e-2cf4d2036f7d&ts=1764098794\"",
          "server": "cloudflare",
          "server-timing": "cfExtPri",
          "vary": "Origin, Accept-Encoding",
          "via": "2.0 heroku-router",
          "x-content-type-options": "nosniff",
          "x-powered-by": "Express",
          "x-ratelimit-limit": "1000",
          "x-ratelimit-remaining": "995",
          "x-ratelimit-reset": "1764098818"
        },
        "responseBody": "{\n  \"userId\": 1,\n  \"id\": 1,\n  \"title\": \"delectus aut autem\",\n  \"completed\": false\n}",
```

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds optional capture of HTTP request/response bodies into session replay events, which increases privacy/security risk if misconfigured despite being opt-in and size-limited. Also changes network observer behavior (async response cloning, URL filtering) and adds new e2e/unit test coverage.
> 
> **Overview**
> **Adds opt-in `fetch` request/response body capture to Session Replay network events.** `NetworkRequestEvent` now includes `requestBody`, `responseBody`, and `responseBodyStatus`, controlled via remote config `sr_logging_config.network.body` with a default 10KB max and UTF-8 byte-accurate truncation.
> 
> Response bodies are read from a cloned response asynchronously, mark binary content types as `skipped_binary`, and emit `error` status if body reads fail. `SessionReplay.recordEvents()` now passes the network config into `NetworkObservers` and filters out SDK/track endpoint URLs so the SDK doesn’t record its own upload traffic.
> 
> Docs are updated with configuration/field details, and new Jest + Playwright e2e tests (plus a package-local Playwright config) validate capture, skipping, and truncation behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbe104725c6a7e9b445d4aba54d4fdd60fd13e57. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/e2e/capture.spec.ts
+++ b/packages/session-replay-browser/e2e/capture.spec.ts
@@ -735,6 +735,24 @@ test.describe('network body capture', () => {
     expect(evt!.responseBodyStatus).toBeUndefined();
   });
 
+  test('does not capture fetch-request events for Amplitude track API calls', async ({ page }) => {
+    await mockRemoteConfig(page, remoteConfigWithNetworkBody());
+    const getFetchEvents = await mockTrackApiWithCapture(page);
+
+    await page.goto(buildUrl('/session-replay-browser/sr-capture-test.html', { sessionId: TEST_SESSION_ID }));
+    await waitForReady(page);
+    await waitForNetworkObservers(page);
+
+    // Trigger a flush, which causes the SDK to POST to api-sr.amplitude.com
+    await page.evaluate(() => window.dispatchEvent(new Event('blur')));
+    await page.evaluate(() => (window as any).sessionReplay.flush(false) as Promise<void>);
+    await page.waitForTimeout(500);
+
+    const fetchEvents = getFetchEvents();
+    const trackApiEvent = fetchEvents.find((e) => String(e.url).includes('api-sr.amplitude.com'));
+    expect(trackApiEvent).toBeUndefined();
+  });
+
   test('truncates response body exceeding maxBodySizeBytes', async ({ page }) => {
     await mockRemoteConfig(page, remoteConfigWithNetworkBody({ maxBodySizeBytes: 5 }));
     const getFetchEvents = await mockTrackApiWithCapture(page);


packages/session-replay-browser/src/observers.ts
```diff
@@ -67,7 +144,32 @@ export class NetworkObservers {
...
         requestEvent.responseHeaders = headers;
 
-        this.notifyEvent(requestEvent);
+        if (bodyConfig?.response) {
+          const contentType = headers['content-type'] || null;
+          if (isBinaryContentType(contentType)) {
+            requestEvent.responseBodyStatus = 'skipped_binary';
+            this.notifyEvent(requestEvent);
+          } else {
+            const cloned = response.clone();
```
jxiwang, line 153: This should be fine for smaller payloads, but I think cloning here can get expensive.

lewgordon-amplitude: Good call flagging this. The clone is unfortunately necessary because a Response body is a ReadableStream that can only be consumed once. If we called .text() on the original response, the page's code would get back a response with an already-consumed stream and break. The clone gives us a second independent stream to read from while returning the original to the caller untouched.

The cost is that the browser has to buffer the full body in memory to feed both streams, so Jesse is right that this gets expensive for large payloads. There is no way to avoid this with the standard Fetch API unless we used response.body.getReader() to stream and stop early, but that would require us to also reconstruct a full Response for the page from the unread remainder, which is significantly more complex.

That said, the Math.min(str.length, maxBytes) cap we added limits the post-read work, and the default 10KB limit means we discard most of a large body quickly. Worth adding a read timeout like analytics-core does (it uses a 500ms Promise.race) to guard against stalled reads. I can add that in this PR or as a follow-up.

jxiwang: Makes sense, I think the only way for us to avoid this is patching fetch.

PR #1590, opened 2026-03-11
fix(analytics-core): remote config should not retry client side error except 429

Mercy811:
### Summary

Apply the same logic of Swift remote config client: https://github.com/amplitude/AmplitudeCore-Swift/pull/66

- 401 indicates invalid api key, prevent future fetch calls and no retry
- No retery for client side errors 4xx except 429 rate limit
image

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes remote-config fetch retry/short-circuit behavior, which can affect whether clients refresh configuration and how they behave under error responses. Risk is moderate due to altered network/backoff semantics but is covered by added unit tests for 400/401/429 and invalid-key short-circuiting.
> 
> **Overview**
> Remote config fetching now **stops retrying on client errors**: `fetch()` will not retry on `401` (marks the API key as invalid) or other `4xx` responses except `429` rate limiting, while still retrying on `5xx` and network/timeout failures.
> 
> Once a `401` is observed, `RemoteConfigClient` **short-circuits future `updateConfigs()`/`getOrCreateFetchPromise()` calls** to avoid repeated requests, returning a null config and logging the skip. Tests were expanded to validate the new retry matrix and the “invalid API key” skip behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 802a182047694b646650d9af94bcf3fcaf72b3ce. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/remote-config/remote-config.ts
+++ b/packages/analytics-core/src/remote-config/remote-config.ts
@@ -357,6 +357,9 @@ export class RemoteConfigClient implements IRemoteConfigClient {
           this.logger.debug(`Remote config client fetch with retry time ${retries} failed with ${res.status}: ${body}`);
 
           if (res.status === CODE_STATUS.INVALID_API_KEY) {
+            this.logger.error(
+              `Remote config client fetch failed with ${CODE_STATUS.INVALID_API_KEY}. Invalid API key; future fetches will be skipped.`,
+            );
             this.isLastFetchInvalidApiKey = true;
             shouldRetry = false;
           } else if (res.status >= 400 && res.status < 500 && res.status !== CODE_STATUS.RATE_LIMIT) {
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -31,6 +31,7 @@ import {
   createIdentifyEvent,
   Logger,
   safeJsonStringify,
+  LogLevel,
 } from '@amplitude/analytics-core';
 import {
   getAttributionTrackingConfig,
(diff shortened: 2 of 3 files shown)


packages/analytics-core/src/remote-config/remote-config.ts
```diff
@@ -332,6 +355,13 @@ export class RemoteConfigClient implements IRemoteConfigClient {
         if (!res.ok) {
           const body = await res.text();
           this.logger.debug(`Remote config client fetch with retry time ${retries} failed with ${res.status}: ${body}`);
+
+          if (res.status === CODE_STATUS.INVALID_API_KEY) {
+            this.isLastFetchInvalidApiKey = true;
+            shouldRetry = false;
+          } else if (res.status >= 400 && res.status < 500 && res.status !== CODE_STATUS.RATE_LIMIT) {
```
daniel-graham-amplitude, line 365: The title of this PR says "should not retry on 401 invalid API key" but it looks like this does a lot more than just that. This is making it so that it doesn't retry on any status code except for [Too Many Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429).

Can I just double-check that this is the desired behaviour. And if it can we update the title?

Mercy811: yes, this is the desired behavior https://github.com/amplitude/AmplitudeCore-Swift/pull/66/changes. Let me change the PR title

PR #1598, opened 2026-03-16
fix(analytics-core): filter invalid event properties instead of rejecting whole object

Mercy811:
## Summary
call `json.stringify()` before calling `isValidObject()` so a single undefined property no longer causes the whole object to be rejected.

### DIFF
--- a/packages/analytics-core/test/revenue.test.ts
+++ b/packages/analytics-core/test/revenue.test.ts
@@ -123,4 +123,14 @@ describe('Revenue class', () => {
 
     expect(event.event_properties).toEqual(expectedProperties);
   });
+
+  test('setEventProperties ignores properties with circular references', () => {
+    const property: { [key: string]: any } = { validKey: 'valid value' };
+    property.circular = property;
+    const revenue = new Revenue();
+    revenue.setEventProperties(property);
+    const event = createRevenueEvent(revenue);
+
+    expect(event.event_properties).toEqual(defaultRevenueProperty);
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/analytics-core/src/utils/valid-properties.ts
```diff
@@ -12,6 +12,18 @@ export const isValidObject = (properties: { [key: string]: any }): boolean => {
   return true;
 };
 
+export const filterValidProperties = (properties: { [key: string]: any }): { [key: string]: any } => {
+  const result: { [key: string]: any } = {};
+  for (const key in properties) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const value: unknown = properties[key];
+    if (isValidProperties(key, value)) {
```
daniel-graham-amplitude: This would mean that if a nested property was invalid then we would drop the whole thing

```
const properties = {
  someProp: {
    validVal: 'val',
    invalidVal: undefined,
  },
}
```
^ `someProp` would be dropped. Are we okay with that?

Mercy811: Good catch. Updated to recursively filter

packages/analytics-core/test/revenue.test.ts
```diff
@@ -112,4 +112,15 @@ describe('Revenue class', () => {
 
     expect(event.event_properties).toEqual(expectedProperties);
   });
+
+  test('setEventProperties filters out undefined properties instead of rejecting the whole object', () => {
```
daniel-graham-amplitude, line 116: Could you also add a test to see what happens when the event has a circular reference.

Mercy811: done!

PR #1626, opened 2026-03-24
fix(session-replay-browser): read window.scrollY on page leave to capture final scroll position

lewgordon-amplitude:
### Summary

rrweb throttles scroll callbacks to 100ms using setTimeout for the trailing edge. When a user scrolls and navigates away before the pending setTimeout fires, the final scroll position is never delivered to ScrollWatcher. This causes scroll map heatmaps to underreport scroll depth by 30-67% for users who scroll and quickly leave.

Fix: In `ScrollWatcher.send()` (called on pagehide), read `window.scrollX/Y` directly and call `update()` one final time before sending. This captures the true final scroll position regardless of rrweb's throttle state.

SR-3039

### Root Cause

rrweb's `initScrollObserver()` wraps the scroll callback in `throttle(cb, sampling.scroll || 100)`. The throttle uses `setTimeout(remaining)` for trailing edge events. When the page unloads before the pending setTimeout fires, the callback is cancelled and the final scroll position is lost. `ScrollWatcher.send()` fires on `pagehide` but only has a stale `_maxScrollY`.

### Testing

Injected the built SDK on a customer's news site via Playwright (headless Chromium) with a mocked remote config endpoint (`capture_enabled: true`). Captured the scroll interaction beacon payload on page navigation to compare reported `maxScrollY` vs actual `window.scrollY`.

#### Before fix (with rrweb throttle, no window.scrollY fallback)

| Scenario | Actual | Reported | Lost |
|---|---|---|---|
| 5000px + 500ms pause | 5000 | 5000 | 0% |
| Quick flick 1500px + immediate leave | 1500 | 500 | **67%** |
| Scroll 1000px + immediate leave | 1000 | 700 | **30%** |
| Scroll 500px (past fold) + immediate leave | 500 | 350 | **30%** |

#### After fix

Tested across **20 pages** on the customer's site (various section pages) x 5 scroll scenarios each = **100 test runs**.

**98/100 passed with 0% data loss.** The 2 "failures" were over-reports (negative miss) on very short pages due to ScrollWatcher state from a prior test leaking across runs in the same browser tab — a test harness artifact, not a code bug.

Representative results (all pages showed the same pattern):

| Scenario | Actual | Reported | Lost |
|---|---|---|---|
| Fast scroll 5000px + 500ms pause | 5000 | 5000 | 0% |
| Fast scroll 5000px + IMMEDIATE leave | 5000 | 5000 | **0%** |
| Quick flick 1500px + IMMEDIATE leave | 1500 | 1500 | **0%** |
| Scroll 1000px + IMMEDIATE leave | 1000 | 1000 | **0%** |
| Scroll 500px (past fold) + IMMEDIATE leave | 500 | 500 | **0%** |

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Adjusts `ScrollWatcher.send()` to derive final scroll metrics from `window.scrollX/Y` during `pagehide`, which can change reported scroll-depth data. Risk is limited to analytics accuracy/edge cases (e.g., zero/undefined scroll values) and is covered by added unit tests.
> 
> **Overview**
> Improves scroll-depth reporting on page exit by having `ScrollWatcher.send()` read `globalScope.scrollX/Y` and call `update()` one last time before emitting the beacon, avoiding missed trailing scroll updates caused by rrweb throttling.
> 
> Updates `scroll` hook tests to mock `scrollX/Y` and adds cases covering stale/missing rrweb scroll callbacks and ensuring maxima don’t regress when the user scrolls back up before leaving.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7021d012c0415dc6ca4573a2666b423620e0a1d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/src/hooks/scroll.ts
+++ b/packages/session-replay-browser/src/hooks/scroll.ts
@@ -123,6 +123,8 @@ export class ScrollWatcher {
       const scrollX = globalScope.scrollX ?? 0;
       const scrollY = globalScope.scrollY ?? 0;
       if (scrollX > 0 || scrollY > 0) {
+        // id is required by the scrollPosition type but is not used by update() —
+        // only x and y are read. 1 is the conventional rrweb mirror ID for the document node.
         this.update({ id: 1, x: scrollX, y: scrollY });
       }
       this.transport.send(deviceId, {


lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/hooks/scroll.ts
```diff
@@ -117,6 +117,14 @@ export class ScrollWatcher {
     const deviceId = deviceIdFn();
     const globalScope = getGlobalScope();
     if (globalScope && deviceId) {
+      // Capture the true final scroll position directly from the window.
+      // rrweb's scroll observer throttles callbacks to 100ms using setTimeout,
+      // so the most recent scroll position may not have been delivered to the hook yet.
+      const scrollX = globalScope.scrollX ?? 0;
+      const scrollY = globalScope.scrollY ?? 0;
+      if (scrollX > 0 || scrollY > 0) {
+        this.update({ id: 1, x: scrollX, y: scrollY });
```
jxiwang, line 128: How come this is `id:1`?

lewgordon-amplitude: Good catch — `id` is required by the `scrollPosition` type but `update()` doesn't actually use it (only `x` and `y` are read). `1` is the conventional rrweb mirror ID for the document node, but it's effectively a dummy value here. I'll add a code comment to make that clear.

PR #1650, opened 2026-04-03
fix(session-replay-browser): wire amp-unmask as default unmaskSelector (SR-2945)

lewgordon-amplitude:
### Summary

Fixes SR-2945. The \`UNMASK_TEXT_CLASS = 'amp-unmask'\` constant existed alongside \`MASK_TEXT_CLASS = 'amp-mask'\` and \`BLOCK_CLASS = 'amp-block'\`, implying all three work by just applying the CSS class. However, \`amp-mask\` and \`amp-block\` are passed to rrweb as \`maskTextClass\` and \`blockClass\` natively, while \`amp-unmask\` was never wired up — it only worked if the user explicitly set \`privacyConfig.unmaskSelector: ['.amp-unmask']\`.

**Fix:** In \`SessionReplayLocalConfig\`, \`.amp-unmask\` is now auto-included as the first entry in \`unmaskSelector\`, merged with any user-provided selectors. This makes all three classes work symmetrically.

### Checklist

* [x] Does your PR title have the correct title format?
* Does your PR have a breaking change?: No — this is additive. Existing \`unmaskSelector\` configs are preserved and extended.

---

> [!NOTE]
> **Medium Risk**
> Changes default privacy masking behavior by ensuring `.amp-unmask` always unmasks content, which could unintentionally expose text if customers already rely on conservative masking without expecting class-based overrides. Scope is contained to config composition and covered by added unit/E2E tests.
> 
> **Overview**
> Fixes SR-2945 by **auto-including `.amp-unmask`** as the first entry in `privacyConfig.unmaskSelector` within `SessionReplayLocalConfig`, with deduping and preservation of any user-provided selectors.
> 
> Updates config-join and session replay tests to expect the injected default (including selector merging/deduping and invalid-selector cleanup behavior), and adds E2E coverage plus a test fixture element in `sr-privacy-test.html` to verify `.amp-unmask` unmasks text under conservative masking from both remote and local privacy config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6784e028358f97eb1e4f59e6c1aa8bacecd15601. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/e2e/privacy.spec.ts
+++ b/packages/session-replay-browser/e2e/privacy.spec.ts
@@ -177,6 +177,66 @@ test.describe('privacy — privacyConfig option', () => {
   });
 });
 
+// ─── amp-unmask CSS class (SR-2945) ──────────────────────────────────────────
+
+test.describe('privacy — amp-unmask CSS class', () => {
+  test('amp-unmask class unmasks text under conservative masking from remote config without explicit unmaskSelector', async ({
+    page,
+  }) => {
+    // Before SR-2945, .amp-unmask had no effect unless the user explicitly added
+    // '.amp-unmask' to privacyConfig.unmaskSelector. Now it's wired as a default.
+    await mockRemoteConfig(page, remoteConfigWithPrivacy({ defaultMaskLevel: 'conservative' }));
+    const getBodies = await captureTrackRequests(page);
+
+    // No local privacyConfig — masking comes entirely from remote config
+    await page.goto(buildUrl(PRIVACY_PAGE, { sessionId: TEST_SESSION_ID }));
+    await waitForReady(page);
+    await page.waitForTimeout(SNAPSHOT_SETTLE_MS);
+    await flushRecording(page);
+
+    const root = getSnapshotRoot(getBodies());
+    expect(root).not.toBeNull();
+
+    // #amp-unmask-text has the .amp-unmask class — should NOT be masked
+    const unmaskEl = findById(root!, 'amp-unmask-text');
+    expect(unmaskEl).toBeDefined();
+    expect(getTextContent(unmaskEl!)).toContain('Unmask class visible text');
+    expect(isMaskedText(getTextContent(unmaskEl!))).toBe(false);
+
+    // #plain-text has no special class — should be masked under conservative
+    const plainEl = findById(root!, 'plain-text');
+    expect(plainEl).toBeDefined();
+    expect(isMaskedText(getTextContent(plainEl!))).toBe(true);
+  });
+
+  test('amp-unmask class unmasks text under conservative masking from local privacyConfig without explicit unmaskSelector', async ({
+    page,
+  }) => {
+    const privacyConfig = JSON.stringify({ defaultMaskLevel: 'conservative' });
+    await mockRemoteConfig(page, remoteConfigRecording);
+    const getBodies = await captureTrackRequests(page);
+
+    await page.goto(buildUrl(PRIVACY_PAGE, { sessionId: TEST_SESSION_ID, privacyConfig }));
+    await waitForReady(page);
+    await page.waitForTimeout(SNAPSHOT_SETTLE_MS);
+    await flushRecording(page);
+
+    const root = getSnapshotRoot(getBodies());
+    expect(root).not.toBeNull();
+
+    // #amp-unmask-text has the .amp-unmask class — should NOT be masked
+    const unmaskEl = findById(root!, 'amp-unmask-text');
+    expect(unmaskEl).toBeDefined();
+    expect(getTextContent(unmaskEl!)).toContain('Unmask class visible text');
+    expect(isMaskedText(getTextContent(unmaskEl!))).toBe(false);
+
+    // #plain-text has no special class — should be masked under conservative
+    const plainEl = findById(root!, 'plain-text');
+    expect(plainEl).toBeDefined();
+    expect(isMaskedText(getTextContent(plainEl!))).toBe(true);
+  });
+});
+
 // ─── Remote privacy config ────────────────────────────────────────────────────
 
 test.describe('privacy — remote sr_privacy_config', () => {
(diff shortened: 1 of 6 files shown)


packages/session-replay-browser/src/config/local-config.ts
```diff
@@ -72,9 +73,12 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
     this.urlChangePollingInterval = options.urlChangePollingInterval ?? DEFAULT_URL_CHANGE_POLLING_INTERVAL;
     this.captureDocumentTitle = options.captureDocumentTitle ?? false;
 
-    if (options.privacyConfig) {
-      this.privacyConfig = options.privacyConfig;
-    }
+    // Auto-include .amp-unmask as a default unmaskSelector entry so it works
+    // symmetrically with amp-mask/amp-block without requiring explicit config (SR-2945).
+    this.privacyConfig = {
```
jpollock-ampl, line 81: Quick question - this flips `this.privacyConfig` from sometimes-undefined to always-defined. Are we confident no downstream code branches on `if (config.privacyConfig)` and treats undefined as a meaningful signal (e.g. "user didn't configure privacy at all")?

The new joined-config test (`should fall back to empty object when config.privacyConfig is undefined`) suggests there's an `?? {}` guard downstream that's now effectively unreachable through the normal constructor path - might be worth a quick grep to confirm nothing else reads `privacyConfig` defensively.

packages/session-replay-browser/src/config/local-config.ts
```diff
@@ -72,9 +73,12 @@ export class SessionReplayLocalConfig extends Config implements ISessionReplayLo
...
+      ...(options.privacyConfig ?? {}),
+      unmaskSelector: [`.${UNMASK_TEXT_CLASS}`, ...(options.privacyConfig?.unmaskSelector ?? [])],
```
jpollock-ampl: Nit/non-blocking - If a user explicitly passes `unmaskSelector: ['.amp-unmask', '.foo']` (the previously documented opt-in), they'll end up with `['.amp-unmask', '.amp-unmask', '.foo']`. Likely harmless for rrweb's matcher, but a `Array.from(new Set([...]))` would keep things tidy. WDYT?

packages/session-replay-browser/test/config/joined-config.test.ts
```diff
@@ -275,14 +275,36 @@ describe('SessionReplayJoinedConfigGenerator', () => {
...
+        );
+        expect(config).toEqual({
+          ...mockLocalConfig,
+          captureEnabled: true,
+          optOut: mockLocalConfig.optOut,
+          privacyConfig: {
+            defaultMaskLevel: 'medium',
+            blockSelector: undefined,
+            maskSelector: undefined,
+            unmaskSelector: ['.amp-unmask', '.localClassName', '.remoteClassName'],
```
jpollock-ampl, line 306: Nit/non-blocking - Worth adding a case where the user explicitly passes `'.amp-unmask'` in `unmaskSelector` to lock in the dedupe behavior (or the lack of it), depending on which way we land on the comment above.

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/e2e/privacy.spec.ts
```diff
@@ -177,6 +177,66 @@ test.describe('privacy — privacyConfig option', () => {
...
+// ─── amp-unmask CSS class (SR-2945) ──────────────────────────────────────────
+
+test.describe('privacy — amp-unmask CSS class', () => {
+  test('amp-unmask class unmasks text under conservative masking from remote config without explicit unmaskSelector', async ({
+    page,
+  }) => {
+    // Before SR-2945, .amp-unmask had no effect unless the user explicitly added
+    // '.amp-unmask' to privacyConfig.unmaskSelector. Now it's wired as a default.
+    await mockRemoteConfig(page, remoteConfigWithPrivacy({ defaultMaskLevel: 'conservative' }));
+    const getBodies = await captureTrackRequests(page);
```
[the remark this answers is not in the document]
lewgordon-amplitude: This is already fixed in commit 2844cc0f (pushed just before bugbot ran this review). Bugbot reviewed against c197742e which was the state before the fix. The current HEAD of SR-2945-wire-amp-unmask-default already has `const { getBodies } = await captureTrackRequests(page)` at both locations (lines 189 and 217).

lewgordon-amplitude:
bugbot run

jpollock-ampl (approved):

packages/session-replay-browser/src/config/local-config.ts
lewgordon-amplitude: Good call — I grepped before this PR landed. The reads are:

- `session-replay.ts:337,649,657-680` — all use optional chaining (`?.`) and treat undefined / partially-populated as the same "no override" case. No defensive `if (privacyConfig)` early-out branches.
- `session-replay.ts:888` — destructures and passes to `maskFn`/`maskAttributeFn`, which accept `config?: PrivacyConfig` and behave identically with `undefined` vs an effectively-empty object.
- `joined-config.ts:167` — the `?? {}` you spotted is the only one. Through the LocalConfig path it's now unreachable, but I left it as defensive coverage in case `SessionReplayJoinedConfigGenerator` is ever instantiated with a non-LocalConfig input. The `should fall back to empty object when config.privacyConfig is undefined` test instantiates the generator directly with `privacyConfig: undefined` to keep that branch covered (added in ca8380e1 after I tried and rejected the cleaner non-null-assertion approach due to lint rules).

The one observable downstream change: `getEffectiveMaskLevel` now runs on what used to be the "no privacyConfig" path. With no `defaultMaskLevel` and no `urlMaskLevels`, it returns `DEFAULT_MASK_LEVEL = 'medium'`, so the `=== 'conservative'` check in `getMaskTextSelectors` still doesn't fire. No behavior change.

packages/session-replay-browser/src/config/local-config.ts
lewgordon-amplitude: Good catch — done in 6784e028. Wrapped the merge with `Array.from(new Set([...]))` so a user explicitly passing `'.amp-unmask'` no longer ends up with a duplicate entry. The `joined-config` selectorMap was already key-deduping the join, but it was leaking through at the LocalConfig layer.

packages/session-replay-browser/test/config/joined-config.test.ts
lewgordon-amplitude: Added in 6784e028 — two tests:

- `joined-config.test.ts` LocalConfig describe: `should dedupe .amp-unmask when user explicitly includes it in unmaskSelector` (locks in `['.amp-unmask', '.foo']` → `['.amp-unmask', '.foo']` at the constructor layer).
- Same file, joined-config describe: `should dedupe .amp-unmask when user explicitly passes it alongside the auto-injected default` (locks in the join-level dedupe through `privacyConfigSelectorMap`).

PR #1676, opened 2026-04-14
ci: allow publish-v2 to skip lerna version

Mercy811:
## Summary
- add a `publishOnly` workflow input to `publish-v2.yml`
- skip the release and prerelease `lerna version` steps when `publishOnly` is enabled
- preserve the existing publish flow so partial publish failures can be retried without re-versioning

## Testing
- Not run (not requested)

---

> [!NOTE]
> **Low Risk**
> Low risk workflow-only change that conditionally skips `lerna version`/tag creation; main risk is misconfiguration leading to publishing without bumping versions or creating releases.
> 
> **Overview**
> Adds a new manual-dispatch input `publishOnly` to the `publish-v2.yml` workflow to support retrying a partial publish.
> 
> When `publishOnly` is enabled, the workflow skips the release and prerelease versioning steps (the `pnpm deploy:version`/`lerna version` phase) while keeping the existing NPM publish steps unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a36adaf5ba30790040247bf250e8538dbd5f844b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-v2.yml
+++ b/.github/workflows/publish-v2.yml
@@ -16,6 +16,11 @@ on:
         type: string
         description: Branch to create pre-release from (only applies to prerelease/dry-run).
         required: false
+      publishOnly:
+        type: boolean
+        description: Skip `lerna version` and only run the publish step. Use this to recover from a partial publish where some packages were already released.
+        required: false
+        default: false
 
 jobs:
   authorize:
@@ -107,6 +112,7 @@ jobs:
       # Only create release version when using from-git (default behavior)
       # from-package mode uses existing package.json versions and doesn't need git tags
       - name: Create release version
+        if: ${{ github.event.inputs.publishOnly != 'true' }}
         run: |
           echo "Running lerna version..."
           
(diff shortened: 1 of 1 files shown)


Mercy811:
bugbot run

daniel-graham-amplitude:
Can you change the title from `Fix(ci)` to `chore`... `fix` is for bugfixes and cause new patches to be released.

.github/workflows/publish-v2.yml
```diff
@@ -16,6 +16,11 @@ on:
         type: string
         description: Branch to create pre-release from (only applies to prerelease/dry-run).
         required: false
+      publishOnly:
```
daniel-graham-amplitude: Can we call it something like `npmPublishOnly` to disambiguate this. `publish` could also mean publishing git tags.

.github/workflows/publish-v2.yml
```diff
@@ -16,6 +16,11 @@ on:
...
+        type: boolean
+        description: Skip `lerna version` and only run the publish step. Use this to recover from a partial publish where some packages were already released.
+        required: false
+        default: false
```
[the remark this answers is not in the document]
Mercy811: > If the previous run already published some package versions, a retry in publishOnly mode may still fail when re-publishing those already-released versions.

This is not correct. pnpm will skip those already-released versions. What it does is that it compare the current version with the one in remote npm registry and decide whether to push to remote.

.github/workflows/publish-v2.yml
Mercy811: I'm renaming it as `skipLernaVersion`

daniel-graham-amplitude (approved):

PR #1685, opened 2026-04-17
fix(session-replay-browser): reduce snapshot size and handle 413s

lewgordon-amplitude:
### Summary

**Reduce snapshot payload size**
- Enable all rrweb `slimDOMOptions` by default: scripts, comments, favicons, head whitespace, and all head meta tags (description/keywords, social, robots, http-equiv, authorship, verification) are now stripped from snapshots. None of these affect replay fidelity.
- Remove the `omitElementTags` config option (was only wiring up `script` and `comment`; now all slim-DOM stripping is unconditional)

**WAF 413 bisect-retry**
- On a WAF 413 (response body contains `"Payload exceeds"`), split the event batch in half and retry each half independently; recurse until batches reach a single event, then stop and warn
- App-layer 413s (no WAF signature) are not retried, preventing infinite loops

**Capture-time drop for oversized single events**
- Add `MAX_SINGLE_EVENT_SIZE` constant (9 MB)
- Guard in `EventCompressor.addCompressedEventToManager` drops any serialized event exceeding the limit before it's queued — avoids a wasted network round trip for snapshots that would fail the server's ~10 MB limit anyway

**Tests**
- Unit tests for all new behaviors
- New e2e spec (`size-limits.spec.ts`): oversized event not delivered, SDK recovers after drop, WAF 413 triggers bisect-retry, app-layer 413 doesn't loop
- Move e2e test server to port 5174 to isolate worktree from main-branch dev server

### Checklist

* Does your PR have a breaking change?:  Yes — `omitElementTags` removed from public config (was undocumented/unused in practice)

---

> [!NOTE]
> **Medium Risk**
> Changes core replay delivery behavior (per-event dropping and 413 retry/bisect logic) across both main-thread and web-worker send paths, which could affect data loss vs. retry behavior if misclassified. Also removes the `omitElementTags` option, which is a small breaking config change.
> 
> **Overview**
> **Reduces session replay payload size and adds stricter size limits.** rrweb `slimDOMOptions` are now enabled broadly by default (scripts/comments plus various `<head>` artifacts), and the public `omitElementTags` option is removed from both plugin and standalone configs.
> 
> **Adds per-event size guarding and safer 413 handling.** A new `MAX_SINGLE_EVENT_SIZE` (9MB, UTF-8 bytes) drops oversized serialized events at capture time and again before send (to protect against older/IDB-stored events). 413 responses are now classified as *WAF vs app-layer* via `WAF_PAYLOAD_TOO_LARGE_PATTERN`: WAF 413 triggers split-and-retry of the batch, while non-WAF 413 is dropped without retry to avoid loops, in both main-thread and worker delivery.
> 
> **Expands test coverage.** Adds unit/e2e coverage for oversized-event dropping, WAF bisect-retry, and ensuring the SDK remains functional after 413s (including new `size-limits.spec.ts`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6639b360d2d03a6b2e1aaa514c367ebfb4b7e68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -353,11 +373,38 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
       case Status.Failed:
         await this.handleOtherResponse(context);
         break;
+      case Status.PayloadTooLarge:
+        this.handlePayloadTooLargeResponse(context, responseBody.includes('Payload exceeds'));
```
jpollock-ampl: How brittle is the WAF detection here? If the upstream message wording changes (vendor update, rule tweak, etc.), `responseBody.includes('Payload exceeds')` silently flips to `false`, every multi-event WAF 413 falls through to the single-event "cannot split further" branch, and bisect-retry effectively turns off without us noticing.

A few options worth considering:
- Match a broader pattern (e.g. case-insensitive `/payload.*exceed/i`) to be resilient to small wording changes.
- Treat any 413 with a non-empty body as WAF-ish for bisect purposes (since app-layer 413 in this stack today is the empty / plain-text case). Trades some precision for resilience.
- Add a comment with the exact WAF rule ID or source so future readers know where the literal came from.

Not a blocker, just want to make sure we're aware of the silent-failure mode.

packages/session-replay-browser/src/constants.ts
```diff
@@ -20,6 +20,7 @@ export const SESSION_REPLAY_EU_URL = 'https://api-sr.eu.amplitude.com/sessions/v
 export const SESSION_REPLAY_STAGING_URL = 'https://api-sr.stag2.amplitude.com/sessions/v2/track';
 export const STORAGE_PREFIX = `${AMPLITUDE_PREFIX}_replay_unsent`;
 export const MAX_EVENT_LIST_SIZE = 1 * 1000000; // ~1 MB limit for JSON serialized events payload
+export const MAX_SINGLE_EVENT_SIZE = 9 * 1000000; // 9 MB, just under server's 10 MB per-event threshold
```
jpollock-ampl: Quick question - the guard in `addCompressedEventToManager` compares `compressedEvent.length` (a JS string char count, UTF-16 code units) against this constant, but the comment frames it as a byte limit ("9 MB ... server's 10 MB threshold"). For ASCII-heavy content these are roughly equivalent, but a snapshot containing a lot of multi-byte UTF-8 (CJK text, emoji, etc.) could serialize to ~2-3x the byte count of its char count, so a 9M-char event could be 18-27 MB on the wire and still slip past this guard.

If the intent is a true byte limit, `new Blob([compressedEvent]).size` (or a `TextEncoder` length) would be more accurate. If the char-count approximation is intentional, might be worth tweaking the comment so the next reader doesn't tighten the threshold under a wrong assumption.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -353,11 +373,38 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+
+    this.loggerProvider.warn(
+      `Session replay event batch rejected by ${source} (${context.events.length} events, ${totalSizeKB} KB total) — splitting and retrying`,
+    );
+
+    // Clean up the original IDB record, then re-enqueue both halves as new in-memory batches
+    void context.onComplete();
+    const noop = (): Promise<void> => Promise.resolve();
+    const mid = Math.floor(context.events.length / 2);
+    this.sendEventsList({ ...context, events: context.events.slice(0, mid), onComplete: noop });
```
jpollock-ampl, line 423: Help me understand - when we re-enqueue both halves we spread `context`, which carries `attempts` along. If a batch was already on attempt N (close to `flushMaxRetries`) when the WAF 413 came back, both halves inherit that count and may be one retry away from being dropped on the next 5xx / network blip. Should we be resetting `attempts: 0` on the halves since they're functionally new requests, or is the carry-over intentional (e.g. to bound total work for a pathologically large batch)?

packages/session-replay-browser/e2e/size-limits.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  mockRemoteConfig,
+  buildUrl,
+  waitForReady,
+  readRouteBody,
+} from './helpers';
+
+// 9.1 M chars: just over MAX_SINGLE_EVENT_SIZE = 9 * 1_000_000.
+// Passed as a number to page.evaluate so only the integer crosses CDP;
+// the browser generates the large string entirely in-process.
+const OVERSIZED_ATTR_LENGTH = 9_100_000;
```
jpollock-ampl: Nit/non-blocking - Could we import `MAX_SINGLE_EVENT_SIZE` here and derive this as `MAX_SINGLE_EVENT_SIZE + 100_000`? If someone bumps the constant in `constants.ts`, the test will silently start exercising the wrong threshold otherwise. Not a blocker.

packages/session-replay-browser/e2e/size-limits.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+    await waitForReady(page);
+    await page.waitForTimeout(SNAPSHOT_SETTLE_MS);
+
+    await page.evaluate(() => window.dispatchEvent(new Event('blur')));
+    await page.evaluate(() => (window as any).sessionReplay.flush(false) as Promise<void>);
+    // Wait long enough for an infinite loop to be detectable
+    await page.waitForTimeout(SNAPSHOT_SETTLE_MS * 10);
+
+    // Bisect bottoms out at single events; total calls must be bounded (O(n) not infinite)
+    expect(callCount).toBeLessThan(20);
```
jpollock-ampl: Nit/non-blocking - `< 20` feels generous for the "no infinite loop" assertion - I think the bisect should bottom out at 1 call for the app-layer 413 case (no WAF signature, goes straight to `completeRequest` with err, no retries). A tighter bound (e.g. `<= 1` or `<= 2`) would catch a regression where someone accidentally enables bisect on app-layer 413s. Up to you.

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -369,6 +392,30 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+    this.sendEventsList({ ...context, events: context.events.slice(mid), onComplete: noop });
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in cf60ed8c. Added an early-return guard in `handlePayloadTooLargeResponse` so non-WAF (app-layer) 413s immediately call `completeRequest` with an error and return, without bisecting. Only WAF 413s (where the rejection is due to compressed payload size, which bisection can help) continue to the bisect-retry path. Updated unit tests to use `isWaf=true` for bisection cases and added a new test covering the non-WAF immediate-drop behavior. Also updated the integration test expecting the new error message.

packages/session-replay-browser/src/worker/track-destination.ts
```diff
@@ -79,6 +94,10 @@ async function sendWithRetry(id: string, payloadJson: string, context: SendConte
       postMessage({ type: 'complete', id });
       return;
     }
+    if (result.payloadTooLarge) {
+      postMessage({ type: 'payload_too_large', id, isWaf: result.isWaf === true });
+      return;
```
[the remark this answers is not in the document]
lewgordon-amplitude: This was addressed in earlier commits on this branch. The worker's `sendWithRetry` now gates the `payload_too_large` message behind `useRetry` at line 97 (`if (result.payloadTooLarge && useRetry)`), so when `useRetry=false` the worker posts a `warn`+`complete` without triggering bisect-retry on the main thread. The main-thread path also short-circuits at the `!useRetry` branch before entering `handleReponse`, so behavior is now consistent between both paths.

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/e2e/capture.spec.ts
```diff
@@ -1295,6 +1295,82 @@ test.describe('web worker mode', () => {
...
+    await mockRemoteConfig(page, remoteConfigRecording);
+
+    let callCount = 0;
+    const deliveredBodies: string[] = [];
+
+    await page.route('https://api-sr.amplitude.com/**', async (route: Route) => {
+      callCount++;
+      if (callCount === 1) {
+        // Simulate the server rejecting the initial payload as too large
+        await route.fulfill({ status: 413, contentType: 'application/json', body: '{"code":413}' });
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in 6a3431ee. Updated both E2E test route mocks from `body: '{"code":413}'` to `body: '{"error":"Payload exceeds the maximum allowed size of 10MB"}'`. The WAF detection checks `responseBody.includes('Payload exceeds')`, so these tests now use a body that triggers `isWaf=true`, which correctly exercises the bisect-retry path. The old non-WAF body would have hit the new immediate-drop path (which is the correct behavior for non-WAF 413s, but not what these tests intend to test).

packages/session-replay-browser/src/track-destination.ts
lewgordon-amplitude: Good call. Switched to a centralised `WAF_PAYLOAD_TOO_LARGE_PATTERN = /payload.*exceed/i` constant and used it from both the main-thread (`track-destination.ts`) and worker (`worker/track-destination.ts`) paths — case-insensitive and resilient to small wording changes.

packages/session-replay-browser/src/constants.ts
lewgordon-amplitude: Good catch — fixed. The capture-time guard in `EventCompressor.addCompressedEventToManager` and the IDB backstop in `events-manager.ts` now compare UTF-8 byte size (`new Blob([str]).size`) against `MAX_SINGLE_EVENT_SIZE`, so a 9 M-char CJK/emoji payload that would serialise to 18-27 MB on the wire is correctly dropped at the boundary instead of slipping through to the network. Updated the constant comment to make the byte vs char distinction explicit.

packages/session-replay-browser/src/track-destination.ts
lewgordon-amplitude: Carry-over is not an issue here — `sendEventsList` calls `addToQueue` which unconditionally sets `attempts: 0, timeout: 0` on the queued context (track-destination.ts:118-124), and `addToQueue` then increments to 1 before the first send. So each bisected half starts with a fresh transient-error retry budget, which I think matches your intuition that they are functionally new requests. The original `attempts` value never propagates.

packages/session-replay-browser/e2e/size-limits.spec.ts
lewgordon-amplitude: Done — `OVERSIZED_ATTR_LENGTH` is now derived as `MAX_SINGLE_EVENT_SIZE + 100_000`, and the in-test threshold assertion (`JSON.stringify(event).length < 9_000_000`) was also reworked to use `new Blob([...]).size <= MAX_SINGLE_EVENT_SIZE` so both stay in sync if the constant is bumped.

packages/session-replay-browser/e2e/size-limits.spec.ts
lewgordon-amplitude: Tightened to `<= 2`. The non-WAF 413 path now short-circuits via `completeRequest` so the dropped batch produces exactly one tracking call; I left a single-call slack to absorb an unrelated follow-up flush during the wait without making the test flaky, and added a comment explaining why the bound is what it is so a future regression that re-enables bisect on app-layer 413s gets caught.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -119,6 +120,30 @@ export const createEventsManager = async <Type extends EventType>({
...
     sequenceId?: number;
   }) => {
+    // Backstop for events that entered IDB before the per-event size guard in
+    // addCompressedEventToManager (e.g. stored by a previous SDK version or via
+    // storeCurrentSequence/sendStoredEvents which bypass the capture-time check).
+    // Compare UTF-8 byte size, not JS char count, to match the server-side limit.
+    const sizedEvents = rawEvents.map((e) => ({ event: e, bytes: new Blob([e]).size }));
+    const oversized = sizedEvents.filter((s) => s.bytes > MAX_SINGLE_EVENT_SIZE);
+    if (oversized.length > 0) {
+      config.loggerProvider.warn(
```
jpollock-ampl: Nit/non-blocking - Worth surfacing oversized-event drops as a telemetry signal in addition to the warn log? Same applies to the capture-time guard in `EventCompressor.addCompressedEventToManager`. Right now the only way to know we're dropping data in production is to scrape logs, and a drop-rate metric would help us tune `MAX_SINGLE_EVENT_SIZE` and prioritize root-cause work upstream in rrweb. Not a blocker.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -119,6 +120,30 @@ export const createEventsManager = async <Type extends EventType>({
...
+          .join(
+            ', ',
+          )}. If this recurs, please open a GitHub issue at https://github.com/amplitude/Amplitude-TypeScript/issues or contact Amplitude support.`,
+      );
+    }
+    const events =
+      oversized.length > 0
+        ? sizedEvents.filter((s) => s.bytes <= MAX_SINGLE_EVENT_SIZE).map((s) => s.event)
+        : rawEvents;
+    if (events.length === 0) {
```
jpollock-ampl, line 142: Nit/non-blocking - Quick question - was the unawaited `void store.cleanUpSessionEventsStore(...)` here intentional? If it rejects we silently swallow the error (no `.catch`), whereas other store calls in this file (e.g. `addEvent`) attach a `.catch` handler. Probably fine in practice, but worth a `.catch(...)` for symmetry.

jpollock-ampl (approved):

packages/session-replay-browser/src/events/events-manager.ts
lewgordon-amplitude: Good call — switched to a .catch that warns via the logger, matching the pattern used elsewhere in this file (addEvent, getStorageSize). Done in c6639b36.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -372,32 +392,36 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
 
-  // Split depth is naturally bounded: each 413 halves the batch until events.length <= 1,
-  // at which point the single oversized event is dropped. A batch of N events produces
-  // at most N sequential retry requests before all are resolved.
-  private handlePayloadTooLargeResponse(context: SessionReplayDestinationContext) {
-    if (context.events.length <= 1) {
-      // Cannot split further — drop the single oversized event.
+  handlePayloadTooLargeResponse(context: SessionReplayDestinationContext, isWaf: boolean): void {
+    const source = isWaf ? 'WAF (compressed payload too large)' : 'server (event too large)';
+    const totalSizeKB = Math.round(context.events.reduce((sum, e) => sum + e.length, 0) / KB_SIZE);
```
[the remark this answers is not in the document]
lewgordon-amplitude: Acknowledged — leaving as-is. This is a debug log message only (no behavioral impact); for ASCII content char count ≈ byte count, and the under-report on multi-byte content is bounded to ~2-3x in a string a developer reads while triaging WAF rejections. Not worth the churn relative to the byte-size guards in the data path.

PR #1696, opened 2026-04-22
fix: duplicate sr instances 

polbins:
### Summary

`Timeline.register()` had a TOCTOU race: the name-dedup check ran synchronously but `this.plugins.push(plugin)` happened only after `await plugin.setup?.()`. Two concurrent `client.add(plugin)` calls with the same plugin name both passed the check before either pushed, so both ran `setup()` and both were registered. For Session Replay RN this produced two native recorders fighting for the Android snapshot budget and double-uploading every event.

### Changes

- **`analytics-core`**: reserve plugin names in a `Set<string>` before `setup()` runs so concurrent callers are skipped immediately. A `_registrationGeneration` counter (incremented on `reset()`) prevents a pre-reset in-flight `catch` from deleting a post-reset name reservation. The plugins array is captured before the async gap so a post-reset resolve pushes into the orphaned array and not the fresh one.
- **`plugin-session-replay-react-native` (Android)**: add an `isActive` guard so a second `setup()` logs a warning and returns early. Reset on `teardown()` / `invalidate()` so legitimate re-init paths keep working.
- **`plugin-session-replay-react-native` (iOS)**: same `isActive` guard pattern — `setup()` skips if already active, `teardown()` guards against calling `stop()` on a nil `sessionReplay` (force-unwrap crash if called before `setup()`).

### How to verify

- `pnpm test` in `packages/analytics-core` — 4 new tests: concurrent race, retry-after-failure, re-entrancy, reset-during-setup.
- Manual (RN example): add `SessionReplayPlugin` twice back-to-back or under `React.StrictMode`. Logcat should show one `setup:` block + one skip warning, not two full setups.

### Checklist

* [x] PR title follows [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)
* Breaking change: **No**

### DIFF
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -19,6 +19,9 @@ export class Timeline {
   // re-entrant register() with the same name is skipped; freed on setup() failure and
   // on deregister() so the name can be reused.
   _reservedPluginNames: Set<string> = new Set();
+  // Incremented on reset() so in-flight catch blocks from before the reset don't
+  // delete name reservations made by a post-reset registration.
+  _registrationGeneration = 0;
   // loggerProvider is set by the client at _init()
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
@@ -43,6 +46,7 @@ export class Timeline {
     }
 
     this._reservedPluginNames.add(name);
+    const generation = this._registrationGeneration;
 
     // Capture the plugins array so a post-reset resolve pushes into the orphaned (old)
     // array instead of the freshly reset one.
@@ -52,7 +56,9 @@ export class Timeline {
       await plugin.setup?.(config, this.client);
       targetPlugins.push(plugin);
     } catch (error) {
-      this._reservedPluginNames.delete(name);
+      if (this._registrationGeneration === generation) {
+        this._reservedPluginNames.delete(name);
+      }
       throw error;
     }
   }
@@ -72,6 +78,7 @@ export class Timeline {
   reset(client: CoreClient) {
     this._clearOptOutListeners();
     this._reservedPluginNames.clear();
+    this._registrationGeneration++;
     this.applying = false;
     const plugins = this.plugins;
     plugins.map((plugin) => plugin.teardown?.());
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -61,7 +61,9 @@ export class Timeline {
     try {
       await pending;
     } finally {
-      this._pendingRegistrations.delete(name);
+      if (this._pendingRegistrations.get(name) === pending) {
+        this._pendingRegistrations.delete(name);
+      }
     }
   }
 
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -15,6 +15,9 @@ export class Timeline {
   // Flag indicates whether timeline is ready to process event
   // Events collected before timeline is ready will stay in the queue to be processed later
   plugins: Plugin[] = [];
+  // Tracks registrations whose setup() is in flight so concurrent register() calls with
+  // the same name dedupe and inherit the winner's outcome.
+  _pendingRegistrations: Map<string, Promise<void>> = new Map();
   // loggerProvider is set by the client at _init()
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
@@ -24,22 +27,42 @@ export class Timeline {
   constructor(private client: CoreClient) {}
 
   async register(plugin: Plugin, config: IConfig) {
-    if (this.plugins.some((existingPlugin) => existingPlugin.name === plugin.name)) {
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-      this.loggerProvider.warn(`Plugin with name ${plugin.name} already exists, skipping registration`);
-      return;
-    }
-
     if (plugin.name === undefined) {
       plugin.name = UUID();
-      this.loggerProvider.warn(`Plugin name is undefined. 
-      Generating a random UUID for plugin name: ${plugin.name}. 
+      this.loggerProvider.warn(`Plugin name is undefined.
+      Generating a random UUID for plugin name: ${plugin.name}.
       Set a name for the plugin to prevent it from being added multiple times.`);
     }
 
+    const name = plugin.name;
+
+    const inflight = this._pendingRegistrations.get(name);
+    if (inflight) {
+      this.loggerProvider.warn(`Plugin with name ${name} already exists, skipping registration`);
+      await inflight;
+      return;
+    }
+    if (this.plugins.some((existingPlugin) => existingPlugin.name === name)) {
+      this.loggerProvider.warn(`Plugin with name ${name} already exists, skipping registration`);
+      return;
+    }
+
     plugin.type = plugin.type ?? 'enrichment';
-    await plugin.setup?.(config, this.client);
-    this.plugins.push(plugin);
+    // Capture the current plugins array so a post-reset resolve pushes into the orphaned
+    // (old) array instead of the freshly reset one. Defer to the next microtask so
+    // _pendingRegistrations is populated before user code runs — otherwise a plugin whose
+    // setup() re-enters register() recurses.
+    const targetPlugins = this.plugins;
+    const pending = Promise.resolve().then(async () => {
+      await plugin.setup?.(config, this.client);
+      targetPlugins.push(plugin);
+    });
+    this._pendingRegistrations.set(name, pending);
+    try {
+      await pending;
+    } finally {
+      this._pendingRegistrations.delete(name);
+    }
   }
 
   async deregister(pluginName: string, config: IConfig) {
@@ -55,6 +78,7 @@ export class Timeline {
 
   reset(client: CoreClient) {
     this._clearOptOutListeners();
+    this._pendingRegistrations.clear();
     this.applying = false;
     const plugins = this.plugins;
     plugins.map((plugin) => plugin.teardown?.());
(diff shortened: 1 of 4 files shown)


packages/plugin-session-replay-react-native/android/src/main/java/com/amplitude/pluginsessionreplayreactnative/PluginSessionReplayReactNativeModule.kt
```diff
@@ -105,12 +119,16 @@ class PluginSessionReplayReactNativeModule(private val reactContext: ReactApplic
 
   @ReactMethod
   fun teardown() {
-    sessionReplay.shutdown()
+    if (isActive) {
+      sessionReplay.shutdown()
+      isActive = false
+    }
```
aliaksandr-kazarez, line 125: nit: can you try to address that for iOS as well, or create a linear ticket to do that.

aliaksandr-kazarez (approved):
LGTM;

packages/analytics-core/src/timeline.ts
```diff
@@ -24,22 +27,44 @@ export class Timeline {
...
-    this.plugins.push(plugin);
+    // Capture the current plugins array so a post-reset resolve pushes into the orphaned
+    // (old) array instead of the freshly reset one. Defer to the next microtask so
+    // _pendingRegistrations is populated before user code runs — otherwise a plugin whose
+    // setup() re-enters register() recurses.
+    const targetPlugins = this.plugins;
+    const pending = Promise.resolve().then(async () => {
+      await plugin.setup?.(config, this.client);
+      targetPlugins.push(plugin);
+    });
```
daniel-graham-amplitude: This works okay, another way to do this is:
```
const pending = (async () => {
  await plugin.setup?.(config, this.client);
  targetPlugins.push(plugin);
})();
```

Mercy811: They are different 
image

packages/analytics-core/src/timeline.ts
```diff
@@ -24,22 +27,44 @@ export class Timeline {
...
     plugin.type = plugin.type ?? 'enrichment';
-    await plugin.setup?.(config, this.client);
-    this.plugins.push(plugin);
+    // Capture the current plugins array so a post-reset resolve pushes into the orphaned
+    // (old) array instead of the freshly reset one. Defer to the next microtask so
+    // _pendingRegistrations is populated before user code runs — otherwise a plugin whose
+    // setup() re-enters register() recurses.
+    const targetPlugins = this.plugins;
+    const pending = Promise.resolve().then(async () => {
+      await plugin.setup?.(config, this.client);
```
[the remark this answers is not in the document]
Mercy811: @copilot everything before line 60 is sync and we check _pendingRegistrations to early return at line 39. Could you elaborate on what you meant? An example would help

daniel-graham-amplitude (approved):
The JS parts, lgtm

packages/analytics-core/src/timeline.ts
aliaksandr-kazarez: this one is tricky 🙂 check out following snippet

```
console.log('start');
const pending = (async () => {
  await null;
  console.log('IIFE');
})();
console.log('end')
```

Mercy811: this is because await also schedules a microtask.

packages/analytics-core/src/timeline.ts
```diff
@@ -24,22 +27,44 @@ export class Timeline {
...
+      targetPlugins.push(plugin);
```
Mercy811, line 57: @copilot I found another possible bug: plugin is pushed to the temporary targetPlugins but not this.plugins is used in many other places like apply()

crleona:
Could we just add the plugin before calling setup so that the second call detects that the same plugin is already there and fails to add it?

packages/analytics-core/src/timeline.ts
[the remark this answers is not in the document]
Mercy811: @copilot reset() is only called by init(). So this will only happen if init() are called twice. The result is that the plugin is not in timeline.plugins and thus will not be used. I'm ok to leave it here as the main issue is fixed.

Mercy811 (approved):
Thanks Paul, analytics-core changes look good to me!

crleona:
There are easier ways to accomplish this. I am not happy with this approach.

polbins:
> There are easier ways to accomplish this. I am not happy with this approach.

Yeah, I agree this still feels a little more complex than ideal. I’m open to a cleaner approach here.

My main hesitation with adding the plugin before `setup()` is that `this.plugins` is also the active event pipeline, so events could hit a plugin before it’s initialized. The current shape is mostly trying to keep the name reserved while setup is in flight, without exposing the plugin until setup finishes.

crleona:
> > There are easier ways to accomplish this. I am not happy with this approach.
> 
> Yeah, I agree this still feels a little more complex than ideal. I’m open to a cleaner approach here.
> 
> My main hesitation with adding the plugin before `setup()` is that `this.plugins` is also the active event pipeline, so events could hit a plugin before it’s initialized. The current shape is mostly trying to keep the name reserved while setup is in flight, without exposing the plugin until setup finishes.

Couple possibilities:
- Keep a separate data structure for name dedup that we could check before setup/adding to the main plugins list
- Keep track of whether the plugin has initialized before sending it events, so you can add it to the main plugins list before initializing.

polbins:
> > > There are easier ways to accomplish this. I am not happy with this approach.
> > 
> > 
> > Yeah, I agree this still feels a little more complex than ideal. I’m open to a cleaner approach here.
> > My main hesitation with adding the plugin before `setup()` is that `this.plugins` is also the active event pipeline, so events could hit a plugin before it’s initialized. The current shape is mostly trying to keep the name reserved while setup is in flight, without exposing the plugin until setup finishes.
> 
> Couple possibilities:
> 
> * Keep a separate data structure for name dedup that we could check before setup/adding to the main plugins list
> * Keep track of whether the plugin has initialized before sending it events, so you can add it to the main plugins list before initializing.

Tried the `Set<string>` simplification, it's simpler and still prevents double `setup()`. But there's a tradeoff: duplicate callers .promise resolves immediately instead of awaiting the winner. We don't know of the exact init pattern in the original ticket yet unfortunately, so I kept the Map to preserve that guarantee. Happy to go `Set` if you think the risk is acceptable.

crleona:
> > > > There are easier ways to accomplish this. I am not happy with this approach.
> > > 
> > > 
> > > Yeah, I agree this still feels a little more complex than ideal. I’m open to a cleaner approach here.
> > > My main hesitation with adding the plugin before `setup()` is that `this.plugins` is also the active event pipeline, so events could hit a plugin before it’s initialized. The current shape is mostly trying to keep the name reserved while setup is in flight, without exposing the plugin until setup finishes.
> > 
> > 
> > Couple possibilities:
> > 
> > * Keep a separate data structure for name dedup that we could check before setup/adding to the main plugins list
> > * Keep track of whether the plugin has initialized before sending it events, so you can add it to the main plugins list before initializing.
> 
> Tried the `Set<string>` simplification, it's simpler and still prevents double `setup()`. But there's a tradeoff: duplicate callers .promise resolves immediately instead of awaiting the winner. We don't know of the exact init pattern in the original ticket yet unfortunately, so I kept the Map to preserve that guarantee. Happy to go `Set` if you think the risk is acceptable.

Why do we need that behavior?

polbins:
> > > > > There are easier ways to accomplish this. I am not happy with this approach.
> > > > 
> > > > 
> > > > Yeah, I agree this still feels a little more complex than ideal. I’m open to a cleaner approach here.
> > > > My main hesitation with adding the plugin before `setup()` is that `this.plugins` is also the active event pipeline, so events could hit a plugin before it’s initialized. The current shape is mostly trying to keep the name reserved while setup is in flight, without exposing the plugin until setup finishes.
> > > 
> > > 
> > > Couple possibilities:
> > > 
> > > * Keep a separate data structure for name dedup that we could check before setup/adding to the main plugins list
> > > * Keep track of whether the plugin has initialized before sending it events, so you can add it to the main plugins list before initializing.
> > 
> > 
> > Tried the `Set<string>` simplification, it's simpler and still prevents double `setup()`. But there's a tradeoff: duplicate callers .promise resolves immediately instead of awaiting the winner. We don't know of the exact init pattern in the original ticket yet unfortunately, so I kept the Map to preserve that guarantee. Happy to go `Set` if you think the risk is acceptable.
> 
> Why do we need that behavior?

I'm not really sure if it happens in the wild, even our docs don't use this `.promise` pattern. Let me just continue with the Set<string> implementation then

crleona (approved):

Mercy811 (changes requested):
Hi @polbins, could you hold the PR? I don't think this is a solid fix by adding `_registrationGeneration`. Here is an example edge case: 
1. register('foo1') and then register('foo2') 
2. both of them await and reset() is called
3. refresh register('foo1') register('foo2') called again and succeed
4. both original register() calls failed but the failing order is diffeerent than the calling order foo2 failed first and then foo1, and the number won't match now

I'm working on a refactor

Mercy811:
> Could we just add the plugin before calling setup so that the second call detects that the same plugin is already there and fails to add it?

Thanks @crleona for offering this idea. I created another pr to solve this issue https://github.com/amplitude/Amplitude-TypeScript/pull/1702. 

It's much cleaner and avoid the race between an in-flight register() and a reset() by having `_reservedPluginNames` alone: 

1. `register("foo")` adds `foo` to `_reservedPluginNames` and awaits `plugin.setup()` 
2. While that await is suspended, `reset()` runs, clearing `_reservedPluginNames` 
3. A fresh `register("foo")` now succeeds and re-reserves the name.
4. The original `setup()` finally rejects. Without the generation check, its catch block wiping the new registration's reservation and letting yet another duplicate register("foo") slip through.

PR #1702, opened 2026-04-28
fix(analytics-core): close TOCTOU race in Timeline.register()

Mercy811:
### Summary

This PR tries to solve the bug originally brought up by https://github.com/amplitude/Amplitude-TypeScript/pull/1696 in a cleaner way.

`Timeline.register()` had a check-then-act race: the name-dedup check (`this.plugins.some(...)`) ran synchronously, but `this.plugins.push(plugin)` happened only after `await plugin.setup?.()`. Concurrent `client.add(plugin)` calls with the same plugin name both passed the check before either pushed.

Time-serialized walkthrough — two concurrent `client.add(plugin)` calls (A and B, same plugin name):

- **t₀** — A enters `register()`, runs `this.plugins.some(...)` synchronously → no match, falls through
- **t₀** — B enters `register()`, runs `this.plugins.some(...)` synchronously → still no match (A hasn't pushed yet), falls through
- **t₁** — A hits `await plugin.setup?.()` and yields
- **t₁** — B hits `await plugin.setup?.()` and yields
- **t₂** — A's setup resolves, A runs `this.plugins.push(plugin)`
- **t₂** — B's setup resolves, B runs `this.plugins.push(plugin)`
- **Result** — `setup()` ran twice, `plugins[]` holds two entries with the same name, every event flows through both copies

In Session Replay React Native this surfaced as **two native recorders fighting for the Android snapshot budget and double-uploading every event** — the symptom that prompted this fix.

### Approach

Track plugin lifecycle state in a parallel `Map<string, 'installing' | 'installed'>` keyed by plugin name. Reserve the name synchronously by writing `'installing'` into the map **before** the `await`, so concurrent calls see the entry and bail out with the existing warning. After setup, defensively check the status hasn't been cleared by `reset()` before pushing.

If `setup()` throws, the entry stays as `'installing'` — re-registration of the same name is blocked. Same name implies same setup logic, so retry would just fail again; `reset()` is the escape hatch for clearing all state.

`Plugin` is a public SDK interface used by external plugin authors, so the status is kept off the plugin object — internal-only `Map` on `Timeline`. `plugins[]` continues to hold only fully-installed plugins, so `apply()`, `flush()`, `onIdentityChanged()`, etc. need no changes — they implicitly only see installed plugins.

### Changes

- `packages/analytics-core/src/timeline.ts`
  - Add `PluginStatus` type (`'installing' | 'installed'`) and `pluginStatus: Map<string, PluginStatus>` field
  - Rewrite `register()` to assign UUID-before-dedup, reserve the slot synchronously via `pluginStatus.set(name, 'installing')`, and defensively check status before pushing
  - `deregister()` deletes the status entry; `reset()` clears the map
- `packages/analytics-core/test/timeline.test.ts`
  - 5 new tests:
    - **Concurrent same-name dedup** — the regression test for this bug (`Promise.all([register(p), register(p)])` with deferred `setup`)
    - **Failed setup re-throws and blocks re-registration** of the same name
    - **`reset()` during in-flight setup** drops the registration
    - **Concurrent unnamed registers** both land with unique UUIDs
    - **`apply()` skips installing plugin** while setup is in flight

### Test plan

- [x] `pnpm --filter @amplitude/analytics-core test` — 787/787 pass, 100% coverage
- [x] `pnpm --filter @amplitude/analytics-core lint` — 0 errors (warnings unchanged)
- [x] `pnpm --filter @amplitude/analytics-core build` — typecheck clean for both ES5 and ESM
- [x] Confirmed regression test fails on `main` and passes with the fix
- [ ] Reviewer: sanity-check Session Replay React Native manual run to confirm only one recorder spins up under repeated `client.add()` calls

### Checklist

### DIFF
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -55,6 +55,9 @@ export class Timeline {
   }
 
   async deregister(pluginName: string, config: IConfig) {
+    // Clear the status first so a name stuck in 'installing' (mid-install, or setup() threw)
+    // can be unblocked via deregister(). Map.delete is a no-op if the key is missing.
+    this.pluginStatus.delete(pluginName);
     const index = this.plugins.findIndex((plugin) => plugin.name === pluginName);
     if (index === -1) {
       config.loggerProvider.warn(`Plugin with name ${pluginName} does not exist, skipping deregistration`);
@@ -62,7 +65,6 @@ export class Timeline {
     }
     const plugin = this.plugins[index];
     this.plugins.splice(index, 1);
-    this.pluginStatus.delete(pluginName);
     await plugin.teardown?.();
   }
 
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -8,7 +8,7 @@ import { Result } from './types/result';
 import { buildResult } from './utils/result-builder';
 import { UUID } from './utils/uuid';
 
-type PluginStatus = 'installing' | 'installed';
+type PluginStatus = 'reserved' | 'installed';
 
 export class Timeline {
   queue: [Event, EventCallback][] = [];
@@ -44,12 +44,12 @@ export class Timeline {
 
     plugin.type = plugin.type ?? 'enrichment';
     // Reserve the name synchronously to close the TOCTOU window across `await setup`.
-    // If setup throws, the entry stays as 'installing' and blocks future re-registration —
+    // If setup throws, the entry stays as 'reserved' and blocks future re-registration —
     // a same-named plugin would just fail again, so retry isn't useful.
-    this.pluginStatus.set(name, 'installing');
+    this.pluginStatus.set(name, 'reserved');
     await plugin.setup?.(config, this.client);
     // reset() may have cleared the status map while setup was awaiting.
-    if (this.pluginStatus.get(name) !== 'installing') {
+    if (this.pluginStatus.get(name) !== 'reserved') {
       return;
     }
     this.plugins.push(plugin);
@@ -57,7 +57,7 @@ export class Timeline {
   }
 
   async deregister(pluginName: string, config: IConfig) {
-    // Clear the status first so a name stuck in 'installing' (mid-install, or setup() threw)
+    // Clear the status first so a name stuck in 'reserved' (mid-install, or setup() threw)
     // can be unblocked via deregister(). Map.delete is a no-op if the key is missing.
     this.pluginStatus.delete(pluginName);
     const index = this.plugins.findIndex((plugin) => plugin.name === pluginName);
--- a/packages/analytics-core/src/timeline.ts
+++ b/packages/analytics-core/src/timeline.ts
@@ -8,7 +8,7 @@ import { Result } from './types/result';
 import { buildResult } from './utils/result-builder';
 import { UUID } from './utils/uuid';
 
-type PluginStatus = 'installing' | 'installed' | 'failed';
+type PluginStatus = 'installing' | 'installed';
 
 export class Timeline {
   queue: [Event, EventCallback][] = [];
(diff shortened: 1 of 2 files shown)


packages/analytics-core/src/timeline.ts
```diff
@@ -24,22 +27,35 @@ export class Timeline {
...
+      Generating a random UUID for plugin name: ${plugin.name}.
       Set a name for the plugin to prevent it from being added multiple times.`);
     }
+    const name = plugin.name;
+
+    const existing = this.pluginStatus.get(name);
+    if (existing === 'installing' || existing === 'installed') {
+      this.loggerProvider.warn(`Plugin with name ${name} already exists, skipping registration`);
+      return;
+    }
```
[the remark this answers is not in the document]
Mercy811: this is expected to early return plugins with the same name.

crleona (approved):

polbins (approved):
LGTM! I'll close the other one

packages/analytics-core/src/timeline.ts
```diff
@@ -24,25 +29,37 @@ export class Timeline {
...
+    if (this.pluginStatus.has(name)) {
+      this.loggerProvider.warn(`Plugin with name ${name} already exists, skipping registration`);
+      return;
+    }
 
     plugin.type = plugin.type ?? 'enrichment';
+    // Reserve the name synchronously to close the TOCTOU window across `await setup`.
+    // If setup throws, the entry stays as 'installing' and blocks future re-registration —
+    // a same-named plugin would just fail again, so retry isn't useful.
+    this.pluginStatus.set(name, 'installing');
```
daniel-graham-amplitude: Having it be stuck forever in the `installing` state feels strange to me, and might introduce some debugging/troubleshooting problems.

Would another option be to add a third state, something like `install-errored`.

```javascript
try {
  this.pluginStatus.set(name, 'installing');
  await plugin.setup?.(config, this.client);
  // reset() may have cleared the status map while setup was awaiting.
  if (this.pluginStatus.get(name) !== 'installing') {
    return;
  }
  this.plugins.push(plugin);
  this.pluginStatus.set(name, 'installed');
} catch (e) {
  // log error here
  this.pluginStatus.set(name, 'install-errored');
}
```

Mercy811: I started with the error status but removed it later because it introduce another race of register() and reset() in https://github.com/amplitude/Amplitude-TypeScript/pull/1696#issuecomment-4354828660 which results in the second plugin not being installed. 

```
t0: A.register('foo')  — pluginStatus = { foo → 'installing' }
                         await A's setup (will throw)
t1: reset()            — pluginStatus.clear()
                         A's setup still running in background
t2: B.register('foo')  — pluginStatus.has('foo') = false (cleared)
                         pluginStatus = { foo → 'installing' }
                         await B's setup (will succeed)
t3: A's setup THROWS   — catch fires: pluginStatus.set('foo', 'install-errored')
                         ← clobbers B's 'installing'!
t4: B's setup resolves — pluginStatus.get('foo') = 'install-errored' ≠ 'installing'
                         → return early, no push
```

daniel-graham-amplitude: Okay makes sense. Could we rename it to something other than `installing` then? It just seems strange that it could be stuck indefinitely with the name `installing`.

Mercy811: How about `reserved`?

daniel-graham-amplitude: That would work. Other options: `blocked`, `locked`.

Mercy811: I will go with `locked`!

daniel-graham-amplitude (approved):

packages/analytics-core/src/timeline.ts
daniel-graham-amplitude: 👍

PR #1705, opened 2026-04-29
fix(session-replay): multi-tab IDB correctness, hang protection, memory fallback

lewgordon-amplitude:
## Problem

Session replay was silently dropping events in production when IDB was used. Four bugs:

1. **Fallback never triggered** — `tx.done.catch` handlers logged errors but never called `recordFailure()`, so the consecutive-failure threshold was never reached and the memory store fallback never fired. Events routed to a broken IDB store were silently dropped.
2. **Cross-tab event loss** — `addEventToCurrentSequence` treated a foreign-tab's `sessionCurrentSequence` record as absent and overwrote it with only the new event, dropping all of the other tab's in-progress events. Two tabs sharing a session would ping-pong overwriting each other.
3. **`storeCurrentSequence` not atomic** — three separate IDB requests allowed concurrent writes to interleave, causing event loss or duplicate uploads.
4. **IDB hangs undetected** — `openDB` can block indefinitely (blocked connection, DB closing state, storage pressure); mid-recording transactions can also stall. No error fires, `recordFailure()` never called, fallback never triggers.

## Fixes

- **`recordFailure()` in `tx.done.catch`** — fallback now actually triggers after threshold failures
- **Foreign-tab event promotion** — when `addEventToCurrentSequence` encounters a record owned by another tab, it promotes that tab's events to `sequencesToSend` before claiming the slot. No events dropped.
- **Atomic `storeCurrentSequence`** — wraps both store operations in a single readwrite transaction, preventing the race
- **`consecutiveFailureThreshold` default → 1** — falls back to memory on first IDB error
- **`withTimeout` + `armTxDoneTimeout`** — `openDB` wrapped with 2s timeout; `getSequencesToSend`, `storeCurrentSequence`, `addEventToCurrentSequence` each have a 5s transaction watchdog. Any hang triggers `recordFailure()` and the memory fallback.

## Tests

- **Jest: 786/786 passing, 100% statement/branch/line/function coverage** on `events-idb-store.ts`
- **Playwright (real Chromium): 12/12 passing**, including:
  - Events from both tabs survive in shared IDB (cross-tab promote)
  - `storeCurrentSequence` atomicity under concurrent `addEvent`
  - SDK falls back to memory immediately on first IDB error
  - `openDB` hang → SDK initializes via memory fallback within 707ms
  - No unhandled promise rejections during normal recording or failure modes

## Test plan
- [ ] Verify Jest suite passes: `pnpm --filter @amplitude/session-replay-browser test`
- [ ] Verify Playwright suite passes in real browser
- [ ] Confirm production IDB-disabled experiment shows same capture behavior as this fix

---

> [!NOTE]
> **High Risk**
> Touches core session-replay persistence/flush logic (multi-tab ownership, transactional writes, and new timeouts), so regressions could impact event capture and delivery despite added coverage.
> 
> **Overview**
> Fixes multi-tab and failure-mode correctness in the session replay IndexedDB event store by **introducing per-tab ownership (`tabId`) for `sessionCurrentSequence`**, promoting foreign-tab in-progress events to `sequencesToSend` instead of overwriting, and making `storeCurrentSequence` an **atomic multi-store transaction**.
> 
> Adds **hang/failure protection**: `openDB` is wrapped with a 2s timeout, each readwrite operation gets a 5s `tx.done` watchdog, `tx.done.catch` now increments the failure counter, and the default `consecutiveFailureThreshold` is lowered to **1** to trigger immediate memory-store fallback.
> 
> Expands regression coverage with new Jest suites for multi-tab contamination and timeout scenarios, adds Playwright E2E tests for multi-tab interleaving + fallback behavior, updates `jest-setup.js` to polyfill `crypto.randomUUID`, and adjusts an opt-out test to assert events exist exactly once across both IDB stores.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 377950b7eb6aa5aa9c1b268bd1162475af2aee16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/e2e/idb.spec.ts
+++ b/packages/session-replay-browser/e2e/idb.spec.ts
@@ -405,3 +405,463 @@ test.describe('IDB transaction abort handling', () => {
     expect(idbDebug).toHaveLength(0);
   });
 });
+
+// ─── Multi-tab and fallback tests ─────────────────────────────────────────────
+
+/**
+ * Reads the unique event-string contents of every record across both IDB stores
+ * (sequencesToSend.events + sessionCurrentSequence.events), regardless of tabId.
+ * Used to verify that no events were dropped across a multi-tab interleaving.
+ */
+function getAllPersistedEvents(page: import('@playwright/test').Page) {
+  return page.evaluate(
+    (dbName: string): Promise<string[]> =>
+      new Promise((resolve, reject) => {
+        const req = indexedDB.open(dbName);
+        req.onerror = () => reject(req.error);
+        req.onsuccess = () => {
+          const db = req.result;
+          const tx = db.transaction(['sequencesToSend', 'sessionCurrentSequence'], 'readonly');
+          const all: string[] = [];
+          const onCursor = (req: IDBRequest<IDBCursorWithValue | null>) => {
+            req.onsuccess = () => {
+              const cursor = req.result;
+              if (cursor) {
+                const value = cursor.value as { events: string[] };
+                for (const ev of value.events ?? []) all.push(ev);
+                cursor.continue();
+              }
+            };
+          };
+          onCursor(tx.objectStore('sequencesToSend').openCursor());
+          onCursor(tx.objectStore('sessionCurrentSequence').openCursor());
+          tx.oncomplete = () => resolve(all);
+          tx.onerror = () => reject(tx.error);
+        };
+      }),
+    IDB_NAME,
+  );
+}
+
+/**
+ * Tags each rrweb event with a unique marker before it is sent to the SDK by
+ * monkey-patching `sessionReplay.addEvent` is too invasive — instead we inject
+ * synthetic events via the public API.  A simpler approach: emit user events
+ * (clicks) which the SDK captures with timestamps — we count unique events.
+ */
+
+test.describe('IDB multi-tab and fallback behaviour', () => {
+  /**
+   * Multi-tab invariant: when two tabs in the same browser context record events
+   * for the same sessionId, ALL events from both tabs eventually appear in the
+   * shared IndexedDB across the two stores.  Pre-fix, Tab A's
+   * addEventToCurrentSequence would silently overwrite Tab B's in-progress events.
+   * Post-fix, Tab A promotes Tab B's events to sequencesToSend before claiming
+   * the slot, so nothing is dropped.
+   *
+   * We can't easily count rrweb events one-by-one because rrweb captures large
+   * sets of mutation events asynchronously.  Instead we count the total number
+   * of stored events across both tabs and verify it matches the sum of what each
+   * tab generated independently — within a small tolerance for non-deterministic
+   * snapshot timing.  More importantly we tag tab-B events with a deterministic
+   * marker (a blur trigger after a known click) and verify those events SURVIVE
+   * the cross-tab claim.
+   */
+  test('multi-tab: events from both tabs survive in shared IDB (cross-tab promote)', async ({ browser }) => {
+    // Single context = shared origin, shared IndexedDB.  Each `page` in the
+    // context is a real browser tab from the SDK's perspective (separate
+    // sessionStorage → distinct tabId).
+    const context = await browser.newContext();
+    const tabA = await context.newPage();
+    const tabB = await context.newPage();
+
+    try {
+      const allErrorsA: Error[] = [];
+      const allErrorsB: Error[] = [];
+      tabA.on('pageerror', (e) => allErrorsA.push(e));
+      tabB.on('pageerror', (e) => allErrorsB.push(e));
+
+      // Mock remote config and SR API on each tab.
+      for (const p of [tabA, tabB]) {
+        await mockRemoteConfig(p, remoteConfigRecording);
+        await p.route('https://api-sr.amplitude.com/**', (route) =>
+          route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(SR_API_SUCCESS) }),
+        );
+      }
+
+      // Load Tab A, then Tab B with the same sessionId.  They will share the
+      // same IDB database (same origin) and pick up unique tabIds via sessionStorage.
+      const url = buildUrl('/session-replay-browser/sr-capture-test.html', {
+        sessionId: TEST_SESSION_ID,
+        logLevel: LOG_LEVEL_DEBUG,
+      });
+
+      await tabA.goto(url);
+      await waitForReady(tabA);
+
+      await tabB.goto(url);
+      await waitForReady(tabB);
+
+      // Confirm distinct tabIds (sessionStorage is per-tab).
+      const tabIdA = await tabA.evaluate(() => sessionStorage.getItem('_amp_sr_tab_id'));
+      const tabIdB = await tabB.evaluate(() => sessionStorage.getItem('_amp_sr_tab_id'));
+      expect(tabIdA).toBeTruthy();
+      expect(tabIdB).toBeTruthy();
+      expect(tabIdA).not.toBe(tabIdB);
+
+      // Generate events on each tab. Interleave clicks so the two tabs ping-pong
+      // writes to the shared sessionCurrentSequence slot.
+      for (let i = 0; i < 3; i++) {
+        await tabA.click('#test-button');
+        await tabB.click('#test-button');
+        await tabA.click('#test-input');
+        await tabB.click('#test-input');
+      }
+
+      // Give rrweb time to flush captured mutations into IDB.
+      await tabA.waitForTimeout(300);
+      await tabB.waitForTimeout(300);
+
+      // Read the union of all persisted events from both stores.  Use either
+      // tab — both see the same database.
+      const allEvents = await getAllPersistedEvents(tabA);
+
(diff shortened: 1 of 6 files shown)


lewgordon-amplitude:
🚀 **Prerelease published** from RC branch \`sr-idb-multitab-test-rc\`

\`\`\`
npm install @amplitude/session-replay-browser@sr-idb-multitab-test-rc
\`\`\`

Pinned version: \`1.39.1-sr-idb-multitab-test-rc.0\`

[Workflow run](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/25335062378)

lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -4,18 +4,104 @@ import { EventType, Events, SendingSequencesReturn } from '../typings/session-re
...
+ * individual put/get.  If the driver accepts a put but the underlying
+ * transaction silently fails to commit (the production stall scenario),
+ * tx.done still hasn't settled, and recordFailure must NOT have been
+ * suppressed.  The fix for that scenario in production is the tx.done.catch
+ * handler attached separately by callers, which catches the eventual abort.
+ *
+ * The soft-cancel pattern only suppresses the timeout for the "all puts
+ * resolved successfully" case — a case where tx.done is overwhelmingly
+ * likely to settle imminently in production.  If it doesn't, callers can
+ * still detect the failure on the NEXT operation (which will fail to open
```
jpollock-ampl, line 81: Could you help me understand the soft-cancel? The doc identifies that if all individual put/get promises resolve but the underlying transaction silently fails to commit (no abort, no error, `tx.done` never settles), the success-path `cancelTimeout()` suppresses the watchdog and we never call `recordFailure`. 

The fallback assumption is "the next op will fail too," but in steady-state recording `addEvent` runs continuously and `recordSuccess()` actively resets the consecutive-failure counter on each apparently-successful op, so we'd keep marking puts as success while data never reaches disk.

Is there a reason not to await `tx.done` before `recordSuccess()` / `cancelTimeout()`? 

My thinking is that would close the gap at the cost of latency on the hot path. 

If latency is the concern, an alternative is to leave the watchdog armed on the success path too - genuine fast commits settle `tx.done` well within 5s and `tx.done.then` clears the timer, so the only behavior change is "silent stall after apparent success" eventually triggering fallback.

packages/session-replay-browser/e2e/idb.spec.ts
```diff
@@ -405,3 +405,456 @@ test.describe('IDB transaction abort handling', () => {
...
+        };
+        return req as IDBOpenDBRequest;
+      };
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      const origOpen = indexedDB.open.bind(indexedDB);
+      // Only stall the SDK's IDB open; let any internal Playwright/page IDB
+      // calls go through.  The SDK's DB name is derived from the API key
+      // (first 10 chars).  Match anything starting with our test key prefix.
+      indexedDB.open = function (name: string, version?: number): IDBOpenDBRequest {
+        if (name.startsWith('d90c5cf09c')) {
```
jpollock-ampl: Pretty sure this hardcodes the first 10 chars of `TEST_API_KEY`. If the test key ever changes, this `startsWith` won't match, the SDK's openDB will go through to real IDB, the hang scenario won't be exercised, and the test will pass for the wrong reason (no hang -> fast init -> `elapsed < 3000` trivially true).

Could we derive the prefix from the same constant the SDK uses?
```ts
const expectedPrefix = TEST_API_KEY.substring(0, 10);
// ...
if (name.startsWith(expectedPrefix)) {
```

packages/session-replay-browser/e2e/idb.spec.ts
```diff
@@ -405,3 +405,456 @@ test.describe('IDB transaction abort handling', () => {
...
+      const allEvents = await getAllPersistedEvents(tabA);
+
+      // Pre-fix this would lose most cross-tab events.  We can't count exact
+      // numbers because rrweb is asynchronous, but the total must be reasonably
+      // large (tens of events from two tabs of clicks + snapshots).  We assert a
+      // floor that is comfortably above what a single tab alone would produce
+      // had cross-tab events been dropped.
+      // (Each click triggers multiple rrweb events; with 6 clicks per tab and
+      //  full snapshots we expect >> 10 total.)
+      expect(allEvents.length).toBeGreaterThan(10);
```
jpollock-ampl, line 529: Nit/non-blocking - Quick question - is there a stronger invariant we could assert here? A single tab in isolation produces well over 10 events from rrweb snapshots alone, so this would pass even if cross-tab promotion was completely broken and Tab B's events were entirely overwritten. The comment block above acknowledges rrweb's nondeterminism, but tagging events with a tab-specific marker before they enter the SDK and asserting both tab markers survive in the final union would give us a real regression guard.

packages/session-replay-browser/test/events-idb-store-multitab.test.ts
```diff
@@ -0,0 +1,684 @@
...
+
+      await store!.getSequencesToSend();
+
+      // Allow microtasks to settle so tx.done.catch fires.
+      await Promise.resolve();
+      await Promise.resolve();
+
+      // EXPECTED (post-fix): onPersistentFailure called once.
+      // ACTUAL (pre-fix): never called → FAILS.
+      expect(onPersistentFailure).toHaveBeenCalledTimes(1); // BUG: currently 0
```
jpollock-ampl: Do we still need this comment?  Can we remove the `// BUG: currently 0`?

jpollock-ampl (approved):

packages/session-replay-browser/src/events/events-idb-store.ts
lewgordon-amplitude: Good catch on the theoretical gap. You're right: if individual puts resolve but `tx.done` silently never commits, the success-path `cancelTimeout()` suppresses the watchdog and `tx.done.catch` never fires — so `recordSuccess()` resets the counter while data never reaches disk.

The practical protection here is IDB's readwrite transaction serialization. All three methods (`getSequencesToSend`, `storeCurrentSequence`, `addEventToCurrentSequence`) use readwrite transactions on the same two stores (`sessionCurrentSequence` + `sequencesToSend`). If T1's `tx.done` never settles, T2 is queued waiting for T1 to commit or abort — T2's put/get requests never resolve, T2 never reaches the soft-cancel, and T2's watchdog fires unconditionally. In steady-state recording, failure is detected on the *immediate* next operation, not just "eventually".

I did try your "leave the watchdog armed" alternative. There's a fundamental incompatibility with the test suite: `jest.runAllTimers()` fires timer callbacks synchronously without yielding to microtasks between them. fake-indexeddb's auto-commit happens via a 0ms setTimeout, so `runAllTimers()` fires both the auto-commit timer and the 5s watchdog in a single synchronous pass — `txDone.then(clearTimeout)` never gets a chance to run before the watchdog fires, causing spurious `recordFailure()` calls in the integration tests that I couldn't fix without a larger test refactor.

I've pushed a commit strengthening the `armTxDoneTimeout` comment to document the serialization guarantee explicitly. Happy to pursue the tx.done-based detection in a follow-up if you'd prefer that over the serialization argument.

packages/session-replay-browser/e2e/idb.spec.ts
lewgordon-amplitude: Good catch — the hardcoded `'d90c5cf09c'` would silently stop matching if the test API key ever changed, letting the hang test pass for the wrong reason.

Fixed in the latest commit: the prefix is now derived from `IDB_NAME.substring(0, 10)` so it tracks the constant automatically.

packages/session-replay-browser/e2e/idb.spec.ts
lewgordon-amplitude: Agreed — `> 10` is a weak floor and wouldn't catch a complete regression where Tab B's events were silently overwritten. Tagging events with a tab-specific marker (e.g. injecting a synthetic event with the page URL or a known prefix before SDK ingestion) and asserting both markers survive in the final union would be a much stronger guard.

I'll track this as a follow-up rather than blocking this PR — the main concern is rrweb's nondeterministic snapshot timing making it hard to inject reliable tab-specific markers without flakiness risk. Happy to pair on it separately.

packages/session-replay-browser/test/events-idb-store-multitab.test.ts
lewgordon-amplitude: Good eye — removed in the latest commit. Both occurrences (line 208 and 260) are cleaned up.

packages/session-replay-browser/test/events-idb-store-multitab.test.ts
```diff
@@ -0,0 +1,684 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
lewgordon-amplitude, line 1: Resolved

PR #1708, opened 2026-04-29
ci(publish-v2): fail if ref advanced since e2e checkout

Mercy811:
## Summary
Fail the `publish-v2` deploy job before `lerna version` runs if the release ref has advanced past the SHA the `e2e` job actually ran Playwright against. Closes the race that let [#1699](https://github.com/amplitude/Amplitude-TypeScript/pull/1699) ship without ever running through Playwright E2E.

Complementary follow-up tracked in [SDKW-5](https://linear.app/amplitude/issue/SDKW-5/block-pr-merges-to-main-while-publish-v2x-is-running): a merge-side gate that blocks PR merges to `main` while a publish run is in flight, so the race window never opens in the first place. This PR is the deploy-side seatbelt and lands first.

## 1. How #1699 bypassed E2E

`publish-v2` has two relevant jobs: `e2e` (runs Playwright via `.github/actions/e2e-test`) and `deploy` (runs `.github/actions/build-and-test`, which is unit tests only — no Playwright). Each one checks out by branch name (`ref: main`), so each one resolves `main` independently at *its* checkout time.

Timeline ([run 25065411243](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/25065411243)):

| Time (UTC) | Event |
| ---------- | ----- |
| 16:36:23 | `publish-v2` dispatched. `github.sha = 634ac979` (pre-#1699). |
| 16:36:47 | `e2e` job starts. |
| 16:37:19 | `e2e` checks out `main` → resolves to `634ac979`. Playwright runs against pre-#1699 code. |
| **16:40:52** | **#1699 (`98ecb9d`) merges to main.** |
| 16:44:18 | `e2e` finishes — passes (it never saw #1699). |
| 16:44:21 | `deploy` job starts. |
| ~16:44:30 | `deploy` checks out `main` → resolves to `98ecb9d` (now includes #1699). |
| ~16:44–16:59 | `Build and Test` runs `pnpm test` (unit tests). #1699 updated those tests itself, so they pass. |
| 16:59:37 | `lerna version` commits `chore(release): publish` (`7c363459`) on top of `98ecb9d` and pushes — fast-forward, succeeds. |
| ~17:00 | `pnpm deploy:publish` ships the SDK including #1699's compression-by-default change. |

The only check that would have caught the gzip behavior change was Playwright E2E, and Playwright never re-ran on the post-#1699 tree. Result: misleading git history (the release commit is a child of #1699) and a published artifact whose behavior change was never E2E-tested.

## 2. How this PR closes the race

The `e2e` job now exposes its post-checkout HEAD SHA as a job output (`needs.e2e.outputs.head_sha`). A new step in the deploy job, `Verify ref has not advanced since e2e checkout`, runs after `Build and Test` and before `Create release version`. It uses `git ls-remote` to read the current `origin/<ref>` SHA and compares it to E2E's HEAD; on mismatch it fails with a clear \"re-dispatch\" message.

Using E2E's actual HEAD as the comparison anchor (rather than `github.sha`) means the guard correctly covers hotfix releases dispatched with a `branch` input override — the anchor is *whatever E2E actually ran against*, regardless of what was current at workflow_dispatch time.

Three possible windows where a PR can land during the workflow, all closed:

| Window | Without this PR | With this PR |
| ------ | --------------- | ------------ |
| (a) Before deploy checkout | Deploy checks out post-PR ref, lerna fast-forwards, **silent bad publish** | Check sees `origin/<ref> ≠ e2e HEAD SHA` → **fails before lerna runs** |
| (b) Between deploy checkout and the new check | Same as (a) | Same as (a) — check fails |
| (c) Between the new check and `lerna` push (during the multi-minute version-hook build) | Same as (a) | Check passing implies `origin/<ref>` was still at e2e's HEAD SHA. Since refs are monotonic, the deploy checkout *also* got that SHA, so the local tree is anchored there. `lerna` commits on top of it; push to a now-advanced `origin/<ref>` **fails non-fast-forward**, `Publish Release to NPM` never runs. |

Skipped only for `skipLernaVersion=true` recovery runs (no lerna push, nothing to guard).

### Out of scope
- Bad code already merged into the release ref *before* `e2e` checkout — that's a test-coverage problem, not a race, and unaffected by this change.
- Force-pushes that rewind the ref to E2E's HEAD SHA would defeat the check, but force-pushes to release branches shouldn't happen.
- Adding `--force-push` to the `lerna version` invocation later would reopen window (c). Worth a code-owner note.

## Test plan
- [x] YAML parses locally
- [x] \`\${{ ... }}\` interpolations of user-controlled context (\`branch\`, \`ref_name\`) moved into \`env:\` to satisfy semgrep \`run-shell-injection\`
- [x] Script smoke-tested locally for: happy path, advanced ref, missing ref, invalid ref name (e.g. \`-evil\`), and \`ls-remote\` auth/network failure
- [ ] Next normal release dispatch succeeds (no PRs merge during the run — most common path, behavior unchanged)
- [ ] Manual race verification: dispatch publish-v2, push a no-op commit to main during the deploy window, confirm the deploy job fails on the new step with the \"re-dispatch\" message and that no \`chore(release): publish\` commit and no npm publish happens

### DIFF
--- a/.github/workflows/publish-v2.yml
+++ b/.github/workflows/publish-v2.yml
@@ -115,12 +115,12 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Verify main has not advanced since dispatch
+      - name: Verify ref has not advanced since dispatch
         if: ${{ github.event.inputs.skipLernaVersion != 'true' && (github.event.inputs.branch == '' || github.event.inputs.branch == github.ref_name) }}
         env:
           DISPATCH_SHA: ${{ github.sha }}
           REF: ${{ github.event.inputs.branch || github.ref_name }}
-        run: bash scripts/publish/check-main-not-advanced.sh
+        run: bash scripts/publish/check-ref-not-advanced.sh
 
       # Only create release version when using from-git (default behavior)
       # from-package mode uses existing package.json versions and doesn't need git tags
--- a/.github/workflows/publish-v2.yml
+++ b/.github/workflows/publish-v2.yml
@@ -115,6 +115,32 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      # Guard against the race where a PR merges to main between dispatch and
+      # the lerna version push. If main has advanced past the dispatch SHA, the
+      # chore(release): publish commit would land on top of code that this run
+      # never tested in E2E, making the git history misleading. Re-dispatch is
+      # the right recovery — it re-runs E2E against the new HEAD.
+      - name: Verify main has not advanced since dispatch
+        if: ${{ github.event.inputs.skipLernaVersion != 'true' && (github.event.inputs.branch == '' || github.event.inputs.branch == github.ref_name) }}
+        run: |
+          DISPATCH_SHA="${{ github.sha }}"
+          REF="${{ github.event.inputs.branch || github.ref_name }}"
+          git fetch origin "$REF"
+          REMOTE_SHA=$(git rev-parse "origin/$REF")
+          if [ "$REMOTE_SHA" != "$DISPATCH_SHA" ]; then
+            echo "❌ $REF has advanced since this publish was dispatched."
+            echo "   Dispatch SHA:      $DISPATCH_SHA"
+            echo "   Current $REF: $REMOTE_SHA"
+            echo ""
+            echo "A commit landed on $REF while this workflow was running. The"
+            echo "chore(release): publish commit would appear after it in history,"
+            echo "falsely implying the new commit was tested by this run's E2E."
+            echo ""
+            echo "Re-dispatch this workflow to include the new commits and re-run E2E."
+            exit 1
+          fi
+          echo "✅ $REF is still at dispatch SHA: $DISPATCH_SHA"
+
       # Only create release version when using from-git (default behavior)
       # from-package mode uses existing package.json versions and doesn't need git tags
       - name: Create release version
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-v2.yml
```diff
@@ -115,6 +115,33 @@ jobs:
...
+            echo "   Current $REF: $REMOTE_SHA"
+            echo ""
+            echo "A commit landed on $REF while this workflow was running. The"
+            echo "chore(release): publish commit would appear after it in history,"
+            echo "falsely implying the new commit was tested by this run's E2E."
+            echo ""
+            echo "Re-dispatch this workflow to include the new commits and re-run E2E."
+            exit 1
+          fi
+          echo "✅ $REF is still at dispatch SHA: $DISPATCH_SHA"
```
daniel-graham-amplitude: Can I request that we move this into a shell script and then have that script be invoked from here. 

Something like:
```
sh ./check-is-main-updated.sh
```

Not for a functional reason, just readability and so the `publish-v2` job doesn't get too unwieldy.

Mercy811:
@semgrep-amplitude. 

UPD: semgrep cannot be triggered by tagging. Pushing a new commit will re-run it

daniel-graham-amplitude (approved):

Mercy811:
@claude can we fix the issue by merging e2e job into publish job so that we only checkout once? Compared to current fix, which one is better? Do not update code, this is just a question

PR #1725, opened 2026-05-04
feat(session-replay): enforce min_session_duration_ms from remote config

bravecod:
## Summary

- Adds `min_session_duration_ms` field to `SamplingConfig` type (parsed from `sr_sampling_config` in remote config)
- Adds `minSessionDurationMs` field to `SessionReplayJoinedConfig`
- Tracks `sessionStartTime` when a new session ID is set in `asyncSetSessionId`
- In `sendEvents`, checks elapsed time against `minSessionDurationMs` before forwarding events — sessions shorter than the threshold are dropped with a log message
- Adds two unit tests covering the below-threshold (suppressed) and at/above-threshold (sent) cases

## Linear

[SR-3858](https://linear.app/amplitude/issue/SR-3858)

## Test plan

- [ ] `it should not send if session duration is below minSessionDurationMs` — mock a 1 s elapsed session with a 5 s threshold; verify `sendCurrentSequenceEvents` is not called
- [ ] `it should send if session duration meets minSessionDurationMs` — mock a 5 s elapsed session with a 1 s threshold; verify `sendCurrentSequenceEvents` is called with correct args
- [ ] Existing `sendEvents` tests still pass (no regression to the no-threshold path)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new remote-config-driven gate that can suppress replay delivery and alters session boundary/flush behavior; mistakes could drop or misattribute replay data despite added clamps and tests.
> 
> **Overview**
> Adds support for remote `sr_sampling_config.min_session_duration_ms`, exposing it as `joinedConfig.minSessionDurationMs` with defensive validation/clamping (60s ceiling) and warnings.
> 
> Enforces this threshold across replay delivery paths: `SessionReplay.sendEvents()` and page-exit `sendBeacon` now suppress sends until the session has run long enough, track suppressed send attempts, and emit a one-time `CustomRRwebEvent.REPLAY_GATE_DECISION` on the first send-after-pass.
> 
> Persists per-session replay start time in `localStorage` (with TTL pruning) so the gate survives reloads/navigation, and hardens session transitions by dropping prior-session beacon buffers and resetting gate state only on real session changes. `createEventsManager` also gains a synchronous `shouldSend` hook plus `dropPendingBeaconEvents` to prevent gated sequences from leaking via IDB/beacon.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 660aa6bc38e7a87ca8de9cc68beda68baf28022b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/src/config/joined-config.ts
+++ b/packages/session-replay-browser/src/config/joined-config.ts
@@ -139,6 +139,10 @@ export class SessionReplayJoinedConfigGenerator {
       if (Object.prototype.hasOwnProperty.call(samplingConfig, 'sample_rate')) {
         config.sampleRate = samplingConfig.sample_rate;
       }
+
+      if (Object.prototype.hasOwnProperty.call(samplingConfig, 'min_session_duration_ms')) {
+        config.minSessionDurationMs = samplingConfig.min_session_duration_ms;
+      }
     } else {
       // If config API response was valid (ie 200), but no config returned, assume that
       // customer has not yet set up config, and use sample rate from SDK options,
--- a/packages/session-replay-browser/src/config/joined-config.ts
+++ b/packages/session-replay-browser/src/config/joined-config.ts
@@ -141,7 +141,7 @@ export class SessionReplayJoinedConfigGenerator {
       }
 
       if (Object.prototype.hasOwnProperty.call(samplingConfig, 'min_session_duration_ms')) {
-        config.minSessionDurationMs = samplingConfig.min_session_duration_ms;
+        config.minSessionDurationMs = this.sanitizeMinSessionDurationMs(samplingConfig.min_session_duration_ms);
       }
     } else {
       // If config API response was valid (ie 200), but no config returned, assume that
@@ -234,7 +234,42 @@ export class SessionReplayJoinedConfigGenerator {
       remoteConfig: sessionReplayRemoteConfig,
     };
   }
+
+  /**
+   * Defensive bounds for the remote-supplied min_session_duration_ms. A misconfigured
+   * value (e.g. 30_000_000) would silently suppress every replay until the config is
+   * pushed again, so we clamp to a sane ceiling and warn on out-of-range inputs.
+   * Returns undefined for clearly invalid values so the gate falls back to disabled
+   * rather than carrying a NaN through downstream comparisons.
+   */
+  private sanitizeMinSessionDurationMs(raw: unknown): number | undefined {
+    if (typeof raw !== 'number' || !Number.isFinite(raw)) {
+      this.localConfig.loggerProvider.warn(
+        `min_session_duration_ms remote value is not a finite number (got ${String(raw)}); ignoring.`,
+      );
+      return undefined;
+    }
+    if (raw < 0) {
+      this.localConfig.loggerProvider.warn(`min_session_duration_ms remote value is negative (${raw}); ignoring.`);
+      return undefined;
+    }
+    if (raw > MAX_MIN_SESSION_DURATION_MS) {
+      this.localConfig.loggerProvider.warn(
+        `min_session_duration_ms remote value ${raw} exceeds ${MAX_MIN_SESSION_DURATION_MS}ms ceiling; clamping.`,
+      );
+      return MAX_MIN_SESSION_DURATION_MS;
+    }
+    return raw;
+  }
 }
+
+/**
+ * Upper bound for the remote-configured replay min duration. 60 seconds is well above
+ * any reasonable bounce threshold; values higher than this are almost certainly typos
+ * (e.g. seconds confused for milliseconds, or an extra zero) and would otherwise
+ * suppress every replay until the config is corrected.
+ */
+export const MAX_MIN_SESSION_DURATION_MS = 60_000;
 
 export const createSessionReplayJoinedConfigGenerator = async (apiKey: string, options: SessionReplayOptions) => {
   const localConfig = new SessionReplayLocalConfig(apiKey, options);
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -17,6 +17,11 @@ export type EventsManagerWithBeacon<Type extends EventType> = AmplitudeSessionRe
    * Used to populate a sendBeacon payload when the page is unloading.
    */
   getBeaconEvents(): string[];
+  /**
+   * Drops all pending beacon events. Used when the session is decided to be below
+   * the min duration threshold so its events don't leak into a later session's beacon.
+   */
+  dropPendingBeaconEvents(): void;
   trackDestination: SessionReplayTrackDestination;
 };
 
@@ -219,8 +224,13 @@ export const createEventsManager = async <Type extends EventType>({
           if (!canSend) {
             // The split atomically moved events to sequencesToSend; without cleanup
             // they would be unconditionally replayed on next page load via
-            // sendStoredEvents, bypassing the min session duration gate.
-            void store.cleanUpSessionEventsStore(sequenceToSend.sessionId, sequenceToSend.sequenceId);
+            // sendStoredEvents, bypassing the min session duration gate. The split
+            // batch must also be dropped from the beacon buffer so it isn't sent
+            // via sendBeacon on page unload (potentially attributed to a later session).
+            store.cleanUpSessionEventsStore(sequenceToSend.sessionId, sequenceToSend.sequenceId).catch((e) => {
+              config.loggerProvider.warn('Failed to clean up dropped session replay sequence:', e);
+            });
+            advanceBeaconWindow(absIdx);
             return;
           }
           // Events before absIdx belong to the split batch being sent; advance window.
@@ -244,12 +254,17 @@ export const createEventsManager = async <Type extends EventType>({
 
   const getBeaconEvents = (): string[] => [...beaconBuffer];
 
+  const dropPendingBeaconEvents = () => {
+    advanceBeaconWindow(beaconWindowStart + beaconBuffer.length);
+  };
+
   return {
     sendCurrentSequenceEvents,
     addEvent,
     sendStoredEvents,
     flush,
     getBeaconEvents,
+    dropPendingBeaconEvents,
     trackDestination,
   };
 };
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -133,9 +133,6 @@ export const createEventsManager = async <Type extends EventType>({
         });
     }
 
-    if (shouldSend && !shouldSend()) {
-      return;
-    }
     trackDestination.sendEventsList({
       events: events,
       sessionId: sessionId,
@@ -155,6 +152,11 @@ export const createEventsManager = async <Type extends EventType>({
 
   const sendCurrentSequenceEvents = ({ sessionId, deviceId }: { sessionId: number; deviceId: string }) => {
     lastKnownDeviceId = deviceId;
+    // Evaluate shouldSend synchronously before the async store read. asyncSetSessionId
+    // updates sessionStartTime immediately after calling sendEvents(), so by the time
+    // storeCurrentSequence resolves the start time would reflect the new session and
+    // the elapsed check would compute ~0ms, silently dropping the previous session's events.
+    if (shouldSend && !shouldSend()) return;
     // Snapshot the absolute end-index before the async store read so that any events
     // pushed after this point are NOT considered sent and remain in the beacon buffer.
     const snapshotAbsIdx = beaconWindowStart + beaconBuffer.length;
@@ -209,6 +211,7 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
+          if (shouldSend && !shouldSend()) return;
           // Events before absIdx belong to the split batch being sent; advance window.
           advanceBeaconWindow(absIdx);
           sendEventsList({
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -28,6 +28,7 @@ export const createEventsManager = async <Type extends EventType>({
   payloadBatcher,
   storeType,
   trackDestinationWorkerScript,
+  shouldSend,
 }: {
   config: SessionReplayJoinedConfig;
   type: Type;
@@ -36,6 +37,7 @@ export const createEventsManager = async <Type extends EventType>({
   payloadBatcher?: PayloadBatcher;
   storeType: StoreType;
   trackDestinationWorkerScript?: string;
+  shouldSend?: () => boolean;
 }): Promise<EventsManagerWithBeacon<Type>> => {
   const trackDestination = new SessionReplayTrackDestination({
     ...config,
@@ -131,6 +133,9 @@ export const createEventsManager = async <Type extends EventType>({
         });
     }
 
+    if (shouldSend && !shouldSend()) {
+      return;
+    }
     trackDestination.sendEventsList({
       events: events,
       sessionId: sessionId,
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -216,7 +216,13 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
-          if (!canSend) return;
+          if (!canSend) {
+            // The split atomically moved events to sequencesToSend; without cleanup
+            // they would be unconditionally replayed on next page load via
+            // sendStoredEvents, bypassing the min session duration gate.
+            void store.cleanUpSessionEventsStore(sequenceToSend.sessionId, sequenceToSend.sequenceId);
+            return;
+          }
           // Events before absIdx belong to the split batch being sent; advance window.
           advanceBeaconWindow(absIdx);
           sendEventsList({
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -202,6 +202,11 @@ export const createEventsManager = async <Type extends EventType>({
     deviceId: string;
   }) => {
     lastKnownDeviceId = deviceId;
+    // Capture shouldSend synchronously before the async store write, for the same
+    // reason as sendCurrentSequenceEvents: asyncSetSessionId updates sessionStartTime
+    // synchronously, so evaluating inside the .then() would use the new session's
+    // start time and compute ~0ms elapsed, silently dropping a valid batch.
+    const canSend = !shouldSend || shouldSend();
     // Record the absolute index of this event in the beacon buffer before the async
     // store operation. If a batch split occurs, we advance the window up to (but not
     // including) this event so that it starts the next pending window.
@@ -211,7 +216,7 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
-          if (shouldSend && !shouldSend()) return;
+          if (!canSend) return;
           // Events before absIdx belong to the split batch being sent; advance window.
           advanceBeaconWindow(absIdx);
           sendEventsList({
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -98,6 +98,13 @@ export class SessionReplay implements AmplitudeSessionReplay {
   pageLeaveFns: PageLeaveFn[] = [];
   sessionStartTime: number | undefined;
   rrwebEventManager: EventsManagerWithBeacon<'replay'> | undefined;
+  /**
+   * Count of sendEvents() calls suppressed by the min-session-duration gate for the
+   * current session. Drives the REPLAY_GATE_DECISION rrweb event on first send-after-pass.
+   */
+  private suppressedSendCount = 0;
+  /** True once REPLAY_GATE_DECISION has been emitted for the current session. */
+  private hasEmittedGateDecision = false;
   private scrollHook?: scrollCallback;
   private clickHandler?: ClickHandler;
   private networkObservers?: NetworkObservers;
@@ -407,6 +414,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
       deviceId: deviceIdForReplayId,
     });
     this.sessionStartTime = Date.now();
+    this.suppressedSendCount = 0;
+    this.hasEmittedGateDecision = false;
     // Persist new session's start time so a page reload mid-session resumes the gate.
     // Also drop the previous session's entry — the localStorage cleanup deliberately
     // happens here (and via TTL prune on init) rather than after a successful send:
@@ -613,7 +622,25 @@ export class SessionReplay implements AmplitudeSessionReplay {
         // Drop pending beacon events so this below-threshold session's data
         // can't leak into a later session's beacon flush on page unload.
         this.rrwebEventManager?.dropPendingBeaconEvents();
+        this.suppressedSendCount++;
         return;
+      }
+      // On the first send-after-pass for the session, emit a custom rrweb event so the
+      // payload itself carries the gate verdict. Lets backend ingestion diff intended
+      // vs actual replay counts to spot start-time-tracking regressions.
+      if (!this.hasEmittedGateDecision && this.config?.minSessionDurationMs !== undefined) {
+        this.hasEmittedGateDecision = true;
+        const elapsedMs = this.sessionStartTime !== undefined ? Date.now() - this.sessionStartTime : undefined;
+        void this.addCustomRRWebEvent(
+          CustomRRwebEvent.REPLAY_GATE_DECISION,
+          {
+            sessionId: sessionIdToSend,
+            suppressedSendCount: this.suppressedSendCount,
+            elapsedMs,
+            minSessionDurationMs: this.config.minSessionDurationMs,
+          },
+          false,
+        );
       }
       this.eventsManager.sendCurrentSequenceEvents({ sessionId: sessionIdToSend, deviceId });
     }
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -190,6 +190,24 @@ export class SessionReplay implements AmplitudeSessionReplay {
     };
   }
 
+  /**
+   * Single source of truth for the min_session_duration_ms gate. Returns true when the
+   * current session has not yet reached the configured threshold (and we have both a
+   * configured value and a recorded start time). Returns false when there's no
+   * threshold, no recorded start time, or the threshold has been met — i.e. it
+   * answers "should this batch be suppressed?".
+   *
+   * Centralizing the check means future changes (clock-skew tolerance, switching to
+   * performance.now(), etc.) are one-line edits.
+   */
+  private isBelowMinSessionDuration(): boolean {
+    const minSessionDurationMs = this.config?.minSessionDurationMs;
+    if (minSessionDurationMs === undefined || this.sessionStartTime === undefined) {
+      return false;
+    }
+    return Date.now() - this.sessionStartTime < minSessionDurationMs;
+  }
+
   private getCurrentPageForTargeting(): SessionReplayTargetingInput['page'] {
     const currentUrl = getGlobalScope()?.location?.href;
     return currentUrl != null ? { url: currentUrl } : undefined;
@@ -266,12 +284,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
         type: 'replay',
         storeType,
         trackDestinationWorkerScript,
-        shouldSend: () => {
-          const { minSessionDurationMs } = this.config ?? {};
-          if (minSessionDurationMs === undefined || this.sessionStartTime === undefined) return true;
-          const elapsed = Date.now() - this.sessionStartTime;
-          return elapsed >= minSessionDurationMs;
-        },
+        shouldSend: () => !this.isBelowMinSessionDuration(),
       });
       this.rrwebEventManager = rrwebEventManager;
       managers.push({ name: 'replay', manager: rrwebEventManager });
@@ -323,10 +336,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
         if (!events.length) return;
         const deviceId = this.getDeviceId();
         if (!deviceId) return;
-        const { minSessionDurationMs } = this.config;
-        if (minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
-          if (Date.now() - this.sessionStartTime < minSessionDurationMs) return;
-        }
+        if (this.isBelowMinSessionDuration()) return;
         rrwebEventManager.trackDestination.sendBeacon({
           events,
           sessionId: this.identifiers.sessionId,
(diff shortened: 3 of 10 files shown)


lewgordon-amplitude (changes requested):
The shape of the change is right — gate at `sendEventsList` via a callback, plus a higher-level early-return for logging — but there are correctness gaps that will cause real data drops in production. I'd block on (1)–(3), take (4)–(5) before merge if cheap, and treat the rest as follow-ups.

### Must fix

**1. Gate by per-payload sessionId, not by `this.sessionStartTime`.**
The `shouldSend` closure (`session-replay.ts:254`) reads `this.sessionStartTime` lazily, which is mutated on every session change and on init. Two concrete failure modes:
- *Cross-session race:* `asyncSetSessionId` calls `sendEvents(previousSessionId)`, then resets `sessionStartTime` on line 373. The IDB read inside `sendCurrentSequenceEvents` is async; by the time `sendEventsList` calls `shouldSend`, the timer points at the new session. Old session's events get dropped against new session's clock.
- *Stored-events flush on init:* `sendStoredEvents` flushes prior sessions' sequences on boot. Each goes through `sendEventsList` → `shouldSend`, which compares against the **current** session's timer (just initialized). Long, legitimate prior sessions get dropped silently.

Fix: track start times per session (`Map<sessionId, number>`), and have `sendEventsList` look up the time for the sessionId it's actually sending.

**2. Beacon path bypasses the gate.**
`session-replay.ts:310` calls `rrwebEventManager.trackDestination.sendBeacon` directly. Short sessions exiting via `pagehide`/`beforeunload` ship anyway. Move the check into `sendBeacon` itself so there's a single source of truth.

**3. `flush()` bypasses the gate.**
`eventsManager.flush()` (`events-manager.ts:227`) calls `trackDestination.flush()` directly, skipping `sendEventsList`. Same fix: gate inside the trackDestination, or have `flush` consult `shouldSend` before forwarding.

### Should fix

**4. The init timer assumption is brittle.**
```ts
this.sessionStartTime = typeof options.sessionId === 'number' ? options.sessionId : Date.now();
```
This silently treats numeric session IDs as wall-clock timestamps. Works for default Amplitude session IDs, breaks for anyone using custom numeric IDs, and is inconsistent with `asyncSetSessionId` (which uses `Date.now()` unconditionally). Pick one rule: either always `Date.now()` (and accept timer reset on re-init), or persist `sessionStartTime` alongside the session ID. Document the choice.

**5. Validate the remote value.**
A bad `min_session_duration_ms` (NaN, negative, string) silently suppresses everything. Reject non-finite/non-negative values at the joined-config layer with a `loggerProvider.warn`, and consider an upper-bound clamp so a fat-fingered `300000` doesn't nuke a customer's data.

### Other

**6. Drop logging is asymmetric.**
Only the `sendEvents` gate logs; the `shouldSend` gate is silent. Add a `loggerProvider.debug` line in the events-manager path so all drops are diagnosable, and consider emitting a custom rrweb event (or metadata field) when a session is suppressed — otherwise customer-support questions like "why did my replay volume drop?" have no signal to point at.

**7. Confirm scope: replay only, or all session-replay traffic?**
`shouldSend` is wired into the replay manager but not the interaction manager (line 270). Short sessions still ship interaction events. If the product intent is "drop sub-threshold sessions entirely," that's a gap. If it's "save replay storage cost only," current scoping is correct — please call this out in the PR description so future readers know.

**8. Test coverage.**
- The `sendEvents` tests mutate `sessionStartTime` directly, which sidesteps the production init path (and hides issue #4). Add a test that goes through real init with a `Date.now()`-style sessionId.
- No regression test for the cross-session race or the stored-events flush path.
- No test for the beacon path (will need one once #2 is fixed).

**Nit: Type placement.** `min_session_duration_ms` lives on `SamplingConfig` but isn't sampling. If this is the first of more session-level filters to come, carve out a separate type now; if it's a one-off, leave it.

**Nit: `willSend` variable.** `sendEvents` introduces `willSend` only to negate it on the next line — `if (sessionDuration < this.config.minSessionDurationMs)` reads cleaner.

---

The two-layer gating design is fine and the events-manager refactor is clean, so once (1)–(3) are addressed I expect this to be a short re-review.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -361,6 +362,7 @@ export class SessionReplay implements AmplitudeSessionReplay {
       sessionId: sessionId,
       deviceId: deviceIdForReplayId,
     });
+    this.sessionStartTime = Date.now();
```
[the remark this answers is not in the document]
bravecod: Addressed in commit ab786fa4 — `sessionStartTime = Date.now()` is now set during `_init` alongside the initial `this.identifiers` assignment, so the first session is gated by `minSessionDurationMs` like any subsequent session. Regression test added in `session-replay.test.ts`.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -248,6 +251,12 @@ export class SessionReplay implements AmplitudeSessionReplay {
         type: 'replay',
         storeType,
         trackDestinationWorkerScript,
+        shouldSend: () => {
+          const { minSessionDurationMs } = this.config ?? {};
+          if (minSessionDurationMs === undefined || this.sessionStartTime === undefined) return true;
+          const elapsed = Date.now() - this.sessionStartTime;
+          return elapsed >= minSessionDurationMs;
+        },
```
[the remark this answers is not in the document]
bravecod: Addressed in commit fc6157bb — `shouldSend` is now evaluated synchronously inside `sendCurrentSequenceEvents` *before* the async `storeCurrentSequence` call, so the previous session ID rollover in `asyncSetSessionId` cannot race the elapsed check. The inline comment on lines 155–158 documents the invariant.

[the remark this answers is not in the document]
bravecod: Addressed in commit fc6157bb — the page-leave beacon path in `session-replay.ts` now performs the same `minSessionDurationMs` elapsed-time check as `sendEvents` and short-circuits before calling `trackDestination.sendBeacon`, so under-threshold sessions are no longer leaked on unload.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -204,6 +211,7 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
+          if (shouldSend && !shouldSend()) return;
```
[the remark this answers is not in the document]
bravecod: Addressed in commit f775ac7c — `shouldSend` is now captured synchronously into `canSend` before `addEventToCurrentSequence` is called, mirroring the fix in `sendCurrentSequenceEvents`. The async `.then()` only consults the snapshot, so an interleaved `asyncSetSessionId` cannot flip the result mid-flight.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -204,6 +216,7 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
+          if (!canSend) return;
```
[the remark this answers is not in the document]
bravecod: Good catch — addressed in commit 1e44eac3. When `canSend` is false during a batch split, the events were already atomically moved to `sequencesToSend` by the IDB transaction, so without cleanup they would be unconditionally replayed via `sendStoredEvents` on next page load. The `addEvent` `.then()` now calls `store.cleanUpSessionEventsStore(sessionId, sequenceId)` on the dropped batch. Regression test added.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -204,6 +216,13 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
+          if (!canSend) {
+            // The split atomically moved events to sequencesToSend; without cleanup
+            // they would be unconditionally replayed on next page load via
+            // sendStoredEvents, bypassing the min session duration gate.
+            void store.cleanUpSessionEventsStore(sequenceToSend.sessionId, sequenceToSend.sequenceId);
```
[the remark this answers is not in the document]
bravecod: Addressed in commit b9716f43 — `cleanUpSessionEventsStore` now has a `.catch()` handler that logs via `config.loggerProvider.warn`, matching the convention for the other async store calls in this file. Regression test added asserting the warn is invoked when the IDB op rejects.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +558,19 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
-      deviceId &&
+    if (this.eventsManager && sessionIdToSend && deviceId) {
+      if (this.config?.minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
+        const sessionDuration = Date.now() - this.sessionStartTime;
+        const willSend = sessionDuration >= this.config.minSessionDurationMs;
+        if (!willSend) {
+          this.loggerProvider.log(
+            `Session ${sessionIdToSend} not sent: duration ${sessionDuration}ms is below minimum ${this.config.minSessionDurationMs}ms.`,
+          );
+          return;
```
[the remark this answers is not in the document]
bravecod: Good catch — addressed in commit b9716f43. Added a `dropPendingBeaconEvents()` method on the events manager that advances the beacon window past the current end. `sendEvents` now calls it on the below-threshold short-circuit path so the dropped sessions events cant leak into a later session as suggested. The `addEvent` `canSend=false` split path also advances the beacon window past the dropped batch (otherwise the split events would still sit in the buffer). Tests added for both paths.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -196,6 +198,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
       this.loggerProvider.enable(options.logLevel as LogLevel);
     this.currentPageUrl = getCurrentUrl();
     this.identifiers = new SessionIdentifiers({ sessionId: options.sessionId, deviceId: options.deviceId });
+    // Use the session ID itself as the start time if it's a numeric timestamp, otherwise Date.now()
```
lewgordon-amplitude: **🚫 Blocker — don't infer session start from `sessionId`.**

There's no contract that a numeric `sessionId` is an epoch-ms timestamp — it could be a counter, hash, or arbitrary numeric ID. A non-timestamp numeric ID produces nonsense elapsed values and either always passes or always fails the gate.

`Date.now()` alone undercounts on resumed sessions. Adding a `sessionStartTime` option puts the burden on the caller and standalone integrations will get it wrong.

**Recommendation:** persist replay start time in `localStorage` keyed by replay ID:
- On `init`, read `replayStart:${apiKey}:${replayId}`. Use it if present; otherwise write `Date.now()` and use that.
- On `asyncSetSessionId`, write the new replay ID's start time.
- Clean up on successful send, on below-threshold drop, and via TTL prune on init.
- `try/catch` around storage access with `Date.now()` fallback for private-mode/quota-exceeded.

This shifts the semantic from "min session duration" to "min replay duration" — please confirm with PM that SR-3858 wants replay-duration semantics (likely yes, since the gate exists to filter unusably-short *recordings*).

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -89,6 +89,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
 
   // Visible for testing only
   pageLeaveFns: PageLeaveFn[] = [];
```
lewgordon-amplitude, line 101: The line above (`pageLeaveFns`) carries a `// Visible for testing only` comment. These two new public fields should match — either add the comment or make them `private` and access via `(sessionReplay as any).field` in tests.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -248,7 +252,14 @@ export class SessionReplay implements AmplitudeSessionReplay {
         type: 'replay',
         storeType,
         trackDestinationWorkerScript,
+        shouldSend: () => {
+          const { minSessionDurationMs } = this.config ?? {};
+          if (minSessionDurationMs === undefined || this.sessionStartTime === undefined) return true;
+          const elapsed = Date.now() - this.sessionStartTime;
```
lewgordon-amplitude: This same `Date.now() - sessionStartTime < minSessionDurationMs` logic appears here, in the beacon path (`pageLeaveFns`), and in `sendEvents`. Extract a private helper (e.g. `private isBelowMinSessionDuration(): boolean`) so future changes — switching to `performance.now`, adding clock-skew tolerance — are one-line edits.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -298,6 +309,10 @@ export class SessionReplay implements AmplitudeSessionReplay {
         if (!events.length) return;
         const deviceId = this.getDeviceId();
         if (!deviceId) return;
+        const { minSessionDurationMs } = this.config;
+        if (minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
+          if (Date.now() - this.sessionStartTime < minSessionDurationMs) return;
```
lewgordon-amplitude: Pre-existing race made slightly worse here: the beacon buffer can still contain events from the *previous* session if `asyncSetSessionId` was called and its async `storeCurrentSequence` hasn't resolved before unload. The gate reads the current (new) `sessionStartTime`, so legitimately-sendable old-session events would now be suppressed.

Not blocking, but please add a comment acknowledging the limitation. The clean fix is tracking start time per buffered batch rather than globally — worth a follow-up.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +560,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
-    this.eventsManager &&
-      sessionIdToSend &&
-      deviceId &&
+    if (this.eventsManager && sessionIdToSend && deviceId) {
+      if (this.config?.minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
+        const sessionDuration = Date.now() - this.sessionStartTime;
+        const willSend = sessionDuration >= this.config.minSessionDurationMs;
+        if (!willSend) {
+          this.loggerProvider.log(
+            `Session ${sessionIdToSend} not sent: duration ${sessionDuration}ms is below minimum ${this.config.minSessionDurationMs}ms.`,
```
lewgordon-amplitude: Per-suppression log lines exist but there's no aggregate signal. If start-time tracking ever breaks in production, replay ingestion drops with no obvious root cause. Consider emitting an internal Amplitude event (e.g. `replay_suppressed_below_min_duration`) so rollout is observable and on-call can diff suppression rate against ingest drops.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +560,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
   sendEvents(sessionId?: string | number) {
     const sessionIdToSend = sessionId || this.identifiers?.sessionId;
     const deviceId = this.getDeviceId();
-    this.eventsManager &&
-      sessionIdToSend &&
-      deviceId &&
+    if (this.eventsManager && sessionIdToSend && deviceId) {
+      if (this.config?.minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
+        const sessionDuration = Date.now() - this.sessionStartTime;
```
lewgordon-amplitude: `sendEvents` is gated, but `flush()` is unconditional — a programmatic `flush()` on a below-threshold session bypasses the gate entirely. Likely correct for `sendStoredEvents` (those passed gates at write time), but please confirm `flush()` behavior against SR-3858. If it should also gate, add the check; if not, drop a comment explaining why.

packages/session-replay-browser/src/config/joined-config.ts
```diff
@@ -139,6 +139,10 @@ export class SessionReplayJoinedConfigGenerator {
       if (Object.prototype.hasOwnProperty.call(samplingConfig, 'sample_rate')) {
         config.sampleRate = samplingConfig.sample_rate;
       }
+
+      if (Object.prototype.hasOwnProperty.call(samplingConfig, 'min_session_duration_ms')) {
+        config.minSessionDurationMs = samplingConfig.min_session_duration_ms;
```
lewgordon-amplitude: No client-side sanity bound — a misconfigured remote value (`min_session_duration_ms: 30000000`) would silently suppress all replays. Consider clamping to a reasonable max (e.g. 60s) or `warn`-logging out-of-range values so a typo is recoverable without a config push.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -150,6 +157,11 @@ export const createEventsManager = async <Type extends EventType>({
 
   const sendCurrentSequenceEvents = ({ sessionId, deviceId }: { sessionId: number; deviceId: string }) => {
     lastKnownDeviceId = deviceId;
+    // Evaluate shouldSend synchronously before the async store read. asyncSetSessionId
+    // updates sessionStartTime immediately after calling sendEvents(), so by the time
+    // storeCurrentSequence resolves the start time would reflect the new session and
+    // the elapsed check would compute ~0ms, silently dropping the previous session's events.
```
lewgordon-amplitude, line 163: 👍 Good catch on the race. The synchronous capture pattern (and the same one in `addEvent` below) is exactly right, and the comment explains *why* clearly enough that someone won't accidentally "clean it up" later.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -204,6 +221,18 @@ export const createEventsManager = async <Type extends EventType>({
       .addEventToCurrentSequence(sessionId, event.data)
       .then((sequenceToSend) => {
         if (sequenceToSend) {
+          if (!canSend) {
+            // The split atomically moved events to sequencesToSend; without cleanup
+            // they would be unconditionally replayed on next page load via
+            // sendStoredEvents, bypassing the min session duration gate. The split
+            // batch must also be dropped from the beacon buffer so it isn't sent
+            // via sendBeacon on page unload (potentially attributed to a later session).
+            store.cleanUpSessionEventsStore(sequenceToSend.sessionId, sequenceToSend.sequenceId).catch((e) => {
```
lewgordon-amplitude, line 230: 👍 The cleanup-on-suppression for the split path closes a real leak — without it, suppressed events would replay unconditionally via `sendStoredEvents` on next load, completely bypassing the gate.

packages/session-replay-browser/test/session-replay.test.ts
```diff
@@ -261,6 +261,19 @@ describe('SessionReplay', () => {
     }
   });
   describe('init', () => {
+    test('should set sessionStartTime from numeric sessionId', async () => {
```
lewgordon-amplitude: Once the blocker is resolved (no more `sessionId`-as-timestamp inference), this test needs to go. Replacement coverage for the localStorage approach should include: fresh-init writes-and-reads, second-init recovers stored value, missing/corrupt storage falls back to `Date.now()`, TTL prune drops stale entries.

lewgordon-amplitude (changes requested):
Solid race-condition handling and thorough beacon-leakage cleanup — the synchronous `canSend` capture and the split-path `cleanUpSessionEventsStore` both address real bugs.

Blocking on one thing: the `sessionStartTime` derivation. Inferring it from `sessionId` assumes the ID is an epoch-ms timestamp, which isn't part of the contract. Falling back to `Date.now()` undercounts resumed sessions. Asking customers to pass `sessionStartTime` is fragile in standalone mode. Recommend persisting replay start time in `localStorage` keyed by replay ID so it survives navigations without caller cooperation.

Plus a few non-blocking nits and a couple of risks worth addressing before rollout — see inline.

packages/session-replay-browser/src/session-replay.ts
bravecod: Addressed in commit 2487d8bf. New `replay-start-time-store.ts` module keys by `${apiKey.substring(0,10)}_${sessionId}` in localStorage, with try/catch fallback to `Date.now()` for private-mode / quota-exceeded / sandboxed iframes. `_init` now prunes stale entries (>24h TTL) then read-or-init the current sessions start time; `asyncSetSessionId` writes the new session and removes the previous. Replaced the `sessionId`-as-timestamp tests with fresh-init / second-init recovery / corrupt entry / TTL prune / storage-throws coverage.

One deviation from your recipe: I deliberately *did not* clear on successful send. If a session passes the threshold, sends a batch, then the page reloads mid-session, clearing on send would force re-init to `Date.now()` and re-suppress an already-eligible session. The TTL prune handles long-term cleanup. Will pull in the PM-confirmation on "min replay duration" semantics as a follow-up if you want — happy to align the wording in the field/comment if PM agrees.

packages/session-replay-browser/src/session-replay.ts
bravecod: Addressed in commit 22155eec. Expanded the `// Visible for testing only` comment to explicitly scope it across all three public fields (`pageLeaveFns`, `sessionStartTime`, `rrwebEventManager`) so future readers don't have to infer.

packages/session-replay-browser/src/session-replay.ts
bravecod: Addressed in commit ee8d1496. Extracted `isBelowMinSessionDuration()` as a private method and replaced all three sites (`createEventsManager` `shouldSend` callback, the `pageLeaveFns` beacon path, and `sendEvents`). Future changes — switching to `performance.now()`, adding clock-skew tolerance — are now one-line edits in the helper.

packages/session-replay-browser/src/session-replay.ts
bravecod: Addressed in commit a06dba33 — added an explicit comment above the `pageLeaveFns` registration block calling out the cross-session race, why the gate would suppress legitimately-sendable old-session events, and flagging "track start time per buffered batch" as the clean follow-up fix.

packages/session-replay-browser/src/session-replay.ts
bravecod: Addressed in commit 7e1d41b9. Added a `CustomRRwebEvent.REPLAY_GATE_DECISION` that fires once per session on the first send-after-pass, carrying `suppressedSendCount`, `elapsedMs`, `minSessionDurationMs`, and `sessionId`. Lets backend diff intended vs actual replay counts to spot start-time-tracking regressions.

One caveat I want to flag: the event lives inside the rrweb stream, so sessions that bounce *before* crossing the threshold never emit (their entire payload is suppressed by design). Backend would have to detect those via the gap between session count and replay count. A proper "internal Amplitude event" bridge would close that gap but requires plumbing through `plugin-session-replay-browser` — happy to do as a follow-up if the gap-style signal isn't enough.

packages/session-replay-browser/src/session-replay.ts
bravecod: Investigated and confirmed flush() is intentionally ungated — added a comment in commit c184cb02. Traced every path that queues into `trackDestination`:

1. `sendCurrentSequenceEvents` evaluates `shouldSend` synchronously *before* calling `storeCurrentSequence`, so suppressed batches never enter the destination queue.
2. `addEvent` batch-split path captures `canSend` synchronously and either queues or cleans up — never queues a suppressed split.
3. `sendStoredEvents` reads from the `sequencesToSend` IDB table, which only receives entries from the two above paths (the IDB `sessionCurrentSequence` table never feeds the destination queue directly).

Result: `flush()` only ever forwards payloads that have already passed the gate at queue time. A duplicate gate would be unreachable. Comment explains the invariant.

packages/session-replay-browser/src/config/joined-config.ts
bravecod: Addressed in commit 5f7c078a. Added `sanitizeMinSessionDurationMs` in `SessionReplayJoinedConfigGenerator`:

- Clamps values above 60s (a new `MAX_MIN_SESSION_DURATION_MS` constant — well past any reasonable bounce threshold; higher is almost certainly a typo) and warns.
- Drops non-finite, NaN, and negative values (returns `undefined` so the gate falls back to disabled rather than carrying a NaN through `Date.now() - sessionStartTime < NaN` comparisons) and warns.

Tests added for each branch.

packages/session-replay-browser/src/events/events-manager.ts
bravecod: Thanks! Agreed — the sync-capture comment is the load-bearing piece here, so I kept it (and tightened the language in the events-manager doc string).

packages/session-replay-browser/src/events/events-manager.ts
bravecod: Thanks — yeah, the unconditional `sendStoredEvents` was the surprising bit. The `cleanUpSessionEventsStore` path now also has a `.catch()` (from your earlier feedback) so an IDB failure during the cleanup itself surfaces as a warn rather than silently leaving the entry behind.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -361,6 +413,20 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    this.suppressedSendCount = 0;
+    this.hasEmittedGateDecision = false;
+    // Persist new session's start time so a page reload mid-session resumes the gate.
+    // Also drop the previous session's entry — the localStorage cleanup deliberately
+    // happens here (and via TTL prune on init) rather than after a successful send:
+    // mid-session reloads would otherwise re-init the start time to Date.now() and
+    // re-suppress already-eligible sessions.
+    if (this.config?.apiKey) {
+      setReplayStartTime(this.config.apiKey, sessionId, this.sessionStartTime, this.loggerProvider);
+      if (previousSessionId !== undefined) {
```
lewgordon-amplitude: Guard against `previousSessionId === sessionId`. If a caller passes the current sessionId redundantly, line 425 writes the entry and this block immediately deletes it — leaving storage empty. On the next page reload, `getOrInitReplayStartTime` re-inits from `Date.now()`, restarting the gate clock for what is supposed to be a continuing session.

The plugin guards against this (`plugin-session-replay-browser/src/session-replay.ts:149,164` check id inequality before calling `setSessionId`), but the standalone SR API at `session-replay.ts:392` has no inequality check, so external standalone consumers could hit this.

```suggestion
      if (previousSessionId !== undefined && previousSessionId !== sessionId) {
```

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +610,40 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+        // can't leak into a later session's beacon flush on page unload.
+        this.rrwebEventManager?.dropPendingBeaconEvents();
+        this.suppressedSendCount++;
+        return;
+      }
+      // On the first send-after-pass for the session, emit a custom rrweb event so the
+      // payload itself carries the gate verdict. Lets backend ingestion diff intended
+      // vs actual replay counts to spot start-time-tracking regressions.
+      if (!this.hasEmittedGateDecision && this.config?.minSessionDurationMs !== undefined) {
+        this.hasEmittedGateDecision = true;
```
lewgordon-amplitude, line 661: Nit: `hasEmittedGateDecision` is set even when the emit silently drops. `addCustomRRWebEvent` is a no-op when `recordCancelCallback`/`recordFunction` aren't set (lines 994-1003), but the flag is still tripped — so the future first-real-send won't re-emit.

Narrow edge case: a reloaded long-lived session whose stored start time puts it immediately above threshold, where a gated `sendEvents` (blur listener registered before `recordEvents()` activates) fires before recording is up. That session's `replay-gate-decision` signal is lost. Could fix by extending the precondition on line 631 to also require `this.recordCancelCallback && this.recordFunction` before tripping the flag.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -87,8 +93,18 @@ export class SessionReplay implements AmplitudeSessionReplay {
   private lastTargetingParams?: SessionReplayTargetingInput;
   private lastShouldRecordDecision?: boolean;
 
-  // Visible for testing only
+  // Visible for testing only — the next three fields are intentionally public so
```
lewgordon-amplitude: Nit: this comment overstates the testing-only framing. `pageLeaveFns` is iterated in `pageLeaveListener` (line 509) and `rrwebEventManager` is dereferenced in `sendEvents` for `dropPendingBeaconEvents` (line 624). Both have load-bearing internal callers — making them private later (per the comment's literal reading) would break the gate. Consider rewording to something like "kept public because they're referenced by internal callers and tests", or narrowing the comment to apply only to `sessionStartTime`.

lewgordon-amplitude (approved):
LGTM!

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +585,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+      if (this.config?.minSessionDurationMs !== undefined && this.sessionStartTime !== undefined) {
+        const sessionDuration = Date.now() - this.sessionStartTime;
+        const willSend = sessionDuration >= this.config.minSessionDurationMs;
+        if (!willSend) {
+          this.loggerProvider.log(
+            `Session ${sessionIdToSend} not sent: duration ${sessionDuration}ms is below minimum ${this.config.minSessionDurationMs}ms.`,
+          );
+          // Drop pending beacon events so this below-threshold session's data
+          // can't leak into a later session's beacon flush on page unload.
+          this.rrwebEventManager?.dropPendingBeaconEvents();
```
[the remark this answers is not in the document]
bravecod: Good catch — addressed in commit cbced118. Moved `dropPendingBeaconEvents()` out of `sendEvents` (which fires from blur/visibility-change too) and into `asyncSetSessionId`, before the new identifiers and start time are installed. Cross-session leak is still prevented (the only path where it could happen), and within-session below-threshold sends now preserve the buffer so legitimately-sendable events can still ship via beacon if the session later crosses the threshold. Page-leave path also independently gates. Test flipped to assert the buffer is *preserved* on within-session below-threshold sends; new test asserts the drop fires on session transition.

packages/session-replay-browser/src/session-replay.ts
bravecod: Good catch — addressed in commit 39f64c49. Added the `previousSessionId !== sessionId` guard exactly where suggested. Standalone callers passing the current sessionId redundantly now keep their entry intact rather than restarting the gate clock on next reload. Regression test added.

packages/session-replay-browser/src/session-replay.ts
bravecod: Good catch — fixed in commit 2e2aaa85. Extended the precondition on the gate-decision emit-and-flip to also require `this.recordCancelCallback && this.recordFunction`, so a blur-listener-fired sendEvents on a reloaded long-lived session no longer trips the flag while addCustomRRWebEvent is silently no-opping. Added a regression test covering exactly that ordering: no-op when recording is inactive, then real emit on the next send after recording activates.

packages/session-replay-browser/src/session-replay.ts
bravecod: Fair — addressed in commit 4662f6b1. Reworded to "Public on purpose" with the load-bearing internal callers named (`pageLeaveListener` iterates `pageLeaveFns`, `asyncSetSessionId` dereferences `rrwebEventManager`, `isBelowMinSessionDuration()` reads `sessionStartTime`). Privatizing them would break the gate, not just tests.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -544,10 +585,22 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
bravecod: This was already addressed before the commit you reviewed — `isBelowMinSessionDuration()` was extracted in commit ee8d1496 (now on the latest branch state, before commit 22155eec). All three sites (the `shouldSend` callback, the `pageLeaveFns` beacon path, and `sendEvents`) call the helper. The bot appears to have read the file before the refactor; the latest branch has a single source of truth.

packages/session-replay-browser/src/session-replay.ts
```diff
@@ -356,11 +411,34 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+    // re-suppress already-eligible sessions.
+    if (this.config?.apiKey) {
+      setReplayStartTime(this.config.apiKey, sessionId, this.sessionStartTime, this.loggerProvider);
+      // Only clear the prior entry when the id actually changed. A caller that passes the
+      // current sessionId redundantly would otherwise have its just-written entry deleted
+      // here, restarting the gate clock from Date.now() on the next page reload.
+      if (previousSessionId !== undefined && previousSessionId !== sessionId) {
+        removeReplayStartTime(this.config.apiKey, previousSessionId, this.loggerProvider);
+      }
+    }
```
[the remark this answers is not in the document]
bravecod: Right — fixed properly this time in commit 660aa6bc. The prior guard only protected `removeReplayStartTime`; the rest of the block (`sessionStartTime = Date.now()`, the two counter resets, `setReplayStartTime` overwrite, and the beacon drop) all still executed unconditionally. Wrapped them in a single `isSessionChange` guard so a redundant `setSessionId(currentId)` is now a true no-op for the gate.

Also strengthened the test that should have caught this: it now pins a known start time before the call and asserts the value survives (the old version just checked non-null, which would pass even after a full overwrite).

PR #1729, opened 2026-05-05
feat(session-replay): add fetch keepalive to reduce 499 errors on page navigation

lewgordon-amplitude:
## Summary

- Adds `keepalive: true` to all session replay `fetch` calls (main-thread and worker paths) so the browser keeps requests alive through page navigation, directly preventing 499 (Client Closed Request) errors
- Only enables keepalive when payload ≤ 64 KB to respect the browser's shared keepalive budget; larger payloads fall back to existing retry logic
- Adds `MAX_KEEPALIVE_BYTES = 64 * 1024` constant and two new unit tests covering the size-threshold behaviour

Closes [SR-4123](https://linear.app/amplitude/issue/SR-4123/featsession-replay-add-fetch-keepalive-to-reduce-499-errors-on-page)

## Context

499 is NGINX's status for "client closed request". It fires when the browser cancels an in-flight fetch during navigation or tab close. Retry logic cannot recover from this because the page context is gone. The Fetch `keepalive` option ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#keepalive)) is the web-standard solution — same guarantee as `sendBeacon` but with full HTTP control. The existing `sendBeacon`-on-pagehide path remains unchanged; `keepalive` fills the gap for requests already in-flight at the time of navigation.

## Test plan

- [ ] Existing unit tests pass (`npx jest --testPathPattern="track-destination"`)
- [ ] New `keepalive` describe block covers: small payload → `keepalive: true`, oversized payload → `keepalive: false`
- [ ] Manual: open a page with session replay, trigger a recording, navigate away mid-flush — verify no 499s in network tab

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes network request behavior for all session-replay uploads (main thread and worker), which could affect delivery reliability in browsers with strict keepalive limits.
> 
> **Overview**
> **Session replay upload requests now use Fetch `keepalive` when possible** to reduce 499 (client-closed) errors during page navigation.
> 
> Adds `MAX_KEEPALIVE_BYTES` (64 KB) and sets `keepalive: true` only when the (gzipped or raw) payload is at/below this threshold in both the main-thread and worker `fetch` paths, with updated unit tests to assert the new `keepalive` behavior and size cutoff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbe511852210768a1c4e8bcae5cc1fd71b937a58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/src/constants.ts
+++ b/packages/session-replay-browser/src/constants.ts
@@ -28,6 +28,7 @@ export const MAX_IDB_STORAGE_LENGTH = 1000 * 60 * 60 * 24 * 3; // 3 days
 export const KB_SIZE = 1024;
 export const MAX_URL_LENGTH = 1000;
 export const RETRY_TIMEOUT_MS = 1000;
+export const MAX_KEEPALIVE_BYTES = 64 * 1024; // browser keepalive budget shared with sendBeacon
 
 export enum CustomRRwebEvent {
   GET_SR_PROPS = 'get-sr-props',
--- a/packages/session-replay-browser/src/track-destination.ts
+++ b/packages/session-replay-browser/src/track-destination.ts
@@ -13,7 +13,7 @@ import {
   SessionReplayDestinationContext,
 } from './typings/session-replay';
 import { VERSION } from './version';
-import { MAX_URL_LENGTH, KB_SIZE } from './constants';
+import { MAX_URL_LENGTH, KB_SIZE, MAX_KEEPALIVE_BYTES } from './constants';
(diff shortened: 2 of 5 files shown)


lewgordon-amplitude:
bugbot run

stevenchien-amplitude (approved):
nice! lgtm

PR #1739, opened 2026-05-07
fix(session-replay-browser): wait for remote config to avoid stale cache race (SR-4234)

lewgordon-amplitude:
## Summary

- joined-config subscribed with `'all'` delivery mode, which races a synchronous localStorage cache read against the network fetch. Cache always wins (it's synchronous), so a **stale cache silently overrides the live config** — sessions get recorded against the previous TRC URL rule even when the network is healthy and would have returned the correct config.
- Switch to `{ timeout: 1500 }` delivery mode so the SDK prefers the remote response and only falls back to cache after the budget elapses.
- Added an e2e spec that reproduces the live-network and blocked-network scenarios, plus a unit test that pins the new delivery mode so a revert to `'all'` would fail loudly.

Linear: [SR-4234](https://linear.app/amplitude/issue/SR-4234/sdk-trc-stale-localstorage-cache-wins-race-in-joined-config-recording)

## Customer impact

Qonto renewal blocked. Sessions captured on URLs the customer's TRC URL rule excludes. Any customer who edits their TRC config has at least one session per active user captured against the previous config; customers behind firewalls / ad blockers / CSPs that block `sr-client-cfg` have stale cache persist indefinitely.

## Why 1500ms

`subscribeWaitForRemote` in analytics-core uses an inner per-attempt fetch abort of 1000ms with up to 3 retries, but the outer timeout is the only thing the SR plugin actually waits on — so only the first fetch attempt can complete in time. 1500ms is set 500ms above the inner abort to avoid a tie at 1000ms and give the first attempt's resolution path room to finish. Anything past that falls through to cache. Going to 3000ms+ to fit a second attempt (~2000-2333ms) would block SR initialization too long for the marginal benefit.

## Behavior matrix after the fix

| Scenario | Behavior |
|---|---|
| Healthy network, cache stale | Uses fresh remote config (bug fixed) |
| Network fails fast (firewall/adblocker) | Fetch returns null → joined-config rejects → **fails closed**, no recording |
| Slow network (>1500ms but eventually returns) | Outer timeout fires → falls back to cached config (potentially stale) |
| No cache, healthy network | Uses fresh remote config |
| No cache, network fails | Fails closed, no recording |

The fail-closed behavior on fast network failure is stronger than originally framed in the ticket — strictly better than recording against a stale config. The remaining residual gap (slow-network fallback to potentially stale cache) is what the defense-in-depth follow-ups (cache TTL, fail-closed targeting default) close. The e2e blocked-network test asserts the fail-closed behavior; the local-build live-network test asserts the SDK uses the fresh remote config.

## Behavior change to call out

`generateJoinedConfig` is awaited inside both `_init` (cold start) and `asyncSetSessionId` (warm session rotation). Previously the synchronous-cache path made these resolve in <1ms; now they wait up to 1500ms when the network is degraded. This is the correctness/speed tradeoff the fix requires. A separate analytics-core change to honor `lastSuccessfulFetch`-style throttling inside `getOrCreateFetchPromise` could short-circuit warm calls when a recent fetch succeeded — that's a cross-package change worth its own review.

## Pre-existing analytics-core issue worth flagging separately

`subscribeWaitForRemote`'s success path calls `setConfig(result)` even when `result.remoteConfig === null`, which means a fast network failure now wipes the customer's localStorage cache. Pre-existing in analytics-core (not introduced here), but more visible after this fix because we route through `subscribeWaitForRemote` instead of `subscribeAll`. Worth its own ticket if SR wants to preserve cache through transient failures.

## Test plan

- [x] Unit tests: 881/881 passing in `@amplitude/session-replay-browser` with 100% coverage maintained
- [x] Unit tests: 793/793 passing in `@amplitude/analytics-core` (no behavior change there, sanity check)
- [x] Lint clean
- [x] Pre-commit hooks pass
- [x] e2e (`trc-url-rule.spec.ts`): live-network stale-cache scenario no longer records on non-matching URL; blocked-network scenario fails closed

---

> [!NOTE]
> **Medium Risk**
> Changes Session Replay initialization to wait up to 1500ms for remote config and fail closed on fast fetch failures, which can alter capture behavior and add startup latency in degraded networks.
> 
> **Overview**
> Fixes a race where a synchronous localStorage remote-config cache could override a fresh network response, causing Session Replay to record against stale TRC URL rules.
> 
> `generateJoinedConfig` now subscribes to remote config with a `{ timeout: 1500 }` delivery mode (instead of `'all'`), preferring the remote response and only falling back to cache after the timeout; on null/failed remote config it disables capture. Adds a Playwright E2E suite covering URL-rule enforcement and stale-cache scenarios (live network + blocked network), plus a unit test that pins the new subscription mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6e3694c3b95c774812b411822fd0a76294e9113. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

jxiwang (approved):

PR #1741, opened 2026-05-08
fix(session-replay-browser): honor X-Session-Replay-Event-Skipped server back-pressure (SR-4280)

lewgordon-amplitude:
## Summary

The Amplitude session-replay server returns `HTTP 200` plus the header `X-Session-Replay-Event-Skipped: <error_code>` for deliberate \"no-retry\" drops (throttle, capture disabled, session out of valid range). The 200-with-header pattern is intentional — it suppresses retry storms — but the SDK never actually read the header. Concrete impact today: a single app sheds ~7.3% of its traffic into the throttle path with no client-side adaptation (~163K req/day from one customer alone).

This PR teaches the SDK to read the header and adapt:

- `429` (throttled) → pause the flush schedule for **60 seconds**.
- `4004` (session_in_invalid_range) or `4005` (capture_disabled) → **hard-kill** that session: drop currently-queued contexts and short-circuit any future adds for the same `sessionId`.
- Clean `200` (no skip header) → clear any throttle pause.

The fix is applied in both fetch paths — main thread (`src/track-destination.ts`) and worker (`src/worker/track-destination.ts`). The worker forwards the parsed `skipCode` back via the existing `complete` message so the directive is applied on the main thread, where the queue/scheduler live.

The killed-session set is bounded at 256 entries with insertion-order eviction so a long-lived SDK instance can't accumulate kill records indefinitely.

Linear: https://linear.app/amplitude/issue/SR-4280

## Test plan

- [x] Unit tests covering each header-value path (`throttled`, `capture_disabled`, `session_in_invalid_range`, clean 200, unknown code, non-2xx) — added in `test/track-destination.test.ts` and `test/worker/track-destination.test.ts`.
- [x] End-to-end lifecycle test: throttle → schedule honors pause → advance time past pause → clean 200 → schedule resumes normal cadence.
- [x] Test that an in-flight queue snapshot (already past `flush()`'s queue grab) is short-circuited by `send()` once a kill arrives mid-batch — only 1 fetch fires for 3 batches.
- [x] Test that the throttle log fires once per pause-state transition, not per response.
- [x] Test that `killedSessions` is bounded under stress (1000 kills → size capped, oldest evicted, newest retained).
- [x] All 904 existing tests pass; lint clean; `pnpm build` succeeds.
- [ ] Post-merge: confirm in Datadog that for sessions where the SDK has seen a throttled 200, subsequent request rate from that session falls.
- [ ] Post-merge: end-to-end smoke against staging — configure aggressive throttle, send replay events with patched SDK, confirm flush cadence slows and `X-Session-Replay-Event-Skipped: 429` rate drops over the session.

---

> [!NOTE]
> **Medium Risk**
> Changes session replay delivery control flow (throttle pausing and per-session kill switch), which can alter flush cadence and drop behavior if misinterpreted; mitigated by extensive unit test coverage across main-thread and worker paths.
> 
> **Overview**
> Adds support for the server’s `X-Session-Replay-Event-Skipped` 200-response header so the browser SDK can honor back-pressure instead of retrying blindly.
> 
> On a `429` skip code, the destination now **pauses the flush schedule for 60s**; on `4004`/`4005`, it **hard-stops capture for that session** by draining queued batches and short-circuiting future enqueues/sends (with a bounded killed-session cache). The worker send path is updated to parse the header and forward `skipCode` via the `complete` message so directives are applied on the main thread, with new unit tests covering throttle, kill, scheduling, idempotency, and non-2xx behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffc63f24f41114983c6b47e96803feac5508a008. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/e2e/back-pressure.spec.ts
+++ b/packages/session-replay-browser/e2e/back-pressure.spec.ts
@@ -41,68 +41,13 @@ async function flushSessionReplay(page: import('@playwright/test').Page) {
 }
 
 test.describe('server back-pressure (X-Session-Replay-Event-Skipped header)', () => {
-  test('hard-kill (4005 capture_disabled): SDK stops POSTing for the killed session', async ({ page }) => {
-    const sentUrls: string[] = [];
-    await mockRemoteConfig(page, remoteConfigRecording);
-    await mockTrackApiWithSkipHeader(page, {
-      skipCode: '4005',
-      onRequest: (route) => sentUrls.push(route.request().url()),
-    });
-
-    await page.goto(
-      buildUrl('/session-replay-browser/sr-capture-test.html', {
-        sessionId: TEST_SESSION_ID,
-      }),
-    );
-    await waitForReady(page);
-    await page.waitForTimeout(SNAPSHOT_SETTLE_MS);
-
-    // First flush — server returns 200 + 4005 → kill switch fires.
-    await flushSessionReplay(page);
-    const sentBeforeKill = sentUrls.length;
-    expect(sentBeforeKill).toBeGreaterThan(0);
-
-    // Generate more activity and flush again. The kill should drop these batches without
-    // issuing another POST for the same session.
-    await page.evaluate(() => {
-      // Click the test button to generate a fresh rrweb event
-      document.getElementById('test-button')?.click();
-      document.getElementById('test-input')?.focus();
-    });
-    await flushSessionReplay(page);
-    await flushSessionReplay(page);
-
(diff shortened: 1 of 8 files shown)


lewgordon-amplitude:
bugbot run

stevenchien-amplitude (approved):
awesome! lgtm

PR #1742, opened 2026-05-08
fix(session-replay-browser): drop empty batches at the store layer (SR-4284)

lewgordon-amplitude:
## Summary

Fixes [SR-4284](https://linear.app/amplitude/issue/SR-4284). `SessionReplayServletV2` is rejecting ~416 POSTs/24h with `Empty request body` (HTTP 400) across the fleet. The SDK is `fetch()`-ing with a body containing zero events. Volume is small but it spans many SDK versions.

## Root cause

`shouldSplitEventsList` returns `true` via the size-constraint branch whenever a single incoming event exceeds `MAX_EVENT_LIST_SIZE` (700 KB), **regardless of the current buffer length**. When `addEventToCurrentSequence` runs that branch with an empty buffer, the split path finalizes a zero-event sequence into `sequencesToSend`. That row is later flushed and POSTed as an empty body.

## Changes

Store-layer hardening for both `InMemoryEventsStore` and `SessionReplayEventsIDBStore`:

- **`addEventToCurrentSequence` split path** — skips finalize when buffer is empty. *Primary root-cause fix.*
- **`storeCurrentSequence`** — returns `undefined` when nothing's buffered (memory store was missing the length check; idb store already guarded but now logs).
- **`getSequencesToSend`** — filters out `events: []` rows. Catches stale persisted records from older SDK builds during IDB drain on init.
- **Sampled (1-in-100) `loggerProvider.warn`** at every filter site — so post-deploy Datadog can confirm the guards are catching real cases.

Defensive `if (!events.length) return` in `track-destination.ts` immediately before `fetch(...)` — cheap belt-and-braces against any future regression.

## Test plan

- [x] `pnpm test` — 913 passing, 100% coverage maintained
- [x] `pnpm lint` — clean (only pre-existing warnings)
- [x] Tests cover all four filter sites (split-with-empty, storeCurrentSequence-on-empty, getSequencesToSend filter, pre-fetch guard) for both store backends
- [ ] After merge + rollout: re-query Datadog `service:session-replay-uploader "Empty request body"` and confirm volume drops materially

---

> [!NOTE]
> **Medium Risk**
> Touches session replay persistence and send paths (IDB/memory stores and final `fetch`), so regressions could affect delivery or flushing behavior, though changes are narrowly scoped to filtering empty sequences with added tests.
> 
> **Overview**
> Prevents session replay from ever persisting or POSTing **empty event batches** (fix for SR-4284).
> 
> Adds store-layer guards in both `SessionReplayEventsIDBStore` and `InMemoryEventsStore` to avoid finalizing `events: []` on split/flush, and to filter+prune any previously persisted empty sequences during `getSequencesToSend` (with sampled debug logging for verification).
> 
> Adds a final pre-`fetch` defensive check in `track-destination.ts` to skip sending if the payload’s `events` array is empty, and expands unit + e2e coverage to reproduce the drained-slot/oversize split scenario and assert no empty-body requests are made.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7047131056fe2db5917bf92a818fdda8997fec58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/e2e/size-limits.spec.ts
+++ b/packages/session-replay-browser/e2e/size-limits.spec.ts
@@ -9,7 +9,7 @@ import {
   waitForReady,
   readRouteBody,
 } from './helpers';
-import { MAX_SINGLE_EVENT_SIZE } from '../src/constants';
+import { MAX_EVENT_LIST_SIZE, MAX_SINGLE_EVENT_SIZE } from '../src/constants';
 
 // Just over the byte cap so the next full snapshot is guaranteed oversized.
 // Derived from the constant so the test stays in sync if the threshold is bumped.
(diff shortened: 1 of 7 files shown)


lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -269,17 +280,25 @@ export class SessionReplayEventsIDBStore extends BaseEventsStore<number> {
...
-          sequenceId: cursor.key,
-          sessionId,
-        });
+        // Skip empty persisted records. These can come from older SDK builds that
+        // wrote zero-event sequences via addEventToCurrentSequence's split path
+        // (when a single oversized event triggered a split with empty buffer);
+        // flushing them produces empty-body POSTs the server rejects with 400.
+        if (events.length === 0) {
+          this.maybeWarnEmptyFiltered('getSequencesToSend');
+        } else {
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in 3397ed2b. Switched the `getSequencesToSend` cursor's transaction to `readwrite` and call `await cursor.delete()` on empty rows so they're pruned in-place rather than just filtered. By construction (the SR-4284 store-layer fix) we never write empty rows anymore, so any empty row encountered is unambiguously stale and safe to delete.

Test now asserts both the prune (via raw-DB `getAll` inspection) and the no-second-warn behaviour (a second `getSequencesToSend` on the cleaned store fires no warn).

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/events/events-memory-store.ts
```diff
@@ -18,16 +19,37 @@ export class InMemoryEventsStore extends BaseEventsStore<number> {
...
-      sequenceId: Number(sequenceId),
-      sessionId,
-      events,
-    }));
+    const result: SendingSequencesReturn<number>[] = [];
+    for (const [sequenceId, { sessionId, events }] of Object.entries(this.finalizedSequences)) {
+      if (events.length === 0) {
+        this.maybeWarnEmptyFiltered('getSequencesToSend');
+        continue;
+      }
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in d658f55c. Applied the same prune-in-place pattern to `InMemoryEventsStore.getSequencesToSend` — `delete this.finalizedSequences[Number(sequenceId)]` inside the `events.length === 0` branch. Tests now assert no second warn on a follow-up call. Memory and IDB store behaviour stay consistent.

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -384,6 +384,15 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
       };
 
       const serverUrl = `${getServerUrl(context.serverZone, this.trackServerUrl)}?${urlParams.toString()}`;
+      // Final defensive guard: never POST a zero-event payload. Upper layers (events-manager
+      // oversize filter, send()'s post-batcher check, store-layer filters) should already
+      // have caught this — but SR-4284 fleet logs show ~416 empty-body 400s/24h slipping
+      // through somehow, so a cheap belt-and-braces check immediately before fetch prevents
+      // any future regression from re-introducing the same server rejection.
+      if (payload.events.length === 0) {
```
jpollock-ampl, line 392: Could we mirror the sampled `loggerProvider.warn` from the store-layer guards here? Right now if this last-line-of-defense ever fires in production, there's no signal it caught something - we'd just see fewer 400s with no way to attribute the saved POST to this guard vs. the upstream filters. Given the PR's stated goal of using the warns to confirm guards are catching real cases, mirroring it here would close the loop.

packages/session-replay-browser/src/events/events-idb-store.ts
```diff
@@ -333,8 +360,10 @@ export class SessionReplayEventsIDBStore extends BaseEventsStore<number> {
         return undefined;
       }
 
-      // Skip empty sequences — no point writing a zero-event row to sequencesToSend.
+      // Skip empty sequences — no point writing a zero-event row to sequencesToSend
+      // (would later POST as an empty body and 400 on the server).
       if (currentSequenceData.events.length === 0) {
```
jpollock-ampl, line 366: Nit/non-blocking - Just to make sure I understand - is `storeCurrentSequence` ever called on a quiet session where the current-sequence row legitimately has `events: []` (e.g. a flush cycle with no rrweb activity in the window)? If so, the sampled warn here would fire on perfectly normal behavior, which would dilute the post-deploy Datadog signal. The `addEventToCurrentSequence` and `getSequencesToSend` warns feel unambiguous; this one might not be.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -384,6 +384,15 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
       };
 
       const serverUrl = `${getServerUrl(context.serverZone, this.trackServerUrl)}?${urlParams.toString()}`;
```
jpollock-ampl, line 386: Nit/non-blocking - Tiny - worth flipping the order so the empty-payload guard runs before we compute `serverUrl`? On the empty path that work is wasted today.

packages/session-replay-browser/e2e/size-limits.spec.ts
```diff
@@ -202,3 +202,104 @@ test.describe('WAF 413 bisect-retry', () => {
...
+      expect(events.length).toBeGreaterThan(0);
+    }
+
+    // Assertion 3: the mid-size event still reached the wire. Without the fix, the
+    // empty row would have been written first and the real event held in the slot;
+    // with the fix, the real event still gets delivered on the next flush.
+    const allEvents = rawBodies.flatMap(decodeEvents);
+    const hasMidSizeEvent = rawBodies.some((b) => b.length > MAX_EVENT_LIST_SIZE);
+    expect(hasMidSizeEvent).toBe(true);
+    expect(allEvents.length).toBeGreaterThan(0);
```
jpollock-ampl, line 305: Nit/non-blocking - If I'm reading this right, this assertion is implied by the loop above (each body has `events.length > 0` and we already asserted `rawBodies.length > 0`), so the flatMap can never be empty. Could drop it, or swap it for something stronger like asserting the mid-sized event's payload made it through (e.g. inspecting `allEvents` for the marker attribute).

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/e2e/size-limits.spec.ts
```diff
@@ -202,3 +202,106 @@ test.describe('WAF 413 bisect-retry', () => {
     expect(callCount).toBeLessThanOrEqual(2);
   });
 });
+
+// ─── Empty-batch leak (SR-4284) ───────────────────────────────────────────────
+
+test.describe('empty batch leak (SR-4284)', () => {
```
jpollock-ampl, line 208: Nit - remove ticket number?

jpollock-ampl (approved):

PR #1743, opened 2026-05-08
fix(session-replay-browser): merge queued sends after throttle pause (SR-4286)

lewgordon-amplitude:
## Summary

When the server throttles us via `X-Session-Replay-Event-Skipped: 429`, the SDK pauses its flush schedule for 60s (`THROTTLED_FLUSH_PAUSE_MS`). During the pause, completed sequences pile up in `trackDestination.queue` — and on release we send each one as its **own POST**. A 60s pause that accumulates 6 sequences releases as 6 sequential POSTs back-to-back, which is exactly the behavior the request-count-based throttle is asking us to avoid.

This change merges same-`(sessionId, deviceId, apiKey, type, serverZone, sampleRate, version)` queued contexts into one POST when the deferred flush fires:

- Group queued contexts by destination identity (anything that affects URL/headers/routing splits the group).
- Greedily concat events with a soft char-length cap (`MERGE_AFTER_THROTTLE_SOFT_CAP = 2 * MAX_EVENT_LIST_SIZE = 1.4MB`), well under the 10MB-compressed 413 ceiling.
- Merged context's `onComplete` fans out so every source IDB sequence is cleaned up.
- Reset `attempts: 0` on the merged context so post-throttle recovery gets a full retry budget instead of collapsing N source budgets into one.
- Gated on `mergeOnNextFlush`, set only when `schedule()` deferred a flush due to the pause — steady-state behavior is unchanged.
- Merge log fires once per pause window (cleared when `applyServerDirective` sees a clean 200), mirroring the throttle log's transition-only gating.

The existing 413 split-and-retry path still works on a merged context; the source-cleanup amplification (page-close mid-retry can lose up to N sequences instead of 1) is documented in code — exceedingly rare given the cap vs ceiling headroom.

Linear: [SR-4286](https://linear.app/amplitude/issue/SR-4286/session-replay-browser-combine-queued-sends-after-throttle-pause)

## Test plan

- [x] All 920 existing + new tests pass with 100% statements/branches/functions/lines coverage
- [x] New tests cover: schedule sets/clears \`mergeOnNextFlush\` based on pause state; merge collapses same-session contexts and fans out \`onComplete\`; different sessions stay separate; soft cap enforced; steady-state untouched; \`attempts: 0\` reset; optional-field grouping fallbacks; version-based grouping; e2e throttle→pause→enqueue→release→merged POST; worker-path complete fans out; kill-mid-merge drains correctly; 413-on-merged uses noop halves; merge log gates per pause cycle
- [x] Lint clean
- [ ] CI green
- [ ] Bugbot review

---

> [!NOTE]
> **Medium Risk**
> Touches session replay delivery/back-pressure behavior by changing how queued batches are flushed after a throttle, which could affect request volume and cleanup semantics despite being gated to throttle-pause scenarios and covered by extensive tests.
> 
> **Overview**
> **Reduces request spikes after server throttling.** When a throttle pause is active, the next `flush()` now merges queued session replay batches that share the same destination identity into fewer POSTs (with a soft size cap and a fanned-out `onComplete`), instead of sending each queued sequence as its own request.
> 
> Adds `MERGE_AFTER_THROTTLE_SOFT_CAP` and merge/log gating in `track-destination`, plus comprehensive unit coverage and a new Playwright e2e that simulates `X-Session-Replay-Event-Skipped: 429` and asserts queued sequences collapse into a single fetch POST on release. Playwright config also allows overriding `baseURL` via `PLAYWRIGHT_BASE_URL`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 333d5b1b2f9455054767edba1165fe23c6f07ce7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/src/track-destination.ts
+++ b/packages/session-replay-browser/src/track-destination.ts
@@ -24,6 +24,7 @@ import {
   EVENT_SKIP_CODE_INVALID_RANGE,
   EVENT_SKIP_CODE_CAPTURE_DISABLED,
   THROTTLED_FLUSH_PAUSE_MS,
+  MERGE_AFTER_THROTTLE_SOFT_CAP,
 } from './constants';
 import { gzipJson } from './utils/gzip';
 
@@ -71,6 +72,10 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
   // The server uses this header (instead of 4xx) to signal a deliberate no-retry drop so SDKs
   // don't retry-storm. We honor it here by slowing or stopping our flush schedule.
   private flushPauseUntilMs = 0;
+  // Set when schedule() defers a flush because we're inside a throttle pause; consumed by
+  // flush() to merge same-session contexts before sending. Throttling is enforced by request
+  // count, so collapsing N queued batches into one POST directly reduces throttle pressure.
+  private mergeOnNextFlush = false;
   private killedSessions = new Set<string | number>();
 
   constructor({
@@ -248,7 +253,13 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
     // If the server signaled throttling on a recent 200, defer the next flush until the
     // pause window ends. This lets us keep batching events without retry-storming the server.
     const pauseRemaining = this.flushPauseUntilMs - Date.now();
+    const isPaused = pauseRemaining > 0;
     const effectiveTimeout = pauseRemaining > timeout ? pauseRemaining : timeout;
+    if (isPaused) {
+      // Mark the upcoming flush for merge: contexts piling up during the pause should
+      // be coalesced into one POST per (session, device, api, type, ...) group.
+      this.mergeOnNextFlush = true;
+    }
     this.scheduled = setTimeout(() => {
       void this.flush(true).then(() => {
         if (this.queue.length > 0) {
@@ -259,7 +270,7 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
   }
 
   async flush(useRetry = false) {
-    const list = this.queue;
+    let list = this.queue;
     this.queue = [];
 
     if (this.scheduled) {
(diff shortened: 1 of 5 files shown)


lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,110 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+        if (current === null) {
+          // Reset attempts to 0 on the merged context so the post-throttle delivery gets a
+          // full retry budget. The throttle pause has already absorbed back-pressure; the
+          // alternative (Math.max of source attempts) would collapse N source budgets into
+          // one and end-of-life all N IDB records on a single retry exhaustion.
+          current = { ...ctx, events: [...ctx.events], attempts: 0 };
+          currentChars = ctxChars;
+          continue;
+        }
+        if (currentChars + ctxChars > MERGE_AFTER_THROTTLE_SOFT_CAP) {
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in 333d5b1b. Switched to UTF-8 byte counting via `new Blob([e]).size` to match how `base-events-store.ts` enforces `MAX_EVENT_LIST_SIZE` upstream. Updated the constant comment to reflect the bytes-based comparison.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,110 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+        }
+        const prevOnComplete = current.onComplete;
+        const ctxOnComplete = ctx.onComplete;
+        current.events = current.events.concat(ctx.events);
+        currentChars += ctxChars;
+        current.onComplete = async () => {
+          // Run both in parallel; an underlying store cleanup failure in one shouldn't
+          // block the other. Errors are surfaced per the source callbacks' own handlers.
+          await Promise.all([prevOnComplete(), ctxOnComplete()]);
+        };
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in 333d5b1b. Switched to `Promise.allSettled` so a single source onComplete rejection doesn't surface as an unhandled rejection (the merged onComplete is invoked fire-and-forget) or block sibling cleanups. Added a regression test asserting one rejecting onComplete doesn't block the other and the merged onComplete itself resolves.

lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,115 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
     }
   }
 
+  /**
+   * Coalesces queued contexts that share the same destination identity so the post-throttle
+   * release sends fewer requests. Identity covers everything that affects the request URL,
+   * routing, or per-request semantics — splitting on any difference keeps each merged POST
+   * indistinguishable from the source contexts it replaced.
+   *
+   * Greedy concat with a soft char-length cap (`MERGE_AFTER_THROTTLE_SOFT_CAP`) keeps merged
```
jpollock-ampl, line 300: Nit/non-blocking - Worth updating "soft char-length cap" to "soft UTF-8 byte cap" here. After the bugbot fix in 333d5b1b the comparison is now measured in bytes (via `new Blob([e]).size`), so the function-level doc is out of sync with the constant comment and the implementation.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -465,7 +575,14 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
       `Session replay event batch rejected by ${source} (${context.events.length} events, ${totalSizeKB} KB total) — splitting and retrying`,
     );
 
-    // Clean up the original IDB record, then re-enqueue both halves as new in-memory batches
+    // Clean up the original IDB record, then re-enqueue both halves as new in-memory batches.
+    // For a merged-on-throttle context (mergeQueueAfterThrottle), this onComplete is the
+    // fanned-out callback covering N source IDB records — they'll all be cleaned up here.
+    // Halves get noop onCompletes, so a page-close between this cleanup and a half delivery
+    // means up to N source sequences are lost. The merge soft cap (1.4MB chars) is well under
```
jpollock-ampl, line 582: Nit/non-blocking - Same stale unit here: "1.4MB chars" should read "1.4MB bytes" now that the cap is enforced against UTF-8 byte size.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,115 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+          await Promise.allSettled([prevOnComplete(), ctxOnComplete()]);
+        };
+      }
+      flushCurrent();
+    }
+
+    if (merged.length < list.length && !this.mergeLogFiredThisPause) {
+      this.mergeLogFiredThisPause = true;
+      this.loggerProvider.log(
+        `Session replay throttle pause ended; merged ${list.length} queued batches into ${merged.length} request(s)`,
```
jpollock-ampl, line 379: Nit/non-blocking - Quick question on the "throttle pause ended" wording. I believe `mergeQueueAfterThrottle` also runs on a manual `flush(false)` invoked while the pause is still active (the e2e test does exactly this: flips throttle off, calls `flush(false)` before `flushPauseUntilMs` clears). In that path the log reads as if the pause has ended when we're really flushing through it. WDYT about something like "merging queued batches due to throttle pause" to cover both the timer-fired and manual-flush paths?

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,115 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+        }
+        if (currentBytes + ctxBytes > MERGE_AFTER_THROTTLE_SOFT_CAP) {
+          flushCurrent();
+          current = { ...ctx, events: [...ctx.events], attempts: 0 };
+          currentBytes = ctxBytes;
+          continue;
+        }
+        const prevOnComplete = current.onComplete;
+        const ctxOnComplete = ctx.onComplete;
+        current.events = current.events.concat(ctx.events);
```
jpollock-ampl, line 363: Nit/non-blocking - Have we considered `current.events.push(...ctx.events)` here? `concat` allocates a new array each step, so a group of N batches does O(N^2) array work where in-place push is O(N). `current` is already a fresh object we own (the `{ ...ctx, events: [...ctx.events], attempts: 0 }` copy), so mutating in place is safe. Not a real perf concern at typical N, just simpler.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -259,19 +273,115 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+        currentBytes += ctxBytes;
+        current.onComplete = async () => {
+          // allSettled (not all): an underlying store cleanup failure in one shouldn't
+          // block the other, and the merged onComplete is invoked fire-and-forget via
+          // `void context.onComplete()` — a rejection from `Promise.all` would surface
+          // as an unhandled rejection. Errors stay encapsulated in the source callbacks.
+          await Promise.allSettled([prevOnComplete(), ctxOnComplete()]);
```
jpollock-ampl, line 370: Nit/non-blocking - Help me understand the wrapping pattern here. Because each merge step wraps the previous `current.onComplete`, awaiting the final merged onComplete walks an N-deep nested `Promise.allSettled` tree (with each level invoking its inner wrapper). Functionally fine, but it'd be a touch cleaner to collect source callbacks into a `sourceOnCompletes: Array<() => Promise<void>>` on `current` and have a single onComplete fan out `Promise.allSettled(sourceOnCompletes.map(cb => cb()))`. Optional, the current shape works and tests cover it.

jpollock-ampl (approved):

PR #1076, opened 2025-05-09
feat(session-replay): add autoStart configuration option

aliaksandr-kazarez:
- Add autoStart parameter to SessionReplayConfig interface
- Update native modules (iOS and Android) to support autoStart option
- Add default value (true) for autoStart in getDefaultConfig
- Add test for default config
- Update dependencies in example/ios/Podfile.lock
- Convert session-replay-config from .tsx to .ts

AMP-129879

### Summary

add autoStart option to session replay react native plugin

### Checklist

* [x] Does your PR title have the correct [title format](https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions)?
* [x] Does your PR have a breaking change?:  no

PR #1771, opened 2026-05-22
feat(plugin-session-replay-react-native): pass maskLevel through to native SessionReplay

aliaksandr-kazarez:
## Summary

Wires a new `maskLevel` configuration field through the plugin's JS → ObjC → Swift and JS → Kotlin bridges so consumers of `@amplitude/plugin-session-replay-react-native` can request `Light` / `Medium` / `Conservative` masking from JS. Previously the plugin always defaulted to `MEDIUM` regardless of JS config because the bridge layers never forwarded the value to the native `SessionReplay` constructor.

Brings the plugin variant to parity with `@amplitude/session-replay-react-native`, which already wires this up correctly.

Tickets:
- Implements [SDKRN-11](https://linear.app/amplitude/issue/SDKRN-11/plugin-sdk-pass-masklevel-through-to-native-sessionreplay-js-ios)
- Parent [SDKRN-9](https://linear.app/amplitude/issue/SDKRN-9/add-conservative-masking-support-for-react-native-text) — see the [findings comment](https://linear.app/amplitude/issue/SDKRN-9#comment-f7884447) for full root cause

Bridge pattern template: [#1076](https://github.com/amplitude/Amplitude-TypeScript/pull/1076) (added `autoStart` the same way).

## Changes

All under `packages/plugin-session-replay-react-native/`:

| Layer | File | Change |
| --- | --- | --- |
| TS config | `src/session-replay-config.ts` | Add `MaskLevel` enum (`Light`/`Medium`/`Conservative`), `maskLevel?: MaskLevel` field, default `Medium` |
| TS index | `src/index.tsx` | Re-export `MaskLevel` |
| JS setup | `src/session-replay.ts` | Pass `maskLevel` as the 9th positional arg to `PluginSessionReplayReactNative.setup(...)` |
| ObjC bridge | `ios/PluginSessionReplayReactNative.mm` | Add `maskLevel:(NSString)maskLevel` to `RCT_EXTERN_METHOD` |
| Swift bridge | `ios/PluginSessionReplayReactNative.swift` | Extend `@objc(setup:...)` selector, add `maskLevel: String` parameter, pass `maskLevel: .fromString(maskLevel)` to `SessionReplay(...)`, add `MaskLevel.fromString(_:)` extension (copied verbatim from sibling standalone package) |
| Kotlin bridge | `android/.../PluginSessionReplayReactNativeModule.kt` | Add `MaskLevel` + `PrivacyConfig` imports, add `maskLevel: String` final parameter to `setup(...)`, map via `when` block, pass `privacyConfig = PrivacyConfig(maskLevel = mappedMaskLevel)` to `SessionReplay(...)` |
| Tests | `test/index.test.ts` | Default value, enum values, mocked-native-arg forwarding for all three masking levels + the undefined-defaults case |

Positional-arg contract (all four layers agree, source of truth is `src/session-replay.ts`):
1. `apiKey`
2. `deviceId`
3. `sessionId`
4. `serverZone`
5. `sampleRate`
6. `enableRemoteConfig`
7. `logLevel`
8. `autoStart`
9. `maskLevel` *(new)*

## Deviation from the original task packet

The task packet listed 7 implementation files. This PR also touches one additional file:

- `packages/plugin-session-replay-react-native/jest.config.js` — added `moduleFileExtensions: ['tsx', 'ts', 'js', 'jsx', 'json']` and added `<rootDir>/example/` to `testPathIgnorePatterns`.

Reason: the existing config only included `['ts', 'js', 'json']` (inherited from the base), which made it impossible for `test/index.test.ts` to import `SessionReplayPlugin` from `src/session-replay.ts` (that file transitively imports `./native-module.tsx`). The fix mirrors the sibling `packages/session-replay-react-native/jest.config.js`. The `example/` ignore preserves the prior baseline (the pre-existing `example/__tests__/App.test.tsx` was already broken at HEAD due to a missing `react-native-webview` install at the workspace root — not in scope to fix here).

No shipped behavior change from this config edit. No `example/`, README, or podspec touched.

## Test results

```
$ pnpm test
PASS @amplitude/plugin-session-replay-react-native test/index.test.ts
  Session Replay default config
    ✓ should have autostart default to true
    ✓ should have maskLevel default to Medium
  MaskLevel enum values
    ✓ exposes the three expected masking levels
  SessionReplayPlugin.setup forwards maskLevel to native
    ✓ forwards Conservative as the 9th positional argument
    ✓ forwards Light as the 9th positional argument
    ✓ forwards Medium as the 9th positional argument
    ✓ defaults to Medium when maskLevel is not provided
    ✓ passes prior positional arguments through unchanged

Test Suites: 1 passed, 1 total
Tests:       8 passed, 8 total
```

```
$ pnpm build
✔ Wrote files to lib/commonjs
✔ Wrote files to lib/module
✔ Wrote definition files to lib/typescript
```

```
$ pnpm lint
All matched files use Prettier code style!
```

Confirmed `lib/typescript/session-replay-config.d.ts` includes `export declare enum MaskLevel` and `maskLevel?: MaskLevel`, and `lib/typescript/index.d.ts` re-exports `MaskLevel`.

## iOS E2E coupling (NOT a blocker for merging)

iOS Conservative behavior cannot be fully E2E-verified until [SDKRN-10](https://linear.app/amplitude/issue/SDKRN-10) (the `session-replay-ios` native fix) ships and the XCFramework is vendored into `AmplitudeSessionReplay-iOS`. Android Conservative behavior IS verifiable today (Android native heuristic already works against `ReactTextView`). This PR can merge independently — full E2E lives under parent SDKRN-9.

## Test plan

- [ ] Reviewer: confirm the 9th positional arg ordering is consistent across `src/session-replay.ts`, `ios/...mm`, `ios/...swift`, and `android/.../...Module.kt`
- [ ] Reviewer: confirm `MaskLevel.fromString` Swift extension and Kotlin `when` mapping match the sibling standalone package
- [ ] Reviewer: confirm `getDefaultConfig().maskLevel === MaskLevel.Medium`
- [ ] CI: `pnpm test` and `pnpm build` pass

---

> [!NOTE]
> **Medium Risk**
> Changes privacy-sensitive masking defaults and native bridge contracts; incorrect resolution could under-mask PII in replays, though behavior is heavily tested and defaults remain `'medium'`.
> 
> **Overview**
> Adds **`privacyConfig.maskLevel`** (`'light' | 'medium' | 'conservative'`) so React Native Session Replay masking is configurable from JavaScript instead of always defaulting to medium at native init.
> 
> For **`@amplitude/plugin-session-replay-react-native`**, config gains `PrivacyConfig` / `MaskLevel` types, **`SessionReplayPlugin.setup`** resolves the level (default **`'medium'`** when omitted or `{}`) and passes it as the **9th bridge argument** into updated **iOS (ObjC/Swift)** and **Android (Kotlin `PrivacyConfig`)** `setup` paths. README documents mask levels and fixes a typo; the example app sets `privacyConfig`.
> 
> For **`@amplitude/session-replay-react-native`**, **`MaskLevel`** moves from an enum to a string-literal union aligned with the browser SDK; **`privacyConfig.maskLevel`** is the supported API while top-level **`maskLevel`** is deprecated. **`normalizeConfig`** folds legacy `maskLevel` into `privacyConfig` (explicit `privacyConfig` wins), and **`nativeConfig`** forwards a flat `maskLevel` string to native with the same **`'medium'`** fallback.
> 
> Jest config for the plugin gains **`tsx`/`jsx` extensions** and ignores **`example/`** tests; both packages get expanded tests for default resolution and native forwarding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9d9715700ba60537cb284cbd4049793c8302888. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/plugin-session-replay-react-native/src/session-replay-config.ts
```diff
@@ -1,10 +1,29 @@
...
+   */
+  Conservative = 'conservative',
+}
+
 export interface SessionReplayConfig {
   sampleRate?: number;
   enableRemoteConfig?: boolean;
   logLevel?: LogLevel;
   autoStart?: boolean;
+  maskLevel?: MaskLevel;
```
aliaksandr-kazarez: hey, let's follow how other Session Replay SDKs are naming that, and follow it. 
I believe it should be PrivacyConfig or similar.

packages/plugin-session-replay-react-native/src/index.tsx
aliaksandr-kazarez, line 1: do you also need to apply those changes to session-replay-react-native package?

packages/session-replay-react-native/src/session-replay.ts
```diff
@@ -207,10 +207,11 @@ export async function stop(): Promise<void> {
 }
 
 function nativeConfig(config: Required<SessionReplayConfig>): NativeSessionReplayConfig {
+  const resolvedMaskLevel = config.privacyConfig?.maskLevel ?? config.maskLevel;
```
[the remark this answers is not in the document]
aliaksandr-kazarez: Good catch. Fixed in 584ab86c: dropped `privacyConfig.maskLevel` (and the matching top-level `maskLevel`) from `getDefaultConfig()` and apply the `Medium` default only at resolution time in `nativeConfig()`. Added test cases covering (a) deprecated `maskLevel` only, (b) `privacyConfig.maskLevel` taking precedence over the deprecated field, (c) default fallback when neither is set.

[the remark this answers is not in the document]
aliaksandr-kazarez: Follow-up: refactored to do the deprecated→`privacyConfig` translation at the config boundary in 6c93e5f6, so `nativeConfig()` no longer carries default-fallback logic and internal code only sees `SessionReplayConfigInternal` (deprecated `maskLevel` stripped). Behavior unchanged — same three cases still pass.

packages/plugin-session-replay-react-native/src/session-replay-config.ts
```diff
@@ -1,17 +1,78 @@
...
+  Light = 'light',
+  /**
+   * Medium masking - balanced approach to content masking
+   */
+  Medium = 'medium',
+  /**
+   * Conservative masking - maximum content masking for privacy
+   */
+  Conservative = 'conservative',
+}
```
Mercy811: can we reuse the one defined at `packages/session-replay-react-native/src/session-replay-config.ts`?

packages/session-replay-react-native/src/session-replay-config.ts
Mercy811: This might be out of scope of this pr but just a heads up that it's better to use const instead of enum because some compilers can treat string-valued enums like const enums. It inlines each member at the call site and strips the enum declaration from the output.  Example: https://github.com/statsig-io/private-js-client-monorepo/pull/878.

SR browser just declares it as a type https://github.com/amplitude/Amplitude-TypeScript/blob/15224b4412852597e0f1d2f5b54c450c02989b2b/packages/plugin-session-replay-browser/src/typings/session-replay.ts#L20-L30

Mercy811 (changes requested):

packages/plugin-session-replay-react-native/src/session-replay-config.ts
aliaksandr-kazarez: Good call. I looked at importing `MaskLevel` directly from `@amplitude/session-replay-react-native`, but that package's entry point pulls in its native module, so reusing its value would add a runtime dependency on a *second* native SR module to the plugin (risking double native-module registration in host apps).

To get the de-dup benefit without that coupling, I converted `MaskLevel` in **both** packages to the exact string-literal union the SR browser SDK uses (`'light' | 'medium' | 'conservative'`) — a pure type with zero runtime. The two definitions are now structurally identical and `import type`-compatible, so if we later pull out a shared types-only package they drop in cleanly. Happy to do that full extraction as a follow-up if you'd prefer. (Commit d9d97157)

packages/session-replay-react-native/src/session-replay-config.ts
aliaksandr-kazarez: Done — switched `MaskLevel` from a string `enum` to a string-literal union type in both RN packages, matching the SR browser declaration you linked. Updated the default config, the native-boundary resolution, the tests, the example app, and the README to use the string literals. (Commit d9d97157)

packages/session-replay-react-native/src/session-replay.ts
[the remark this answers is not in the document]
aliaksandr-kazarez: Fixed in d9d97157. `getDefaultConfig()` no longer bakes in `privacyConfig: { maskLevel: Medium }`; the effective default is resolved once at the native boundary as `privacyConfig.maskLevel ?? 'medium'`. With the default removed, a user passing only the deprecated `maskLevel: 'conservative'` is normalized into `privacyConfig` and reaches native as `conservative` — no silent downgrade. Added a regression test.

packages/plugin-session-replay-react-native/src/session-replay.ts
```diff
@@ -19,7 +20,7 @@ export class SessionReplayPlugin implements EnrichmentPlugin<ReactNativeClient,
   config: ReactNativeConfig;
   isInitialized = false;
 
-  sessionReplayConfig: SessionReplayConfig;
+  sessionReplayConfig: ResolvedSessionReplayConfig;
```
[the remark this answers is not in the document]
aliaksandr-kazarez: Fixed in d9d97157. With the baked-in default removed, both packages coalesce at the native boundary (`?? 'medium'`) instead of an unsafe `as` cast, so a partial `privacyConfig: {}` resolves to `medium` rather than forwarding `undefined` across the native bridge. Added tests covering the empty-`privacyConfig` case in both packages.

aliaksandr-kazarez:
Pushed `d9d97157` addressing the open review feedback:

**Functional fix (Bugbot high-sev findings)**
- Removed the baked-in `privacyConfig: { maskLevel: Medium }` default from `getDefaultConfig()` in **both** the standalone and plugin RN packages.
- The effective mask level is now resolved **once at the native boundary** as `privacyConfig.maskLevel ?? 'medium'`. This fixes (a) the silent privacy downgrade where the deprecated top-level `maskLevel: 'conservative'` was overridden by the `Medium` default, and (b) the `undefined`-to-native-bridge path when a partial `privacyConfig: {}` is supplied.

**Style / reuse (reviewer comments)**
- Converted `MaskLevel` from a string `enum` to a string-literal union type (`'light' | 'medium' | 'conservative'`), matching the SR browser SDK declaration. This sidesteps the const-enum inlining concern and keeps the two RN packages structurally identical as pure types — without coupling the plugin's runtime to the standalone package's native module (which a direct value import would do). See thread replies for the reuse rationale.

**Tests / docs**
- Standalone: 49/49 jest pass, incl. new coverage for the deprecated-`maskLevel` fallback and the empty-`privacyConfig` → `medium` case.
- Plugin: 8/8 jest pass, incl. new empty-`privacyConfig` → `medium` case and `getDefaultConfig` no longer baking in a level.
- Updated the example app and the plugin README to the string-literal API.
- Local `build` + `typecheck` + `eslint` + `prettier` green for both packages.

Not merging — re-requested review.

packages/plugin-session-replay-react-native/src/session-replay-config.ts
Mercy811: Thanks Alex. I thought the standalone SR RN is a dependency of SR RN plugin like SR browser plugin. But the plugin reuses the native SR libraries directly, not the standalone RN JS package and the reuse happens at the
 platform-level dep layer, not the npm/JS layer.

Mercy811 (approved):
Lgtm!

sojingle (approved):
LGTM, thanks!

Issue #1595, opened 2026-03-13
Remove AsyncStorage dependency for react-native and let us pass our own Storage


ferrannp:
It is 2026, you are hardcoding an ancient version: `"@react-native-async-storage/async-storage": "^1.17.11"`. Let us just pas our own storage, we want to use https://github.com/mrousavy/react-native-mmkv.

Thank you!

josemariaeliasMM:
We are in the same situation we would like to use react-native-mmkv as well.

I found there is an old feature request about the same topic:

https://github.com/amplitude/Amplitude-TypeScript/issues/1283

Issue #1283, opened 2025-09-09
Feature Request: Remove the @amplitude/analytics-react-native dependency with async storage and allow a custom storage implementation


sebastian-herrera-itti:
## Summary

In `@amplitude/analytics-react-native` sdk, consider making `@react-native-async-storage/async-storage` an optional peer dependency instead of a required dependency. This would allow developers to provide their own storage implementation, such as `react-native-mmkv` or `expo-sqlite/kv-store`, without being forced to include AsyncStorage. This change would reduce APK size and avoid maintaining multiple local data sources.

## Motivations

The `@amplitude/analytics-react-native` SDK has `@react-native-async-storage/async-storage` as a required dependency. This forces us to include an additional local data source, even though we are already using `mmkv` for persistent local storage. It also adds unnecessary weight to the APK.

ottob:
Any updates on this?

Issue #540, opened 2023-08-15
Remove AsyncStorage package dependency for React Native


nuramirahs:
## Summary

After setting up our own custom storage provider (we are currently using `react-native-mmkv-storage`), we intend to remove the AsyncStorage library. However, we run into the error when trying to do so: `Error: [@RNC/AsyncStorage]: NativeModule: AsyncStorage is null`

Would like to know if there's a way we can completely remove the dependency on `@react-native-async-storage/async-storage` from the Amplitude package or if there's a recommended workaround for this issue?

justin-fiedler:
Hi @nuramirahs thanks for using Amplitude. At this time there is no way to remove the dependency since it is [required in the main RN SDK package](https://github.com/amplitude/Amplitude-TypeScript/blob/502a080b6eca2bc390b5d8076f24b9137d213f89/packages/analytics-react-native/package.json#L65).

We could improve this by separating out the storage implementation to a separate package and publishing it as a separate NPM module e.g. `@amplitude/storage-react-native`.

I will see if I can get this on our roadmap but it may be a while we have a number of other fixes/updates in progress. I will keep you updated.

Thanks!

lucaslozz:
I'm commenting this issue to keep seeing the notifications.

Noitham:
Hey @justin-fiedler, is there any update on the work, or the possibility of extracting the storage implementation into a separate package?

A-Yatsyk:
Hi,

Are there any updates on separating AsyncStorage from @amplitude/analytics-react-native?
We are now using AsyncStorage in our project and would like to avoid keeping an unnecessary dependency.

Thanks!

Vadko:
+1 for this, hardcoding storage into package dependencies is not a good pattern, there should be an option to pass storage instance and methods from any storage library

Issue #266, opened 2022-11-08
Don't force use of AsyncStorage for React Native


mmmoussa:
## Summary

In the React Native library, allow passing in a custom AsyncStorage library that is compatible with the AsyncStorage API.

## Motivations

In order to use this package in React Native we currently have to install `react-native-async-storage`. We don't currently use that library and would prefer to use `react-native-mmkv-storage` instead, which supports the same API interface as AsyncStorage. The reason for doing so is to achieve better performance.

You can even see an example of how this would be done here: https://github.com/Flagsmith/flagsmith-js-client/blob/main/lib/react-native-flagsmith/src/flagsmith-core.ts#L31

mmmoussa:
Thanks @liuyang1520 for the guidance! I was confused by this in the README:

> In addition, this package relies on `@react-native-async-storage/async-storage`, which must also be installed

nuramirahs:
Hi @liuyang1520,

I'm currently facing the same issue as well. Even after setting up our own custom storage provider, we're still running into the error: `Error: [@RNC/AsyncStorage]: NativeModule: AsyncStorage is null`

Eager to know if there's a way we can fully remove the dependency on `@react-native-async-storage/async-storage` from the Amplitude package or if there's a recommended workaround for this issue.

Some guidance would be greatly appreciated, thanks!

sumersao:
Has anyone solved these problems?

PR #1772, opened 2026-05-22
fix(analytics-react-native): lazy-load AsyncStorage so customers can opt out (SDKRN-8)

Mercy811:
## Summary

Replace the static `import AsyncStorage from '@react-native-async-storage/async-storage'` at the top of `packages/analytics-react-native/src/storage/local-storage.ts` with a lazy `require` wrapped in try/catch. When the package can't be resolved, `LocalStorage.isEnabled()` returns `false` and the existing fallback chain in `createFlexibleStorage` drops down to `MemoryStorage` instead of throwing at module-load.

Linear: [SDKRN-8](https://linear.app/amplitude/issue/SDKRN-8)

## Problem

`@react-native-async-storage/async-storage` is currently a regular `dependency` and the SDK statically imports it. Customers who:

- provide their own `config.storageProvider` (e.g. `react-native-mmkv`), and
- want to exclude AsyncStorage from their native build (to drop the redundant native module / shrink APK / avoid Expo Doctor duplicate warnings / avoid pnpm autolinking edge cases)

…still hit:

```
Error: [@RNC/AsyncStorage]: NativeModule: AsyncStorage is null
```

Reported across [#266](https://github.com/amplitude/Amplitude-TypeScript/issues/266), [#540](https://github.com/amplitude/Amplitude-TypeScript/issues/540), [#1283](https://github.com/amplitude/Amplitude-TypeScript/issues/1283), [#1595](https://github.com/amplitude/Amplitude-TypeScript/issues/1595) and several customer support threads.

### Reproduce steps

1. pull latest main
2. cd examples/react-native-app
3. remove `@react-native-async-storage/async-storage` in `package.json`
4. `cd ios && bundle install && bundle exec pod install`
5. `cd examples/react-native-app && pnpm start` to start metro
6. in another terminal `pnpm ios` 

Screenshot 2026-05-25 at 6 06 39 PM

## What changed

- `packages/analytics-react-native/src/storage/local-storage.ts`: lazy-resolve AsyncStorage inside a `getAsyncStorage()` helper. Each method checks for availability before calling. The require is wrapped in try/catch so a missing JS package no longer crashes module-load. The existing try/catch around individual storage calls still handles the "JS present but native module null" case.
- `packages/analytics-react-native/README.md`: document the opt-out recipe (custom `storageProvider` + `react-native.config.js` autolinking exclusion).
- `packages/analytics-react-native/test/storage/local-storage.test.ts`: new test verifying `LocalStorage` degrades to a no-op when the package can't be resolved (`jest.resetModules` + `jest.doMock` with a throwing factory).

No `package.json` changes — `@react-native-async-storage/async-storage` remains a regular `dependency`, so the default install path is unchanged and existing customers keep their identity continuity.

## Customer opt-out recipe (now documented in README)

```ts
// 1. Provide a custom storage provider
import { init } from '@amplitude/analytics-react-native';
init(API_KEY, { storageProvider: myMMKVProvider });
```

```js
// 2. Exclude AsyncStorage from native autolinking
// react-native.config.js
module.exports = {
  dependencies: {
    '@react-native-async-storage/async-storage': {
      platforms: { ios: null, android: null },
    },
  },
};
```

The JS package stays in `node_modules` (~1 KB) but is never invoked; the iOS and Android binaries no longer include AsyncStorage.

## Out of scope

- Widening the async-storage version range pin — tracked separately in [SDKRN-3](https://linear.app/amplitude/issue/SDKRN-3).
- Major version bump / removing async-storage entirely — explicitly decided against in the [design discussion](https://amplitude.slack.com/archives/C04HSDQA2VB/p1779484264172449); would break user/device ID continuity for the default-install majority.

## Test plan

- [x] `pnpm test` — 88 tests pass, coverage thresholds met (statements 86.66%, branches 84.32%, functions 87%, lines 86.77%)
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean (8 pre-existing warnings in unrelated files)
- [ ] Manual: install in the RN example app with `storageProvider` override + `react-native.config.js` exclusion, confirm no `NativeModule: AsyncStorage is null` error and that events still queue/flush.
- [ ] Manual: install in the RN example app without any overrides, confirm identity + event persistence still work as on `main`.

---

> [!NOTE]
> **Medium Risk**
> Changes default persistence resolution for all RN customers when AsyncStorage is missing or native-linked incorrectly; wrong error handling could silently drop identity/event persistence without crashing.
> 
> **Overview**
> Replaces the top-level AsyncStorage import in `local-storage.ts` with a cached lazy `require`, so apps can exclude the native module without crashing at SDK load. **`LocalStorage`** now treats a missing package, load failures (with selective `console.warn`), and **null native bridge** errors as disabled persistence—`isEnabled()` is false and read/write paths no-op or return undefined, including **`getRaw`** for identity migration callers.
> 
> Documents the full **AsyncStorage opt-out** (override **`storageProvider`** and **`cookieStorage`**, autolinking exclusion, RN Web stub/alias patterns) in the package README.
> 
> Adds **unit tests** for missing package, broken installs, and native-bridge rejection; wires the **RN example app** with in-memory storage, autolinking exclusion, expanded **Maestro smoke** (boot + track/identify), and removes **RNCAsyncStorage** from the example **Podfile.lock**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36c1d53dac2a25ec79a72082361d1a4c25cd0de7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analytics-react-native/src/storage/local-storage.ts
```diff
@@ -1,12 +1,50 @@
...
+  /* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
+  try {
+    const mod = require('@react-native-async-storage/async-storage');
+    // Handles both ES-module (`{ default: AsyncStorage }`) and direct-export
+    // shapes — e.g. `jest.mock(..., () => mockAsyncStorage)` returns the mock
+    // directly without a `default` wrapper.
+    asyncStorage = mod?.default ?? mod;
+  } catch {
+    asyncStorage = undefined;
+  }
```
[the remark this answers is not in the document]
Mercy811: Thanks — Comments 1 and 2 are addressed in the next commit (README now explicitly tells customers to override both `storageProvider` and `cookieStorage`, and the "never invoked" wording is qualified).

Pushing back on this third one and leaving MODULE_NOT_FOUND silent. Reasoning:

- The supported full-opt-out story is "override both storage slots **and** keep async-storage's JS package resolvable (the `react-native.config.js` exclusion path)." In that path the `require()` succeeds and MODULE_NOT_FOUND never fires.
- The case where MODULE_NOT_FOUND actually triggers is a customer who's gone further — `npm overrides: false`, `patch-package`, or some other JS-level removal. That's an intentional choice; logging on every app launch would just be noise for them.
- The "broken install" / "syntax error in the package" failure mode (the one most likely to mask a real bug) is already covered: non-MODULE_NOT_FOUND errors hit the `console.warn` branch. So the bug-masking concern is addressed for the realistic failure case.
- If async-storage is silently missing because of a genuine install accident (typo in overrides, etc.), the customer will notice fast: identity resets every launch, and the "Opting out" README section spells out the trade-off so they can map the symptom back to the cause.

If we ever see a real support case where a silent MODULE_NOT_FOUND masked a real install bug, I'm happy to revisit and log on all errors. For now, keeping it silent on the supported opt-out path seems like the right trade-off.

Mercy811:
@Bug Bot

Mercy811:
bugbot run

Mercy811:
Re: code-reviewer Should-fix on `InMemoryStorage.getRaw` (App.tsx) — pushing back on this one.

The reviewer flagged `JSON.stringify(value)` as a "subtly wrong reference implementation" that would double-encode strings. But this is the canonical pattern used by the SDK's own `MemoryStorage<T>` at `packages/analytics-core/src/storage/memory.ts:13-16`:

```ts
async getRaw(key: string): Promise<string | undefined> {
  const value = await this.get(key);
  return value ? JSON.stringify(value) : undefined;
}
```

My `InMemoryStorage` in the example app mirrors that pattern exactly, which is intentional — the example should match how the SDK's own in-memory backend behaves, not invent a different shape. Customers copying this pattern get the same `getRaw` semantics as the SDK's built-in fallback.

The "double-encoded string" concern is also moot in practice for this example app: `storageProvider` stores `Event[]` (objects) and `cookieStorage` stores `UserSession` (an object), so the string-primitive case doesn't apply to either slot.

Addressed Nit 1 (toast lingering between Track Event and Track Identify) in the next commit by waiting for the first toast to dismiss before tapping. Locally re-verified — all Maestro steps pass.

Keeping Nit 2 (README "Common gotchas" SSR note) as-is — RN Web in Next.js is a real customer setup and the gotcha is genuine for those users. Trimming the note would lose useful information.

Mercy811:
bugbot run

packages/analytics-react-native/src/storage/local-storage.ts
```diff
@@ -1,12 +1,63 @@
...
+// the SDK throwing at module-load time.
+//
+// The result is cached for the app lifetime: `require()` is synchronous in
+// React Native and the module registry is stable, so retrying after a failed
+// resolution would never produce a different result.
+const getAsyncStorage = (): AsyncStorageLike | undefined => {
+  if (resolved) {
+    return asyncStorage;
+  }
+  resolved = true;
```
daniel-graham-amplitude: Having a "resolved" variable is redundant. You can just say:

```
if (asyncStorage) {
  return asyncStorage;
}
```

packages/analytics-react-native/src/storage/local-storage.ts
```diff
@@ -1,12 +1,63 @@
...
+  /* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
+  try {
+    const mod = require('@react-native-async-storage/async-storage');
```
daniel-graham-amplitude, line 34: Did you consider async ESM imports? (example: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#import_a_module_for_its_side_effects_only)

packages/analytics-react-native/src/storage/local-storage.ts
```diff
@@ -1,12 +1,63 @@
...
+let resolved = false;
+let asyncStorage: AsyncStorageLike | undefined;
+
+// Resolve AsyncStorage lazily so the module can be opted out of via custom
+// `storageProvider` + `react-native.config.js` autolinking exclusion without
+// the SDK throwing at module-load time.
+//
+// The result is cached for the app lifetime: `require()` is synchronous in
+// React Native and the module registry is stable, so retrying after a failed
+// resolution would never produce a different result.
```
daniel-graham-amplitude: I think these comments are a tad verbose. A lot of this information could and should be gleaned from just looking at the source code.

packages/analytics-react-native/src/storage/local-storage.ts
daniel-graham-amplitude: I think I understand now, resolved could also mean it failed to load and we don't wish to try again.

packages/analytics-react-native/src/storage/local-storage.ts
```diff
@@ -1,12 +1,63 @@
...
+  resolved = true;
+  /* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
+  try {
+    const mod = require('@react-native-async-storage/async-storage');
+    // Handles both ES-module (`{ default: AsyncStorage }`) and direct-export
+    // shapes — e.g. `jest.mock(..., () => mockAsyncStorage)` returns the mock
+    // directly without a `default` wrapper.
+    asyncStorage = mod?.default ?? mod;
+  } catch (e) {
+    asyncStorage = undefined;
```
daniel-graham-amplitude: Optional: you could set this to `null` to indicate that it's not available at all, then you wouldn't need the resolved variable.

packages/analytics-react-native/README.md
```diff
@@ -27,6 +27,174 @@ yarn add @amplitude/analytics-react-native
 yarn add @react-native-async-storage/async-storage
 ```
 
+## Opting out of AsyncStorage
+
+If you'd rather use your own storage backend (for example `react-native-mmkv`, an encrypted store, or SQLite), you can supply your own storage and exclude AsyncStorage from your native build.
```
daniel-graham-amplitude: I would replace `storage backend` with `storage client`. `backend` suggests it's coming from a service.

packages/analytics-react-native/README.md
```diff
@@ -27,6 +27,174 @@ yarn add @amplitude/analytics-react-native
...
+
+The SDK uses two separate storage slots:
+
+- `storageProvider` — the event queue (events waiting to be flushed to Amplitude).
+- `cookieStorage` — identity / session state (device ID, user ID, session ID).
+
+To fully opt out, override **both**. If you only override `storageProvider`, the SDK still tries to read/write identity through the default chain, which falls back to AsyncStorage on native — and if you've also removed AsyncStorage, identity degrades to in-memory and resets on every app launch.
+
+1. Implement the `Storage` interface and pass both slots. Note that `init`'s signature is `(apiKey, userId, options)` — pass `undefined` for `userId` so the overrides land in the options slot, not on `userId`:
```
daniel-graham-amplitude: > pass `undefined` for `userId` 

Unless the user wants to pass a `userId`.

packages/analytics-react-native/src/storage/local-storage.ts
Mercy811: that's async, it's better to be sync to keep the existing storage methods clean.

packages/analytics-react-native/README.md
```diff
@@ -27,6 +27,174 @@ yarn add @amplitude/analytics-react-native
...
+       '@react-native-async-storage/async-storage': {
+         platforms: { ios: null, android: null },
+       },
+     },
+   };
+   ```
+
+   AsyncStorage is no longer linked into your iOS or Android binaries. The JS package stays in `node_modules` so `require()` still resolves, but with both storage slots overridden, the SDK never invokes any AsyncStorage methods.
+
+3. For React Native Web, exclude AsyncStorage from your web bundle.
```
daniel-graham-amplitude: This feels like overkill to me. From looking at [Bundlephobia](https://bundlephobia.com/package/@react-native-async-storage/async-storage@3.1.0) it looks like async-storage has a min+GZ size of 0.25 kilobytes. Could we at least maybe move this to another .md that's something like an Advanced Usage markdown file?

packages/analytics-react-native/README.md
daniel-graham-amplitude: > Implement the `Storage` interface

Can we provide a link to the Storage interface so that the user knows what to implement?

packages/analytics-react-native/src/storage/local-storage.ts
Mercy811: In a clean full-opt-out the scenario it is indeed dead code because getAsyncStorage() will never be called. But in the partial-opt-out case where, for example, it gets called from the cookieStorage fallback and async-storage is also missing, Metro doesn't cache failed resolutions, so without the flag we'd re-throw and re-log on every storage operation.

daniel-graham-amplitude:
@Mercy811 can I confirm that you were able to:
1. successfully run the example app without async-storage autolinking
2. successfully reproduce the import error
3. successfully observe that your changes caused the import error to stop happening

packages/analytics-react-native/src/storage/local-storage.ts
Mercy811: Yes!

packages/analytics-react-native/src/storage/local-storage.ts
Mercy811: I'd still like to keep them just for readability if you don't have strong opinion on this. No harm of having comments IMHO

packages/analytics-react-native/README.md
Mercy811: @cursoragent fix it

packages/analytics-react-native/README.md
Mercy811: @cursoragent fix it to: 

pass `undefined` or the desired user Id for `userId`

Mercy811:
> @Mercy811 can I confirm that you were able to:
successfully run the example app without async-storage autolinking
successfully reproduce the import error
successfully observe that your changes caused the import error to stop happening

@daniel-graham-amplitude that's right, by the example app. And also added a maestro test to prevent regression too. Enriched PR description.

packages/analytics-react-native/README.md
Mercy811: I will move them to dev doc center. It doesn't affect the mini file and the gzip size. It only affects the tarball size which adds download cost of npm install this package.

packages/analytics-react-native/src/storage/local-storage.ts
Mercy811: I like this for simplicity! Updated at https://github.com/amplitude/Amplitude-TypeScript/pull/1772/changes/79f9d9a09c3552f07491881974948537eda1a1fe

packages/analytics-react-native/README.md
Mercy811: https://github.com/amplitude/amplitude-docs-next/pull/471

daniel-graham-amplitude (approved):
Lgtm

PR #1775, opened 2026-05-27
ci: skip RN smoke test if analytics-react-native not affected

daniel-graham-amplitude:
### Summary

CI/CD optimization that skips the RN Smoke test if the analytics-react-native package was not affected.

This job takes ~15 minutes and this reduces it to <1 minute for when package not affected

### Checklist

---

> [!NOTE]
> **Low Risk**
> CI-only change; smoke tests still run on main and on PRs that touch the React Native package, with a small risk of false skips if Nx affected detection misclassifies dependencies.
> 
> **Overview**
> **Pull requests** now run a lightweight **`check-affected`** job before the Maestro iOS smoke test. It uses Nx (`nx show projects --affected` between the PR base and head) via new **`scripts/is-package-unaffected.sh`** to see whether **`@amplitude/analytics-react-native`** is in the affected set.
> 
> When the package is **not** affected, the workflow sets `unaffected=true` and **`ios-smoke` is skipped** (`needs` + `if` on that output). **Pushes to `main`** still run **`ios-smoke`** without waiting on the skip gate (`github.event_name != 'pull_request'`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8358ac7055b745d8c37b88b512b784cd1e9b3dc9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mercy811:
Can we also run the test when the core package is updated? I remember we had a bug from `getGlobalScope()` from core.

daniel-graham-amplitude:
> Can we also run the test when the core package is updated? I remember we had a bug from getGlobalScope() from core.

@Mercy811 it will. Whenever a dependency of analytics-react-native is updated, `nx affected` will show `analytics-react-native` as an affected package.

Mercy811:
Can we group steps needs `steps.package-unaffected.outputs.unaffected == 'false'` together? 

To collapse the repeated if checks, this can split into two jobs and gate at the job level instead of per-step. 

```
  jobs:
    check-affected:
      runs-on: ubuntu-latest          # cheap; no need for macos-14 just to check
      outputs:
        unaffected: ${{ steps.check.outputs.unaffected }}
      env:
        NX_BASE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || 'origin/main~1' }}
        NX_HEAD: ${{ github.sha }}
      steps:
        - uses: actions/checkout@v4
          with: { fetch-depth: 0 }
        - uses: pnpm/action-setup@v4
        - uses: actions/setup-node@v4
          with: { node-version: 20, cache: 'pnpm' }
        - run: pnpm install --frozen-lockfile
        - id: check
          run: |
            if sh scripts/is-package-unaffected.sh @amplitude/analytics-react-native; then
              echo "unaffected=true" >> "$GITHUB_OUTPUT"
            else
              echo "unaffected=false" >> "$GITHUB_OUTPUT"
            fi

    smoke:
      needs: check-affected
      if: needs.check-affected.outputs.unaffected == 'false'
      runs-on: macos-14
      env: { ... }                    # existing env block
      steps:
        # all current heavy steps with NO `if:` clauses
        # "Dump simulator logs on failure" goes back to plain `if: failure()`
```

ubuntu-latest is ~10× cheaper per minute than macos-14. Today even unaffected runs still spin up a Mac runner just to learn there's nothing to do.

daniel-graham-amplitude:
That's a good idea @Mercy811 . Done!

That will also be a good gate for if we want to add any other React Native specific tests.
 
Screenshot 2026-05-27 at 3 26 44 PM

Mercy811 (approved):
Thanks Dan, lgtm!

PR #1781, opened 2026-05-28
feat(analytics-browser): enable fetch keepalive by default to survive page navigation

Mercy811:
## Summary

Linear: [SDKW-14](https://linear.app/amplitude/issue/SDKW-14/investigate-initoptionsmore-causing-gtm-template-upgrade-error-log-in)

This PR adds a new capability to the browser SDK: the default `fetch` transport now sends uploads with **`keepalive: true`**, so in-flight event requests can **complete even when the page is navigating away** (redirects, link clicks, tab close).

Previously, delivering events at navigation time meant opting into `transport: 'beacon'` (`navigator.sendBeacon`). That works, but `sendBeacon` is limited: POST-only, no custom headers, **no gzip** (`Content-Encoding`), no access to the server response, and a hard 64 KiB cap. `fetch` with `keepalive` gives the same "survives navigation" behavior on the **default** transport while keeping everything we already rely on — gzip compression, custom headers, response handling, and the existing retry path. In short, customers no longer have to trade away those features to get navigation-time delivery.

### What changed
- `packages/analytics-browser/src/transports/fetch.ts`
  - New exported constant `KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024`.
  - Set `keepalive` when the outgoing body is within the cap. Size is measured with `ArrayBuffer.byteLength` for the gzip-compressed body, and the cheap JSON string length as a rough estimate for the uncompressed body (no per-request `Blob`/`TextEncoder` allocation).
- New opt-out: `enableKeepalive?: boolean` on `TransportOptions`. keepalive stays on by default; it is only disabled when explicitly set to `false`. It applies to the `fetch` transport (no effect on `xhr`/`beacon`).
- Tests added/updated for the enabled (small body) and disabled (oversized body) paths, including the compressed-body path and the `enableKeepalive` opt-out.

### Opting out
keepalive is on by default. To disable it (e.g. to avoid competing for the shared keepalive budget):

```js
amplitude.init(API_KEY, { transport: { type: 'fetch', enableKeepalive: false } });
// or at runtime:
amplitude.setTransport({ type: 'fetch', enableKeepalive: false });
```

## Why a 16 KB cap?

`keepalive` comes with a hard, browser-enforced budget, and opting an oversized request into it is **worse** than leaving it off — the browser rejects the request outright instead of sending it. The cap is a graceful fallback: bodies within it get navigation survival; bodies above it send as a normal `fetch` (no size limit, just no survival).

**The limit is normative in the Fetch spec.** WHATWG Fetch Standard, "HTTP-network-or-cache fetch":

> If the sum of `contentLength` and `inflightKeepaliveBytes` is greater than **64 kibibytes**, then return a network error.

Source: https://fetch.spec.whatwg.org/#http-network-or-cache-fetch

Three things drive the 16 KB choice:

1. **It's a network error, not a silent downgrade.** A plain `fetch()` has no body-size limit; the moment you add `keepalive: true`, a body over budget makes the browser **reject the request entirely**. So we only opt a request into keepalive when its body fits.

2. **The 64 KiB budget is *shared*, not per-request.** `inflightKeepaliveBytes` is the combined total of **all** in-flight keepalive requests in the document — our analytics flushes, session-replay uploads (which already use keepalive), and any keepalive/`sendBeacon` traffic from the customer's own code. Headers count toward it too. A cap well below the full budget leaves headroom and reduces contention-induced rejections.

3. **We gate on string length, which can undercount UTF-8 bytes.** To avoid allocating a `Blob`/`TextEncoder` on every send, the uncompressed path uses `body.length` (UTF-16 code units) as a rough proxy for byte size — mirroring the existing gzip threshold (`bodyString.length >= MIN_GZIP_UPLOAD_BODY_SIZE_BYTES`). A code unit expands to at most 3 UTF-8 bytes (a surrogate pair is 2 units → 4 bytes, i.e. 2/unit), so using `64 KiB / 4 = 16 KB` conservatively keeps even a fully multi-byte payload within the real budget.

For reference, Sentry adopted keepalive-by-default and added in-flight queue accounting after early issues with >64 KB requests ([sentry-javascript#7553](https://github.com/getsentry/sentry-javascript/pull/7553), [#2548](https://github.com/getsentry/sentry-javascript/issues/2548)), and `session-replay-browser` already uses keepalive in this repo ([commit fa5f84a](https://github.com/amplitude/Amplitude-TypeScript/commit/fa5f84af9a499e7bc12fbd5a1b8ab06ac89c2271)).

## Real-world payload sizes

In practice, event upload payloads are tiny relative to the 16 KB cap because the body is **gzip-compressed**. Observed request payload size is ~**0.2 KB** whether the batch contains a few events (e.g. 3–4) or many (e.g. 30) — so virtually all real traffic stays well under the cap and benefits from keepalive.

Screenshot 2026-05-28 at 10 12 20 PM

## Behavior & blast radius
- Bodies within the cap (effectively all real, gzipped traffic) now survive page navigation.
- Bodies over the cap send exactly as they do today: a plain `fetch` with no size limit, no navigation survival.
- In-session, if a keepalive request is ever rejected on the shared budget, it falls into the destination retry path (`catch` → `handleOtherResponse` → re-schedule), bounded by `flushMaxRetries` — so the outcome is **delay, not loss**.
- Browser coverage: `keepalive` is Baseline *Newly available* (Firefox 133, Nov 2024; Chromium/Safari since ~2019). On browsers without support the option is **silently ignored** — no error, the request just behaves as a normal `fetch`.

## Test plan
- `pnpm --filter @amplitude/analytics-browser test -- test/transport/fetch.test.ts test/config.test.ts` → pass.
  - keepalive enabled for bodies ≤ cap; disabled for oversized uncompressed and oversized compressed bodies.
  - keepalive disabled when `enableKeepalive: false`; enabled when `undefined`/`true`; option wired through `createTransport`.
  - Existing header/gzip assertions updated to include `keepalive`.

---

> [!NOTE]
> **Medium Risk**
> Changes default upload behavior on every browser fetch and interacts with the shared 64 KiB keepalive budget; oversized keepalive requests fail at the network layer, though in-session retries should limit data loss.
> 
> **Overview**
> Browser **`FetchTransport`** now sets **`keepalive: true`** on POST uploads when the outgoing body is within a **32 KiB** cap (`KEEPALIVE_MAX_BODY_SIZE_BYTES`), so typical event batches can finish after redirects or navigation instead of being cancelled.
> 
> Sizing uses **`ArrayBuffer.byteLength`** for gzip bodies and **JSON string length** as a cheap upper bound for uncompressed payloads (no per-request `Blob`). Bodies above the cap omit keepalive and behave like today’s plain `fetch`.
> 
> Tests assert keepalive on small/uncompressed and small/compressed requests, and **`keepalive: false`** for oversized JSON and mocked oversized gzip output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c18000f4e9240bcab9a1d1efa4543efb3ed71fc0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -8,18 +8,8 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
-// keepalive lets a request survive page navigation, but the browser caps the *combined*
-// size of all in-flight keepalive requests in a document at 64 KiB and rejects anything
-// over it (https://fetch.spec.whatwg.org/#http-network-or-cache-fetch). Since that budget
-// is shared (e.g. with session-replay), we gate at half to leave headroom; larger bodies
-// fall back to a plain fetch that won't survive navigation. The headroom also covers using
-// the cheap string length as a rough byte estimate (like the gzip threshold below) instead
-// of measuring exact UTF-8 bytes.
-export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 32 * 1024;
+export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
 
-// Temporary browser-specific fetch transport with gzip support.
-// TODO: Merge this implementation back into @amplitude/analytics-core FetchTransport
-// once React Native SDK supports request body gzip.
 export class FetchTransport extends BaseTransport implements Transport {
   private customHeaders: Record<string, string>;
 
@@ -58,8 +48,6 @@ export class FetchTransport extends BaseTransport implements Transport {
       ...headers,
     };
 
-    // Use string length as a rough byte estimate (the compressed path already has an exact
-    // byte count) to avoid allocating a Blob on every request.
     const bodySize = typeof body === 'string' ? body.length : body.byteLength;
 
     const options: RequestInit = {
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -16,10 +16,13 @@ export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
 // once React Native SDK supports request body gzip.
 export class FetchTransport extends BaseTransport implements Transport {
   private customHeaders: Record<string, string>;
+  private enableKeepalive: boolean;
 
-  constructor(customHeaders: Record<string, string> = {}) {
+  constructor(customHeaders: Record<string, string> = {}, enableKeepalive?: boolean) {
     super();
     this.customHeaders = customHeaders;
+    // Enabled unless explicitly disabled.
+    this.enableKeepalive = enableKeepalive !== false;
   }
 
   async send(serverUrl: string, payload: Payload, shouldCompressUploadBody = false): Promise<Response | null> {
@@ -58,7 +61,7 @@ export class FetchTransport extends BaseTransport implements Transport {
       headers,
       body,
       method: 'POST',
-      keepalive: bodySize <= KEEPALIVE_MAX_BODY_SIZE_BYTES,
+      keepalive: this.enableKeepalive && bodySize <= KEEPALIVE_MAX_BODY_SIZE_BYTES,
     };
 
     const response = await fetch(serverUrl, options);
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -56,13 +56,11 @@ export class FetchTransport extends BaseTransport implements Transport {
       ...headers,
     };
 
-    const bodyByteLength = typeof body === 'string' ? new Blob([body]).size : body.byteLength;
-
     const options: RequestInit = {
       headers,
       body,
       method: 'POST',
-      keepalive: bodyByteLength <= MAX_KEEPALIVE_BYTES,
+      keepalive: isWithinKeepaliveBudget(body),
     };
 
     const response = await fetch(serverUrl, options);
@@ -75,3 +73,12 @@ export class FetchTransport extends BaseTransport implements Transport {
     }
   }
 }
+
+const isWithinKeepaliveBudget = (body: string | ArrayBuffer): boolean => {
+  if (typeof body !== 'string') {
+    return body.byteLength <= MAX_KEEPALIVE_BYTES;
+  }
+  // UTF-8 byte length is always >= the JS (UTF-16) string length, so a string longer than the
+  // budget is already over it — skip the Blob allocation in that case.
+  return body.length <= MAX_KEEPALIVE_BYTES && new Blob([body]).size <= MAX_KEEPALIVE_BYTES;
+};
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -8,6 +8,20 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
+/**
+ * Upper bound on the request body size (in bytes) for which `keepalive` is enabled.
+ *
+ * The Fetch spec caps the *combined* size of all in-flight `keepalive` requests in a
+ * document at 64 KiB; exceeding it makes fetch return a network error rather than
+ * sending the request (see "HTTP-network-or-cache fetch":
+ * https://fetch.spec.whatwg.org/#http-network-or-cache-fetch). That budget is shared
+ * with anything else using keepalive in the same document (e.g. session-replay, the
+ * customer's own beacons), so we gate at half the budget to leave headroom and reduce
+ * the chance of a contention-induced rejection. Larger bodies fall back to a plain
+ * fetch (no size limit) that simply won't survive a page navigation.
+ */
+export const MAX_KEEPALIVE_BYTES = 32 * 1024;
+
 // Temporary browser-specific fetch transport with gzip support.
 // TODO: Merge this implementation back into @amplitude/analytics-core FetchTransport
 // once React Native SDK supports request body gzip.
@@ -49,10 +63,17 @@ export class FetchTransport extends BaseTransport implements Transport {
       ...headers,
     };
 
+    // keepalive lets the request survive page navigation (e.g. a redirect), preventing
+    // the browser from cancelling the in-flight upload. It is gated on body size because
+    // the browser rejects keepalive requests that would exceed the shared 64 KiB budget;
+    // larger bodies fall back to a plain fetch. See MAX_KEEPALIVE_BYTES above.
+    const bodyByteLength = typeof body === 'string' ? new Blob([body]).size : body.byteLength;
+
     const options: RequestInit = {
       headers,
       body,
       method: 'POST',
+      keepalive: bodyByteLength <= MAX_KEEPALIVE_BYTES,
     };
 
     const response = await fetch(serverUrl, options);
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -10,6 +10,9 @@ import {
 
 export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
 
+// Temporary browser-specific fetch transport with gzip support.
+// TODO: Merge this implementation back into @amplitude/analytics-core FetchTransport
+// once React Native SDK supports request body gzip.
 export class FetchTransport extends BaseTransport implements Transport {
   private customHeaders: Record<string, string>;
 
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -8,6 +8,7 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
+// Conservative 1/4 of the shared 64 KiB keepalive budget. Why 16K: https://github.com/amplitude/Amplitude-TypeScript/pull/1781
 export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
 
 // Temporary browser-specific fetch transport with gzip support.
--- a/packages/analytics-browser/src/transports/fetch.ts
+++ b/packages/analytics-browser/src/transports/fetch.ts
@@ -8,18 +8,11 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
-/**
- * Upper bound on the request body size (in bytes) for which `keepalive` is enabled.
- *
- * The Fetch spec caps the *combined* size of all in-flight `keepalive` requests in a
- * document at 64 KiB; exceeding it makes fetch return a network error rather than
- * sending the request (see "HTTP-network-or-cache fetch":
- * https://fetch.spec.whatwg.org/#http-network-or-cache-fetch). That budget is shared
- * with anything else using keepalive in the same document (e.g. session-replay, the
- * customer's own beacons), so we gate at half the budget to leave headroom and reduce
- * the chance of a contention-induced rejection. Larger bodies fall back to a plain
- * fetch (no size limit) that simply won't survive a page navigation.
- */
+// keepalive lets a request survive page navigation, but the browser caps the *combined*
+// size of all in-flight keepalive requests in a document at 64 KiB and rejects anything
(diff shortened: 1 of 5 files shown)


Mercy811:
Addressed review round 2:
- **S1 (decouple compressed-body test):** done — the over-budget compressed test now mocks `compressToGzipArrayBuffer`/`isCompressionStreamAvailable` directly rather than intercepting `Response`, so it no longer couples to the helper's internals.

Not changing, with reasoning:
- **S2 (try/finally around global cleanup):** leaving as-is. The manual set/delete of `CompressionStream`/`Response`/`Blob.prototype.stream` mirrors the existing gzip tests in this file; converting to `beforeEach`/`afterEach` or `try/finally` would be a broader refactor of pre-existing tests, out of scope for this PR. (The new S1 test no longer touches those globals at all.)
- **N1 (under-budget keepalive test is redundant):** keeping it as an explicit, self-documenting assertion of the enabled path. It's cheap and the intent is clearer than relying on the header/gzip tests to incidentally cover keepalive.
- **N2 (Blob sizing is browser-scoped):** acknowledged; the `MAX_KEEPALIVE_BYTES` comment already scopes this transport to the browser, and jsdom returns the correct UTF-8 byte count.

Mercy811:
@Bug Bot

Mercy811:
bugbot run

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -49,10 +56,13 @@ export class FetchTransport extends BaseTransport implements Transport {
       ...headers,
     };
 
+    const bodyByteLength = typeof body === 'string' ? new Blob([body]).size : body.byteLength;
+
     const options: RequestInit = {
       headers,
       body,
       method: 'POST',
+      keepalive: bodyByteLength <= MAX_KEEPALIVE_BYTES,
```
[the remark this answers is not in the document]
Mercy811: Good call — done in ddd40510. Extracted `isWithinKeepaliveBudget()` and short-circuit on `body.length` before allocating a Blob, since UTF-8 byte length is always ≥ the JS string length.

Mercy811:
bugbot run

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -8,6 +8,15 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
+// keepalive lets a request survive page navigation, but the browser caps the *combined*
+// size of all in-flight keepalive requests in a document at 64 KiB and rejects anything
+// over it (https://fetch.spec.whatwg.org/#http-network-or-cache-fetch). Since that budget
+// is shared (e.g. with session-replay), we gate at half to leave headroom; larger bodies
+// fall back to a plain fetch that won't survive navigation. The headroom also covers using
+// the cheap string length as a rough byte estimate (like the gzip threshold below) instead
```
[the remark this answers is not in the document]
Mercy811: Renamed to `KEEPALIVE_MAX_BODY_SIZE_BYTES` in c18000f4 to make clear it's a conservative per-request body cap (half the 64 KiB budget), not the budget itself.

Mercy811:
bugbot run

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -49,10 +58,15 @@ export class FetchTransport extends BaseTransport implements Transport {
...
+    // Use string length as a rough byte estimate (the compressed path already has an exact
+    // byte count) to avoid allocating a Blob on every request.
+    const bodySize = typeof body === 'string' ? body.length : body.byteLength;
+
     const options: RequestInit = {
       headers,
       body,
       method: 'POST',
+      keepalive: bodySize <= KEEPALIVE_MAX_BODY_SIZE_BYTES,
     };
```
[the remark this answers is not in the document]
Mercy811: Intentional tradeoff, not changing. `body.length` is a deliberate rough proxy mirroring the existing gzip threshold a few lines up (`bodyString.length >= MIN_GZIP_UPLOAD_BODY_SIZE_BYTES`), and the gate is at 32 KB — half the 64 KiB budget — specifically to absorb UTF-16→UTF-8 expansion. Also note that requests to Amplitude's default endpoints are gzipped, so the keepalive gate uses the exact compressed `byteLength` there; `.length` only governs uncompressed bodies (small, or compression unavailable). In the worst case (heavily multi-byte, uncompressed, near the cap) a request could be mis-enabled and rejected — but that lands in the destination retry path in-session (delay, not loss), and at page unload it's no worse than a plain fetch, which is cancelled anyway. Avoiding a TextEncoder/Blob allocation on every send is the goal here.

packages/analytics-browser/test/transport/fetch.test.ts
```diff
@@ -125,12 +128,69 @@ describe('fetch transport', () => {
...
+      const transport = new FetchTransport();
+      const url = 'http://localhost:3000';
+      const payload = { api_key: '', events: [{ event_type: 'test', device_id: 'test_device_id' }] };
+
+      const fetchSpy = jest.spyOn(window, 'fetch').mockReturnValueOnce(Promise.resolve(new Response('{}')));
+      await transport.send(url, payload, false);
+
+      const [, options] = fetchSpy.mock.calls[0];
+      expect(options?.keepalive).toBe(true);
+    });
```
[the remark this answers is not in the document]
Mercy811: Skipping for the same reason as the source-side reply: the gate intentionally uses string length as a rough proxy (like the gzip threshold), with the 32 KB half-budget cap as headroom. A multi-byte regression test would assert behavior we're deliberately not guaranteeing at byte precision on the uncompressed path.

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -49,10 +51,13 @@ export class FetchTransport extends BaseTransport implements Transport {
       ...headers,
     };
 
+    const bodySize = typeof body === 'string' ? body.length : body.byteLength;
+
     const options: RequestInit = {
       headers,
       body,
       method: 'POST',
+      keepalive: bodySize <= KEEPALIVE_MAX_BODY_SIZE_BYTES,
```
daniel-graham-amplitude: My concern about this is that we're potentially competing with the combined keep-alive budget of other requests so we should only use this when necessary.

I think we should add another condition to this:

```javascript
let isPageExiting = false;
function beforeUnloadHandler() {
  isPageExiting = true;
}
globalScope?.addEventListener('beforeunload', beforeUnloadHandler);
//...
keepalive: bodySize <= KEEPALIVE_MAX_BODY_SIZE_BYTES && isPageExiting
```

Mercy811: The request is canceled if the user closes the tab at any point between when the request is initiated and the tab is closed. Listening to a page exit flag isn't helpful here. 

The risk is small here because the payload size is less than 1 KB. See the screenshot in the PR description. The size is released as long as the request succeeds.

daniel-graham-amplitude: I'm going to need to research how keepalive works better. It sounds like there's some tradeoffs between using `keepalive` and not using `keepalive` meaning there's a risk of nasty surprises. I feel like this implementation does make sense and I'll probably approve it, but I need to know what it is we're shipping first.

Mercy811: No problem, take your time 🙏

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -8,6 +8,8 @@ import {
   Transport,
 } from '@amplitude/analytics-core';
 
+export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
```
lewgordon-amplitude, line 12: It may be worth adding a comment here on why 16K.

lewgordon-amplitude (approved):
LGTM, though I'm curious if @daniel-graham-amplitude finds any gotchas during his research.

packages/analytics-browser/src/transports/fetch.ts
Mercy811: I have a section to explain this in pr description. It's long so I'd keep it there for more details

Mercy811: Let me add a few lines of comments too!

lewgordon-amplitude: I'd probably do a shorter comment here and then just link to the PR perhaps in the code. Sometimes it can be hard to track down PRs/PR comments.

Mercy811: Yes! https://github.com/amplitude/Amplitude-TypeScript/pull/1781/commits/f127af682d0e903d0f58118676bd5b24dde664dc

daniel-graham-amplitude:
> LGTM, though I'm curious if @daniel-graham-amplitude finds any gotchas during his research

@lewgordon-amplitude I think we're good. The risk that I was worried about was this starving other fetch requests. But I think the fact that 1) we keep a modest budget; 2) we don't run fetch requests in parallel; mitigates the risk substantially and the benefits make it worthwhile

packages/analytics-browser/test/config.test.ts
```diff
@@ -431,6 +431,16 @@ describe('config', () => {
...
+
+    test('should pass enableKeepalive: false through to the fetch transport', async () => {
+      const transport = createTransport({ type: 'fetch', enableKeepalive: false });
+      const fetchMock = jest.fn().mockResolvedValue({ text: () => Promise.resolve('{}') });
+      (global as { fetch?: unknown }).fetch = fetchMock;
+      await transport.send('http://localhost:3000', { api_key: '', events: [] });
+      const [, options] = fetchMock.mock.calls[0];
+      expect(options?.keepalive).toBe(false);
+      delete (global as { fetch?: unknown }).fetch;
+    });
```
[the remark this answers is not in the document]
Mercy811: Fixed — now saves the original `global.fetch` and restores it in a `finally` block instead of deleting it.

packages/analytics-core/src/types/transport.ts
```diff
@@ -10,6 +10,9 @@ export type TransportType = 'xhr' | 'beacon' | 'fetch';
 export interface TransportOptions {
   type?: TransportType;
   headers?: Record<string, string>;
+  // Whether the fetch transport sets `keepalive` so uploads survive page navigation.
+  // Enabled unless explicitly set to false. Has no effect on the xhr/beacon transports.
+  enableKeepalive?: boolean;
```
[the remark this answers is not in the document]
Mercy811: `TransportOptions` is only used for analytics-browser. Keep it as it now.

packages/analytics-browser/src/transports/fetch.ts
```diff
@@ -8,15 +8,21 @@ import {
...
 
+// Conservative 1/4 of the shared 64 KiB keepalive budget. Why 16K: https://github.com/amplitude/Amplitude-TypeScript/pull/1781
+export const KEEPALIVE_MAX_BODY_SIZE_BYTES = 16 * 1024;
+
 // Temporary browser-specific fetch transport with gzip support.
 // TODO: Merge this implementation back into @amplitude/analytics-core FetchTransport
 // once React Native SDK supports request body gzip.
 export class FetchTransport extends BaseTransport implements Transport {
   private customHeaders: Record<string, string>;
+  private enableKeepalive: boolean;
```
daniel-graham-amplitude, line 19: Can we capitalize the `A` in `alive` here so that it's consistent with camelCase standards (first word: all lowercase; other words: first letter capitalized only)

Mercy811: But the property is just `keepalive` https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive

daniel-graham-amplitude: Okay, it just looks weird to me, but I understand the reasoning and we can keep it this way. I'm just a bit concerned about typos for non-TS users who don't have type-checking.

Mercy811: Oh yeah I have the same concern but on the opposite way that people knowing it's `keepalive` will also have typos if we use camel case here. The mdn naming is unconventional 😔

daniel-graham-amplitude (approved):

PR #1779, opened 2026-05-27
chore(element-selector): scaffold element selector types

jxiwang:
## Summary

Second PR in the v1 element-selector groundwork chain. Lands the static + primitive pieces of `@amplitude/element-selector`: type definitions, both regex pattern packs, helper functions, and the two v1 strategies. Each piece is independently unit-tested — nothing drives them together yet. The orchestrator, fallback, config resolver, and engine factory land in the follow-up PR (SR-4609).

Closes [SR-4608](https://linear.app/amplitude/issue/SR-4608).

## What's in this PR

**Types** (`src/types.ts`)
- `Strategy`, `StrategyContext` — pluggable-strategy interface
- `ResolvedSelectorConfig` — runtime config consumed by strategies / orchestrator / fallback
- `ElementSelectorRemoteConfig` — remote-config payload shape (every field optional)
- `SelectorEngine` — public engine interface (implementation lands in the orchestration PR)

**Pattern packs**
- `patterns/autogenerated-ids.ts` — `DEFAULT_AUTOGENERATED_ID_PATTERNS` (React `useId`, Radix, Headless UI, MUI, hex hashes, trailing-digit timestamps, 4+ consecutive digits) + `compile()` + `isStableId()`
- `patterns/unstable-classes.ts` — `DEFAULT_UNSTABLE_CLASS_PATTERNS` (Tailwind utilities + variants, CSS-in-JS / build hashes, Swiper / MUI / Radix runtime state classes) + `compile()` + `filterClasses()`

**Helpers**
- `helpers/get-stable-id.ts` — single source of truth for "is this element's id usable?" (honors the explicit-tracking-attribute suppression signal + the autogen filter)
- `helpers/describe-relative.ts` — positional descent builder, emits `tag:nth-of-type(n)`

**Strategies**
- `strategies/explicit-tracking-attribute.ts` — anchor on `data-amp-track-id` (or the customer-configured attribute name)
- `strategies/stable-id.ts` — anchor on `tag#id` when the id survives `getStableId`

**Tests** — one file per source file, all running on jsdom; pattern packs are table-driven with positive and negative cases drawn from real React / Radix / Headless UI / MUI / Swiper / Tailwind / CSS-modules / Emotion / styled-components samples.

## What's NOT in this PR

- Orchestrator with ancestor walk → SR-4609
- `fallbackCssPath` → SR-4609
- `resolveSelectorConfig` → SR-4609
- `createSelectorEngine` factory + the runtime `SelectorEngine` instance → SR-4609
[...]

PR #1783, opened 2026-06-01
feat(element-selector): orchestrator, fallback, config resolver, engine factory

jxiwang:
### Summary

Third PR in the v1 element-selector groundwork chain. Composes the orchestration layer on top of the primitives from [SR-4608](https://github.com/amplitude/Amplitude-TypeScript/pull/1779) — orchestrator, fallback, config resolver, and engine factory. After this lands, the package has a complete runtime engine that consumers can instantiate.

Closes [SR-4609](https://linear.app/amplitude/issue/SR-4609).

## What's in this PR

**Orchestrator** (`src/orchestrator.ts`)
- `runOrchestrator(el, config, options)` — two-pass walk over `[target, ...ancestors]`
- Pass 1: tries `explicitTrackingAttribute` at every level; Pass 2: tries `stableId` at every level
- This gives an "explicit anchor at any depth beats a structural anchor at any depth" priority — matches the design doc's "customer intent overrides structural inference" rule
- On anchor found, descends back to the target via `describeRelative` and returns `anchorSelector > nth-of-type(N) > ...`
- Honors `maxAncestorWalkDepth` defensively (caller is the engine; uses `config.maxAncestorWalkDepth`)
- Uniqueness-checks every candidate via `querySelectorAll` so a strategy that emits a non-unique selector falls through to the next pass instead of being trusted
- Exports `DEFAULT_STRATEGIES` for diagnostics

**Fallback walker** (`src/fallback-css-path.ts`)
- `fallbackCssPath(el, config)` — runs when the strategy chain returns null
- Hardened cssPath: filters ids through `getStableId` (honors both the autogen filter AND the empty-attribute suppression signal), uses `:nth-of-type(N)` for sibling disambiguation (never classes, which are runtime-volatile)
- Terminates the walk early once it hits an id that survives `getStableId` — no walking past stable anchors
- Honors `maxAncestorWalkDepth`

**Config resolver** (`src/config/resolve-config.ts`)
- `resolveSelectorConfig(remote?, logger?)` — collapses the optional, string-based remote payload into a fully-typed `ResolvedSelectorConfig`
- Defaults to `enabled: false` (the engine ships dormant — autocapture stays on legacy `cssPath` until remote config flips it on)
- Customer pattern lists are a **full replacement** of the defaults, not a merge — an empty array is an explicit "disable filtering" signal
- Clamps `maxAncestorWalkDepth`: only positive finite integers honored; zero / negative / NaN / Infinity → undefined
- Exports `DEFAULT_RESOLVED_CONFIG` (frozen snapshot) for consumers and tests to introspect the baseline

**Engine factory** (`src/engine.ts`)
- `createSelectorEngine(config, options?): SelectorEngine` — composes the orchestrator with the fallback behind the documented `SelectorEngine` interface
- Implements `generate(el)`, `getConfig()`, `updateConfig(next)`, `onConfigChange(cb)` (with unsubscribe)
- Per-instance state and subscriber list — two engines have independent config and listeners
- Subscriber exceptions during `updateConfig` are isolated so one bad listener can't break the others

**Shared helpers**
- `src/helpers/escape-id.ts` — `escapeIdForCss(id)` shared by `stableId` strategy and `fallback-css-path`. Uses native `CSS.escape` when available, falls back to a manual hex-escape when not (jsdom historically omits it). Single source of truth means strategy and fallback always emit byte-identical anchors for the same input.

**Public exports** in `src/index.ts`
- Re-exports the orchestrator, fallback, resolver, factory, and types
- Orchestrator and fallback are exposed so the dashboard / Chrome extension can call them ad-hoc for previews — not just through the engine

**Tests**
- `test/orchestrator.test.ts` — default chain, two-pass priority, autogen-id filtering, uniqueness fallthrough on ambiguous candidates, depth limit, custom chain, malformed-selector defense
- `test/fallback-css-path.test.ts` — anchor termination, autogen filtering, suppression signal, `nth-of-type` disambiguation, walk-to-root, depth limit, querySelector round-trip
- `test/config/resolve-config.test.ts` — defaults, partial overrides, full-replacement semantics, invalid-regex skip, depth clamp edge cases, frozen snapshot
- `test/engine.test.ts` — happy path through each strategy + fallback, `updateConfig` / `onConfigChange` fan-out + unsubscribe + listener isolation, instance isolation, custom strategy chain
- Updated `test/index.test.ts` with assertions for the new public exports
- 173 tests total (was ~71 on SR-4608)

## Preexisting primitive fixes folded in

Three bugs in the primitives were blocking the SR-4609 test suite — fixed in-line because they're tiny and CI was gated on them:

1. `stableId` strategy used bare `CSS.escape` which throws in jsdom → swapped to the new `escapeIdForCss` helper that polyfills when the global is missing
2. Radix id pattern `/^radix-:/` only matched colon-suffixed forms (`radix-:r3:`) → broadened to `/^radix-/` so `radix-A1B2C3` also matches
3. CSS modules class pattern required `{5,}` chars in the middle segment, rejecting `Button_root__abc123` → loosened middle to `{3,}`

These would have come up the moment anyone ran the SR-4608 tests against the full corpus; surfacing them here.

## What's NOT in this PR

- Logger integration with `@amplitude/analytics-core` → SR-4641
- Property tests, scenarios corpus, JSON schema, README → SR-4641
- Interactive testbed → SR-4644
- Autocapture plugin consumption → SR-4659 (Phase 2)

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how DOM selectors are generated for autocapture when enabled via remote config; mistakes could mis-tag elements, though v1 ships disabled by default and is heavily tested in-package.
> 
> **Overview**
> Adds the **runtime composition layer** for `@amplitude/element-selector`: remote config is resolved via `resolveSelectorConfig`, selectors are produced through a two-pass `runOrchestrator` (explicit tracking attribute, then stable id) with **uniqueness checks**, and `createSelectorEngine` wires that to a hardened **`fallbackCssPath`** when the strategy chain returns null.
> 
> **Config** defaults to `enabled: false`, treats customer pattern arrays as **full replacements** (empty list disables filtering), and clamps invalid `maxAncestorWalkDepth` to no limit. The engine exposes `generate`, `getConfig`, `updateConfig`, and `onConfigChange` (isolated subscriber errors).
> 
> Also adds shared **`escapeIdForCss`**, exports orchestrator/fallback/resolver/factory from the package entry, tightens **Radix** and **CSS modules** default patterns, and adds broad jsdom coverage (~173 tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e2941fd4584d0c5492b98da74da6ffca39fc64c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jxiwang:
bugbot run

jxiwang:
bugbot run

bravecod (approved):

PR #1801, opened 2026-06-04
feat(plugin-autocapture-browser): integrate element selector

jxiwang:
## Summary

Fifth PR in the v1 element-selector chain. Lands the **SDK-side integration**: the autocapture plugin now imports `@amplitude/element-selector`, instantiates a `SelectorEngine` per SDK instance, and routes selector generation through it when the customer's remote config opts in. Customers with `enabled: true` in their `configs.analyticsSDK.autocapture.elementSelector` remote-config payload get v1 selectors; everyone else stays on the existing `cssPath`.

The engine is **private to the plugin instance**. Cross-realm consumers (Amplitude dashboard, Chrome extension) instantiate their own engines from the shared package against the same customer's remote config. Same-page consumers (the visual-tagging-selector widget) will receive the engine reference as an init-options argument in a follow-up ticket — not via a public global. Consistency comes from "same package + same config payload," not from sharing a runtime instance across realms.

Closes [SR-4659](https://linear.app/amplitude/issue/SR-4659).

## What's in this PR

**Workspace dependency**

- Add `@amplitude/element-selector: workspace:*` to `packages/plugin-autocapture-browser/package.json` dependencies.

**Engine wiring in plugin `setup()`**

- Resolve the initial config at setup via `resolveSelectorConfig(undefined, config.loggerProvider)` so the engine has documented defaults even before remote config lands.
- Construct one engine per plugin instance with `createSelectorEngine(resolved, { logger: config.loggerProvider })`. Stable identity for the lifetime of the SDK.
- When `config.fetchRemoteConfig` is on, subscribe to `configs.analyticsSDK.autocapture.elementSelector` via the same `remoteConfigClient.subscribe(...)` pattern the plugin already uses for `pageActions`. On every payload, call `selectorEngine.updateConfig(resolveSelectorConfig(payload, logger))`.

**Kill-switch routing in `DataExtractor.getElementPath`**

- `getElementPath` branches: when `selectorEngine.getConfig().enabled === true`, route through `selectorEngine.generate(element)`; otherwise fall through to the legacy `cssPath(element)`.
- The kill switch lives in remote config, not in code — flipping it on for an org is a runtime decision, not a redeploy.
- Performance instrumentation around the branch unchanged (`autocapturePlugin.getElementPath` histogram still recorded).


jxiwang:
bugbot run

PR #1802, opened 2026-06-04
feat(analytics-react-native): offline connectivity plugin + JS wiring (SDKRN-5) [1/5]

Mercy811:
## Summary
**PR 1 of 5** — splitting the SDKRN-5 offline-mode feature (originally #1796) into a reviewable, **non-breaking** stack. Each PR leaves the SDK runnable when merged in order.

This PR adds the **cross-platform JS** pieces + **no-op native placeholders**:
- `networkConnectivityCheckerPlugin` (`BeforePlugin`): seeds and flips `config.offline` from the native connectivity module (or `navigator.onLine` on web); flushes on reconnect; best-effort falls back to online when no connectivity source exists.
- Client wiring in `_init` (installs before `Destination`, skips on the `OfflineDisabled` sentinel).
- Plugin unit tests + end-to-end `offline-integration.test.ts` (native bridge mocked in the jest setups).
- **No-op `AmplitudeReactNativeConnectivity` native stubs** (iOS Swift + Android Kotlin) that register the module and always report connected, so JS has a real module to bind to.

### Why non-breaking
After this PR, offline mode is a **no-op** (the stubs always report "online") — behavior is identical to pre-feature. The real platform monitors arrive in the follow-ups:

| PR | Adds | Base |
|----|------|------|
| **#1 (this)** | JS plugin + wiring + tests + no-op native stubs | `main` |
| #2 | Real iOS module (`NWPathMonitor`/`SCNetworkReachability`) — replaces iOS stub | #1 |
| #3 | Real Android module (`ConnectivityManager`) + `ACCESS_NETWORK_STATE` — replaces Android stub | #2 |
| #4 | Robolectric tests for the Android module | #3 |
| #5 | Example `expo-app` (RN dedupe + offline test buttons) — independent | `main` |

## Test plan
- ✅ `analytics-react-native` typecheck + lint clean
- ✅ jest **106/106 pass** (web + mobile; the native bridge is mocked, so the stubs don't affect the suite)

---

> [!NOTE]
> **Medium Risk**
> Changes init plugin order and `config.offline` driving Destination behavior; production impact is limited while native stubs always report online, but react-native-web uses real `navigator.onLine` and mis-detection could affect sends.
> 
> **Overview**
> Adds **offline-mode plumbing** for React Native analytics: a `networkConnectivityCheckerPlugin` that seeds and updates `config.offline` from `AmplitudeReactNativeConnectivity` (native) or `navigator.onLine` on react-native-web, flushes on reconnect (immediate on native, delayed on web), and **defaults to online** when no connectivity source exists (including a regression guard for native apps missing the module).
> 
> `AmplitudeReactNative` now installs that plugin **before** `Destination` by default and skips it when `offline: OfflineDisabled`. **iOS/Android no-op native modules** register the bridge and always return connected so shipping this alone does not suppress sends; real monitors land in follow-up PRs.
> 
> `ReactNativeConfig` no longer omits `offline` from `IConfig`. Jest mocks, plugin unit tests, client init tests, and an offline integration test cover queue-while-offline / flush-on-reconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b5a73be90844ce3196d03d402a564fccce50842. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/types/config/react-native-config.ts
+++ b/packages/analytics-core/src/types/config/react-native-config.ts
@@ -6,7 +6,7 @@ type HiddenOptions = 'apiKey' | 'lastEventId';
 
 export type ReactNativeOptions = Omit<Partial<ReactNativeConfig>, HiddenOptions>;
 
-export interface ReactNativeConfig extends Omit<IConfig, 'offline' | 'requestMetadata'> {
+export interface ReactNativeConfig extends Omit<IConfig, 'requestMetadata'> {
   trackingOptions: ReactNativeTrackingOptions;
   trackingSessionEvents?: boolean;
   migrateLegacyData?: boolean;
--- a/packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
+++ b/packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
@@ -1,0 +1,51 @@
+package com.amplitude.reactnative
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import com.facebook.react.module.annotations.ReactModule
+
+const val CONNECTIVITY_MODULE_NAME = "AmplitudeReactNativeConnectivity"
+
+/**
+ * No-op placeholder for the connectivity native module, bridged back to JS as
+ * [CONNECTIVITY_MODULE_NAME]. It exists so the JS
+ * `networkConnectivityCheckerPlugin` can bind to a real native module and seed
+ * an initial "online" state; it never reports going offline.
+ *
+ * Real `ConnectivityManager` monitoring is added in a follow-up PR that replaces
+ * this file. Until then, offline mode is a no-op on Android —
+ * [getNetworkConnectivityStatus] always reports connected, so the SDK never
+ * wrongly suppresses sends.
+ */
+@ReactModule(name = CONNECTIVITY_MODULE_NAME)
+class AmplitudeReactNativeConnectivityModule(
+    reactContext: ReactApplicationContext
+) : ReactContextBaseJavaModule(reactContext) {
+
+    override fun getName(): String {
+        return CONNECTIVITY_MODULE_NAME
+    }
+
+    // Required so `NativeEventEmitter` doesn't warn on the JS side.
+    @ReactMethod
+    fun addListener(eventName: String) {
+        // No-op placeholder.
+    }
+
+    @ReactMethod
+    fun removeListeners(count: Int) {
+        // No-op placeholder.
+    }
+
+    /**
+     * JS reads this once on setup to seed the initial `offline` value. The
+     * placeholder always reports connected.
+     */
+    @ReactMethod
+    fun getNetworkConnectivityStatus(promise: Promise) {
+        promise.resolve(Arguments.createMap().apply { putBoolean("isConnected", true) })
+    }
+}
--- a/packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativePackage.java
+++ b/packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativePackage.java
@@ -17,6 +17,7 @@ public class AmplitudeReactNativePackage implements ReactPackage {
     public List<NativeModule> createNativeModules(@NonNull ReactApplicationContext reactContext) {
         List<NativeModule> modules = new ArrayList<>();
         modules.add(new AmplitudeReactNativeModule(reactContext));
+        modules.add(new AmplitudeReactNativeConnectivityModule(reactContext));
         return modules;
     }
 
--- a/packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
+++ b/packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
@@ -48,22 +48,25 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
   const setup = async (config: ReactNativeConfig, amplitude: ReactNativeClient) => {
     const nativeModule = NativeModules.AmplitudeReactNativeConnectivity as ConnectivityNativeModule | undefined;
 
-    const handleConnectivityChange = (isConnected: boolean) => {
+    const handleConnectivityChange = (isConnected: boolean, { deferFlush = false } = {}) => {
+      config.loggerProvider.debug(`Network connectivity status: ${isConnected ? 'online' : 'offline'}.`);
       const offline = !isConnected;
       if (config.offline === offline) {
         return;
       }
       config.offline = offline;
-      if (isConnected) {
-        config.loggerProvider.debug('Network connectivity changed to online.');
-        // No `ERR_NETWORK_CHANGED` on native, so flush immediately.
-        void amplitude.flush();
+      if (!isConnected) {
+        return;
+      }
+      if (deferFlush) {
+        setTimeout(() => {
+          void amplitude.flush();
+        }, config.flushIntervalMillis);
       } else {
-        config.loggerProvider.debug('Network connectivity changed to offline.');
+        void amplitude.flush();
       }
     };
 
-    // Native (iOS/Android): use the bridged module.
     if (nativeModule) {
       const eventEmitter = new NativeEventEmitter(nativeModule);
       // `startObserving` only fires on changes, so seed the initial state.
@@ -71,7 +74,7 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
         const status = await nativeModule.getNetworkConnectivityStatus();
         config.offline = !status.isConnected;
       } catch (e) {
-        config.loggerProvider.debug(`Failed to read initial network connectivity status: ${String(e)}`);
+        config.loggerProvider.warn(`Failed to read initial network connectivity status: ${String(e)}`);
         config.offline = false;
       }
       subscription = eventEmitter.addListener(CONNECTIVITY_EVENT_NAME, (event: ConnectivityEvent) => {
@@ -83,33 +86,15 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
     // Web (react-native-web): use `navigator.onLine`. Gate on `isWeb()` because
     // RN's `navigator` polyfill has no `onLine` — without this, a native app
     // missing the module would read `!undefined === true` and pin itself offline.
-    if (isWeb() && typeof navigator !== 'undefined') {
-      config.offline = !navigator.onLine;
-
-      addWebNetworkListener('online', () => {
-        if (config.offline === false) {
-          return;
-        }
-        config.loggerProvider.debug('Network connectivity changed to online.');
-        config.offline = false;
-        // Immediate flush on web causes ERR_NETWORK_CHANGED, so delay it.
-        setTimeout(() => {
-          void amplitude.flush();
-        }, config.flushIntervalMillis);
-      });
-
-      addWebNetworkListener('offline', () => {
-        if (config.offline === true) {
-          return;
-        }
-        config.loggerProvider.debug('Network connectivity changed to offline.');
-        config.offline = true;
-      });
+    const nav = globalScope?.navigator;
+    if (isWeb() && nav) {
+      config.offline = !nav.onLine;
+      addWebNetworkListener('online', () => handleConnectivityChange(true, { deferFlush: true }));
+      addWebNetworkListener('offline', () => handleConnectivityChange(false));
       return;
     }
 
-    // No connectivity source (no native module, not web) → assume online.
-    config.loggerProvider.debug(
+    config.loggerProvider.warn(
       'Network connectivity checker plugin found no connectivity source; defaulting to online.',
     );
     config.offline = false;
--- a/packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
+++ b/packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
@@ -4,14 +4,10 @@ import { isWeb } from '../utils/platform';
 
 export const CONNECTIVITY_EVENT_NAME = 'AmplitudeNetworkConnectivityChanged';
 
-/**
- * Shape of the native `AmplitudeReactNativeConnectivity` module bridged from
- * iOS (`RCTEventEmitter` subclass) and Android (`ReactContextBaseJavaModule`).
- */
+/** Native `AmplitudeReactNativeConnectivity` module bridged from iOS/Android. */
 interface ConnectivityNativeModule {
   getNetworkConnectivityStatus(): Promise<{ isConnected: boolean }>;
-  // `addListener`/`removeListeners` are required by `NativeEventEmitter` so it
-  // doesn't warn; they are no-ops on the JS side here.
+  // Required by `NativeEventEmitter`; no-ops on the JS side.
   addListener(eventName: string): void;
   removeListeners(count: number): void;
 }
@@ -52,10 +48,8 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
   const setup = async (config: ReactNativeConfig, amplitude: ReactNativeClient) => {
     const nativeModule = NativeModules.AmplitudeReactNativeConnectivity as ConnectivityNativeModule | undefined;
 
-    // Apply a connectivity change. Comparing against the current `config.offline`
-    // value naturally debounces repeated same-state signals (e.g. Android's
-    // `onCapabilitiesChanged` firing multiple times) so we don't flush-storm on
-    // transient reconnect flapping.
+    // Comparing against current `config.offline` debounces repeated same-state
+    // signals so we don't flush-storm on reconnect flapping.
     const handleConnectivityChange = (isConnected: boolean) => {
       const offline = !isConnected;
       if (config.offline === offline) {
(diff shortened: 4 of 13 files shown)


Mercy811:
bugbot run

packages/analytics-react-native/test/plugins/network-connectivity-checker.test.ts
```diff
@@ -0,0 +1,291 @@
...
+};
+
+const createAmplitudeMock = () =>
+  ({
+    flush: jest.fn(() => ({ promise: Promise.resolve() })),
+  } as unknown as ReactNativeClient);
+
+// `offline` is omitted from the public ReactNativeConfig type but exists on the
+// underlying core Config instance at runtime.
+const offlineOf = (config: object): boolean | null | undefined => (config as { offline?: boolean | null }).offline;
```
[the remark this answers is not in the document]
Mercy811: Good catch — addressed in 37c5419a. Removed the stale comment and deleted the `offlineOf` cast helper entirely; the tests now read `config.offline` directly since it's typed on `ReactNativeConfig` in this PR.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,139 @@
...
+        // Flushing immediately on web causes ERR_NETWORK_CHANGED, so delay it.
+        setTimeout(() => {
+          void amplitude.flush();
+        }, config.flushIntervalMillis);
+      });
+
+      addWebNetworkListener('offline', () => {
+        config.loggerProvider.debug('Network connectivity changed to offline.');
+        config.offline = true;
+      });
```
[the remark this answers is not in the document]
Mercy811: Addressed — guarded both web handlers so they only act on real transitions, mirroring the native path: `online` skips when already online (37c5419a) and `offline` skips when already offline (7b5a73be), each with a test. Note: browsers only fire `online`/`offline` on an actual `navigator.onLine` change, so a true flush storm is unlikely in practice, but this keeps the web path consistent with native and defends against duplicate events.

packages/analytics-react-native/test/react-native-client.test.ts
```diff
@@ -221,6 +222,39 @@ describe('react-native-client', () => {
...
+    test('should not install the network connectivity checker when offline is OfflineDisabled', async () => {
+      jest.spyOn(CookieMigration, 'parseOldCookies').mockResolvedValueOnce({ optOut: false });
+      const installSpy = jest.spyOn(NetworkChecker, 'networkConnectivityCheckerPlugin');
+      const client = new AmplitudeReactNative();
+      const addSpy = jest.spyOn(client, 'add');
+      await client.init(API_KEY, USER_ID, {
+        ...attributionConfig,
+        // `offline` is omitted from the public ReactNativeOptions type but is
+        // honored by the underlying core Config.
+        offline: core.OfflineDisabled,
```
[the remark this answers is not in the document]
Mercy811: Addressed in 37c5419a — removed the stale comment and the `as core.ReactNativeOptions` cast; `offline` is now a typed option on `ReactNativeOptions`, so the object is directly assignable.

Mercy811:
bugbot run

packages/analytics-react-native/test/plugins/offline-integration.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    DeviceEventEmitter.emit(CONNECTIVITY_EVENT_NAME, { isConnected: false });
+
+    // Track events while offline; their promises won't resolve until flushed,
+    // so don't await them here.
+    void client.track('event_while_offline_1').promise;
+    void client.track('event_while_offline_2').promise;
+
+    // Allow the events to be enqueued and persisted to storage.
+    await new Promise((resolve) => setTimeout(resolve, 50));
+
```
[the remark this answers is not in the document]
Mercy811: Addressed in ebe4adeb. Instead of the fixed `setTimeout(50)` wait, the test now awaits an event-driven signal: it spies on `storageProvider.set` and resolves once both offline events have been persisted (the Destination persists as it queues each event). Fully deterministic — no fixed delay, no timer polling, and it completes as soon as the work is done (~20ms).

I first tried fake timers as you suggested, but the async helpers (`runAllTimersAsync`/`advanceTimersByTimeAsync`) deadlock under the `react-native` jest preset, so awaiting the actual persistence turned out cleaner than driving timers manually.

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import com.facebook.react.module.annotations.ReactModule
+
+const val CONNECTIVITY_MODULE_NAME = "AmplitudeReactNativeConnectivity"
+
+/**
+ * No-op placeholder for the connectivity native module, bridged back to JS as
```
daniel-graham-amplitude, line 13: Can you confirm that this will work in [React Native Fabric](https://reactnative.dev/architecture/fabric-renderer)? Fabric replaced JS Bridging with Native Modules.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,130 @@
...
+      addWebNetworkListener('online', () => {
+        if (config.offline === false) {
+          return;
+        }
+        config.loggerProvider.debug('Network connectivity changed to online.');
+        config.offline = false;
+        // Immediate flush on web causes ERR_NETWORK_CHANGED, so delay it.
+        setTimeout(() => {
+          void amplitude.flush();
+        }, config.flushIntervalMillis);
```
daniel-graham-amplitude: What would happen if the `flushIntervalMillis` was set to `0`?

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
Mercy811: it should. The existing context plugin is also on Native Modules. The new architecture is backward compatible

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,130 @@
...
+      return;
+    }
+
+    // Web (react-native-web): use `navigator.onLine`. Gate on `isWeb()` because
+    // RN's `navigator` polyfill has no `onLine` — without this, a native app
+    // missing the module would read `!undefined === true` and pin itself offline.
+    if (isWeb() && typeof navigator !== 'undefined') {
+      config.offline = !navigator.onLine;
+
+      addWebNetworkListener('online', () => {
```
daniel-graham-amplitude: To make it less repetitive can we use `handleConnectivityChange` here too

```typescript
addWebNetworkListener('online', () => {
  // deferFlush option to tell handleConnectivityChange to delay the flush by flushIntervalMillis
  handleConnectivityChange(true, { deferFlush: true });
});


addWebNetworkListener('online', () => {
  // deferFlush option to tell handleConnectivityChange to delay the flush by flushIntervalMillis
  handleConnectivityChange(false, { deferFlush: true });
});
```

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
Mercy811: Worth adding a smoke test of a sample app with the new architecture in a separate PR.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    // Native (iOS/Android): use the bridged module.
+    if (nativeModule) {
+      const eventEmitter = new NativeEventEmitter(nativeModule);
+      // `startObserving` only fires on changes, so seed the initial state.
+      try {
+        const status = await nativeModule.getNetworkConnectivityStatus();
+        config.offline = !status.isConnected;
+      } catch (e) {
+        config.loggerProvider.debug(`Failed to read initial network connectivity status: ${String(e)}`);
```
daniel-graham-amplitude: This feels worthy of an `error` log level. We'd want customers reporting this if it reaches this path.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,130 @@
...
+      subscription = eventEmitter.addListener(CONNECTIVITY_EVENT_NAME, (event: ConnectivityEvent) => {
+        handleConnectivityChange(event.isConnected);
+      });
+      return;
+    }
+
+    // Web (react-native-web): use `navigator.onLine`. Gate on `isWeb()` because
+    // RN's `navigator` polyfill has no `onLine` — without this, a native app
+    // missing the module would read `!undefined === true` and pin itself offline.
+    if (isWeb() && typeof navigator !== 'undefined') {
```
daniel-graham-amplitude: Can we scope it to `globalScope.navigator`. I just generally think it's good practice to not have unbound global variables.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: flush() will be called in next marcotask then. This aligns this existing browser offline logics.

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativePackage.java
```diff
@@ -17,6 +17,7 @@ public class AmplitudeReactNativePackage implements ReactPackage {
     public List<NativeModule> createNativeModules(@NonNull ReactApplicationContext reactContext) {
         List<NativeModule> modules = new ArrayList<>();
         modules.add(new AmplitudeReactNativeModule(reactContext));
+        modules.add(new AmplitudeReactNativeConnectivityModule(reactContext));
```
daniel-graham-amplitude, line 20: Is there a specific benefit to having a new module just for network connectivity? I'm wondering if it's better to just add the network network connectivity methods to the existing `AmplitudeReactNativeModule`.

```kotlin
fun addNetworkConnectivityListener();
fun removeNetworkConnectivityListener();
fun getNetworkConnectivityStatus()
```

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
```diff
@@ -0,0 +1,130 @@
...
+          return;
+        }
+        config.loggerProvider.debug('Network connectivity changed to offline.');
+        config.offline = true;
+      });
+      return;
+    }
+
+    // No connectivity source (no native module, not web) → assume online.
+    config.loggerProvider.debug(
```
daniel-graham-amplitude: I think this should have a `warn` log level?

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: how about warn? I want to keep it at the same level as https://github.com/amplitude/Amplitude-TypeScript/pull/1802#discussion_r3365839924 because both of them fail to get initial seed and fall back to offline false.

daniel-graham-amplitude: Warn should be okay.

packages/analytics-react-native/test/react-native-client.test.ts
```diff
@@ -221,6 +222,37 @@ describe('react-native-client', () => {
...
   }
 
+  describe('network connectivity checker plugin', () => {
+    const NETWORK_CHECKER_PLUGIN_NAME = '@amplitude/plugin-network-checker-react-native';
+
+    test('should install the network connectivity checker by default', async () => {
+      jest.spyOn(CookieMigration, 'parseOldCookies').mockResolvedValueOnce({ optOut: false });
+      const installSpy = jest.spyOn(NetworkChecker, 'networkConnectivityCheckerPlugin');
+      const client = new AmplitudeReactNative();
+      const addSpy = jest.spyOn(client, 'add');
```
daniel-graham-amplitude: Can we use `beforeEach` and `afterEach` in these tests.

```
let installSpy, client, addSpy;
beforeEach(async () => {
  jest.spyOn(CookieMigration, 'parseOldCookies').mockResolvedValueOnce({ optOut: false });
  installSpy = jest.spyOn(NetworkChecker, 'networkConnectivityCheckerPlugin');
  client = new AmplitudeReactNative();
  addSpy = jest.spyOn(client, 'add');
});

afterEach(async () => {
  installSpy?.mockRestore();
});
```

packages/analytics-react-native/test/plugins/network-connectivity-checker.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      expect(config.offline).toBe(true);
+
+      DeviceEventEmitter.emit(CONNECTIVITY_EVENT_NAME, { isConnected: true });
+      DeviceEventEmitter.emit(CONNECTIVITY_EVENT_NAME, { isConnected: true });
+      expect(amplitude.flush).toHaveBeenCalledTimes(1);
+    });
+
+    test('teardown removes the native subscription', async () => {
+      const config = useDefaultConfig();
+      const plugin = networkConnectivityCheckerPlugin();
```
daniel-graham-amplitude: Looks like every test starts with 

```
const config = useDefaultConfig();
const plugin = networkConnectivityCheckerPlugin();
```

So they can be moved into `beforeEach`.

packages/analytics-react-native/test/plugins/network-connectivity-checker.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      expect(config.offline).toBe(false);
+    });
+  });
+
+  /*
+   * Web fallback exercises `navigator.onLine` + online/offline events. The
+   * global scope and navigator are mocked so this runs under both the node
+   * (test:mobile) and jsdom (test:web) suites.
+   */
+  describe('web fallback', () => {
```
daniel-graham-amplitude: Nitpick: is it really a fallback? Seems like a standard functionality?

packages/analytics-core/src/types/config/react-native-config.ts
```diff
@@ -6,7 +6,7 @@ type HiddenOptions = 'apiKey' | 'lastEventId';
 
 export type ReactNativeOptions = Omit<Partial<ReactNativeConfig>, HiddenOptions>;
 
-export interface ReactNativeConfig extends Omit<IConfig, 'offline' | 'requestMetadata'> {
```
daniel-graham-amplitude, line 9: If we merge this now, and then this gets published before the subsequent PR's are published, then that version will be in a state where the offline configuration is available, but doesn't do anything.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: Done in 1b3e94d5. Both web handlers now delegate to `handleConnectivityChange`, with a `deferFlush` option: `online` → `handleConnectivityChange(true, { deferFlush: true })` (defers the flush by `flushIntervalMillis`), `offline` → `handleConnectivityChange(false)`. Native still flushes immediately. The dedup guard now lives in one place.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: Done in 1b3e94d5 — now reads `globalScope?.navigator` off the cached `globalScope` (from `getGlobalScope()`) instead of the bareword global.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: Raised in 1b3e94d5 — went with `warn` rather than `error` (the SDK recovers by falling back to online, so it's degraded-but-functional, not fatal). Easy to bump to `error` if you'd prefer it louder.

packages/analytics-react-native/src/plugins/network-connectivity-checker.ts
Mercy811: Done in 1b3e94d5 — this no-connectivity-source log is now `warn`.

packages/analytics-react-native/test/react-native-client.test.ts
Mercy811: Done in 1b3e94d5 — moved the shared setup (parseOldCookies mock, installSpy, client, addSpy) into `beforeEach`, with `installSpy.mockRestore()` in `afterEach`; each test now just calls `init` and asserts.

packages/analytics-react-native/test/plugins/network-connectivity-checker.test.ts
Mercy811: Done in 1b3e94d5 — hoisted `config`/`amplitude`/`plugin` into a `beforeEach` for the native-mode describe. I scoped it there rather than the top level because the web-mode tests must mock `getGlobalScope` *before* the plugin is created (the plugin caches `globalScope` at creation), so those keep their inline setup.

packages/analytics-react-native/test/plugins/network-connectivity-checker.test.ts
Mercy811: you're right. The real fallback is offline false. Updated.

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativeConnectivityModule.kt
Mercy811: https://linear.app/amplitude/issue/SDKRN-28/smoke-test-sample-app-on-react-native-new-architecture

packages/analytics-core/src/types/config/react-native-config.ts
Mercy811: That's right. The plan is to merge all stacked PRs together. Splitting into smaller PRs just for easier review

daniel-graham-amplitude: But if this gets merged into `main`, and then somebody hits publish, it will be in that non-functional state. Is your plan to just not merge to main until they're all ready?

packages/analytics-react-native/android/src/main/java/com/amplitude/reactnative/AmplitudeReactNativePackage.java
Mercy811: Separate it for cleaner logic and ownership for each module. There is no performance cost. For the old architecture, It will create a lightweight module instance at app launch, and the new architecture lazy loads it.

packages/analytics-core/src/types/config/react-native-config.ts
Mercy811: Yes, that's the plan.

daniel-graham-amplitude: Okay 👍

aliaksandr-kazarez (approved):
LGTM;

aliaksandr-kazarez:
@Mercy811 are other PRs ready for review? 
Seems like links in the description are broken (pointing to old PRs)

Mercy811:
> @Mercy811 are other PRs ready for review? 
> 
> Seems like links in the description are broken (pointing to old PRs)

Hi @aliaksandr-kazarez, thanks for reviewing it. No I'm still working on them. Hopefully they will be out this week 😀

PR #1806, opened 2026-06-05
feat(session-replay-browser): add eagerFullSnapshotSend, captureFullSnapshotOnFocus, and max event size config options

lewgordon-amplitude:
## Summary

GAs the Session Replay "canary-only" performance tuning knobs (from `sr-4646-rc2`) into the mainline `@amplitude/session-replay-browser` config so consuming apps no longer need to pass them as untyped/unknown options. All four are now first-class, documented (TSDoc), tested config options.

**Every default preserves today's behavior** — this is a no-op behavior change unless a consumer explicitly opts in.

### Options added

| Option | Type | Default (= current behavior) | Effect when changed |
| --- | --- | --- | --- |
| `eagerFullSnapshotSend` | `boolean` | `true` (eager send on every full snapshot) | `false` keeps the snapshot compressed/buffered but defers the network send to the normal interval/size cadence — avoids focus/checkout-driven request storms when many SDK instances run per page |
| `captureFullSnapshotOnFocus` | `boolean` | `true` (snapshot on window `focus`) | `false` skips the on-focus `takeFullSnapshot(true)` entirely on high focus-churn pages |
| `maxPersistedEventsSizeBytes` | `number` | `MAX_EVENT_LIST_SIZE` (700,000) | raw UTF-8 byte cap before the store splits a batch into its own request; clamped to `[1KB, 8MB]` and logged |
| `maxSingleEventSizeBytes` | `number` | `MAX_SINGLE_EVENT_SIZE` (9,000,000) | raw UTF-8 byte cap for a single rrweb event (dropped above this, at capture time and pre-send); clamped to `[1KB, 10MB]` and logged |

### Wiring
- Types + TSDoc in `config/types.ts`; parsed/sanitized in `local-config.ts` (`sanitizeByteSize` clamps + logs).
- `eagerFullSnapshotSend` gates the `onFullSnapshotProcessed` callback passed to `EventCompressor`.
- `captureFullSnapshotOnFocus` early-returns in `focusListener`.
- `maxPersistedEventsSizeBytes` flows through `createEventsManager` into the memory/IDB stores (`BaseEventsStore.shouldSplitEventsList`).
- `maxSingleEventSizeBytes` enforced in `EventCompressor` (capture time) and `events-manager` (pre-send backstop).

### Placement note
[...]

PR #1798, opened 2026-06-04
fix(session-replay-browser): coalesce style attribute mutations when mergeMutations is enabled

lewgordon-amplitude:
## Summary

- When `performanceConfig.mergeMutations` is enabled, `mergeGroup()` in `merge-mutation-events.ts` concatenated every attribute mutation in the merged window with **no per-node dedup**. A burst of N inline `style` updates on the same node id stayed N attribute entries even after merging — which made animated text (e.g. the GoFundMe ticker crossfade) replay as stacked text.
- Added `coalesceStyleAttributes()`: within a merged group, `style` attribute mutations are collapsed using **last-write-wins per node id**. Only the final `style` write per id survives; any non-style attribute keys are always preserved (a superseded entry keeps its other keys, and is dropped only if it was style-only). This is replay-safe because rrweb applies attribute mutations in order, so only the final `style` affects the rendered end state.
- Existing transient add/remove/cascade logic is untouched — only the `attributes` assembly changed.
- No new public flag: the whole merge path already only runs when `mergeMutations` is enabled (via `event-compressor.ts` `mergeMutationTasks`).
- Exposed `mergeMutations` on the plugin's `SessionReplayPerformanceConfig` typing (it was previously omitted) with a doc comment, and documented it in the plugin README `performanceConfig` table.

## Test plan

- [x] Added unit tests in `merge-mutation-events.test.ts` mirroring the ticker pattern: two consecutive mergeable mutations, same node id, `style` "opacity: 0.5…" then "opacity: 1; transform: …" → single attribute entry with the **last** style value.
- [x] Test that non-style attribute changes are NOT collapsed/lost.
- [x] Test that style is coalesced while non-style keys on superseded entries are preserved.
- [x] `pnpm test -- merge-mutation-events` → 34 passed.
- [x] `pnpm typecheck` (session-replay-browser) → clean for changed files (one unrelated pre-existing error in `e2e/idb.spec.ts`).
- [x] `pnpm build:esm` for both `@amplitude/session-replay-browser` and `@amplitude/plugin-session-replay-browser` → pass.
- [x] `eslint` + `prettier --check` on changed files → clean.

Linear: https://linear.app/amplitude/issue/SR-4654

_This PR's contribution to the bundle:_ the `mergeMutations` style-attribute coalescing path is active in the canary build and is part of the bundle that produced the reliability numbers above.

## Production canary validation

[...]

PR #1807, opened 2026-06-05
feat(session-replay-browser): coalesce persisted-sequence page-load backlog drain

lewgordon-amplitude:
## Summary

Linear: **SR-4660**

On page load the Session Replay SDK drains every persisted event sequence from a prior session via `sendStoredEvents` and enqueues them back-to-back. With no coalescing and no spacing this flushed as **N separate POSTs** in a single flush — a request flood on page load (the "page-load backlog drain") that feeds request-volume spikes and server-side throttling.

This PR coalesces that drain by **reusing the proven post-throttle merge** instead of inventing a second mechanism:

- Extracted the existing identity-grouped merge body into a shared `coalesceByIdentity(list)` helper.
- The throttle path (`mergeQueueAfterThrottle`) now delegates to it — **byte-for-byte identical merge behavior**, plus its existing once-per-pause log.
- Added a drain path (`mergeDrainBacklog`) that uses the same helper, gated by a new `markCoalesceNextFlush()` flag.
- `events-manager.sendStoredEvents` calls `markCoalesceNextFlush()` before enqueuing the backlog (only when >1 sequence). Because those enqueues are synchronous and the flush is deferred to the next tick via `schedule(0)`, the whole backlog lands in the queue before the flag is consumed, so many small same-identity persisted batches collapse into far fewer POSTs.

### onComplete fan-out safety
The merged context's `onComplete` fans out to **every** source context's callback (via `Promise.allSettled`), so each underlying IDB sequence record is still cleaned up **exactly once** on success — preserving the existing recovery semantics. The 413 split-and-retry path already handles a merged context's fanned-out cleanup.

### Behavior preservation
- Steady-state live capture is untouched — it never sets the flag, so its sending behavior is unchanged.
- Different destination identities (session/device/api/type/zone/sampleRate/version) are **never** merged.
- Oversized merges still split safely via the soft byte cap + existing 413 bisect path.
- If a throttle merge and a drain land on the same flush, the throttle path runs and clears the drain flag too (no redundant double-merge).

### Spacing — deferred
Spacing the drained requests over the flush cadence is **deferred as a follow-up**. The coalescing alone removes the per-sequence fan-out (the dominant source of the flood); spacing adds timer/retry-interaction complexity for marginal additional benefit and is best shipped/measured separately.

## Test plan
Unit tests added/extended in `packages/session-replay-browser/test/` (package gate is ~100% coverage; all new branches covered):

- [x] Page-load drain of multiple same-identity persisted batches → **one** coalesced POST.
- [x] Each source `onComplete` still fires **exactly once**.
- [x] Different-identity batches are **not** merged (separate POSTs, no drain log).
- [x] Oversized merge still splits into multiple POSTs.
- [x] `markCoalesceNextFlush` flag is set and consumed by the next flush; throttle path also clears it.
- [x] `events-manager` marks coalescing only when >1 sequence; single sequence is left alone.
- [x] End-to-end: enqueue via `sendEventsList` → single coalesced fetch with merged events.
- [x] `pnpm --filter @amplitude/session-replay-browser test` → 1017 passing, 100% statements/branches/functions/lines.
- [x] `pnpm lint` (package) clean, `pnpm docs:check` passes.

---

> [!NOTE]
> **Medium Risk**
> Changes flush/merge behavior on init for multi-sequence IDB drains; logic is gated and reuses existing merge paths, but incorrect coalescing could affect delivery or IDB cleanup via onComplete fan-out.
> 
> **Overview**
> Fixes **SR-4660** by coalescing multiple persisted IndexedDB sequences on page load into fewer track API POSTs instead of one request per stored sequence.
> 
> **`sendStoredEvents`** now calls **`markCoalesceNextFlush()`** when more than one sequence is drained, before enqueuing each batch synchronously.
> 
> **`SessionReplayTrackDestination`** adds a drain-specific **`coalesceNextFlush`** flag (separate from throttle **`mergeOnNextFlush`**), **`markCoalesceNextFlush()`** (sets the flag and **`schedule(0)`** so it cannot stick), and **`mergeDrainBacklog`** on flush. Post-throttle merging is refactored to share **`coalesceByIdentity`** with the drain path so merge semantics and **`onComplete`** fan-out stay the same; throttle wins if both flags are set on one flush.
> 
> Coverage includes unit tests for events-manager and track-destination, plus a Playwright e2e that builds a hung-POST backlog, reloads, and asserts a single coalesced POST with all markers and the drain log.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60422bd0d90b751b4bdf4ad53d1205564d9b967a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -222,7 +222,7 @@ export const createEventsManager = async <Type extends EventType>({
       return;
     }
     config.loggerProvider.log(`Draining ${sequencesToSend.length} stored sequence(s) from previous session.`);
-    // SR-4660: these persisted sequences are about to be enqueued back-to-back. Without
+    // These persisted sequences are about to be enqueued back-to-back. Without
     // coalescing they flush as N separate POSTs — a request flood on page load. Mark the
     // imminent flush to merge same-identity batches (the enqueues below are synchronous, so
     // the whole backlog lands in the queue before the deferred flush consumes the flag). Skip
--- a/packages/session-replay-browser/src/events/events-manager.ts
+++ b/packages/session-replay-browser/src/events/events-manager.ts
@@ -222,6 +222,14 @@ export const createEventsManager = async <Type extends EventType>({
       return;
     }
     config.loggerProvider.log(`Draining ${sequencesToSend.length} stored sequence(s) from previous session.`);
+    // SR-4660: these persisted sequences are about to be enqueued back-to-back. Without
+    // coalescing they flush as N separate POSTs — a request flood on page load. Mark the
+    // imminent flush to merge same-identity batches (the enqueues below are synchronous, so
+    // the whole backlog lands in the queue before the deferred flush consumes the flag). Skip
+    // the single-sequence case where there's nothing to merge.
+    if (sequencesToSend.length > 1) {
+      trackDestination.markCoalesceNextFlush();
+    }
     sequencesToSend.forEach((sequence) => {
       sendEventsList({
         sequenceId: sequence.sequenceId,
--- a/packages/session-replay-browser/src/track-destination.ts
+++ b/packages/session-replay-browser/src/track-destination.ts
@@ -101,7 +101,7 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
   // count, so collapsing N queued batches into one POST directly reduces throttle pressure.
   private mergeOnNextFlush = false;
   // Set by markCoalesceNextFlush() before the page-load backlog is enqueued; consumed by
-  // flush() to coalesce the drained persisted sequences (SR-4660). Distinct from
+  // flush() to coalesce the drained persisted sequences. Distinct from
   // mergeOnNextFlush so the drain isn't conflated with a throttle pause for logging.
   private coalesceNextFlush = false;
   // Gates the merge log to once per throttle pause window — mirroring the throttle log's
@@ -226,7 +226,7 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
    * sequences replayed back-to-back on init via sendStoredEvents). Because those enqueues are
    * synchronous and the flush is deferred to the next tick via schedule(0), the whole backlog
    * lands in the queue before the flag is consumed — collapsing N small POSTs into far fewer
-   * and avoiding the request flood observed on page load (SR-4660). Steady-state live capture
+   * and avoiding the request flood observed on page load. Steady-state live capture
    * never sets this flag, so its sending behavior is unchanged.
    *
    * Schedules a flush so the flag is always consumed by the next flush, even when every
@@ -396,7 +396,7 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
   }
 
   /**
-   * Page-load backlog drain path (SR-4660): on init the SDK replays every persisted sequence
+   * Page-load backlog drain path: on init the SDK replays every persisted sequence
    * from a prior session via sendStoredEvents. Enqueued back-to-back they would flush as N
    * separate POSTs — a request flood on page load that feeds volume spikes and throttling.
    * Reuses the exact same identity-grouped merge as the post-throttle path so the backlog
--- a/packages/session-replay-browser/src/track-destination.ts
+++ b/packages/session-replay-browser/src/track-destination.ts
@@ -100,6 +100,10 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
   // flush() to merge same-session contexts before sending. Throttling is enforced by request
   // count, so collapsing N queued batches into one POST directly reduces throttle pressure.
   private mergeOnNextFlush = false;
+  // Set by markCoalesceNextFlush() before the page-load backlog is enqueued; consumed by
+  // flush() to coalesce the drained persisted sequences (SR-4660). Distinct from
+  // mergeOnNextFlush so the drain isn't conflated with a throttle pause for logging.
+  private coalesceNextFlush = false;
   // Gates the merge log to once per throttle pause window — mirroring the throttle log's
   // transition-only gating — so a sustained throttle scenario doesn't spam logs every cycle.
   private mergeLogFiredThisPause = false;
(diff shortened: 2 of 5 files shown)


lewgordon-amplitude:
bugbot run

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -217,6 +217,14 @@ export const createEventsManager = async <Type extends EventType>({
...
     }
     config.loggerProvider.log(`Draining ${sequencesToSend.length} stored sequence(s) from previous session.`);
+    // SR-4660: these persisted sequences are about to be enqueued back-to-back. Without
+    // coalescing they flush as N separate POSTs — a request flood on page load. Mark the
+    // imminent flush to merge same-identity batches (the enqueues below are synchronous, so
+    // the whole backlog lands in the queue before the deferred flush consumes the flag). Skip
+    // the single-sequence case where there's nothing to merge.
+    if (sequencesToSend.length > 1) {
+      trackDestination.markCoalesceNextFlush();
+    }
```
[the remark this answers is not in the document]
lewgordon-amplitude: Fixed in 44b050ef1. `markCoalesceNextFlush()` now self-schedules a flush (`this.schedule(0)`), so `coalesceNextFlush` is always consumed by the next flush even when every backlog sequence is dropped before reaching the queue (e.g. all events oversized) and no enqueue schedules a flush. The self-scheduled flush runs with an empty queue (no POST) and clears the flag, so a later unrelated live flush can't be mis-coalesced as a page-load drain. Added a regression test covering the all-dropped/nothing-enqueued case; package tests pass at 100% coverage.

lewgordon-amplitude:
bugbot run

lewgordon-amplitude:
🚀 **Prerelease published** from RC branch `sr-perf-reliability-rc` (also includes the other two SR perf/reliability PRs).

```
npm install @amplitude/session-replay-browser@1.45.0-sr-perf-reliability-rc.0
```

dist-tag: `sr-perf-reliability-rc` · [Workflow run](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/27021214938)

lewgordon-amplitude:
🚀 **Prerelease published** from RC branch `sr-perf-reliability-rc2`

```
npm install @amplitude/session-replay-browser@1.45.0-sr-perf-reliability-rc2.0
npm install @amplitude/plugin-session-replay-browser@1.32.0-sr-perf-reliability-rc2.0
```

dist-tag: `sr-perf-reliability-rc2` · [workflow run](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/27145091112)

packages/session-replay-browser/test/track-destination.test.ts
```diff
@@ -1486,6 +1488,157 @@ describe('SessionReplayTrackDestination', () => {
...
+        expect(drainLogs).toHaveLength(0);
+      });
+
+      test('drain still splits an oversized merge across multiple POSTs', async () => {
+        const trackDestination = new SessionReplayTrackDestination({ loggerProvider: mockLoggerProvider });
+        const sendSpy = jest.spyOn(trackDestination, 'send').mockResolvedValue(undefined);
+
+        // Each event is just over half the soft cap, so any two together exceed it — the
+        // greedy merge must flush one-per-context, mirroring the throttle path's behavior.
+        const big = 'x'.repeat(800_000);
```
jpollock-ampl: Mind deriving this from `MERGE_AFTER_THROTTLE_SOFT_CAP` like the sibling throttle test (~line 1183) now does?

```ts
const big = 'x'.repeat(Math.floor(MERGE_AFTER_THROTTLE_SOFT_CAP / 2) + 100_000);
```

That test was just changed off the hardcoded `800_000` "so it tracks MAX_EVENT_LIST_SIZE changes", and I think this new one wants the same guarantee. `800_000` only exceeds half the cap (700_000 today) by coincidence; if `MAX_EVENT_LIST_SIZE` ever dropped below ~533k, two of these would stop exceeding the cap and the split assertion would silently pass without testing the split path. Sharing the formula also keeps the two tests' intent matched.

packages/session-replay-browser/src/events/events-manager.ts
```diff
@@ -217,6 +217,14 @@ export const createEventsManager = async <Type extends EventType>({
       return;
     }
     config.loggerProvider.log(`Draining ${sequencesToSend.length} stored sequence(s) from previous session.`);
+    // SR-4660: these persisted sequences are about to be enqueued back-to-back. Without
```
jpollock-ampl: Nit - remove ticket number?

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -299,19 +328,53 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+      this.mergeLogFiredThisPause = true;
+      this.loggerProvider.log(
+        `Session replay throttle pause ended; merged ${list.length} queued batches into ${merged.length} request(s)`,
+      );
+    }
+    return merged;
+  }
+
+  /**
+   * Page-load backlog drain path (SR-4660): on init the SDK replays every persisted sequence
```
jpollock-ampl: Nit - remove ticket number?

jpollock-ampl (approved):

lewgordon-amplitude:
🚀 **Prerelease published** from RC branch `sr-perf-reliability-rc3` (bundles #1806, #1814, #1807, #1808, #1798)

```
npm install @amplitude/plugin-session-replay-browser@sr-perf-reliability-rc3
npm install @amplitude/session-replay-browser@sr-perf-reliability-rc3
```

Versions: plugin `1.32.0-sr-perf-reliability-rc3.0`, standalone `1.45.0-sr-perf-reliability-rc3.0`

[Workflow run](https://github.com/amplitude/Amplitude-TypeScript/actions/runs/27283453804)

packages/session-replay-browser/test/track-destination.test.ts
lewgordon-amplitude: Done in 60422bd — now uses the same `Math.floor(MERGE_AFTER_THROTTLE_SOFT_CAP / 2) + 100_000` formula as the sibling throttle test.

packages/session-replay-browser/src/events/events-manager.ts
lewgordon-amplitude: Removed SR-4660 from the comment in 60422bd.

packages/session-replay-browser/src/track-destination.ts
lewgordon-amplitude: Removed SR-4660 from the comment in 60422bd.

PR #1812, opened 2026-06-05
fix(analytics-browser): only flush on actual offline->online transition

Mercy811:
## Summary

The browser SDK's network connectivity checker plugin scheduled a flush on **every** `online` event, even when the SDK was already online. Repeated `online` events (e.g. the browser firing `online` while connectivity never actually dropped) therefore scheduled redundant flushes, each of which could surface `ERR_NETWORK_CHANGED`.

This adds an early-return guard so the `online` handler only acts on a genuine offline→online transition (it skips when `config.offline` is already `false`).

This mirrors the same guard added to the react-native connectivity plugin in PR #1802 (flagged by Copilot review) — the browser SDK had the identical gap.

```ts
addNetworkListener('online', () => {
  if (config.offline === false) {
    return;
  }
  config.loggerProvider.debug('Network connectivity changed to online.');
  config.offline = false;
  // Flush immediately will cause ERR_NETWORK_CHANGED
  setTimeout(() => {
    amplitude.flush();
  }, config.flushIntervalMillis);
});
```

The `offline` handler is unchanged.

## Tests

- Added `should not flush on online event when already online` to `network-connectivity-checker.test.ts`: starts online, fires `online` twice, advances timers by `flushIntervalMillis`, asserts `flush` was not called.
- Added `should flush when transitioning from offline to online` to the same file to lock in the real-transition path (the guard must not break it).
- Added per-test plugin teardown (`createPlugin` tracker + `afterEach`) so stale `online` listeners from earlier tests don't leak into the flush-count assertions.
- Updated the existing `browser-client.test.ts` "should listen for network change to online" test: it previously relied on the now-removed redundant-flush behavior (it dispatched `online` while already online). It now starts offline (`navigator.onLine === false`) so it exercises a genuine offline→online transition.

## Test plan

- `pnpm --filter @amplitude/analytics-browser run typecheck` — passes
- `pnpm --filter @amplitude/analytics-browser run test` — 508 passed, 508 total (24 suites), 100% coverage maintained

### DIFF
--- a/packages/analytics-browser/src/plugins/network-connectivity-checker.ts
+++ b/packages/analytics-browser/src/plugins/network-connectivity-checker.ts
@@ -42,6 +42,9 @@ export const networkConnectivityCheckerPlugin = (): BeforePlugin => {
     config.offline = !navigator.onLine;
 
     addNetworkListener('online', () => {
+      if (config.offline === false) {
+        return;
+      }
       config.loggerProvider.debug('Network connectivity changed to online.');
       config.offline = false;
       // Flush immediately will cause ERR_NETWORK_CHANGED
--- a/packages/analytics-browser/test/browser-client.test.ts
+++ b/packages/analytics-browser/test/browser-client.test.ts
@@ -876,6 +876,8 @@ describe('browser-client', () => {
 
     test('should listen for network change to online', async () => {
       jest.useFakeTimers();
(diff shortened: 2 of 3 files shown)


daniel-graham-amplitude (approved):
Lgtm!

PR #1816, opened 2026-06-08
ci(examples): smoke-test the React Native sample app on the New Architecture

Mercy811:
## Summary

Follow-up from the [PR #1802 review comment](https://github.com/amplitude/Amplitude-TypeScript/pull/1802#discussion_r3365788055) — ticket [SDKRN-28](https://linear.app/amplitude/issue/SDKRN-28).

The SDK's connectivity native modules added for offline mode (`AmplitudeReactNativeConnectivity` — iOS `RCTEventEmitter`, Android `ReactContextBaseJavaModule`), along with the existing `AmplitudeReactNative` module, are **legacy-bridge** native modules. They should work under React Native's New Architecture (Fabric / TurboModules, bridgeless) via the legacy interop layer, but this hadn't been verified on a New-Architecture build.

This adds New-Architecture build coverage to the existing iOS smoke test by matrixing `rn-smoke.yml` over both architectures. It **reuses the existing sample app** (`examples/react-native/app`) and its `workspace:*` local-SDK linkage, so when the offline-mode PRs (SDKRN-5) merge to `main`, this smoke test picks up the connectivity module automatically.

Per the ticket, the goal is only to confirm the SDK **won't crash** customer apps on the New Architecture — it does not test feature behavior. The SDK degrades gracefully on any native-module failure (`context.ts` uses optional chaining; `remnant-data-migration.ts` guards + try/catch), so a clean launch plus `track()`/`identify()` succeeding is the crash check.

## Changes

- **`.github/workflows/rn-smoke.yml`** — matrix the `ios-smoke` job over `new-arch: [false, true]`. The `true` leg sets `RCT_NEW_ARCH_ENABLED=1`, which `pod install` reads to bake Fabric/TurboModules + bridgeless into the generated Pods project; the `false` leg preserves the existing legacy-bridge coverage. CocoaPods cache key scoped per architecture.
- **`examples/react-native/app/README.md`** — document building each architecture locally.
- **Removed the example app's Jest render test** (`__tests__/App.test.tsx`, `jest.config.js`, and the `test` script). It only verified that `<App/>` mounts in JS with a fully-mocked SDK; the Maestro smoke flow (`.maestro/smoke.yaml`) already covers app launch + `init`/`track`/`identify` on a real simulator. Removing the `test` script also drops the inferred nx `test` target, so editing this example no longer pulls a (broken-without-a-built-SDK) Jest run into `nx affected --target=test`. Jest-only devDependencies are left in place to avoid an unrelated full `pnpm-lock.yaml` reformat (better pruned separately).

## Verification

Built and ran the sample app locally on the New Architecture (iPhone 16 simulator, iOS 18.6):

- `RCT_NEW_ARCH_ENABLED=1 bundle exec pod install` → codegen ran; `RCT_NEW_ARCH_ENABLED=1` set 46× in the generated Pods project (genuinely bridgeless, not silently old-arch).
- `xcodebuild` Release (simulator) → `** BUILD SUCCEEDED **`.
- Maestro smoke flow → app launches; `track()` / `identify()` run without crashing.

Add a single trailing empty line to **`packages/analytics-react-native/README.md`** to trigger the GHA. 

## Notes

- **iOS-only**, matching the existing workflow. Android New Architecture is documented as a local/manual build; there is no Android CI job.
- The new-arch leg is a cold build, so keep an eye on the 45-minute job timeout.
- `matrix.new-arch` uses a hyphenated key on purpose — dot dereference of hyphenated property names is valid per the [Contexts reference](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions) (and matches the existing `matrix.node-version` usage in `ci-nx.yml`).

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI matrixing, example-app docs/test removal, and a README whitespace tweak; no SDK runtime or auth/data logic is modified.
> 
> **Overview**
> **CI now smoke-tests the React Native example on both legacy and New Architecture (Fabric/TurboModules).** The `ios-smoke` job in `rn-smoke.yml` uses a `new-arch: [false, true]` matrix, sets `RCT_NEW_ARCH_ENABLED` before `pod install` so Pods/codegen match the architecture, and scopes the CocoaPods cache key per matrix leg.
> 
> The example app README adds **local New Architecture** build steps (iOS `RCT_NEW_ARCH_ENABLED=1 pod install`, Android `gradle.properties`) and notes that CI runs Maestro once per architecture.
> 
> **Jest coverage for the sample app is removed** (`App.test.tsx`, `jest.config.js`, and the `package.json` `test` script), so edits to the example no longer pull a mocked render test into nx affected `test` runs; Maestro remains the real simulator crash check. A trivial trailing newline was added to `packages/analytics-react-native/README.md` to exercise the workflow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 462ec11ead18e674f060004e1f1167b2b4215d30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mercy811:
Thanks @Copilot — I verified this against the official GitHub docs and it's a false positive, so I'm keeping the `new-arch` matrix key as-is (this applies to all four comments, which share the same claim).

Per the [Contexts reference](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions):

> In order to use property dereference syntax, the property name must start with a letter or `_` and contain only alphanumeric characters, `-`, or `_`.

`new-arch` starts with a letter and contains only alphanumerics and `-`, so `matrix.new-arch` is a valid property dereference — the `-` is lexed as part of the property name, **not** as subtraction.

In-repo precedent: [`.github/workflows/ci-nx.yml`](../blob/main/.github/workflows/ci-nx.yml) already references `${{ matrix.node-version }}` (also hyphenated) with dot notation, and its `Build` matrix legs run correctly on 20.x / 22.x / 24.x.

So no rename is needed. (Aside: this job is gated on `@amplitude/analytics-react-native` being affected, so the matrix doesn't execute on this CI-only PR — it runs on SDK-affecting PRs.)

daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude (approved):

.github/workflows/rn-smoke.yml
```diff
@@ -54,16 +54,25 @@ jobs:
...
+        new-arch: [false, true]
+
     env:
       APP_DIR: examples/react-native/app
       BUNDLE_ID: org.reactjs.native.example.app
       SIM_NAME: iPhone 15
+      # Baked into the generated Pods project by `pod install` (codegen + the
+      # RCT_NEW_ARCH_ENABLED compile flag), so it must be set for that step;
+      # xcodebuild only compiles what pod install produced.
+      RCT_NEW_ARCH_ENABLED: ${{ matrix.new-arch && '1' || '0' }}
```
daniel-graham-amplitude, line 75: Are you sure that this variable is taking effect in `pod install`? I just copied the results of both `pod install` (new arch, old arch) and did a diff and the contents were identical.

Mercy811: whether to run an app via new arch and use interop is decided at runtime by checking `RCT_NEW_ARCH_ENABLED` value

Mercy811: the app sizes are different. Now sure which files you compared and found them identical
image

daniel-graham-amplitude: I compared the log outputs of `pod install`.

daniel-graham-amplitude: > whether to run an app via new arch and use interop is decided at runtime by checking RCT_NEW_ARCH_ENABLED value

Isn't at also a build-time operation? https://reactnative.dev/architecture/landing-page#ios

Mercy811: I didn't compare logs but the app size is pretty straight forward evidence that they are not identical

Mercy811: that's right setting the configuration is at build time like what we updated in this PR. The result compiled app differs on the value of the flag

Mercy811: > whether to run an app via new arch and use interop is decided at runtime by checking RCT_NEW_ARCH_ENABLED value

UPD: `RCT_NEW_ARCH_ENABLED` is evaluated at build time to decide the value of `newArchEnabled` at https://github.com/facebook/react-native/blob/31f9d45074f069151c155836335e0edff87f70bd/packages/react-native/Libraries/AppDelegate/RCTAppDelegate.mm#L151-L158 which is inferred at runtime

daniel-graham-amplitude: Okay should be good then. I just had a look and one way to check it is like this: `!!global.nativeFabricUIManager`

daniel-graham-amplitude: I'll leave that as optional though.

daniel-graham-amplitude (approved):

.github/workflows/rn-smoke.yml
Mercy811: `!!global.nativeFabricUIManager` seems to be indirect check on web. But should do the work. Fabric is for UI. But once new arch is on, everything should be on including Fabric, Turbo, JSI and interop.

PR #1817, opened 2026-06-08
feat(session-replay-browser): custom transport hooks for authenticated proxies (SR-4497)

bravecod:
## Summary

Adds two optional callbacks to the Session Replay Browser SDK that let customers fully own the outbound HTTP call — for example, attach a JWT `Authorization` header and route through an authenticating proxy — while the SDK keeps batching, retry/backoff, URL resolution, compression, serialization and error handling.

- **`handleSendEvents`** — replaces the SDK's internal `fetch` for replay event uploads (POST).
- **`handleFetchConfig`** — replaces the SDK's internal `fetch` for the remote-config fetch (GET).

Both default to the existing internal `fetch`, so **existing integrations are unchanged**. The callback receives a fully-formed request (resolved URL respecting `trackServerUrl` / `configServerUrl`, default headers, serialized body) and its only job is to execute it and return a `Response`. The callback is invoked **once per attempt**; retry stays in the SDK around it.

Closes [SR-4497](https://linear.app/amplitude/issue/SR-4497).

### Motivation

A customer is blocked deploying Session Replay for SaaS because their security team requires JWT auth on all outbound SR requests. `trackServerUrl` / `configServerUrl` can only redirect the endpoint, not modify the request. These hooks let the customer attach auth and forward through their own proxy, which validates the JWT and forwards to Amplitude. Amplitude never validates the JWT — auth happens entirely in the customer's proxy.

## API

```ts
type SessionReplayOptions = {
  // ...existing options
  handleSendEvents?: (request: SendEventsRequest) => Promise<Response>;
  handleFetchConfig?: (request: FetchConfigRequest) => Promise<Response>;
};
```

```ts
sessionReplay.init(API_KEY, {
  trackServerUrl: 'https://my-proxy.example.com/sr-events',
  handleSendEvents: async ({ url, method, headers, body, keepalive }) =>
    fetch(url, { method, headers: { ...headers, Authorization: `Bearer ${getJwt()}` }, body, keepalive }),
  handleFetchConfig: async ({ url, method, headers, signal }) =>
    fetch(url, { method, headers: { ...headers, Authorization: `Bearer ${getJwt()}` }, signal }),
});
```

## Notable design points

- **Web Worker mode (`useWebWorker`)** — a function can't cross `postMessage`, so the worker keeps compressing off the main thread and delegates **only the network call** back to the main thread via a `fetch-request` / `fetch-response` protocol. Retry stays inside the worker, so worker and main-thread retry semantics are identical and the callback is still invoked once per attempt.
- **Page exit** — `navigator.sendBeacon` can't carry custom headers, so when `handleSendEvents` is set the final unload batch is routed through the callback with `keepalive: true` instead, preserving auth on exit. With no callback, the beacon path is unchanged.
- **`handleFetchConfig`** is wired through `RemoteConfigClient` via a new optional, **instance-scoped** `customFetch` constructor param in `analytics-core`. Other `RemoteConfigClient` consumers pass nothing and are unaffected.
- **Compressed body** — when transport compression is active, `body` is the gzipped `Uint8Array` and `headers` already includes `Content-Encoding: gzip`; the callback forwards both unchanged.
- **Plugin pass-through** — added to `plugin-session-replay-browser`; the segment plugin inherits it via the standalone options type (spread).

## Backward compatibility

- All new options are optional and default to the verbatim existing `fetch` path.
- New `RemoteConfigClient` constructor param is appended and optional.
- The full existing test suites pass with **no assertion changes** (the only test edit adds the new optional 5th constructor arg to two `RemoteConfigClient` call assertions).

## Testing

- **Unit** — new `custom-transport.test.ts` (callback substitution, retry contract, page-exit path, worker delegation main-thread side), worker delegation protocol tests, and `RemoteConfigClient` `customFetch` tests. Both packages pass at **100% coverage** (session-replay-browser: 1029 tests; analytics-core: 799 tests). Plugin suites green.
- **End-to-end** — added `test-server/session-replay-browser/sr-custom-transport.html` and a `/api/sr-echo` echo endpoint in `mock-api.js`. Verified in a real browser that both the config GET and event-track POST carry the injected `Authorization: Bearer <jwt>` header, in **both** main-thread and `useWebWorker: true` modes (42/42 outbound requests carried the JWT, 0 missing).

## Out of scope (follow-ups)

- Same hooks in `@amplitude/unified`, the Node SDK, and other SR transports.
- A `getHeaders()` convenience wrapper on top of `handleSendEvents`.

---

> [!NOTE]
> **Medium Risk**
> Changes the network path for all Session Replay outbound traffic when hooks are enabled (including unload and worker delegation), but defaults preserve existing `fetch`/beacon behavior and coverage is extensive.
> 
> **Overview**
> Adds optional **`handleSendEvents`** and **`handleFetchConfig`** hooks so customers can run replay uploads and remote-config fetches through their own HTTP layer (e.g. JWT + proxy) while the SDK still handles batching, compression, retries, and URL resolution.
> 
> **`analytics-core`:** `RemoteConfigClient` accepts an optional `customFetch` and uses it for the config GET when provided; types are exported.
> 
> **`session-replay-browser`:** Main-thread sends, page-exit batches, and interaction beacons that used `sendBeacon`/XHR now go through `handleSendEvents` with `keepalive` when set so custom headers work on unload. With **`useWebWorker`**, the worker delegates each network attempt to the main thread via a `fetch-request` / `fetch-response` protocol.
> 
> **`plugin-session-replay-browser`:** Forwards both hooks into standalone init options and documents them on plugin options.
> 
> Also adds README proxy guidance, unit tests (`custom-transport.test.ts`, worker delegation, `RemoteConfigClient` custom fetch), and a test-server echo page for manual JWT verification. Monorepo packages get prerelease version bumps and CDN snippet integrity updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3e369bd8ee2ead536e7316d133c281139df3df7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/session-replay-browser/src/worker/track-destination.ts
```diff
@@ -31,9 +31,85 @@ interface SendContext {
...
+  type: 'fetch-response';
+  requestId: string;
+  status: number;
+  skipHeader: string | null;
+  body: string;
+  // True when the main-thread transport threw / rejected.
+  error?: boolean;
+}
+
+let delegationCounter = 0;
```
jpollock-ampl, line 78: Just to make sure I understand - `delegationCounter` and `pendingDelegations` are module-level globals inside the worker bundle. This is fine as long as there's only ever one worker instance. Is there a path where `SessionReplayTrackDestination` could be instantiated more than once within the same worker global (e.g. during `shutdown` and re-`init` in a SPA)?

If the old worker is terminated and a new one is spun up it gets a fresh module scope, so that'd be fine - but if the same worker persists across re-inits, pending delegations from the first session could match request IDs from the second. Worth adding a short comment confirming the assumption, or a `pendingDelegations.clear()` on worker teardown.

packages/session-replay-browser/src/worker/track-destination.ts
```diff
@@ -31,9 +31,85 @@ interface SendContext {
...
+    pendingDelegations.set(requestId, (resp) => {
+      if (resp.error) {
+        reject(new Error(resp.body));
+        return;
+      }
+      // Reconstruct just the slice of Response that doFetch consumes. get() is only ever
+      // queried for EVENT_SKIPPED_HEADER, so it returns the forwarded skip header directly.
+      resolve({
+        status: resp.status,
+        headers: { get: () => resp.skipHeader },
```
jpollock-ampl: The reconstructed `headers` object returns `resp.skipHeader` for any header name, not just `EVENT_SKIPPED_HEADER`. The comment above says `get()` is only ever queried for that one header, which is true today - but this is a latent bug if `doFetch`'s header-reading ever expands. Could we narrow it to match the comment?

```ts
headers: { get: (name: string) => name === EVENT_SKIPPED_HEADER ? resp.skipHeader : null },
```

This also makes it self-documenting so the next reader doesn't have to cross-reference the comment to understand the invariant.

```suggestion
        headers: { get: (name: string) => name === EVENT_SKIPPED_HEADER ? resp.skipHeader : null },
```

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -456,6 +518,43 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+  // Runs the custom transport on the main thread for a single attempt the worker delegated,
+  // then posts the result back into the worker's retry loop as a 'fetch-response'. We read only
+  // what the worker's status handling needs: the skip header on a 2xx and the body text on a
+  // 413 (for WAF detection). A thrown/rejected transport is reported as an error so the worker
+  // surfaces it the same way a thrown fetch would (no retry) — matching the main-thread path.
+  private async handleDelegatedFetch(worker: Worker, msg: WorkerFetchRequestMessage): Promise<void> {
+    const { requestId, url, method, headers, body, keepalive } = msg;
+    try {
+      const res = this.handleSendEvents
+        ? await this.handleSendEvents({ url, method, headers, body, keepalive })
```
jpollock-ampl, line 544: Is the intent here for `handleDelegatedFetch` to be reachable when `this.handleSendEvents` is null, or is it always called with a transport configured?

If I'm reading this right, `useCustomTransport` is only set to `true` when `!!this.handleSendEvents` (line ~495), so a `fetch-request` message should never arrive when the transport is absent. That makes the `else` branch (falling back to bare `fetch`) look unreachable in practice.

Could we either assert/throw instead of silently falling back, or document why the defensive fallback is intentional? Silent fallback to unauthed fetch when the customer expected their transport would be confusing to debug.

packages/session-replay-browser/src/config/types.ts
```diff
@@ -92,6 +92,64 @@ export type UGCFilterRule = {
...
+   * SDK-supplied default headers. When the body was gzipped, this includes
+   * `Content-Encoding: gzip` — drop it only if you also stop forwarding the gzipped body,
+   * otherwise the payload will not decode server-side.
+   */
+  headers: Record<string, string>;
+  /**
+   * The serialized payload, ready to send. A gzipped `Uint8Array` when transport compression
+   * is active, otherwise the raw JSON string. Forward it unchanged — do not re-serialize.
+   */
+  body: BodyInit;
```
jpollock-ampl: The `body: BodyInit` type allows `Blob`, `FormData`, `URLSearchParams`, and `ReadableStream` - none of which are structured-cloneable, so they can't cross `postMessage`. In practice the body is always a `string` or `Uint8Array`, both of which transfer fine.

Would we be willing to tighten this to `string | Uint8Array` (or `string | ArrayBufferView`)? That would let TypeScript catch a future regression where someone passes a non-transferable body type, and it matches the constraint documented in the README.

packages/plugin-session-replay-browser/src/typings/session-replay.ts
```diff
@@ -189,4 +189,15 @@ export interface SessionReplayOptions {
    * @see {@link StandaloneSessionReplayOptions.flushIntervalConfig}
    */
   flushIntervalConfig?: FlushIntervalConfig;
+  /**
+   * Optional custom transport for replay event uploads. Use to attach custom auth (e.g. a JWT
+   * `Authorization` header) and route through an authenticated proxy.
+   * @see {@link StandaloneSessionReplayOptions.handleSendEvents}
+   */
+  handleSendEvents?: StandaloneSessionReplayOptions['handleSendEvents'];
```
jpollock-ampl: Using indexed access (`StandaloneSessionReplayOptions['handleSendEvents']`) means this type silently becomes `undefined` if the property is ever renamed or removed from the standalone options - the option would still exist in the plugin interface but accept any value, which would be a hard-to-spot regression.

Could we import `SessionReplaySendEventsHandler` and `SessionReplayFetchConfigHandler` directly from `@amplitude/session-replay-browser` and reference them here? That makes the dependency explicit and surfaces rename/removal errors immediately.

packages/session-replay-browser/test/custom-transport.test.ts
```diff
@@ -0,0 +1,352 @@
...
+  timeout: 0,
+  flushMaxRetries: 2,
+  deviceId: '1a2b3c',
+  sampleRate: 1,
+  serverZone: ServerZone.US,
+  type: 'replay',
+  onComplete: jest.fn(),
+});
+
+describe('custom transport (handleSendEvents)', () => {
```
jpollock-ampl, line 31: Mind adding a test that verifies `handleFetchConfig` flows end-to-end from `SessionReplayOptions` through `createSessionReplayJoinedConfigGenerator` down to `RemoteConfigClient`? The analytics-core tests cover the client side, and the joined-config test checks that the constructor arg is passed - but there's no test in the SR browser package confirming the option actually wires through when you call the top-level `init()` with `handleFetchConfig` set.

Rough sketch (adjust imports/fixtures as needed):

```ts
it('passes handleFetchConfig to RemoteConfigClient when set', async () => {
  // Arrange
  const handleFetchConfig = jest.fn(() =>
    Promise.resolve({ ok: true, json: () => Promise.resolve({}) } as Response)
  );

  // Act
  await createSessionReplayJoinedConfigGenerator('test-key', {
    ...mockOptions,
    handleFetchConfig,
  });

  // Assert
  expect(MockedRemoteConfigClient).toHaveBeenCalledWith(
    'test-key',
    expect.anything(),
    expect.anything(),
    expect.anything(),
    handleFetchConfig,
  );
});
```

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -147,6 +171,10 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
               pending.resolve();
               this.pendingWorkerRequests.delete(msg.id);
             }
+          } else if (msg.type === 'fetch-request') {
+            // Worker is delegating one network attempt to the custom transport (which lives on
+            // the main thread). Run it and post the result back into the worker's retry loop.
+            void this.handleDelegatedFetch(worker, msg);
```
jpollock-ampl: Building on the unhandled rejection flagged by cursor[bot] at line 177 - since `this.loggerProvider` is in scope here, we could log the failure rather than swallowing it silently:

```ts
void this.handleDelegatedFetch(worker, msg).catch((e) => {
  this.loggerProvider.warn('Failed to handle delegated session replay fetch:', e);
});
```

This means errors show up in SDK debug logs rather than disappearing, which aligns with how the rest of the retry/error path surfaces failures.

```suggestion
            void this.handleDelegatedFetch(worker, msg).catch((e) => {
              this.loggerProvider.warn('Failed to handle delegated session replay fetch:', e);
            });
```

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -456,6 +518,43 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+      if (status === 413) {
+        try {
+          responseBody = await res.text();
+        } catch {
+          // best effort
+        }
+      }
+      worker.postMessage({ type: 'fetch-response', requestId, status, skipHeader, body: responseBody });
+    } catch (e) {
+      worker.postMessage({
```
jxiwang, line 563: Some debug log here might be useful

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -210,6 +238,36 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+      try {
+        // Fire-and-forget: we cannot await during unload. The request is well under 64 KB
+        // (trimmed above) and keepalive: true is requested so it survives page teardown.
+        void this.handleSendEvents({ url: exitUrl, method: 'POST', headers, body: payload, keepalive: true }).catch(
+          () => {
+            // best effort on exit
+          },
+        );
+      } catch {
+        // best effort on exit
```
jxiwang: debug log here would be useful

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -210,6 +238,36 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+        'Content-Type': 'application/json',
+        Accept: '*/*',
+        Authorization: `Bearer ${apiKey}`,
+      };
+      try {
+        // Fire-and-forget: we cannot await during unload. The request is well under 64 KB
+        // (trimmed above) and keepalive: true is requested so it survives page teardown.
+        void this.handleSendEvents({ url: exitUrl, method: 'POST', headers, body: payload, keepalive: true }).catch(
+          () => {
+            // best effort on exit
```
jxiwang: debug log would be useful here

jxiwang (approved):

packages/session-replay-browser/src/config/types.ts
```diff
@@ -92,6 +92,64 @@ export type UGCFilterRule = {
...
+  method: 'GET';
+  /** SDK-supplied default headers. */
+  headers: Record<string, string>;
+  /** Abort signal the SDK uses to enforce the remote-config fetch timeout. Honor it in your fetch. */
+  signal?: AbortSignal;
+}
+
+/**
+ * Custom transport for replay event uploads. Invoked by the SDK in place of its internal
+ * `fetch`. Must return a `Response` (or Response-like with `ok`, `status`, `text()`). Called
```
jpollock-ampl, line 145: Thinking out loud - Does needing to return a `Response` tightly couple us to `fetch()`?  What if they want to use `axios` or another library?

packages/session-replay-browser/src/track-destination.ts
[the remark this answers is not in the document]
bravecod: Fixed in `583df80d`. The fire-and-forget call now has a `.catch` that surfaces the failure through `loggerProvider.warn` instead of leaving an unhandled rejection.

packages/session-replay-browser/src/track-destination.ts
```diff
@@ -210,6 +238,36 @@ export class SessionReplayTrackDestination implements AmplitudeSessionReplayTrac
...
+          },
```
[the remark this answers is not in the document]
bravecod: Fixed in `583df80d`. Both the async `.catch` and the synchronous `catch` on the exit path now call `loggerProvider.warn` so a failed exit upload leaves a diagnostic signal. Added debug logging on this path too in `007e1187`.

packages/session-replay-browser/src/worker/track-destination.ts
bravecod: Confirmed and documented in `583df80d`. Each `init()` constructs a new `SessionReplayTrackDestination`, which spins up a fresh `Worker` (and a SPA `shutdown()`+re-`init()` likewise creates a new one), so every session gets a brand-new module scope — `delegationCounter`/`pendingDelegations` are never shared across sessions and requestIds cannot collide. Added a comment stating this so the assumption is explicit; no `clear()` needed.

packages/session-replay-browser/src/worker/track-destination.ts
bravecod: Done in `583df80d`. Narrowed `get` to `name === EVENT_SKIPPED_HEADER ? resp.skipHeader : null` per your suggestion. Since `doFetch` only ever queries that one header, the `: null` branch is currently unreachable, so it is marked `/* istanbul ignore next */` to keep the 100% branch-coverage gate green while staying self-documenting.

packages/session-replay-browser/src/track-destination.ts
bravecod: Documented in `583df80d`. `useCustomTransport` is `!!this.handleSendEvents`, and the worker only emits `fetch-request` when that is true, so the built-in `fetch` branch is unreachable in practice. Added a comment clarifying it is a defensive guard (so we never silently drop a delegated send), not a supported "run without a transport" path.

packages/session-replay-browser/src/config/types.ts
bravecod: Done in `583df80d`. Tightened `body` from `BodyInit` to `string | Uint8Array` on `SendEventsRequest` and the internal worker request types, with a comment noting the structured-clone (postMessage) constraint. TypeScript will now flag a future regression that introduces a non-cloneable body type.

packages/plugin-session-replay-browser/src/typings/session-replay.ts
bravecod: Done in `583df80d`. The plugin typings now import `SessionReplaySendEventsHandler` / `SessionReplayFetchConfigHandler` directly from `@amplitude/session-replay-browser` instead of using indexed access, so a rename/removal in the standalone options surfaces as a compile error.

packages/session-replay-browser/test/custom-transport.test.ts
bravecod: Added in `583df80d`. New test in `joined-config.test.ts` asserts that `handleFetchConfig` is forwarded as the `customFetch` (5th) constructor arg to `RemoteConfigClient` from `createSessionReplayJoinedConfigGenerator`.

packages/session-replay-browser/src/track-destination.ts
bravecod: Done in `583df80d` — applied your suggested `this.loggerProvider.warn(...)` in the catch.

packages/session-replay-browser/src/track-destination.ts
bravecod: Added in `007e1187`. `handleDelegatedFetch` now logs at debug on both the success (status) and failure paths of the delegated fetch.

packages/session-replay-browser/src/track-destination.ts
bravecod: Added in `007e1187`. The exit path now emits a debug log when routing a batch through the custom transport, plus `warn` on failure (per Bugbot).

packages/session-replay-browser/src/track-destination.ts
bravecod: Added in `007e1187` — debug log on the custom-transport exit path (and warn on failure).

packages/session-replay-browser/src/config/types.ts
bravecod: Good question. Today the contract asks for a `Response` (or a Response-like object), but the SDK only ever reads three things from it: `status`, `headers.get("X-Session-Replay-Event-Skipped")` on a 2xx, and `text()` on a 413. So it is not strictly tied to `fetch` — a customer using axios (or any client) can return a thin adapter, e.g. `{ status: r.status, headers: { get: (n) => r.headers[n] ?? null }, text: async () => JSON.stringify(r.data) }`. The README documents the "Response-like with ok/status/text()" shape for exactly this reason.

That said, requiring callers to shim does couple the ergonomics to the WHATWG `Response`. If we want to fully decouple, we could accept a structured result instead (e.g. `{ status, skipHeader?, body? }`) and reconstruct internally. That is a broader API change, so I left it out of this PR — happy to track it as a follow-up if you think the axios/other-client case is common enough to warrant a first-class shape. Leaving this thread open for your call.

PR #1823, opened 2026-06-10
test(analytics-react-native): native iOS XCTest coverage for connectivity module (SDKRN-5) [2.1/5]

Mercy811:
## Summary
Part of [SDKRN-23](https://linear.app/amplitude/issue/SDKRN-23/25-ios-native-connectivity-module) (iOS native connectivity module).

**PR 2.1 of 5** in the SDKRN-5 offline-mode stack. **Stacked on #1819** (base branch `sdkrn-5-offline-2-ios`).

Adds **native iOS XCTest coverage** for `AmplitudeReactNativeConnectivity` — the follow-up promised in #1819. The key behavior under test: **`NWPathMonitor` delivers the current path as its first update once monitoring starts**, the contract the optimistic `{isConnected: true}` seed in `getNetworkConnectivityStatus` relies on.

### Where the tests live (and why)
The test source lives in the SDK package — `packages/analytics-react-native/ios/Tests/AmplitudeConnectivityTests.swift` — but is compiled and run by the example app's existing `appTests` XCTest target, which acts purely as host/runner. A standalone run inside the package isn't possible: the module subclasses `RCTEventEmitter`, so the tests need React-Core compiled/linked, and RN library pods can't build outside an RN app's pod ecosystem (this is why `create-react-native-library` runs native tests via the example app too). `s.exclude_files = "ios/Tests/**"` keeps the XCTest sources out of the shipped pod (verified: absent from the generated Pods project).

### Changes
- **New tests** (4): seed resolves connected; `startObserving()` → initial `AmplitudeNetworkConnectivityChanged` with `isConnected == true` (real `NWPathMonitor`); no events after `stopObserving()`; `supportedEvents`. An `EventCapturingConnectivity` subclass overrides `sendEvent(withName:body:)` to capture events without a live `RCTBridge`.
- **Module visibility only, no behavior change**: `class` → `open class`; overridden members marked `open`/`public` so the test target can subclass/call them (Swift also requires the static `requiresMainQueueSetup` override to be `public` once the class is open).
- **Example app**: stale RN template `appTests.m` ("Welcome to React", never rendered by this app, Metro-dependent) deleted; the SDK test file wired into `appTests` by file reference; `SWIFT_VERSION = 5.0` on the `appTests` configs. `Podfile.lock` refreshed (podspec checksum + stale 1.5.54 → 1.5.56 version catch-up).
- **CI**: `rn-smoke.yml` runs `xcodebuild test` after the iOS build on **both arch legs**, reusing the same `-derivedDataPath` so pod compilation is shared; Release config so the host app uses the embedded JS bundle (no Metro).

| PR | Adds | Base |
|----|------|------|
| #1802 | JS plugin + wiring + tests + no-op native stubs | `main` |
| #1819 [2/5] | Real iOS module (replaces iOS stub) | #1802 |
| **this [2.1/5]** | **Native iOS XCTest coverage** | **#1819** |
| (next) [3/5] | Real Android module (replaces Android stub) | #1819 |
| (next) [4/5] | Robolectric tests | ↑ |
| ~~[5/5]~~ | Example app — **already merged (#1803)** | — |

## Test plan
- Local `xcodebuild test` (Release, iPhone 16 simulator): **all 4 tests pass**.
- Negative check: with `monitor.start(queue:)` commented out, `testEmitsInitialPathUpdateAfterStartObserving` fails at its 10s timeout — the suite genuinely exercises the monitor contract.
- `pnpm --filter @amplitude/analytics-react-native test`: jest suite unaffected (108 tests / 11 suites pass).
- rn-smoke CI must pass the new XCTest step on both `new-arch` legs.


.github/workflows/rn-smoke.yml
```diff
@@ -161,6 +161,26 @@ jobs:
...
+        env:
+          AMPLITUDE_API_KEY: ${{ secrets.AMPLITUDE_API_KEY }}
+        run: |
+          xcodebuild test \
+            -workspace ios/app.xcworkspace \
+            -scheme app \
+            -configuration Release \
+            -sdk iphonesimulator \
+            -destination "platform=iOS Simulator,name=$SIM_NAME" \
+            -derivedDataPath build \
```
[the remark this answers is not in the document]
Mercy811: Good catch — fixed in 63a35b33. Dropped the unconditional `simctl boot` and rely on `simctl bootstatus "$UDID" -b`, which boots only if the device isn't already booted and waits for boot to complete either way, so the step works whether or not the XCTest step left the simulator booted.

examples/react-native/app/ios/app.xcodeproj/project.pbxproj
```diff
@@ -45,6 +44,7 @@
 		5DCACB8F33CDC322A6C60F78 /* libPods-app.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "libPods-app.a"; sourceTree = BUILT_PRODUCTS_DIR; };
 		81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = file.storyboard; name = LaunchScreen.storyboard; path = app/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		89C6BE57DB24E9ADA2F236DE /* Pods-app-appTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-app-appTests.release.xcconfig"; path = "Target Support Files/Pods-app-appTests/Pods-app-appTests.release.xcconfig"; sourceTree = "<group>"; };
+		C2A1D4E80F3B4C6A9D2E5F70 /* AmplitudeConnectivityTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = AmplitudeConnectivityTests.swift; path = "../../../../packages/analytics-react-native/ios/Tests/AmplitudeConnectivityTests.swift"; sourceTree = SOURCE_ROOT; };
```
crleona: nit: prefer to use group relative locations vs source root relative

Image

packages/analytics-react-native/ios/Tests/AmplitudeConnectivityTests.swift
```diff
@@ -0,0 +1,112 @@
...
+    // Guards both fields: events are emitted from the module's monitor
+    // queue while tests read and swap the handler from the main thread.
+    private let lock = NSLock()
+    private var capturedEvents: [CapturedEvent] = []
+    private var handler: ((CapturedEvent) -> Void)?
+
+    var events: [CapturedEvent] {
+        lock.lock()
+        defer { lock.unlock() }
+        return capturedEvents
```
crleona: nit  (here and elsewhere in this pr):
```suggestion
        lock.withLock {
            return capturedEvents
        }
```

crleona (approved):

examples/react-native/app/ios/app.xcodeproj/project.pbxproj
Mercy811: Good catch. Fixed.

packages/analytics-react-native/ios/Tests/AmplitudeConnectivityTests.swift
Mercy811: fixed

PR #1814, opened 2026-06-08
fix(session-replay-browser): restore 2 MB MAX_EVENT_LIST_SIZE default

lewgordon-amplitude:
## Summary
- Restores `MAX_EVENT_LIST_SIZE` to **2 MB** (mainline had drifted to 700 KB). This is the events-list byte cap that decides when the store splits a high-activity flush into its own request.
- A 700 KB cap splits a busy-page flush into ~3× more requests than the 2 MB cap the `sr-4646` SR build shipped. SR ingest throttling is **request-count based** (per org/app/**device** rate — nova `SessionReplayServletV2` increments a Redis counter by 1/request, device threshold ~5 rps), so the smaller cap drove extra per-device requests and triggered throttling on high-activity pages (e.g. HubSpot CRM).
- The time-based flush (`flushIntervalConfig`) still controls send cadence; this only affects whether a single large flush is split. 2 MB gzips to ~0.2 MB — far under the server's 10 MB decompressed split threshold — and stays a safe POST body on the no-`CompressionStream` fallback path.
- Tests that pinned the old literal now derive from the constant so they track future cap changes.

## Context
The `sr-perf-reliability` RC was cut from `main`, which had independently reduced this cap to 700 KB, so it missed the `sr-4646` build's 2 MB bump (its "primary steady-state lever for request volume"). This aligns mainline so future GA cuts don't reintroduce the request-rate regression. Runtime override via `maxPersistedEventsSizeBytes` is unchanged.

## Test plan
- [x] `pnpm --filter @amplitude/session-replay-browser test` — 1003 passed, 100% coverage
- [ ] CI green

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Changes default upload batching and ingest request volume for all clients using the default cap; behavior is intentional but can affect throttling, memory/IDB buffering, and uncompressed POST size on older browsers.
> 
> **Overview**
> Raises **`MAX_EVENT_LIST_SIZE`** from **700 KB to 2 MB** in session-replay-browser so the event store splits high-activity flushes into fewer ingest POSTs. The smaller cap had increased per-device request count and contributed to SR ingest **request-rate throttling** on busy pages; flush cadence is unchanged—only the batch byte limit before a split moves.
> 
[...]

PR #1829, opened 2026-06-12
feat: SR-4646 default SR perf/reliability knobs to validated amp-on-amp values [session-replay-browser]

lewgordon-amplitude:
## Summary

Promote the Session Replay perf/reliability configuration validated in the amp-on-amp canary to be the SDK **defaults**. These knobs previously defaulted to conservative/legacy behavior and the validated values were applied at runtime via the `session-replay-sdk-perf-config` Experiment flag. This change bakes the flag's `onenav-prod` payload into the SDK so all consumers get them by default.

Source of truth: `session-replay-sdk-perf-config` **onenav-prod** variant payload, which is exactly:

```
{ storeType: "memory", useWebWorker: true, mergeMutations: true,
  flushIntervalConfig: { minIntervalMs: 1000, maxIntervalMs: 10000 },
  eagerFullSnapshotSend: false, captureFullSnapshotOnFocus: false,
  maxPersistedEventsSizeBytes: 6000000 }
```

`sendTimeoutMs` is **not** present in the payload, so it is left unchanged. After this PR the SDK out-of-the-box defaults match all 7 settings of the validated payload exactly.

Blast radius: all SDK consumers that don't set these explicitly. This is intended; we'll tune per-customer (via the perf-config flag) if issues arise.

## Default changes (old → new)

| Knob | Old default | New default | Source |
|---|---|---|---|
| `eagerFullSnapshotSend` | `true` | `false` | flag (onenav-prod) |
| `performanceConfig.mergeMutations` | `false` | `true` | flag (onenav-prod) |
| `maxPersistedEventsSizeBytes` | `2_000_000` (`MAX_EVENT_LIST_SIZE`, 2 MB) | `6_000_000` (6 MB) | flag (onenav-prod, exact byte count `6000000`) |
| `storeType` | `'memory'` (already on `main`) | `'memory'` (unchanged) | flag (onenav-prod) |
| `useWebWorker` | `false` | `true` | flag (onenav-prod) |
| `captureFullSnapshotOnFocus` | `true` | `false` | flag (onenav-prod) |
| `flushIntervalConfig` | `{ minIntervalMs: 500, maxIntervalMs: 10_000 }` (`MIN_INTERVAL`/`MAX_INTERVAL`) | `{ minIntervalMs: 1000, maxIntervalMs: 10_000 }` | flag (onenav-prod) |
| `sendTimeoutMs` | `10_000` | `10_000` (unchanged) | flag did **not** set it → fallback, left as-is |

### Notes per knob

- **`eagerFullSnapshotSend`** — effective default resolves in `session-replay.ts` (`onFullSnapshotProcessed`). Flipped so the eager per-snapshot network send only fires when explicitly `true`; default-off defers full-snapshot sends to the normal interval/size cadence (snapshot is still compressed + buffered immediately, so ordering and page-exit beacon coverage are preserved).
- **`maxPersistedEventsSizeBytes`** — introduced a dedicated `DEFAULT_MAX_PERSISTED_EVENTS_SIZE_BYTES = 6_000_000` constant rather than editing `MAX_EVENT_LIST_SIZE`, because `MAX_EVENT_LIST_SIZE` is still reused (e.g. `MERGE_AFTER_THROTTLE_SOFT_CAP = 2 * MAX_EVENT_LIST_SIZE`) and as a documented per-batch reference. **This reverses PR #1814's deliberate 2 MB default.** Larger batches mean fewer requests and materially less ingest throttling.
- **`storeType`** — already `'memory'` on `main` (`local-config.ts`), so no functional code change; only the TSDoc was updated to document the default and its trade-off. ⚠️ **`storeType: 'memory'` drops cross-navigation persistence** (unsent events do not survive a full page reload). This is the **highest-risk** change in the bundle.
- **`mergeMutations`** — lives under `performanceConfig`. Changed the default *resolution* (only skip when explicitly `false`) rather than forcing `performanceConfig.enabled`.
- **`useWebWorker`** — resolved in `local-config.ts`. Now defaults to `true` (compression runs off the main thread) when the consumer sets neither the top-level `useWebWorker` nor the legacy `experimental.useWebWorker`. The top-level option wins over the legacy one, and an explicit `false` on either is still honored. The e2e harness was updated so an absent `useWebWorker` param exercises the new default (worker on).
- **`captureFullSnapshotOnFocus`** — class-property default flipped to `false` in `local-config.ts`; `focusListener` honors it by skipping the on-focus full snapshot unless the value is explicitly `true`. Removes focus-driven full-snapshot (and, with eager send, request) storms.
- **`flushIntervalConfig`** — introduced `DEFAULT_FLUSH_MIN_INTERVAL_MS = 1000` / `DEFAULT_FLUSH_MAX_INTERVAL_MS = 10_000` constants; applied as the default when the consumer omits `flushIntervalConfig`. The `minIntervalMs` floor moves from `500` → `1000` to cut request volume on busy pages; `maxIntervalMs` stays `10_000`. `MIN_INTERVAL`/`MAX_INTERVAL` remain the absolute clamp bounds and the partial-config cross-validation defaults, so the existing `min <= max` validation is preserved for consumer-supplied configs.
- **`sendTimeoutMs` / `SEND_TIMEOUT_MS`** — flag did not specify a value, so left at the existing 10000ms default.

TSDoc/`@defaultValue` comments next to each option were updated to state the new default and that it reflects the validated amp-on-amp config.

## Canary validation

amp-on-amp canary (appid 187520): rc3 **99.8%** success / **0.02%** throttle vs prod 1.31.0 **96.7%** success / **2.95%** throttle.

These are now **defaults that can be overridden** per-consumer or via the `session-replay-sdk-perf-config` flag.

## Test plan

- [x] `pnpm --filter @amplitude/session-replay-browser test` — 1090 passing, **100%** statements/branches/functions/lines coverage
- [x] `pnpm --filter @amplitude/plugin-session-replay-browser test` — 57 passing, 100% coverage
- [x] `pnpm --filter @amplitude/session-replay-browser lint` — 0 errors
- [x] `pnpm docs:check` — passes
- [x] Updated existing unit/e2e tests that asserted the old defaults (eager-send wiring, merge-mutations default, batch-split cap, worker-path, focus-snapshot, flush-interval)

Refs SR-4646, SR-4657.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ packages/analytics-browser/playground/react-spa/public/amplitude.js
 .cursor/rules/nx-rules.mdc
 .github/instructions/nx.instructions.md
 .pnpm-store/
+packages/session-replay-browser/e2e/results.json
--- a/packages/session-replay-browser/e2e/README.md
+++ b/packages/session-replay-browser/e2e/README.md
@@ -86,7 +86,7 @@ accepts URL params to configure the SDK:
 | `sampleRate` | `1.0` | Overrides the SDK-level sample rate (remote config mock takes precedence) |
 | `deviceId` | `test-device-id` | Device ID |
 | `logLevel` | `0` | SDK log level (4 = DEBUG, useful for troubleshooting) |
-| `useWebWorker` | _unset_ | Passes `useWebWorker` (`true`/`false`) to `init()`; omitted preserves the SDK default (now `true` post SR-4646, so the worker send path is on by default). Pass `false` explicitly to force the main-thread path. |
+| `useWebWorker` | `false` | Passes `useWebWorker: true` to `init()`, enabling the web worker send path. Note: the SDK default is now `true` (SR-4646); the harness explicitly pins `false` when the param is absent to keep the behavior-focused e2e suite on the deterministic main-thread path. The worker path has dedicated coverage (e.g. `capture.spec.ts`, `send-timeout.spec.ts`). |
 | `eagerFullSnapshotSend` | _unset_ | Passes `eagerFullSnapshotSend` (`true`/`false`) to `init()`; omitted preserves the SDK default (now `false` post SR-4646). Delivery tests that need the prompt initial send pass `true` explicitly. |
 | `captureFullSnapshotOnFocus` | _unset_ | Passes `captureFullSnapshotOnFocus` (`true`/`false`) to `init()`; omitted preserves the SDK default (now `false` post SR-4646). Tests that need the on-focus snapshot pass `true` explicitly. |
 | `maxPersistedEventsSizeBytes` | _unset_ | Passes `maxPersistedEventsSizeBytes` (number) to `init()`; omitted preserves the SDK default |
--- a/packages/session-replay-browser/e2e/README.md
+++ b/packages/session-replay-browser/e2e/README.md
@@ -87,7 +87,7 @@ accepts URL params to configure the SDK:
 | `deviceId` | `test-device-id` | Device ID |
 | `logLevel` | `0` | SDK log level (4 = DEBUG, useful for troubleshooting) |
 | `useWebWorker` | `false` | Passes `useWebWorker: true` to `init()`, enabling the web worker send path |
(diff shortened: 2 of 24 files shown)


stevenchien-amplitude (approved):
lgtm!

PR #1830, opened 2026-06-12
fix(session-replay): no-op setSessionId when session id unchanged

lewgordon-amplitude:
## Summary
- Early-return asyncSetSessionId when sessionId and deviceId are unchanged, avoiding spurious sendEvents/recordEvents/targeting resets.
- Verified: `pnpm --filter @amplitude/session-replay-browser test -- session-replay.test.ts` (288 passed)

Linear: https://linear.app/amplitude/issue/SR2-3635

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -418,6 +418,11 @@ export class SessionReplay implements AmplitudeSessionReplay {
   ) {
     const previousSessionId = this.identifiers?.sessionId;
     const currentDeviceId = this.getDeviceId();
+    // Standalone SDK callers may poll setSessionId with a stable bucket id (e.g. hour-aligned
+    // timestamps) and only need a no-op when the bucket hasn't rolled. Without this guard,
+    // the rest of asyncSetSessionId still runs: sendEvents, targeting reset, config refetch,
+    // and recordEvents (stop + restart rrweb). Proceed when deviceId changes or
+    // userProperties are passed so targeting can re-evaluate on Identify.
     if (
       previousSessionId !== undefined &&
       previousSessionId === sessionId &&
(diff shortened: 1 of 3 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -416,12 +416,21 @@ export class SessionReplay implements AmplitudeSessionReplay {
     deviceId?: string,
     options?: { userProperties?: { [key: string]: any } },
   ) {
+    const previousSessionId = this.identifiers?.sessionId;
+    const currentDeviceId = this.getDeviceId();
+    if (
+      previousSessionId !== undefined &&
+      previousSessionId === sessionId &&
+      (deviceId === undefined || deviceId === currentDeviceId) &&
+      !options?.userProperties
```
jpollock-ampl: Is the intent here to no-op only when no userProperties were passed at all, or to treat any explicit value (even an empty object) as a reason to proceed with the full reset?

Since `!options?.userProperties` is a truthiness check, `{ userProperties: {} }` is truthy and skips the no-op, while `undefined` no-ops. I think that's defensible either way, but an empty object arguably carries no properties and could still no-op. Just want to confirm it's deliberate - please correct me if I'm misreading the intended semantics.

lewgordon-amplitude:
bugbot run

jpollock-ampl (approved):

packages/session-replay-browser/src/session-replay.ts
lewgordon-amplitude: Good catch — empty `{}` should no-op. The intent is to only proceed when there are actual user properties to re-evaluate targeting with (e.g. after an Identify). Updated the guard to:

`(options?.userProperties === undefined || Object.keys(options.userProperties).length === 0)`

Added a unit test for the `{ userProperties: {} }` case in 4902500.

PR #1836, opened 2026-06-15
refactor(analytics-core): add Delayed Event type

daniel-graham-amplitude:
### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core event upload paths, queue removal, and parallel flush behavior; mistakes could drop or duplicate events, though coverage is extensive.
> 
> **Overview**
> Adds **delayed event** handling in the destination plugin: events with `event.delay.id` are grouped by delay id, sent to `{serverUrl}/delayed` (in parallel with normal batches), and serialized as `DelayedPayload` (`id`, `timeout`, `events` vs `instant_events`) after stripping `delay` from wire events.
> 
> **Deduplication / updates:** Re-queuing the same `delay.id` + `insert_id` drops the older queued copy with code `0` and message `Stale event overwritten`, unless that delay id is already in flight—then the replacement is marked `delay.isFresh` so `removeEvents` does not remove it when the in-flight request completes. `request_metadata` is not rotated on delayed sends.
> 
> Types gain `Delay.isFresh` and `DelayedPayload`; tests cover stale/in-flight replacement, mixed flush, and invalid-response index mapping after payload reordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57bc78ff173419e90c79fb6a7491cbfef0125d82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,8 +199,18 @@ export class Destination implements DestinationPlugin {
...
-    this.queue.forEach((context) => (context.timeout === 0 ? list.push(context) : later.push(context)));
+    this.queue.forEach((context) => {
+      if (context.event.delay_id) {
+        delayed[context.event.delay_id] = delayed[context.event.delay_id] || [];
+        delayed[context.event.delay_id].push(context);
+      } else if (context.timeout === 0) {
+        list.push(context);
+      } else {
+        later.push(context);
+      }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is resolved.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,34 +204,77 @@ export class Destination implements DestinationPlugin {
...
-  async send(list: Context[], useRetry = true) {
+  getDelayedEventsBatches(delayed: Record<string, [Delay, Context[]]>, useRetry: boolean) {
+    const eventPromises = [];
+    try {
+      for (const [delay, contexts] of Object.values(delayed)) {
+        const delayedBatches = chunk(contexts, this.config.flushQueueSize);
+        const delayedEventBatch = delayedBatches.reduce(async (promise, batch) => {
+          await promise;
+          return await this.send(batch, useRetry, delay);
+        }, Promise.resolve());
```
Mercy811: If I have different delay IDs, it will send multiple delay requests. IIRC, we wanna consolidate to at most one payload per heartbeat interval?

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,34 +204,77 @@ export class Destination implements DestinationPlugin {
...
+        eventPromises.push(delayedEventBatch);
+      }
+    } catch (e) {}
+    return eventPromises;
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is just to be on the safe side. This should never throw an exception.

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: There could be a future case where we have 2 heartbeats going at once (ie: to get around payload size limits). So I think it's still worthwhile to have partitioning by delay ID.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,50 +215,100 @@ export class Destination implements DestinationPlugin {
...
+    const regularEventBatch = batches.reduce(async (promise, batch) => {
       await promise;
       return await this.send(batch, useRetry);
     }, Promise.resolve());
+    const eventPromises = [regularEventBatch];
+
+    const delayedEventBatches = this.getDelayedEventsBatches(delayed, useRetry);
+    eventPromises.push(...delayedEventBatches);
+
+    await Promise.all(eventPromises);
```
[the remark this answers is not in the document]
daniel-graham-amplitude: The new contract here is that Delayed Events aren't part of the same ordering as regular events and don't get the same guarantees.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -125,6 +127,10 @@ export class Destination implements DestinationPlugin {
         callback: (result: Result) => resolve(result),
         timeout: 0,
       };
+      // remove any delayed events with the same insert_id
+      if (event.delay?.id) {
+        this.queue = this.queue.filter((queuedContext) => queuedContext.event.insert_id !== event.insert_id);
+      }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This has been updated

packages/analytics-core/src/plugins/destination.ts
Mercy811: We've already split events into batches so no need to worry about payload size limit. My concern is that if we have multiple heartbeats going on, it will X times requests

daniel-graham-amplitude: I think preventing multiple heartbeats should be enforced by the Heartbeat class. IMO destination shouldn't think about delay ID.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -125,6 +127,21 @@ export class Destination implements DestinationPlugin {
...
+        this.queue.forEach((queuedContext) => {
+          if (queuedContext.event.insert_id === event.insert_id || queuedContext.event.delay?.id === event.delay?.id) {
+            duplicatedEvents.push(queuedContext);
+          } else {
+            queue.push(queuedContext);
+          }
+        });
+        duplicatedEvents.forEach((context) => context.callback(buildResult(context.event, 0, Status.Overwritten)));
+        this.queue = queue;
+      }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is okay. If it's already in-flight, no need to cancel it. The second payload will just overwrite it on the server side.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,50 +226,103 @@ export class Destination implements DestinationPlugin {
...
 
     this.scheduleEvents(this.queue);
   }
 
-  async send(list: Context[], useRetry = true) {
+  getDelayedEventsBatches(delayed: DelayedEventsById, useRetry: boolean) {
+    const eventPromises = [];
+    try {
+      for (const { delay, contexts } of Object.values(delayed)) {
+        eventPromises.push(this.send(contexts, useRetry, delay));
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is by design. We can't split payloads in the delayed events endpoint. The heartbeat.ts class (next PR) will manage the payload size to ensure that it's well under the limit.

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: This is resolved now. Now it's only request per delay ID.

packages/analytics-core/src/types/event/base-event.ts
```diff
@@ -1,6 +1,11 @@
 import { Plan } from './plan';
 import { IngestionMetadataEventProperty } from './ingestion-metadata';
 
+export interface Delay {
+  id: string;
+  timeout?: number;
```
Mercy811, line 6: Do I understand it right that timeout here is optional because
1. if event.delay.id && !event.delay.timeout, it's an instant event
2. if event.delay.id && event.delay.timeout, it's a delayed event

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,50 +222,101 @@ export class Destination implements DestinationPlugin {
...
+      }
+    } catch (e) {
+      // swallow error
```
Mercy811, line 293: what error can we catch here?

packages/analytics-core/src/types/payload.ts
```diff
@@ -12,3 +12,9 @@ export interface Payload {
   client_upload_time?: string;
   request_metadata?: RequestMetadata;
 }
+
+export interface DelayedPayload extends Payload {
+  id: string;
+  timeout: number;
+  instant_events?: readonly Event[];
+}
```
Mercy811, line 20: we might not be able to extend `Payload` if we need `[{ id: 1, events: [...], instant_events: [...]}, { id: 2, events: [...], instant_events: [...]}]`. Have we decided which way to go? 

I'm more into the consolidated payload structure to prevent "spammy" requests.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -40,6 +42,13 @@ export interface Context {
   timeout: number;
 }
 
+type DelayedEventGroup = {
+  delay: Delay;
+  contexts: Context[];
+};
+
+type DelayedEventsById = Record<string, DelayedEventGroup>;
```
Mercy811: How about this? 

```diff
export interface Context {
  event: Event;
  attempts: number;
  callback: EventCallback;
  timeout: number;
  + delay: Delay; // add delay here
}
// delayed id: context
type DelayedEventsById = Record<string, Context>;
```

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: There's no errors that can be caught, this is just to be extra safe in case a future change introduces an exception.

packages/analytics-core/src/types/event/base-event.ts
daniel-graham-amplitude: That's correct yes

packages/analytics-core/src/types/payload.ts
daniel-graham-amplitude: The first version of the service is is going to just be `{ id: ..., events: [...], ...}`

Mercy811: Synced offline. Backend won't support array in the first version.

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: Change made, thanks!

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -125,12 +130,27 @@ export class Destination implements DestinationPlugin {
...
+      return;
+    }
+    /* istanbul ignore next */
+    this.queue = this.queue.filter((context) => {
+      if (context.event.insert_id === incomingEvent.insert_id && context.event.delay?.id === incomingEvent.delay?.id) {
+        context.callback(buildResult(context.event, 0, 'Stale event overwritten'));
+        return false;
+      }
+      return true;
+    });
```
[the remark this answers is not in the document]
daniel-graham-amplitude: I've fixed this with this commit: https://github.com/amplitude/Amplitude-TypeScript/pull/1836/changes/03bd6e94bcbe28ea1b484e6842b16e3eaca2ffb2

I don't think the impact would have had a high severity, but still incorrect behaviour

(cc: @Mercy811 )

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -125,12 +131,32 @@ export class Destination implements DestinationPlugin {
...
+    if (!incomingEvent.delay?.id) {
+      return;
+    }
+    /* istanbul ignore next */
+    this.queue = this.queue.filter((context) => {
+      if (
+        incomingEvent.delay &&
+        context.event.delay?.id === incomingEvent.delay?.id &&
+        context.event.insert_id === incomingEvent.insert_id &&
+        !this.inFlightDelayedEvents[incomingEvent.delay.id]
```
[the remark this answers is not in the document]
daniel-graham-amplitude: I don't think this is possible. If it's in flight, that delay ID is in flight.

Mercy811:
**Review focus:** Can this PR break delivery of *regular* events (events with no `event.delay.id`)? The bar is: events without a delay id must be routed, batched, and uploaded exactly as on `main`.

**Result: regular events are safe.** I traced every branch the PR touches:

- **`execute`** → `removeStaleDelayedEvents` returns immediately when `!incomingEvent.delay?.id`, so non-delayed events hit the original `queue.push` / `schedule` / `saveEvents` path unchanged.
- **`flush` partition** — a context with `timeout === 0` and no `delay?.id` still lands in `list`, gets chunked by `flushQueueSize`, and is sent by the same sequential reduce as before. That reduce is byte-for-byte the old logic, now just wrapped in `Promise.all([...])`; when there are no delayed events `eventPromises` is `[regularEventBatch]`, so the awaited behavior is identical.
- **`send` with `delay` falsy** — `request_metadata` is still populated and `requestMetadata` is still rotated, `serverUrl` is unchanged, and the new `delay` in the destructure (`const { extra, delay, ...rest }`) is a no-op for events that carry no `delay`. The payload is identical.
- **Mixed flush** — the delayed `send` is invoked synchronously inside `getDelayedEventsBatches` *before* the regular batch's `send` runs (the reduce defers it a microtask), and delayed sends pass `request_metadata: undefined` without resetting it. So the regular send still reads the intact `requestMetadata` and rotates it. No cross-contamination.

So the core invariant holds. ✅

While verifying, I did find real bugs — **all in the delayed path, none affecting regular events:**

**1. [High] In-flight stale events silently drop the replacement event** — `destination.ts` `removeEvents`
`removeEvents` matches by `insert_id`, and the whole stale feature is premised on same-`insert_id` replacement. When the predecessor is **in flight**, `removeStaleDelayedEvents` intentionally keeps *both* the old and new context in the queue — both with the same `insert_id`. When the in-flight send resolves, `fulfillRequest([oldCtx]) → removeEvents([oldCtx])` filters out **every** queued context with that `insert_id`, including the newer event that was deliberately retained. The replacement is never sent and its callback never fires. The "should not return stale result if the event is in flight" test only asserts the *old* event's result, so it passes while masking the drop.

**2. [Low] `request_metadata` is never rotated on delayed-only flushes** — `destination.ts` `send`
Delayed sends pass `request_metadata: undefined` and skip `this.config.requestMetadata = new RequestMetadata()`. If a stretch of flushes contains only delayed events, accumulated metadata is neither sent nor cleared, so the next regular flush carries stale/accumulated metadata. Diagnostic-only, not a delivery issue.

**Nit:** the `/* istanbul ignore next */` on the `removeStaleDelayedEvents` filter suppresses coverage on the core stale logic even though new tests exercise it — worth removing so the branch is actually measured (it would likely have surfaced bug #1).

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -198,50 +232,111 @@ export class Destination implements DestinationPlugin {
...
     // Sequentially resolve to make sure backend receives events in order
-    await batches.reduce(async (promise, batch) => {
+    const regularEventBatch = batches.reduce(async (promise, batch) => {
       await promise;
       return await this.send(batch, useRetry);
     }, Promise.resolve());
+    const eventPromises = [regularEventBatch];
+
+    if (Object.keys(delayed).length > 0) {
+      eventPromises.push(this.getDelayedEventsBatches(delayed, useRetry));
```
Mercy811: eventPromises is an array of promise array. Confirm that promises are still resolved in order.
Image

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -125,12 +131,40 @@ export class Destination implements DestinationPlugin {
         callback: (result: Result) => resolve(result),
         timeout: 0,
       };
+      this.removeStaleDelayedEvents(event);
       this.queue.push(context);
       this.schedule(this.config.flushIntervalMillis);
```
Mercy811, line 136: If we mix delayed events and regular events in the pipeline and only split them when flush, will it affect flush interval? This is hard to observe as default flush interval is 1000 ms

Example: 
timeline 0, delayed event
timeline 800, regular event, expected flush should be 800+1000
timeline 1000, flush

daniel-graham-amplitude: It shouldn't make a difference. The flush interval and flush timeout will be the same for all events.

Mercy811: I don't think I understand it. Don't we call schedule(flush interval) here? The example above demonstrates that the current implementation will result in unexpected flush interval.

daniel-graham-amplitude: I don't think we need to distinguish between delayed events versus regular events to decide the flush interval though? In your example, I would say that it flushes at the expected time which is 1000ms after the first event (delayed or not-delayed).

Mercy811: If we don't distinguish them,flush interval configuration always equals to the fixed heart beat interval. This is a regression

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -438,9 +538,14 @@ export class Destination implements DestinationPlugin {
    */
   removeEvents(eventsToRemove: Context[]) {
     this.queue = this.queue.filter(
-      (queuedContext) => !eventsToRemove.some((context) => context.event.insert_id === queuedContext.event.insert_id),
+      (queuedContext) =>
+        !eventsToRemove.some(
+          (context) => context.event.insert_id === queuedContext.event.insert_id && !queuedContext.event.delay?.isFresh,
+        ),
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This edge case would only cause the same event to be sent 2x which would not cause any issues because it would just overwrite the event with the same value.

packages/analytics-core/src/plugins/destination.ts
daniel-graham-amplitude: The heartbeat interval and the flush interval are different. 

* A heartbeat means that every 60_000ms (1 minute), the delayed events will be added to the queue in memory
* A flush interval means that every 1_000ms (by default), the contents of the queue will be flushed and sent, and the queue will be emptied

So the order of events is:
* Init (queue = [])
* Track `eventA` (queue = [eventA])
* A heartbeat occurs for delayedEventA (queue = [eventA, delayedEventA])
* The flush interval timeout is reached and queue is flushed

daniel-graham-amplitude: With delayed events there's two queues:
1. There's a queue where the event sits and waits until the next heartbeat. While it's waiting, the contents of the event can be updated.
2. When a heartbeat is reached, the delayed event is moved from that queue into the destination queue where it awaits being sent.

Mercy811 (approved):
lgtm since this is merged to a feature branch. Would like to have another review when merging to main.

daniel-graham-amplitude:
bugbot run

daniel-graham-amplitude:
bugbot run

PR #1837, opened 2026-06-16
refactor(analytics-core): implement a Heartbeat class

daniel-graham-amplitude:
### Summary

_Example usage_
```
const startEvent = { event_type: 'Start Event', event_properties: { is_updated: false }};
heartbeat.trackNoDelay(startEvent); // this will be ingested right away 
heartbeat.track({ event_type: 'Stop Event' }); // this will be queued up in a data store
heartbeat.update({ ...startEvent, event_properties: { is_updated: true }}); // update queued up data store
// after 1 hour, startEvent will be ingested due to TTL expiration
```

### Checklist

---

> [!NOTE]
> **Medium Risk**
> New event-ingestion path affects how events are sent and deduplicated via delay metadata; behavior is isolated to opt-in Heartbeat usage but mistakes could drop or duplicate events if misintegrated.
> 
> **Overview**
> Adds a **`Heartbeat`** helper in `analytics-core` that queues events by `insert_id`, attaches a shared **`delay`** (`id` + optional `timeout`) on **`EventOptions`**, and periodically re-sends the queue through **`CoreClient.track`** so delayed ingestion can stay in sync until flush or TTL.
> 
> **`track`** requires `insert_id`, applies default pulse (60s) and delay timeout (1h), coalesces concurrent **`resetHeartbeat`** calls, and enforces a ~4KB cap on the queued payload (warn + reject/rollback on **`update`**). **`trackNoDelay`** strips the timeout for immediate ingest while keeping the same delay group for atomic payloads; **`update`** mutates queued events; **`stop`** clears the timer and queue. **`getHeartbeatInstance`** returns one **`Heartbeat`** per **`CoreClient`**.
> 
> The public API exports **`getHeartbeatInstance`**; callers are not wired in this diff (e.g. video capture still uses plain **`track`** placeholders).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e75fac64d0078d241c91f017db29c729ac103fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analytics-core/src/heartbeat.ts
```diff
@@ -0,0 +1,140 @@
...
+
+type HeartbeatMap = Map<CoreClient, Heartbeat>;
+
+const heartbeatMap: HeartbeatMap = new Map<CoreClient, Heartbeat>();
+
+export function getHeartbeatInstance(client: CoreClient, logger: ILogger): Heartbeat {
+  if (heartbeatMap.has(client)) {
+    return heartbeatMap.get(client)!;
+  } else {
+    // TODO: make the interval and delay timeout configurable via config options
```
Mercy811: If we are sharing a single heartbeat instance, do we still need to make delay time configurable and expose its configuration to customers?

packages/analytics-core/src/heartbeat.ts
```diff
@@ -0,0 +1,140 @@
...
+      const eventPromise = this.client.track(event_type, event_properties, eventOptions).promise;
+
+      // if the event has no timeout, then it's instant ingested and we can
+      // delete it after it was successfully ingested
+      if (this.isInstantEvent(event)) {
+        eventPromise.then(() => this.events.delete(event.insert_id!));
+      }
+      trackedEvents.push(eventPromise);
+    }
+    return await Promise.all(trackedEvents);
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is okay, if it gets sent again when it's already ingested, it will be dropped due to duplicate insert ID.

Mercy811 (commented):
Could you add example usage in PR description? eg: how to track video played as an instant event and video end as a delayed event. I don't quite understand the designed usage of these public methods: `track()`, `trackNoDelay()`, `update()`, `heartbeat()`

packages/analytics-core/src/heartbeat.ts
daniel-graham-amplitude: Probably not actually. We probably need to make this strict.

packages/analytics-core/src/heartbeat.ts
```diff
@@ -0,0 +1,145 @@
...
+    const trackedEvents = [];
+    for (const event of this.events.values()) {
+      const { event_type, event_properties, ...eventOptions } = event;
+      const eventPromise = this.client.track(event_type, event_properties, eventOptions).promise;
+
+      // if the event has no timeout, then it's instant ingested and we can
+      // delete it after it was successfully ingested
+      if (this.isInstantEvent(event)) {
+        eventPromise.then(() => this.events.delete(event.insert_id!));
+      }
```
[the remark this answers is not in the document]
daniel-graham-amplitude: This is fine, this is how we deal with failed ingestion.

packages/analytics-core/src/heartbeat.ts
daniel-graham-amplitude: Done

packages/analytics-core/src/heartbeat.ts
```diff
@@ -0,0 +1,163 @@
...
+import { CoreClient } from './types/client/core-client';
+import { BaseEvent, Delay } from './types/event/base-event';
+import { Result } from './types/result';
+import { UUID } from './utils/uuid';
+
+type DelayedEvent = BaseEvent & {
+  delay: Delay;
+};
+
+const EVENTS_SIZE_LIMIT = 4 * 10_000; // 4KB
```
Mercy811, line 11: could you explain more about this? Looks like we take the json string length as a rough estimation of the size.  standard English letters and numbers are 1 byte, some other language characters are 2 or 3 bytes and emojis are 4 bytes. I'd take `const EVENTS_SIZE_LIMIT = 4 * 1_000` or `const EVENTS_SIZE_LIMIT = 4 * 1_000 / 4; // 4KB` to be safer

daniel-graham-amplitude: That was a typo, it should be 4* 10_0000... and yeah it's not really 4kb just a string length of 4k which should be close enough (even in worst case, it could only reach 16kb which would still be under the limit).

packages/analytics-core/src/heartbeat.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    return await this.resetPromise;
+  }
+
+  /**
+   * An instant event is one that has no delay and gets ingested immediately
+   * @param event
+   * @returns
+   */
+  private isInstantEvent(event: DelayedEvent): event is DelayedEvent & { insert_id: string } {
```
Mercy811, line 77: 👍 like the returns type predicate to propagate the narrower type.

Mercy811 (approved):
Lgtm!

PR #1840, opened 2026-06-17
feat(element-selector): consolidate legacy cssPath and kill-switch routing

jxiwang:
## Summary

Extracts the `@amplitude/element-selector` package changes from #1832 into a standalone PR that can land on `main` independently of the autocapture plugin integration work.

### What's included

- **Canonical `legacyCssPath` walker** — moves the Chromium-DevTools-derived cssPath algorithm into `@amplitude/element-selector` so SDK, dashboard, and extension consumers can share one implementation
- **`generateSelector` top-level helper** — centralizes routing for call sites where the engine may be `null` (dashboard contexts, Chrome extension boot)
- **Kill-switch + safety-net in `engine.generate`** — when `config.enabled` is `false`, emit legacy selectors; catch strategy-chain throws and fall back gracefully
- **Publish package** — removes `"private": true` from `package.json`
- **Tests** — coverage for legacy walker, `generateSelector`, and engine fallback paths
- **Size limit** — bumps `@amplitude/element-selector` cap from 5kb → 7kb to reflect consolidated walker

### What's intentionally excluded (stays in #1832)

- `packages/plugin-autocapture-browser` integration (consuming the package, lifecycle wiring, `element-path.ts` refactor)
- `pnpm-lock.yaml` dependency wiring for autocapture

### Relationship to #1832

This PR contains only the `packages/element-selector/` changes (plus `.size-limit.js`). Once merged, #1832 can be rebased to pick up these changes and focus solely on the autocapture plugin wiring.

### Checklist

- [x] PR title follows [conventional commit](https://www.conventionalcommits.org/) format
- [x] Does your PR have a breaking change?: No
- [x] `pnpm build` and `pnpm test` pass for `@amplitude/element-selector`

<div><a href="https://cursor.com/agents/bc-8613daf7-b133-44de-929f-6ca442178423"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8613daf7-b133-44de-929f-6ca442178423"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/element-selector/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@amplitude/element-selector",
   "version": "0.1.0",
-  "private": true,
```
jpollock-ampl, line 4: Is the intent here to start publishing element-selector standalone (e.g. so the Chrome extension can depend on it directly), or did this slip in? Dropping `private: true` makes the package publishable, but the size-limit comment a few lines over still says "The package isn't published as a standalone bundle; this entry exists to catch raw-library bloat." If we are publishing it, we'd probably want to update that note so the two don't contradict each other.

packages/element-selector/src/generate-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+    // legacy cssPath) and its own try/catch safety net. Trust it.
+    return engine.generate(el);
+  }
+  // Engine-null branch. We don't have a live engine to consult, so we route
+  // straight to the legacy walker — the same selector the SDK would be
+  // emitting if `config.enabled === false`. `config` is accepted for symmetry
+  // with the engine signature and so future versions can extend behavior
+  // without a breaking change.
+  void config;
+  try {
```
jpollock-ampl: Would you mind adding a test that actually exercises this `catch`? I'm fairly sure the existing "returns empty string" test passes a text node, which hits `legacyCssPath`'s internal `nodeType` early-return (returns `''` without throwing), so the catch here never runs. To cover the real throw path we'd need the walker to throw - e.g. stub `legacyCssPath`:
```ts
it('returns empty string when the legacy walker throws', () => {
  const el = document.createElement('div');
  const spy = jest.spyOn(require('../src/legacy-css-path'), 'legacyCssPath').mockImplementation(() => { throw new Error('boom'); });
  expect(generateSelector(el, null, DEFAULT_RESOLVED_CONFIG)).toBe('');
  spy.mockRestore();
});
```

packages/element-selector/src/generate-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+    // The engine has its own kill switch (`config.enabled === false` →
+    // legacy cssPath) and its own try/catch safety net. Trust it.
+    return engine.generate(el);
+  }
+  // Engine-null branch. We don't have a live engine to consult, so we route
+  // straight to the legacy walker — the same selector the SDK would be
+  // emitting if `config.enabled === false`. `config` is accepted for symmetry
+  // with the engine signature and so future versions can extend behavior
+  // without a breaking change.
+  void config;
```
jpollock-ampl, line 76: Just confirming we're comfortable that `config` is intentionally inert in this branch. The way it reads now, a caller could pass a fully-resolved `enabled: true` config with a null engine and silently get a legacy selector, with no signal that their config was ignored. I think that's fine for the known null-engine call sites (boot-time extension, inert dashboard context), but please correct me if there's a case where a caller would expect the config to actually take effect here.

packages/element-selector/src/engine.ts
```diff
@@ -107,4 +127,19 @@ export function createSelectorEngine(
       };
     },
   };
+
+  // Local wrapper so a thrown error in the legacy walker (extremely unlikely
+  // — Element inputs only — but defensible if a host environment is missing
+  // `CSS.escape` or `Node` constants) doesn't escape `generate` and crash the
+  // caller. Returns the empty string in that pathological case; consumers
+  // already coalesce empty results to a sentinel.
+  function safeLegacyCssPath(el: Element): string {
```
jpollock-ampl: WDYT about exporting a single `safeLegacyCssPath(el, logger?)` from legacy-css-path.ts and having both call sites use it? This swallow-warn-fallback-to-empty-string logic is nearly identical to the `try { legacyCssPath } catch { return '' }` block in `generateSelector` - both centralize "never let the legacy walker throw," just with/without a logger. Keeping the guarantee in one place would stop the two from drifting the same way the original cssPath copies did, which is exactly the duplication this PR is killing.

packages/element-selector/package.json
jxiwang: Was `private: true`, so we don't deploy the package but can merge. We are ready to deploy this package now.

bravecod (approved):

PR #1842, opened 2026-06-17
feat(autocapture): route Element Path through the element-selector engine

jxiwang:
### Summary

Integrates `@amplitude/element-selector` into the autocapture plugin at the single seam where selectors are produced: `DataExtractor.getElementPath()`. This is the engine integration only — **no config hashing** (tracked separately).

`getElementPath` → `cssPath` is the one chokepoint that generates every selector string in the package: `[Amplitude] Element Path` (via `getEventProperties` → Element Clicked, action-click, Element Changed, Dead Click, Rage Click, Error Click) and the viewport `[Amplitude] Element Exposed` paths (via `track-exposure.ts`). Swapping it here changes selector output for all of them at once, without touching the individual trackers.

#### Changes
- **`DataExtractor` owns a `SelectorEngine`.** `getElementPath` now calls `engine.generate(element)` instead of the local `cssPath`. It ships **dormant**: with the default config (`enabled: false`) the engine routes through element-selector's legacy walker, which is byte-identical to the existing `libs/element-path.ts#cssPath` (same Chromium-DevTools algorithm; `escapeCssIdentifier` delegates to native `CSS.escape`). So **selector output is unchanged** until an org opts in via remote config.
- **`DataExtractor.updateSelectorConfig(remote, logger)`** applies an `ElementSelectorRemoteConfig` payload to the engine (`resolveSelectorConfig` → `engine.updateConfig`).
- **`subscribeToElementSelectorConfig(config, dataExtractor)`** wires the engine to remote config (key `configs.analyticsSDK.elementSelector`, mirroring the existing `pageActions` subscription). Both the **autocapture** and **frustration** plugins subscribe — so when an org enables the engine, every selector-bearing event (including Rage/Dead/Error Click from the frustration plugin) stays consistent.

#### Behavior / rollout
- Dormant by default → zero behavior change on merge.
- Flip `enabled: true` in remote config → `getElementPath` emits the new strategy-based selectors everywhere; flip back → reverts to legacy `cssPath`. One-config-fetch toggle.
- `libs/element-path.ts` is left in place for now; it can later be deleted in favor of re-exporting element-selector's legacy walker (per the consolidation note in that file).

#### Open items
- **Remote-config key/namespace** (`configs.analyticsSDK.elementSelector`) is assumed from the `pageActions` precedent — confirm against the backend contract.
- Autocapture and frustration each subscribe (two subscriptions); a shared provider could de-dupe later.

#### Verification
- `pnpm --filter @amplitude/plugin-autocapture-browser test` — 409 passed, **100% coverage** maintained (added unit tests for the helper, `updateSelectorConfig`, and the frustration subscription/teardown).
- **New e2e (Playwright) coverage** — `e2e/element-path.spec.ts` + `test-server/element-selector-test.html` click real elements and assert the emitted `[Amplitude] Element Path` in a real browser (exercises `getElementPath` → `engine.generate` → dormant legacy `cssPath`, incl. real `CSS.escape`/DOM). Asserts `button#cta-button` for a stable id and `div#container > section > button:nth-child(2)` for a positional, id-less element. Passing on Chromium.
- `pnpm build`, `pnpm lint` (0 errors), `pnpm docs:check` — pass.
- `size-limit`: analytics-browser and unified bundles within budget (engine adds ~2.2 kB gz).
- Merged latest `main` (which now contains the `element-selector` package); lockfile diff reduced to the single workspace-link entry.

### Checklist

* Does your PR have a breaking change?: No — dormant by default; selector output is byte-identical until an org enables the engine via remote config.

<div><a href="https://cursor.com/agents/bc-2ad772f3-66e6-4cfe-bf21-08c6d1477d25"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2ad772f3-66e6-4cfe-bf21-08c6d1477d25"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/plugin-autocapture-browser/test/data-extractor.test.ts
```diff
@@ -450,6 +450,35 @@ describe('data extractor', () => {
...
+        <div id="container">
+          <button id="test-button">Click me</button>
+        </div>
+      `;
+    });
+
+    test('keeps generating selectors after enabling the engine via remote config', () => {
+      const button = document.getElementById('test-button');
+      dataExtractor.updateSelectorConfig({ enabled: true });
+      const result = dataExtractor.getElementPath(button);
```
jpollock-ampl, line 466: WDYT about asserting the exact selector the enabled engine emits for `#test-button` here, rather than just `length > 0`?

As written, this would pass even if the enabled strategy chain silently returned the wrong selector (or the same legacy output), so it's effectively a "doesn't throw" check. Pinning it to the expected string (e.g. `button#test-button`, or whatever the enabled strategy produces for a stable id) would make it a real guard against the chain regressing. The null-payload test below already does exactly this for the dormant case.

packages/plugin-autocapture-browser/src/element-selector-config.ts
```diff
@@ -0,0 +1,39 @@
...
+  config: BrowserConfig,
+  dataExtractor: DataExtractor,
+): (() => void) | undefined {
+  const remoteConfigClient = config.remoteConfigClient;
+  if (!config.fetchRemoteConfig || !remoteConfigClient) {
+    return undefined;
+  }
+
+  const subscriptionId = remoteConfigClient.subscribe(ELEMENT_SELECTOR_REMOTE_CONFIG_KEY, 'all', (remoteConfig) => {
+    dataExtractor.updateSelectorConfig(remoteConfig as ElementSelectorRemoteConfig | null, config.loggerProvider);
```
jpollock-ampl: Could we confirm `configs.analyticsSDK.elementSelector` is the exact key the backend publishes under before this merges?

The whole feature hinges on it: if the real key differs, the subscribe callback never fires, `updateSelectorConfig` is never called, and the engine stays dormant forever with no error surfaced - a silent no-op in prod. The PR body already flags this as open against the `pageActions` precedent, so confirming against the backend contract (or landing it with a tracking ticket) would close the one gap I'd worry about here.

packages/plugin-autocapture-browser/src/data-extractor.ts
```diff
@@ -24,14 +24,32 @@ import type { BaseTimestampedEvent, ElementBasedTimestampedEvent, TimestampedEve
...
+   * engine. It ships dormant: with the default config (`enabled: false`) the
+   * engine routes through the byte-identical legacy walker, so behavior is
+   * unchanged until remote config flips an org onto the new algorithm via
+   * {@link updateSelectorConfig}.
+   */
+  private readonly selectorEngine: SelectorEngine;
+
   constructor(options: ElementInteractionsOptions, context?: { diagnosticsClient: IDiagnosticsClient }) {
     this.diagnosticsClient = context?.diagnosticsClient;
+    this.selectorEngine = createSelectorEngine(resolveSelectorConfig());
```
jpollock-ampl: Is dropping the engine-level `logger` here intentional?

The engine is created without the optional `logger`, so per `createSelectorEngine`'s contract it stays silent on malformed selectors (debug) and listener exceptions (warn) for the life of the extractor. `updateSelectorConfig` does pass `config.loggerProvider` into `resolveSelectorConfig`, but I believe that only surfaces bad-regex warnings during config resolution, not the engine's own `generate`-time diagnostics. If keeping it fire-and-forget is deliberate, no change needed - otherwise we might want to thread `config.loggerProvider` into the engine too so malformed-selector telemetry shows up once an org enables it.

packages/plugin-autocapture-browser/src/frustration-plugin.ts
```diff
@@ -235,6 +236,14 @@ export const frustrationPlugin = (options: FrustrationInteractionsOptions = {}):
     // Create observables for events on the window
     const allObservables = createObservables();
 
+    // Keep the data extractor's selector engine in sync with element-selector
+    // remote config so frustration events emit the same selectors as the rest
+    // of autocapture when an org enables the engine.
+    const elementSelectorConfigCleanup = subscribeToElementSelectorConfig(config, dataExtractor);
```
jpollock-ampl, line 242: This subscribe-and-push block is now identical in both `frustrationPlugin` and `autocapturePlugin`. Worth pulling the `if (cleanup) subscriptions.push(...)` block into a small helper so the two stay in sync? Not a blocker.

jxiwang:
bugbot run

packages/plugin-autocapture-browser/src/frustration-plugin.ts
jxiwang: I can follow up if necessary later.

jxiwang:
bugbot run

bravecod (approved):

PR #1843, opened 2026-06-17
feat(session-replay-react-native): add Fabric SRMaskView foundation

chungdaniel:
## Summary

- Adds the dual-arch Fabric foundation for layout-transparent masking in `@amplitude/session-replay-react-native`: codegen `SRMaskView`, a C++ `display: contents` ShadowNode, the recorder-agnostic `SRMaskingRegistry` seam, and instrumented canary tests.
- No public API changes — `SRMaskView` is internal plumbing only; the Paper path (`AMPMaskComponentView`) is untouched.
- Part of SDKRN-29 / SDKRN-32. Blocks SDKRN-33 (`<AmpMask>` public API).

## Implementation notes

- **Zero-box layout via `adopt()`**: `display:contents` + `traits_.unset(ForceFlattenView)` is enforced in `SRMaskViewComponentDescriptor::adopt()`, which runs after `updateYogaProps()` on every create/clone, rather than a fragile constructor `initialize()`. This survives prop-driven Yoga style resets.
- **iOS registration**: `+load` registers the view with `RCTComponentViewFactory`, and `+componentDescriptorProvider` binds `SRMaskViewComponentDescriptor`; `codegenConfig.ios.componentProvider` maps `SRMaskView`. (Does not rely on auto-gen for npm-installed packages.)
- **Android codegen**: `patchComponentDescriptorsHeader()` replaces the generated descriptor typedef with our custom descriptor, and `patchCodegenCMakeLists()` compiles `cpp/SRMaskViewShadowNode.cpp` into the codegen target so the descriptor links into `libappmodules.so`.
- **Masking lifecycle**: masking re-applies on prop changes; `clearForView` fires on unmount (`unmountChildComponentView` on iOS, `removeView`/`removeViewAt` on Android).

## Differences from the original plan (POC)

The ticket body documents the POC plan; the shipped code diverges in a few places. All are hardening with the same external behavior:

| Area | Plan (ticket) | Shipped | Why |
|---|---|---|---|
| `display:contents` enforcement | explicit ShadowNode constructors calling `initialize()` (DIV-1) | enforced in `SRMaskViewComponentDescriptor::adopt()` | constructor-only init is clobbered by prop-driven `updateYogaProps()`; `adopt()` runs after every create/clone (react-native-screens pattern) and is authoritative |
| Android codegen patch | `patchShadowNodesHeader()` regex on generated `ShadowNodes.h` (DIV-3) | `patchComponentDescriptorsHeader()` (swap the descriptor typedef) + `patchCodegenCMakeLists()` (compile `cpp/SRMaskViewShadowNode.cpp` into the codegen target) | the **descriptor** typedef is the real binding, not the shadow-node typedef; the cpp must link into `libappmodules.so` |
| iOS registration | `+load` → `RCTThirdPartyComponentsProvider` | `+load` → `RCTComponentViewFactory registerComponentViewClass:` + `+componentDescriptorProvider` + `codegenConfig.ios.componentProvider` | factory registration + componentProvider is the reliable binding for npm packages; the generated `registerComponentDescriptorsFromCodegen` path is effectively dead code |
| Naming | `SRMaskViewCustomComponentDescriptor` | `SRMaskViewComponentDescriptor` (descriptor) / `SRMaskViewContentsShadowNode` (node — named distinctly from the codegen typedef to avoid an ODR clash) | — |

## Test plan

- [x] `pnpm --filter @amplitude/session-replay-react-native typecheck && build && test` (98 tests)
- [x] Android + iOS new-architecture example builds + launch (no duplicate-symbol/link errors)
- [x] Manual layout repro (both platforms): native `SRMaskView` (row 5, **no** JS `display:contents` override) matches the no-wrapper baseline (row 1) across sections A/B/C, while plain `<View>`/`AmpMaskView` collapse `flex:1` and shift absolutely-positioned children
- [x] Android `SRMaskViewTest` + iOS `SRMaskViewTests` assert `mask(child)` fires for every direct child (run **locally**)
- [ ] CI (lint, build, test matrix)

> **CI caveat:** the Android instrumented (`SRMaskViewTest`) and iOS (`SRMaskViewTests`) suites are **not** run by any workflow today — `ci.yml` runs Jest only, and `rn-smoke.yml` covers `analytics-react-native`, not this package. They were verified locally this session. There is an open question on `SRMaskViewTest.kt` about whether to wire a native test job into CI or treat these as local canaries.

## Follow-ups

- SDKRN-33: public `<AmpMask>` / `<AmpUnmask>` built on this foundation. Note: today `unmask` shares `AMPMaskComponentView` with `mask`, so it has the same layout-boundary problem whenever the child is layout-dependent — this foundation is what lets the new components avoid that.
- SDKRN-33 also registers the production `SRMaskingPrimitive` (`SRMaskingRegistry.setPrimitive(...)`) — until then the Fabric masking path is inert (no-op) by design.
- SDKRN-14: plugin package Fabric backport


chungdaniel:
bugbot run

chungdaniel:
bugbot run

packages/session-replay-react-native/android/src/androidTest/java/com/amplitude/sessionreplayreactnative/SRMaskViewTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  @Before
+  fun setUp() {
+    SRMaskingRegistry.setPrimitive(null)
+  }
+
+  @Test
+  fun maskChildFiresForEveryDirectChild() {
```
chungdaniel, line 36: Open question for reviewers: is this test worth keeping as-is?

It verifies the `SRMaskingRegistry.mask(child)` dispatch (and the iOS `SRMaskViewTests` equivalent) via a test-double — the seam SDKRN-33's production adapter will depend on. Two caveats worth weighing:

1. It does **not** run in CI (no native/instrumented test job exists today; `ci.yml` runs Jest only, `rn-smoke.yml` covers `analytics-react-native`). So it can't actually guard against regressions — it's a local-only canary right now.
2. It mocks the riskiest part by calling `addView` / `mountChildComponentView` manually. It does **not** prove that Fabric mounts real `display:contents` children under `SRMaskView` at runtime — which is the thing that would actually break masking end-to-end.

Options: (a) keep and add a native test job to CI, (b) keep as a documented local smoke test, or (c) drop until SDKRN-33 adds a true end-to-end test (real RN subtree → recorder receives mask signal). Preferences?

PR #1571, opened 2026-03-04
feat(session-replay): enhance targeting evaluation with page URL

bravecod:
### URL-based targeting
- Re-evaluate session-replay targeting when the URL changes (SPA navigation).
- Use shared `subscribeToUrlChanges` from the URL tracking plugin; on URL change call `evaluateTargetingAndCapture(..., false, true)` and skip IDB cache in `evaluateTargetingAndStore` when `urlChange` is true.
- Stop recording when targeting no longer matches after re-evaluation.

### session-replay-browser
- `setupUrlChangeListenerForTargeting()` subscribes to URL changes and cleans up on shutdown.
- Removed redundant `globalScope.location?.href` in URL plugin (already guarded by early return).
- Removed debug `console.log` from URL plugin.

### Tests
- URL-change listener: no location / null scope, callback invokes evaluateTargetingAndCapture, cleanup on shutdown, no subscribe when no targetingConfig.
- forceRestart + stop recording when no longer matching.
- `evaluateTargetingAndCapture` with empty/undefined page.
- URL plugin: polling, second subscriber, no history, pushState/replaceState with null url, getHref with empty href.
- Fixed unbound-method lint in url-tracking-plugin tests.

### Coverage
- 100% statements, branches, functions, and lines for session-replay-browser, targeting, and plugin-session-replay-browser.
- Addressed unreachable branches and removed dead code in URL plugin.

---

> [!NOTE]
> **Medium Risk**
> Changes when/why session replay starts recording by adding URL-driven targeting re-evaluations and new shared history/listener patching; regressions could affect capture rates or introduce subtle listener/async ordering issues.
> 
> **Overview**
> Session Replay targeting now receives `page.url` context and can re-evaluate on SPA navigation: `SessionReplay.evaluateTargetingAndCapture` accepts `page`, passes it through to `@amplitude/targeting`, and `SessionReplay` subscribes to URL changes (with cleanup on re-init/shutdown and stale-evaluation protection) to queue forced re-evaluations.
> 
> The URL tracking plugin is refactored around a new shared `subscribeToUrlChanges` helper (history patching + popstate/hashchange or polling) to avoid double-patching and to share subscriptions across consumers, and targeting’s IDB cache is bypassed on URL-change re-evaluations. Tests are expanded substantially across `plugin-session-replay-browser`, `session-replay-browser`, `url-tracking-plugin`, and `targeting` (including `tsconfig` test inclusion) to cover the new behaviors and race conditions.
> 
[...]

PR #1844, opened 2026-06-18
feat(targeting): publish @amplitude/targeting so the page-url targeting fix ships

bravecod:
## Why

`@amplitude/targeting` is marked `private: true`, so the release pipeline (`lerna version --no-private` + `pnpm publish`, which both skip private packages) **never publishes it**. As a result, the page-URL targeting fix (#1571, merged to `main` in March) has never reached npm — **every published `@amplitude/targeting` version through `0.2.3` omits `page` from the evaluation context.**

Consequence: URL-based Session Replay targeting rules (e.g. "URL contains `price-plan`") can never match at runtime, because the SDK passes `page: { url }` but the published targeting drops it before calling the engine. This is the root cause of the "TRC enabled but not recording" reports.

Verified end-to-end against the published packages: the broken `0.2.x`/`0.2.3` returns `off` for a matching URL, while a build with the fix returns `on`.

## What

- **Remove `private: true`** so targeting publishes through the normal release flow.
- **Bump `0.2.0` → `0.2.3`** — the repo version was stale behind npm's latest published (`0.2.3`); aligning it means the next release bumps forward cleanly (the page-URL change is a `feat`, so lerna will land it at `0.3.0`) with no version collision against already-published `0.2.x`.
- **Exclude the auto-generated `version.ts` from coverage** — when lerna versions targeting, its `version` lifecycle writes `src/version.ts`, which targeting doesn't import (so it's uncovered) and would otherwise fail the package's 100% coverage threshold once it's no longer skipped by the pipeline.

The page-fix code itself is already on `main` (from #1571); this PR just makes the package publishable so it actually ships.

## Testing

- `pnpm --filter @amplitude/targeting test` → **19/19 pass at 100%**, including with a generated `version.ts` present (simulating the publish-time lifecycle).
- No dependent version bumps needed — `session-replay-browser` references targeting via `workspace:*`, resolved at publish.

## Release

Merging + the normal v2 release will publish **`@amplitude/targeting@0.3.0`** (first published version containing the page-URL targeting fix).

---

> [!NOTE]
> **Low Risk**
> Release and test-config only; no application logic changes in the diff, though publishing will expose a new npm version of a dependency used by Session Replay targeting.
> 
> **Overview**
> Makes **`@amplitude/targeting`** part of the normal npm release by **removing `private: true`** and aligning the package version to **`0.2.3`** so the next lerna bump can ship without colliding with what’s already on npm.
> 
> Adds **`version.ts`** to Jest **`coveragePathIgnorePatterns`** so the lerna-generated version file (unused by this package) doesn’t break the **100% coverage** gate once publishing runs. **No runtime targeting logic changes in this diff**—it unblocks shipping fixes (e.g. page URL in evaluation context) that are already on `main`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 228b1ea4053a92942d2d3333453a74e7206e4d64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


stevenchien-amplitude (approved):
lgtm

jxiwang (approved):

PR #1854, opened 2026-06-24
feat(session-replay-react-native): add New Architecture support

chungdaniel:
### Summary

Adds React Native **New Architecture (TurboModule)** support to `@amplitude/session-replay-react-native`, while keeping the legacy bridge fully working. The existing Swift/Kotlin Session Replay implementation is unchanged — this PR is a thin, additive compatibility layer around it:

- **JS**: a Codegen TurboModule spec (`src/specs/NativeAmpSessionReplay.ts`) + `codegenConfig`. The rich, hand-written types stay in `native-module.ts`; the spec is intentionally loose (`UnsafeObject`) at the bridge boundary. Resolution via `TurboModuleRegistry.getEnforcing` works on both architectures.
- **iOS**: keep `RCT_EXTERN_MODULE` for the legacy bridge; add the TurboModule conformance only under `#ifdef RCT_NEW_ARCH_ENABLED`.
- **Android**: dual source sets — `oldarch` keeps a plain `ReactPackage` (only APIs present in every supported RN version), `newarch` uses `BaseReactPackage` and is compiled solely when `newArchEnabled=true`.

**Backward compatibility:** old-architecture consumers are unaffected (no peer-dependency floor raised; explicit handling for RN < 0.71 retained on both platforms). The New Architecture path effectively requires **RN ≥ 0.74** (Android `BaseReactPackage`).

**Why it's safe:** new-arch code is gated behind `#ifdef` / source-set switches, so it is never referenced by legacy-arch builds. No Session Replay capture logic changed.

**Out of scope (intentional):** this PR does **not** add native field masking — `secureTextEntry` in the example only hides characters on screen; masking is a follow-up PR.

**Verification:** built and ran the example app across all four combinations — iOS (old + new arch) and Android (old + new arch). In every case the SDK initializes and `getSessionReplayProperties()` returns a live session. On the New Architecture the app reports `TurboModule: true · Fabric: true`. Android logcat confirms the full replay pipeline (full + incremental snapshots with real bitmap content, and `rrweb` event uploads to Amplitude), attributed to the run's derived `deviceId`. The example app was expanded to a multi-screen harness (Home/Form/Gallery/WebView) to generate non-trivial replay frames.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how the native module is registered and resolved on both platforms; mis-gating or codegen wiring could break linking or New Arch builds, though session replay capture logic is untouched.
> 
> **Overview**
> Adds **React Native New Architecture** support for `@amplitude/session-replay-react-native` while keeping the legacy bridge path unchanged. Consumers still use the same public JS API; the native module is resolved through a **Codegen TurboModule spec** instead of `NativeModules` directly.
> 
> On **JavaScript**, a new `src/specs/NativeAmpSessionReplay.ts` spec plus `codegenConfig` in `package.json` define the native contract. `native-module.ts` now loads the module via `TurboModuleRegistry.get` (with a lazy proxy when linking fails), and the hand-written types in `native-module.ts` remain what app code uses.
> 
> On **Android**, Gradle switches between `src/oldarch` (plain `ReactPackage` + hand-written `SessionReplayReactNativeSpec` base) and `src/newarch` (`BaseReactPackage`, TurboModule metadata, codegen `NativeAmpSessionReplaySpec`) when `newArchEnabled` is true. The shared `SessionReplayReactNativeModule` now extends the architecture-specific spec base and `override`s bridge methods; Session Replay capture logic is not modified.
> 
> On **iOS**, legacy `RCT_EXTERN_MODULE` stays as-is; under `RCT_NEW_ARCH_ENABLED`, `AMPNativeSessionReplay.mm` adds TurboModule conformance wired to the codegen JSI class.
> 
> README documents dual-arch behavior and unconstrained `react-native` peers; Jest mocks mock both `NativeModules` and `TurboModuleRegistry`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 513777592141050ccfe684d0c642cbb564b8dd0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/session-replay-react-native/package.json
+++ b/packages/session-replay-react-native/package.json
@@ -78,6 +78,14 @@
     "react": "*",
     "react-native": "*"
   },
+  "codegenConfig": {
+    "name": "AmpSessionReplaySpec",
+    "type": "modules",
+    "jsSrcsDir": "src/specs",
+    "android": {
+      "javaPackageName": "com.amplitude.sessionreplayreactnative"
+    }
+  },
   "react-native-builder-bob": {
     "source": "src",
     "output": "lib",
--- a/packages/session-replay-react-native/src/native-module.ts
+++ b/packages/session-replay-react-native/src/native-module.ts
@@ -1,4 +1,5 @@
-import { NativeModules, Platform } from 'react-native';
+import { Platform } from 'react-native';
+import SessionReplaySpec from './specs/NativeAmpSessionReplay';
 
 const LINKING_ERROR =
   `The package '@amplitude/session-replay-react-native' doesn't seem to be linked. Make sure: \n\n` +
@@ -6,16 +7,21 @@ const LINKING_ERROR =
   '- You rebuilt the app after installing the package\n' +
   '- You are not using Expo Go\n';
 
-export const NativeSessionReplay = NativeModules.AMPNativeSessionReplay
-  ? (NativeModules.AMPNativeSessionReplay as NativeSessionReplaySpec)
-  : (new Proxy(
-      {},
-      {
-        get() {
-          throw new Error(LINKING_ERROR);
-        },
+// JS callers use the rich, hand-written types below for full type safety; the
+// codegen `Spec` is intentionally loose (`UnsafeObject`) at the native boundary.
+// The spec resolves the native module via `TurboModuleRegistry.get` (both
+// architectures). When the module isn't linked it resolves to null, so we fall
+// back to a proxy that throws a descriptive linking error on first use rather
+// than crashing app startup at import time.
+export const NativeSessionReplay = (SessionReplaySpec ??
+  new Proxy(
+    {},
+    {
+      get() {
+        throw new Error(LINKING_ERROR);
       },
-    ) as NativeSessionReplaySpec);
+    },
+  )) as unknown as NativeSessionReplaySpec;
 
 /**
  * Configuration interface for setting up the native iOS and Android session replay modules.
--- a/packages/session-replay-react-native/README.md
+++ b/packages/session-replay-react-native/README.md
@@ -8,6 +8,19 @@ Amplitude Session Replay for React Native
 npm install @amplitude/session-replay-react-native
 ```
 
+## React Native New Architecture
+
+This SDK supports both the New Architecture (Bridgeless / TurboModules) and the
+legacy architecture. On the New Architecture the native module is exposed as a
+TurboModule; on the legacy architecture it continues to work as a standard bridge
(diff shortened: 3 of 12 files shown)


packages/session-replay-react-native/package.json
```diff
@@ -69,15 +69,23 @@
     "@amplitude/analytics-types": "^1.3.4"
   },
   "devDependencies": {
-    "@types/react": "^18.0.26",
-    "react": "18.2.0",
-    "react-native": "0.73.0",
+    "@types/react": "^18.3.1",
+    "react": "18.3.1",
+    "react-native": "0.77.2",
```
aliaksandr-kazarez: do we need those for the new Arch, or for the example app?

packages/session-replay-react-native/src/native-module.ts
```diff
@@ -1,21 +1,11 @@
-import { NativeModules, Platform } from 'react-native';
+import SessionReplaySpec from './specs/NativeAmpSessionReplay';
 
-const LINKING_ERROR =
-  `The package '@amplitude/session-replay-react-native' doesn't seem to be linked. Make sure: \n\n` +
-  Platform.select({ ios: "- You have run 'pod install'\n", default: '' }) +
-  '- You rebuilt the app after installing the package\n' +
-  '- You are not using Expo Go\n';
```
aliaksandr-kazarez, line 7: is codegen eliminates this linking problem for old arch?

chungdaniel: Not codegen — it's TurboModuleRegistry.getEnforcing. On old arch it falls back to the legacy NativeModules lookup, and getEnforcing throws RN's built-in "could not be found / not registered in the native binary" invariant if it's missing. So the same not-linked cases are still caught, just by RN instead of our proxy. Only downside: the built-in message is terser than our old one — can wrap it in a try/catch to restore the pod-install/rebuild hint if you want.

aliaksandr-kazarez: if `getEnforcing` throws, it would crash the app on import if native module is not linked properly. I think we should handle that in a backwards compatible way.

aliaksandr-kazarez (approved):

packages/session-replay-react-native/src/native-module.ts
aliaksandr-kazarez: but up to you

PR #1873, opened 2026-07-09
fix(analytics-core): queued functions not being flushed when init not awaited

daniel-graham-amplitude:
### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core client initialization ordering used by all SDKs; behavior change is intentional but could affect subtle timing during startup.
> 
> **Overview**
> Fixes a race where **`init()`** is not awaited and callers **`add()`** plugins (or other pre-ready APIs) in the same turn or via microtasks: work could land on **`q`** after **`runQueuedFunctions`** finished but before **`isReady`**, so it was never executed.
> 
> **`_init`** now loops with **`do { await runQueuedFunctions('q') } while (this.q.length > 0)`** so the pre-ready queue is drained before **`isReady = true`**. **`runQueuedFunctions`** also re-invokes itself when new entries appear on the queue while earlier items’ promises are still settling.
> 
> Adds regression tests in **analytics-core** and **analytics-node** for late-queued plugins during init.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42ce88f23b539c26f5c47f37b2e3f4515112f8e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/analytics-core/src/core-client.ts
+++ b/packages/analytics-core/src/core-client.ts
@@ -51,7 +51,9 @@ export class AmplitudeCore implements CoreClient, PluginHost {
     this.config = config;
     this.timeline.reset(this);
     this.timeline.loggerProvider = this.config.loggerProvider;
-    await this.runQueuedFunctions('q');
+    do {
+      await this.runQueuedFunctions('q');
+    } while (this.q.length > 0);
     this.isReady = true;
   }
 
@@ -69,6 +71,7 @@ export class AmplitudeCore implements CoreClient, PluginHost {
         await val;
       }
     }
+
     // Rerun queued functions if the queue has accrued more while awaiting promises
     if (this[queueName].length) {
       await this.runQueuedFunctions(queueName);
--- a/packages/analytics-core/test/core-client.test.ts
+++ b/packages/analytics-core/test/core-client.test.ts
@@ -253,6 +253,36 @@ describe('core-client', () => {
       expect(_setOptOut).toHaveBeenCalledWith(true);
       expect(client.config.optOut).toBe(true);
     });
+
+    test('should flush queue items added after runQueuedFunctions returns before isReady', async () => {
+      // Covers the do-while in _init: work can land on q after runQueuedFunctions
+      // returns (e.g. a microtask) but before isReady is set.
+      const client = new AmplitudeCore();
+      const setup = jest.fn().mockResolvedValue(undefined);
+      const plugin: EnrichmentPlugin = {
+        name: 'late-plugin',
+        type: 'enrichment',
+        setup,
+        execute: jest.fn(),
+      };
+
+      let flushCount = 0;
+      const originalRunQueuedFunctions = client.runQueuedFunctions.bind(client);
+      jest.spyOn(client, 'runQueuedFunctions').mockImplementation(async (queueName) => {
+        await originalRunQueuedFunctions(queueName);
+        flushCount += 1;
+        if (flushCount === 1 && queueName === 'q') {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+          (client as any).q.push(client._addPlugin.bind(client, plugin));
+        }
+      });
+
+      await (client as any)._init(useDefaultConfig());
+
+      expect(flushCount).toBeGreaterThanOrEqual(2);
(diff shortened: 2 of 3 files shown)


Mercy811 (commented):
Thanks Dan, LGTM!

```
  protected async _init(config: IConfig) {
    this.config = config;
    this.timeline.reset(this);
    this.timeline.loggerProvider = this.config.loggerProvider;
    await this.runQueuedFunctions('q');
    this.isReady = true;
  }
```
The issue is in CoreClient, `this.isReady = true` is put into a microtask. When calling init without await, it yields while isReady is still false and the plugin will be put into the queue which will never be flushed. 
```
client.init()
client.add(plugin)
```

```
  protected async _init(config: IConfig) {
    this.config = config;
    this.timeline.reset(this);
    this.timeline.loggerProvider = this.config.loggerProvider;
    this.isReady = true; // swap order to set isReady first
    await this.runQueuedFunctions('q');
  }
```
Swapping won't work and will fail track() because by the time track() runs, destination plugin is not yet registered. 
```
client.init(...);                     
client.add(plugin);
await client.track('test event').promise;
```

I will ask agent reviews before approving it

Mercy811:
bugbot run

Mercy811:
@claude review

Mercy811:
Other solutions i'm thinking of are 
1.  a do-while loop semantically looks nicer. I'm reading this as when setting isReady to true, this.q is guaranteed to be empty. It won't yield because of microtask 
```
do {
  await this.runQueuedFunctions('q');
} while (this.q.length);
this.isReady = true;
```
2. Another one is like you mentioned before to redesign to promise based like `readyPromise`. But this requires more work.

Let me think about which one fits us best

daniel-graham-amplitude:
@Mercy811 I think I found a better way. I just added `await Promise.resolve()`

```
await Promise.resolve();
if (this[queueName].length) {
  await this.runQueuedFunctions(queueName);
}
```

What the new await does is yields so that the incoming microtasks can be run first (like the `add(plugin)`)  and then the queue will be drained (`await this.runQueuedFunctions(queueName);`) afterwards instead of before.

Mercy811:
> @Mercy811 I think I found a better way. I just added `await Promise.resolve()`
> 
> ```
> await Promise.resolve();
> if (this[queueName].length) {
>   await this.runQueuedFunctions(queueName);
> }
> ```
> 
> What the new await does is yields so that the incoming microtasks can be run first (like the `add(plugin)`) and then the queue will be drained (`await this.runQueuedFunctions(queueName);`) afterwards instead of before.

Hi @daniel-graham-amplitude, IMHO the race is because in CoreClient `_init()`, setting `isReady` and calling `runQueuedFunctions('q');` are not in the same microtask. The above fix in `runQueuedFunctions()` won't fix the root cause. A super edge case would be 

```
init()
Promise.resolve().then(() => add(plugin))
```

I'm still leaning towards the do-while fix.

daniel-graham-amplitude:
@Mercy811 you're correct yes, good catch. I tried to reproduce with `Promise.resolve().then(() => client.add(plugin));` and it failed the test.

What do you think about this:
```
await this.runQueuedFunctions('q');
this.isReady = true;
if (this.q.length) {
  await this.runQueuedFunctions('q');
}
```

If we only drain the queue before setting `isReady=true`, isn't there always a risk that another microtask can be scheduled in between when `runQueuedFunctions` finishes and when `isReady` is set to `true`?

Mercy811:
> If we only drain the queue before setting isReady=true, isn't there always a risk that another microtask can be scheduled in between when runQueuedFunctions finishes and when isReady is set to true?

@daniel-graham-amplitude yes if we only attempt to drain it once.  And that's why I think do-while can prevent this 

```
do {
  await this.runQueuedFunctions('q');
} while (this.q.length);
this.isReady = true;
```

If we consider do-while block as a whole, it runs uninterrupted with setting isReady. Even if there is another microtask in between runQueuedFunctions, as long as I find the queue is not empty, I will drain it again until it's empty. And these two (drain queue & setting isReady) is guaranteed to be on same task.

daniel-graham-amplitude:
Okay I understand now thanks.

Mercy811 (approved):
lgtm!

PR #1878, opened 2026-07-12
fix(analytics-client-common): guard document before writing cookie

### LINKED ISSUE #925: In service worker environment, globalScope.document.cookie setting will throw
## Expected Behavior

In service worker, `globalScope.document.cookie` should be checked before accessing it's value

## Current Behavior

It will throw and a console error will be logged

## Possible Solution

- guard against the possibility of undefined in service worker environment

## Steps to Reproduce

1.
2.
3.
4.

## Environment
- JS SDK Version: 
- Installation Method: 
- Browser and Version:

sarmah-rup:
## Bug

In a service worker, `CookieStorage.set()` throws on every cookie write:

```
Amplitude Logger [Error]: Failed to set cookie for key: <k>. Error: Cannot set properties of undefined (setting 'cookie')
```

so cookies never persist and the console fills with errors in service-worker / non-DOM environments.

## Root cause

`set()` guards only the global scope before writing:

```ts
if (globalScope) {
  globalScope.document.cookie = str;   // globalScope exists in a SW, but document does not
}
```

In a `ServiceWorkerGlobalScope`, `self` exists while `document` is `undefined`, so `globalScope.document.cookie = ...` throws (caught by the surrounding try/catch and logged). The sibling `getRaw()` already reads correctly via `globalScope?.document?.cookie`.

## Solution

Mirror the read-path guard in `set()`:

```ts
if (globalScope?.document) {
  globalScope.document.cookie = str;
}
```

The equivalent `CookieStorage` in `@amplitude/analytics-core` already guards this, so only `analytics-client-common` needed the change.

Closes #925

## Testing

Added two cases to the existing `describe('set')` suite (reusing its `getGlobalScope` mock): global scope present but no `document`, and global scope absent. Both assert no throw and no error logged. Before the fix the first fails with the exact runtime error above. Full package suite passes (137/137) with 100% coverage maintained; lint clean.

---

> [!NOTE]
> **Low Risk**
> Single conditional change in cookie persistence with tests; no auth or data-path impact beyond skipping invalid writes in non-DOM runtimes.
> 
> **Overview**
> Fixes **service worker / non-DOM** environments where `CookieStorage.set()` logged errors on every write because `getGlobalScope()` is truthy while `document` is missing.
> 
> **`CookieStorage.set()`** now only assigns `document.cookie` when `globalScope?.document` exists, matching the read path in `getRaw()` and the implementation in `@amplitude/analytics-core`.
> 
> Adds unit tests that mock missing `document` or missing global scope and assert `set()` resolves without throwing or logging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 904347ab5096ad6d0aa5213cc6bbb2bd7884b566. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


daniel-graham-amplitude:
@sarmah-rup this looks good, but question, which package is it that is importing from `analytics-client-common`? `analytics-core` is meant to replace that package.

sarmah-rup:
Appreciate you taking a look, and the steer toward `analytics-core`.

Four packages still depend on `analytics-client-common` (the web-attribution and session-replay plugins, `session-replay-browser` and `targeting`), but none of them instantiate this `CookieStorage`. The only consumer, `WebAttribution`, reuses `config.cookieStorage`, which in `analytics-browser` and `react-native` is already `analytics-core`'s version, and `analytics-core` guards for a missing `document`. So the main SDK path is safe like you said.

I hit it as a direct consumer, importing `CookieStorage` from `analytics-client-common` in a service worker where `self` exists but `document` does not. It is a one line parity fix so I am glad to leave it in, but just as happy to close if you would rather deprecate the package. Either way, thanks for the quick look.

daniel-graham-amplitude (approved):

daniel-graham-amplitude:
Thanks for this @sarmah-rup . I'm just going to merge and publish this to get you unblocked. We just made a ticket to move everything to analytics-core for the future.

PR #1881, opened 2026-07-13
refactor(analytics-core): move translateRemoteConfigToLocal from analytics-browser to analytics-core

daniel-graham-amplitude:
### Checklist

---

> [!NOTE]
> **Low Risk**
> Logic-preserving module move with tests relocated; public API adds an export on analytics-core rather than changing runtime browser merge behavior.
> 
> **Overview**
> **`translateRemoteConfigToLocal`** is relocated from `analytics-browser` into **`analytics-core`** (`config/joined-config.ts`) and re-exported from the core package index. Browser remote-config merging still calls the same helper, but now imports it from `@amplitude/analytics-core` instead of defining it locally.
> 
> Unit coverage for the translator moves with it: the large `translateRemoteConfigToLocal` suite is removed from browser tests and added under **`analytics-core/test/config/joined-config.test.ts`**, including dedicated **network tracking header** cases. The core logger test mock gains a **`debug`** stub. The element-interactions test server page flips **`fetchRemoteConfig`** from `false` to `true`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18c2392c7c3d4b3ca1e547bf8564dccd6be20039. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analytics-browser/src/config/joined-config.ts
```diff
@@ -196,13 +192,13 @@ function mergeUrls(urlsExact: (string | RegExp)[], urlsRegex: string[] | undefin
...
+ * Primarily merges autocapture settings from the remote config into the browser (or client) config.
  *
  * @param remoteConfig - The remote configuration to apply, or null if none available
- * @param browserConfig - The browser config object to update (modified in place)
+ * @param browserConfig - The browser (or client) config object to update (modified in place)
  */
-export function updateBrowserConfigWithRemoteConfig(
+export function updateClientConfigWithRemoteConfig(
   remoteConfig: RemoteConfig | null,
   browserConfig: BrowserConfig,
```
Mercy811, line 207: Why do we wanna move this to core package? It takes BrowserConfig and should be in browser package only

daniel-graham-amplitude: A future iteration will accept `ReactNativeConfig`

`browserConfig: BrowserConfig` --> `clientConfig: BrowserConfig | ReactNativeConfig`

Mercy811: how many autocapture fields are overlapped? Looking at the "Sample config" section of https://amplitude.atlassian.net/wiki/spaces/GOV/pages/3832119297/React+Native+Autocapture, it looks like that none of them overlaps. I'm a little bit hesitate to move everything to the core package unless it's necessary to prevent core package size exposure. How about starting RN RC  without reusing it and then we could refactor to merge them together

daniel-graham-amplitude: > how many autocapture fields are overlapped

`sessions` and `networkTracking`.

Here's 2 places where code would be shared between Browser and RN

1. networkTracking parsing

```javascript
if (
    typeof typedRemoteConfig.autocapture.networkTracking === 'object' &&
    typedRemoteConfig.autocapture.networkTracking !== null &&
    typedRemoteConfig.autocapture.networkTracking.captureRules?.length
  ) {
    transformedAutocaptureRemoteConfig.networkTracking = {
      ...typedRemoteConfig.autocapture.networkTracking,
    };
    const transformedRcNetworkTracking = transformedAutocaptureRemoteConfig.networkTracking;
    /* istanbul ignore next */
    const captureRules = transformedRcNetworkTracking.captureRules ?? [];
    for (const rule of captureRules) {
      rule.urls = mergeUrls(rule.urls ?? [], rule.urlsRegex, browserConfig);
      delete rule.urlsRegex;
    }
  }
```

2. The `translateRemoteConfigToLocal` function translates remote objects with an "enabled" property into the `boolean|Object` type

```javascript
// remote config
autocapture = { sessions: { enabled: true } };

// local
autocapture = { sessions: true };
```


My feeling is that the complexity of these 2 functionalities is high enough that we're better off making it re-usable between the 2 clients (browser, RN) rather than copying and pasting and then maintaining both separately.

Is there a specific case or concern where having 'joined-config' be in `analytics-core` could cause problems that I've perhaps overlooked?

Mercy811: > Is there a specific case or concern where having 'joined-config' be in analytics-core could cause problems that I've perhaps overlooked?

No, I'm concerned about supporting `ReactNativeConfig` in `updateClientConfigWithRemoteConfig` adds complexity to this already long and complex function and it will become even harder to maintain. Do you have any idea how the function will be after? I'd assume it will be even longer. For example https://github.com/amplitude/Amplitude-TypeScript/blob/b9f33de78623a4c8eb39914e92ffac247580159e/packages/analytics-browser/src/config/joined-config.ts#L279-L299, we have to distinguish browser config and RN config and generate different autocapture object.

`translateRemoteConfigToLocal` is very generic. All good to move to analytics core.

Mercy811 (approved):
lgtm! Could you update the PR title to something like "move translateRemoteConfigToLocal from analytics-browser to analytics-core"

packages/analytics-browser/src/config/joined-config.ts
daniel-graham-amplitude: Updated

PR #1892, opened 2026-07-16
fix(element-selector): resolve config race condition

jxiwang:
# fix(element-selector): resolve config race condition

Fixes a race condition in the element-selector remote-config delivery path that could silently disable a live selector engine and revert autocapture element paths back to the legacy `cssPath` format.

## Problem

The element-selector engine is controlled by remote config. In `'all'` delivery mode, the client delivers cache-then-remote payloads. Incomplete deliveries — `null`/absent (failed fetch), `{}`, or partial payloads without a boolean `enabled` flag — were being passed through `resolveSelectorConfig`, which defaults `enabled` to `false`.

This caused a flip-flop: an org could be enabled via an explicit `{ enabled: true }` delivery, then immediately disabled again by a subsequent incomplete delivery, with no observable signal. Autocapture would silently revert from the new strategy-based selectors back to legacy `cssPath`.

### `data-extractor.ts`
- `updateSelectorConfig` now no-ops unless the delivery explicitly carries a boolean `enabled` flag, preserving the current engine state for incomplete payloads
- Added debug logging when:
  - An incomplete delivery is ignored
  - The engine transitions from disabled → enabled
  - The engine transitions from enabled → disabled

### `element-selector-config.ts`
- The remote-config subscription callback now logs each delivery with its **source** (`cache` | `remote`) and the payload for diagnosis

### `data-extractor.test.ts`
- Added tests covering:
  - `null` delivery does not disable a previously-enabled engine
  - Incomplete payloads (no boolean `enabled`) are ignored
  - Explicit `{ enabled: false }` correctly reverts to legacy `cssPath`
  - Enable/disable transitions emit expected debug logs

## Trade-off

Disabling the engine now requires an explicit `{ enabled: false }` delivery — removing the config or sending a partial payload will no longer turn it off.

## Files changed

| File | Changes |
|------|---------|
| `packages/plugin-autocapture-browser/src/data-extractor.ts` | Guard logic + transition logging |
| `packages/plugin-autocapture-browser/src/element-selector-config.ts` | Diagnostic logging on each delivery |
| `packages/plugin-autocapture-browser/test/data-extractor.test.ts` | 5 new test cases |

**+131 / −11 lines** across 3 files. No breaking API changes.

## Checklist

- [x] PR title follows conventional commit format
- [x] No breaking changes
- [ ]


daniel-graham-amplitude (approved):
Lgtm!

PR #1896, opened 2026-07-20
feat(session-replay-browser): apply body masking rules in SR network logging

lewgordon-amplitude:
## Summary
- Apply allowlist/excludelist JSON pointer masking to `fetch-request` bodies captured via `sr_logging_config`, matching autocapture network tracking semantics.
- Export `pruneJson` from `@amplitude/analytics-core` for shared masking logic.
- Pairs with https://github.com/amplitude/javascript/pull/136906 (SR logging settings UI).

## Test plan
- [x] `pnpm test` in `packages/session-replay-browser` (100% coverage)
- [ ] Verify masked request/response bodies in SR console after UI + SDK ship together

### DIFF
--- a/packages/session-replay-browser/test/observers.test.ts
+++ b/packages/session-replay-browser/test/observers.test.ts
@@ -158,6 +158,29 @@ describe('NetworkObservers', () => {
       expect(events[0].requestBody).toBe('{"key":"value"}');
     });
 
+    it('should apply excludelist masking to JSON request bodies', async () => {
+      const mockResponse = {
+        status: 200,
+        headers: { forEach: jest.fn() },
+      };
+      mockFetch.mockResolvedValue(mockResponse);
+      networkObservers.start(callback, {
+        enabled: true,
+        body: {
+          request: {
+            excludelist: ['/password'],
+          },
+        },
+      });
+
+      await globalScope.fetch('https://api.example.com/data', {
+        method: 'POST',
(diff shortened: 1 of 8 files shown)


stevenchien-amplitude (approved):
LGTM. thanks lew. i'm assuming that bugbot comment can be ignored and is mainly for autocapture parity vs actually required as a fallback in your implementation right?

packages/session-replay-browser/test/observers.test.ts
```diff
@@ -158,6 +158,29 @@ describe('NetworkObservers', () => {
       expect(events[0].requestBody).toBe('{"key":"value"}');
     });
 
+    it('should apply excludelist masking to JSON request bodies', async () => {
```
jpollock-ampl: Would you mind adding a couple more observer-level cases here? This covers the request-body excludelist path nicely, but the response path and allowlist behavior are the parts most likely to regress and aren't exercised end-to-end yet.

Rough scenarios worth covering:
- excludelist masking on a JSON response body (via `clone().text()`)
- allowlist keeps only the listed fields and drops the rest
- a top-level array body with `/*/password` (ties into the array question above)
- `request: true` still captures the full body (boolean back-compat)

The unit tests in `network-body-capture.test.ts` cover the masking logic, so these can stay light - mostly confirming the observer wiring passes the config through.

Mercy811 (approved):
Changes to analytics-core lgtm!

Issue #1630, opened 2026-03-26
Browser SDK retries all HTTP errors including not retryable ones


BelkinMax:
## Expected Behavior
According to the documentation, `flushMaxRetries` is described as: "Sets the maximum number of retries for failed upload attempts. **This is only applicable to errors that the SDK can retry.**"

Based on this documentation and standard HTTP retry best practices, the SDK should:
- **Retry** on transient/retryable errors
- **NOT retry** on permanent/client errors

When a non-retryable error (like 501) is returned, the SDK should fail immediately without exhausting retry attempts.

## Current Behavior
The SDK makes up to `flushMaxRetries` attempts even for errors that indicate permanent failures (such as 400 or 501) and will never succeed on retry.

<img width="583" height="115" alt="Image" src="https://github.com/user-attachments/assets/e3311ee8-b7dd-4e3e-a9c3-26a3187e7804" />

## Possible Solution
Update the retry logic to only retry truly transient errors.

## Steps to Reproduce
1. Configure Amplitude Browser SDK 2 with a custom serverUrl.
2. Set up the endpoint to return a non-retryable HTTP status code (e.g., 501).
3. Track an event.
4. Observe network requests in browser - you will see 5 retry attempts to the endpoint, all returning 501, even though this is a permanent error that will never succeed.

## Environment
- JS SDK Version: `analytics-browser@2.36.1`
- Installation Method: NPM
- Browser and Version: Chrome 145.0.7632.159

PR #1908, opened 2026-07-24
feat(analytics-core): retain events and go offline after max retries on client SDKs

Mercy811:
## Summary

Follow-up to the **2026-07-24 AWS us-west-2 event-server outage** ([SDKW-42](https://linear.app/amplitude/issue/SDKW-42), [SDKRN-55](https://linear.app/amplitude/issue/SDKRN-55)).

When the event server is unreachable (5xx / network errors / timeouts), the Browser and React Native SDKs dropped events once the retry budget was exhausted. Offline mode is only entered on a *device-network* signal (`navigator.onLine` / NetInfo), so a **server-only** outage — where the device network stays up — never trips it and events are dropped after retries.

This change makes `analytics-core`'s `Destination` **mirror the mobile SDKs** (Amplitude-Swift `EventPipeline`) — **for client SDKs only**.

## What changes (`packages/analytics-core/src/plugins/destination.ts`)

1. **Exponential backoff** for transient-failure retries — `retryTimeout * 2^(attempts-1)` (was linear `attempts * retryTimeout`). No interval cap: the retry count (`flushMaxRetries`, default **5** on browser/RN) already bounds it (largest wait ~8s before going offline at defaults).
2. **Go offline + retain instead of drop.** Once an event exhausts `flushMaxRetries` on a transient failure, set `config.offline = true` and **keep the event** (it stays in the queue + storage) rather than dropping it via `removeEventsExceedFlushMaxRetries`.

## Client vs. server gating (important)

The `Destination` plugin lives in `analytics-core` and is shared by the **Node (server) SDK** too. Node is a long-lived process with **no reload/relaunch and no connectivity checker**, so flipping it offline would strand it forever. The go-offline behavior is therefore gated to **client SDKs only**, via a new `isClientSide()` helper (`isReactNative() || isBrowser()`, alongside the existing `isReactNative()` in `utils/environment.ts`), resolved once as a field initializer (the runtime environment is fixed for the process lifetime):

- **Browser / React Native** → go offline + retain (new behavior).
- **Node** → unchanged: drop events past the retry budget.

## Recovery (client SDKs — mobile parity)

Deliberately **no server polling**. Client SDKs come back online exactly like mobile:
- the connectivity checker flips `config.offline = false` on a `navigator` / NetInfo **online** event, **or**
- the **page reloads / app relaunches** — `config.offline` is in-memory so it resets, and because `context.attempts` is **not persisted**, every stored event is retried from a fresh budget.

Events are retained across the outage and delivered on recovery — only storage exhaustion would lose them.

## Scope / behavior notes

- Only the **transient-failure path** (`handleOtherResponse` — 5xx / network / timeout) changes. Terminal per-event drops are untouched: 400 invalid API key, server-flagged invalid events, and 413 single-event still drop via `removeEventsExceedFlushMaxRetries`.
- `OfflineDisabled` is not special-cased in the destination (it only gates installing the connectivity plugin, `browser-client.ts:312`); the destination just reads/writes `config.offline`. On a client SDK that sets `offline: OfflineDisabled`, a max-retry still flips it to `true` and retains — same as mobile.

## Open question

- Storage-exhaustion eviction ("retry forever unless out of space") is **not** implemented here — events are retained in the in-memory queue + best-effort storage. Follow-up or in-scope?

Refs SDKW-42, SDKRN-55.

---

> [!NOTE]
> **Medium Risk**
> Changes core event upload retry/drop behavior for all browser and React Native customers during outages; Node path is explicitly preserved but shared Destination code is security-sensitive for data loss.
> 
> **Overview**
> **Client SDKs (browser / React Native)** no longer drop events when transient upload failures exhaust `flushMaxRetries`. `Destination.handleOtherResponse` now uses **exponential backoff** (`getRetryBackoff`), and on the last failed attempt sets **`config.offline = true`**, keeps events in the queue/storage, and records `offline.by.max.retries`—pausing flush until connectivity recovery or reload (no server polling).
> 
> **Node** keeps prior behavior: linear backoff and drop via `removeEventsExceedFlushMaxRetries`. Branching uses new **`isBrowser` / `isClientSide`** helpers (exported from `analytics-core`).
> 
> Tests cover the new retry/offline paths; the browser integration test that expected **max-retry rejection** was removed. RN tests disable attribution where needed for stability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5597e116b7151d38b1c35dcebfd923706a902eb0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/analytics-browser-test/test/index.test.ts
+++ b/packages/analytics-browser-test/test/index.test.ts
@@ -800,47 +800,6 @@ describe('integration', () => {
       second.done();
     });
 
-    test('should exhaust max retries', async () => {
-      const scope = nock(url).post(path).times(3).reply(500, {
-        code: 500,
-      });
-
-      await client.init(apiKey, {
-        logLevel: 0,
-        flushMaxRetries: 3,
-        defaultTracking,
-        fetchRemoteConfig: false,
-      }).promise;
-      const response = await client.track('test event').promise;
-      expect(response.event).toEqual({
-        user_id: undefined,
-        device_id: uuid,
-        session_id: number,
-        time: number,
-        platform: 'Web',
-        language: 'en-US',
-        ip: '$remote',
-        insert_id: uuid,
-        partner_id: undefined,
-        event_type: 'test event',
-        event_properties: {
-          '[Amplitude] Page Domain': '',
-          '[Amplitude] Page Location': '',
-          '[Amplitude] Page Path': '',
-          '[Amplitude] Page Title': '',
-          '[Amplitude] Page URL': '',
-          '[Amplitude] Previous Page Location': '',
-          '[Amplitude] Previous Page Type': 'direct',
-        },
-        event_id: 0,
-        library: library,
-        user_agent: userAgent,
-      });
-      expect(response.code).toBe(500);
-      expect(response.message).toBe('Event rejected due to exceeded retry count');
-      scope.done();
-    }, 10000);
-
     test('should handle missing api key', async () => {
       await client.init('', undefined, {
         logLevel: 0,
--- a/packages/analytics-core/src/plugins/destination.ts
+++ b/packages/analytics-core/src/plugins/destination.ts
@@ -32,6 +32,7 @@ import { EventCallback } from '../types/event-callback';
 import { IDiagnosticsClient } from '../diagnostics/diagnostics-client';
 import { isSuccessStatusCode } from '../utils/status-code';
 import { getStacktrace } from '../utils/debug';
+import { isClientSide } from '../utils/environment';
 
 export interface Context {
   event: Event;
@@ -95,6 +96,11 @@ export class Destination implements DestinationPlugin {
   flushId: ReturnType<typeof setTimeout> | null = null;
   queue: Context[] = [];
   diagnosticsClient: IDiagnosticsClient | undefined;
+  // True for client SDKs (browser / React Native), which can recover from an offline
+  // state via a network reconnect or a page reload / app relaunch. False for the Node
+  // (server) SDK — a long-lived process with no such recovery — where events past the
+  // retry budget are dropped as before.
+  isClientSide = isClientSide();
 
   constructor(context?: { diagnosticsClient: IDiagnosticsClient }) {
     this.diagnosticsClient = context?.diagnosticsClient;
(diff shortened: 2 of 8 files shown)


Mercy811:
bugbot run

packages/analytics-core/src/utils/environment.ts
```diff
@@ -9,3 +9,12 @@ export function isReactNative(): boolean {
   const globalScope = getGlobalScope() as { navigator?: { product?: string } };
   return globalScope?.navigator?.product === 'ReactNative';
 }
+
+export function isBrowser(): boolean {
+  const globalScope = getGlobalScope() as { document?: unknown } | undefined;
+  return typeof globalScope?.document !== 'undefined';
```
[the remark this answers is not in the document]
Mercy811: Good catch — fixed. `isClientSide()` now covers document-less browser contexts:
`isReactNative() || isBrowser() || isWebWorker() || isChromeExtension()`, where
`isWebWorker()` checks that the global scope is an instance of its own
`WorkerGlobalScope`. MV3 background workers now take the retain-and-go-offline
path instead of the Node drop path.

Known limitation: `localStorage` is `[Exposed=Window]`, so no worker — dedicated,
shared, or service — can reach it. The SDK falls back to `MemoryStorage` there,
so retained events aren't persisted across a worker restart.

Retain + go offline is still the right behavior for them. Node is excluded only
because it's a long-lived process that would stay offline forever; workers are
short-lived either way — a dedicated worker is bounded by its host page, a
service worker by idle termination — so in-memory `config.offline` resets on the
next start, which is the same recovery the page-reload path already relies on.
Retaining in memory strictly beats dropping.

packages/analytics-core/src/plugins/destination.ts
```diff
@@ -390,13 +396,49 @@ export class Destination implements DestinationPlugin {
...
+      list.forEach((context) => {
+        context.attempts += 1;
+        if (context.attempts < this.config.flushMaxRetries) {
+          context.timeout = this.getRetryBackoff(context.attempts);
+          tryable.push(context);
+        } else {
+          isExceedingMaxRetries = true;
+        }
+      });
+      if (isExceedingMaxRetries) {
```
daniel-graham-amplitude, line 424: By my spreadsheet calculations, it would (by default) take ~34 minutes before it stops trying and then goes "offline".  

To confirm, we [still save events](https://github.com/amplitude/Amplitude-TypeScript/blob/xinyiye/sdkw-42-browser-sdk-go-offline-persist-after-max-retries/packages/analytics-core/src/plugins/destination.ts#L136) to offline storage (`storageProvider`) before they go `offline`? It's just that marking the client as "offline" prevents the unsent events from being flushed whenever there's a failure?

Mercy811: could you share your calculation, I get 16s from event tracked to offline with exponential backoff:
1. track 1s flush, 1s
2. failure 1, 2^0 = 1s
3. failure 2, 2s
4. failure 3, 4s
5. failure 4, 8s
6. failure 5, offline
total = 1 + 1 + 2 + 4 + 8 = 16s

Mercy811: > To confirm, we [still save events](https://github.com/amplitude/Amplitude-TypeScript/blob/xinyiye/sdkw-42-browser-sdk-go-offline-persist-after-max-retries/packages/analytics-core/src/plugins/destination.ts#L136) to offline storage (storageProvider) before they go offline? It's just that marking the client as "offline" prevents the unsent events from being flushed whenever there's a failure?

Yes, `saveEvents()` is called every time on `destination.execute()`

daniel-graham-amplitude: Isn't it [12 retries](https://github.com/amplitude/Amplitude-TypeScript/blob/xinyiye/sdkw-42-browser-sdk-go-offline-persist-after-max-retries/packages/analytics-core/src/config.ts#L21)? When you count up to 12 that's how I calculate that.

Mercy811: that's probably for node, browser default is at https://github.com/amplitude/Amplitude-TypeScript/blob/ba2a05505ba88a79d31046ba886fe10f8b6a36e0/packages/analytics-browser/src/config.ts#L75

packages/analytics-browser-test/test/index.test.ts
```diff
@@ -800,47 +800,6 @@ describe('integration', () => {
       second.done();
     });
 
-    test('should exhaust max retries', async () => {
-      const scope = nock(url).post(path).times(3).reply(500, {
-        code: 500,
```
crleona, line 805: it looks like every response that goes through `handleOtherResponse` gets the new offline logic - are there any responses that we should not retry?

Mercy811: good catch!  

- before:  the logic is always retry except knowing to drop (invalid, missing fields etc. identified by server) and then drop. 
- after: the logic is retry and then offline so a bad event not identified by server can cause the SDK offline forever. But it seems that this logic works well for mobile SDKs. So we could safely take the assumption that server will tell SDKs which events to drop and the should retry rest events

We shouldn't retry non-tryable 5xx for example https://github.com/amplitude/Amplitude-TypeScript/issues/1630. Thanks @daniel-graham-amplitude for bring this up.

PR #1909, opened 2026-07-24
build(deps): add knip to detect unused dependencies

Mercy811:
Closes [SDKW-43](https://linear.app/amplitude/issue/SDKW-43/add-lint-rule-to-detect-unused-dependencies).

## What

Adds [knip](https://knip.dev) to flag dependencies declared in a `package.json` but never used by that package. Run with `pnpm lint:deps`; gated in CI as **Check Unused Dependencies**.

ESLint can't do this — a rule sees one file at a time and can never conclude that *no* file uses a dependency. (`import/no-extraneous-dependencies` catches only the inverse.)

Also removes the 30 dead dependencies it found.

## Note on size

The size-limit report shows **0% change, and that's the expected result** — an unused dependency was never imported, so it was never in the bundle. If the bundle had shrunk, it would mean I deleted something real.

The win is **install size**, not bundle size. Runtime dependency closure — what a consumer actually installs:

| Package | Before | After |
|---|---|---|
| `session-replay-browser` | 17 | **13** |
| `targeting` | 10 | **4** |
| `plugin-session-replay-browser` | 19 | **15** |

~1.5 MB off the consumer install path, plus `@aws-sdk/client-s3` (3.2 MB) out of 5 packages' devDeps. `targeting` drops 10 → 4 because removing `@amplitude/analytics-core` cascades to its own deps.

No load-time benefit for consumers. The lasting value is the CI gate stopping re-accumulation — it already caught one dead dependency my own config had masked.

## `knip.config.js`

A few dependencies are genuinely used without being imported, so they're ignored — each verified individually, with the reason in a comment:

- `tslib` — injected by `"importHelpers": true`
- the rollup plugins — imported by the shared `scripts/build/rollup.config.js`, but must resolve from each package's own `node_modules`
- a handful named as string literals in build configs, or read as built artifacts off disk

The ignore lists are **derived**, not hand-written: a dependency is ignored only if the package declares it, doesn't import it directly, and knip can't already resolve it. That keeps knip's configuration-hint output at zero, so a future hint means something.

## Also here

- `@rollup/plugin-replace` moved `dependencies` → `devDependencies` in `session-replay-browser`. Knip didn't flag it (it's used, by `rollup.config.js`), but a build-only tool in `dependencies` ships to consumers. Happy to drop this.
- Shared pnpm/Node CI setup extracted into `.github/actions/setup-node-pnpm` (−45 lines). `check-size-limits` keeps its steps inline — it checks out the base branch, which breaks a local composite action's post step.
- Corrected stale Node versions in `AGENTS.md` and `.nvmrc` (CI runs 20/22/24; `lerna` already requires ≥20.19).

## Before merging

**Branch protection** needs `Check Unused Dependencies` added to the required checks on `main`, or the job runs without gating.

## Follow-ups

- `examples/**` and `packages/analytics-react-native-test` are excluded (vendored `.yarn` releases; on-device harness wired through metro configs by name).
- `packages/unified` and `plugin-experiment-browser` import the shared rollup config but declare none of its plugins — they build only via pnpm store fallback. Separate ticket: that's undeclared deps, not unused ones.

## Verification

`pnpm lint:deps`, `pnpm build`, and `pnpm test:unit` all pass. Injecting an unused dependency makes the check exit 1 and name it.


package.json
```diff
@@ -66,22 +66,19 @@
...
     "eslint-plugin-import": "^2.32.0",
     "eslint-plugin-jest": "^27.1.6",
-    "eslint-plugin-playwright": "^2.2.0",
     "express": "^4.18.2",
     "fake-indexeddb": "4.0.2",
-    "glob": "^10.3.10",
     "husky": "^8.0.2",
     "jest": "^29.3.1",
     "jest-environment-jsdom": "^29.3.1",
+    "knip": "^6.27.0",
```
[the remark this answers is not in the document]
Mercy811: Good catch on the inconsistency, but the premise is the stale doc rather than the Knip version — so I fixed it the other way, in 36103dd26.

Three things:

1. **No workflow on `main` runs Node 18.** Both matrices in `ci.yml` and `ci-nx.yml` are `[20.x, 22.x, 24.x]`. The `18.17.x` in `AGENTS.md:28` was simply out of date; the only Node 18 workflow is `publish-v1.yml`, which checks out `ref: v1.x` where this change won't exist.

2. **Node 18.17 already cannot run the mandated validation, independent of Knip.** `lerna` — already a root devDependency, and what backs `pnpm build`, `pnpm test` and `pnpm lint` — requires `^20.19.0 || ^22.12.0 || >=24.0.0`. That's the same floor Knip 6 asks for, so pinning an older Knip would not make any of the documented steps work on 18.

3. Downgrading would also cost accuracy. I evaluated the Node-18-compatible alternative (`depcheck`) and rejected it: it reports `invalidFiles` for nearly every `jest.config.js` in this repo (each does `const package = require('./package')`, and `package` is a reserved word under its strict-mode parser) and every `tsconfig.json` (trailing commas). Anything referenced only from those files reads as unused, which would have produced ~15 false positives on test-only dependencies.

So instead of changing the Knip version I corrected `AGENTS.md` to state the real matrix and why Node 18 is unsupported, and bumped `.nvmrc` from `18` to `20`. Nothing in CI reads `.nvmrc`, so it was purely a local-dev trap: `nvm use` handed contributors a Node that couldn't run any of the documented commands.

Mercy811:
Addressed the review in 1190b5055.

**Should-fix 1 — 163 config hints.** Fair catch, and a real blind spot: I ran knip with `--no-config-hints` throughout my own verification, which suppressed exactly the diagnostics telling me the ignore list was imprecise.

Rather than hand-scope 20+ workspace entries or paper over it with `--no-config-hints`, the lists are now **derived**. A dependency is ignored only if the package declares it, doesn't import it directly in its own `rollup.config.js`, and knip can't already resolve it. Hint output is **zero**, and the reasoning still lives in one commented place. It's also self-maintaining — a package that drops a rollup plugin, or starts importing one directly, stops ignoring it automatically. `gtm-snippet` now correctly gets only the three plugins its bundle actually uses instead of all nine.

Chasing the last 3 hints turned up something I'd gotten wrong: **`@babel/plugin-transform-modules-commonjs` never needed ignoring** — knip's jest plugin does resolve the nested babel plugin string. My original PR description claimed otherwise; I've corrected it. The `tslib` residue in the two web-worker packages is recorded in `RESOLVED_WITHOUT_IGNORE`, each verified by dropping the entry and confirming knip still reports nothing.

**Should-fix 2 — dead `jest.setup.examples.js` entry.** Dropped. Thanks for tracing it to `ecece43fd`. I did *not* touch the stale `test:examples` script that still references it — that's pre-existing breakage unrelated to this PR, so it's better as its own change than buried here.

**Nit 1 — matrix redundancy.** Hoisted out of the 3-way Node matrix in `ci.yml` into its own single-run job, matching `ci-nx.yml`.

**Nit 2 — redundant entry patterns.** Removed `vite.config.js`, `playwright.config.ts` and `.size-limit.js`; knip's built-in plugins detect them.

**Nit 3 — branch protection.** Can't verify or change that from the diff, and it needs repo-admin rights. Flagging for whoever merges: `Check Unused Dependencies` needs adding to the required checks on `main`, or the job runs without gating.

Separately, I've since confirmed the rule actually fails on a regression rather than just passing on a clean tree — injecting an unused dep makes it exit 1 and name the exact package and dependency.

Thanks for the `workspace:*` vs pinned-version distinction on the build graph — that's a cleaner argument than the one I had, since `rrweb-packer`/`rrweb-utils` were never in the local nx graph at all.

Mercy811:
Blocking finding fixed in `aad9a9b61`. Good catch — that was a real hole, and the irony isn't lost on me: the deriver I added to reduce noise was masking exactly the kind of dead entry this check exists to find.

**Root cause.** My filter assumed "in the shared list ∧ declared ∧ not self-imported ⇒ used via `scripts/build/rollup.config.js`". That premise is false for `segment-session-replay-plugin` — as you found, the only package whose `rollup.config.js` never imports the shared config.

**Fix.** Split the one list into two, each gated on the right condition:

- `IMPLICIT_TS_DEPS` (`tslib`) applies to **every** package, since `"importHelpers": true` injects it into any TypeScript output. Worth noting this is why the narrower fix you suggested — gating the whole list on shared-config usage — wouldn't quite work: 13 packages have no `rollup.config.js` at all and would have started failing on `tslib`.
- `SHARED_ROLLUP_PLUGINS` (the eight plugins) applies **only** to packages that actually import the shared config. A self-contained config now gets no free pass.

**Verified in the order that proves it.** Before removing anything, I re-ran knip with the new deriver and confirmed it now *flags* the dependency and exits 1:

```
Unused devDependencies (1)
rollup-plugin-execute  packages/segment-session-replay-plugin/package.json:61:6
```

Then removed it — it's simply dead, your recommended fix — and knip returns to zero findings and zero hints. `build` and `test` pass for that package (29 tests, 100% coverage). That makes 30 removals total.

**Nit 1** — added a comment on `importedBy` documenting that it's a regex, not a parser, and spelling out that both failure directions surface loudly: a missed import leaves a redundant ignore that knip reports as a hint, and a matched comment drops a needed ignore and fails CI with the dependency named.

One process note: your round-2 investigation left a temporary `RESOLVED_WITHOUT_IGNORE` entry in my working tree mid-run. It was reverted and nothing stray reached any commit — I diffed the working tree, `HEAD`, and the remote to confirm — but flagging it since a reviewing agent editing files under review can confuse the author's own verification.

Mercy811:
bugbot run

Mercy811:
bugbot run

.github/actions/setup-node-pnpm/action.yml
```diff
@@ -0,0 +1,29 @@
+name: 'Setup Node and pnpm'
+description: 'Set up pnpm and Node.js with a pnpm cache, and install workspace dependencies'
+
+# Callers still need their own actions/checkout step: a local action can only be
+# resolved once the repository is on disk. For the same reason, a job that
+# replaces the working tree mid-run cannot use this action — its post step would
+# fail once this file is gone. See the check-size-limits job in ci-nx.yml.
```
daniel-graham-amplitude: What is this comment in reference to?

daniel-graham-amplitude:
Lgtm!

daniel-graham-amplitude (approved):

.github/actions/setup-node-pnpm/action.yml
Mercy811: This is for https://github.com/amplitude/Amplitude-TypeScript/pull/1909/changes/fa34f876de67bf5786fd29c8d8626024a21e071f. The job failed, with this error:
```
 Can't find 'action.yml' under '.../.github/actions/setup-node-pnpm'. Did you forget to run actions/checkout before running your local action?
```
At the end of every job, GitHub runs cleanup steps for the actions it used. This action's cleanup step saves the pnpm cache. size-limit-action swaps the checked-out code back to the base branch. When cleanup time comes, GitHub goes looking for action.yml, doesn't find it, and fails the job. This is a bootstrap problem as the file is not on the main branch. I'll create anther pr to use it in size-limit-action after it's merged.

jxiwang (approved):

PR #1914, opened 2026-07-29
chore(analytics-browser): add platform and web_environment tags to diagnostics

Mercy811:
### Summary

Adds two diagnostics tags on the browser SDK, set at init in `browser-client.ts` next to the existing `library` and `user_agent` tags:

| Tag | Value |
|---|---|
| `platform` | `Web` — reuses the `BROWSER_PLATFORM` constant backing the analytics event `platform` field (`src/plugins/context.ts`), and matches the mobile diagnostics values ([Swift](https://github.com/amplitude/AmplitudeCore-Swift/blob/ff6c51468788daa82e5cc559ce519581f21dcf6e/Sources/AmplitudeCore/Diagnostics/DiagnosticsClient.swift#L268): `iOS`/`tvOS`/`visionOS`/`macOS`/`watchOS`, [Kotlin](https://github.com/amplitude/Amplitude-Kotlin/blob/45ead59d3acf1459c82bfc3718b67cc2066bedc3/android/src/main/java/com/amplitude/android/diagnostics/AndroidDiagnosticsContextProvider.kt#L17): `Android`) |
| `web_environment` | `browser` \| `web_worker` \| `service_worker` \| `chrome_extension` \| `chrome_extension_service_worker` \| `node` \| `unknown` |

`web_environment` refines `platform: Web` by JS runtime — each value is a distinct event-loss regime (workers have no `localStorage` so the event queue falls back to `MemoryStorage`; service workers are additionally killed after ~30s idle and can't use `xhr`/`beacon` transports). Named `web_environment` because Datadog already auto-tracks an `environment` tag. Full design incl. the API-availability-per-environment table: [SDKW-47 design comment](https://linear.app/amplitude/issue/SDKW-47/diagnostics-add-platform-web#comment-d282be3e).

Detection is internal to `analytics-browser` (`src/utils/environment.ts`, `getRuntimeEnvironment()`) — it's only consumed here, and core stays light since it's shared by all platforms. `analytics-core` is untouched. Check order is load-bearing:

1. service worker before web worker — `ServiceWorkerGlobalScope` inherits from `WorkerGlobalScope`
2. `browser` after the `chrome.runtime.id` split — extension pages and content scripts have a `document`
3. `browser` before `node` — jsdom exposes both `document` and `process.versions.node`

### Out of scope

Tracking the *exact* event-drop reason (storage fallback, retry exhaustion, transport unsupported): SDKW-48.

### Testing

* `analytics-browser` — 499 tests pass, coverage 100% on all four metrics (repo gate); new suite covers every `getRuntimeEnvironment` branch, incl. service-worker-vs-web-worker precedence, extension variants, and exposed-but-not-instance worker constructors
* `analytics-core` — untouched, 869 tests pass
* `eslint` clean on changed files

Closes SDKW-47 — https://linear.app/amplitude/issue/SDKW-47/diagnostics-add-platform-web

### Checklist

---

> [!NOTE]
> **Low Risk**
> Observability-only metadata at init; no changes to event pipeline, storage, or transports.
> 
> **Overview**
> Adds **`platform`** (`Web`, from exported `BROWSER_PLATFORM`) and **`web_environment`** (from new `getRuntimeEnvironment()`) as diagnostics tags during browser SDK init, alongside existing `library` and `user_agent` tags.
> 
> **`getRuntimeEnvironment()`** in `src/utils/environment.ts` classifies the JS runtime (browser, workers, Chrome extension contexts, Node, or `unknown`) using `getGlobalScope()` with an explicit check order (service worker before web worker, extension split on main thread, browser before Node for jsdom). Tests cover init tag wiring and every classification branch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 806320be0adf2cce759f10da98e5402ffa5d3563. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mercy811:
bugbot run

packages/analytics-browser/src/utils/environment.ts
```diff
@@ -0,0 +1,75 @@
...
+  return typeof (scope as { process?: { versions?: { node?: unknown } } }).process?.versions?.node === 'string';
+};
+
+/**
+ * Classifies the JS runtime for diagnostics. The check order is load-bearing:
+ * - service worker before web worker: ServiceWorkerGlobalScope inherits from WorkerGlobalScope
+ * - browser after the extension split: extension pages and content scripts have a `document`
+ * - browser before node: jsdom exposes both `document` and `process.versions.node`
+ */
+export function getWebEnvironment(): WebEnvironment {
```
daniel-graham-amplitude: Maybe a bit pedantic, but I feel like WebEnvironment isn't quite correct, considering that some of these aren't web (like Node). Could we call it something like JsEnvironment or RuntimeEnvironment?

Mercy811: I see. I named it as is because the platform is always web and the environment means where the web SDK runs on. But I got what you mean. Renamed it to `getRuntimeEnvironment()`

daniel-graham-amplitude (approved):

PR #1919, opened 2026-07-31
fix(analytics-browser): tolerate nullish querySelectorAll in mutation observers

### LINKED ISSUE #1918: formInteractionTracking: unhandled TypeError and silent form loss when an added node's querySelectorAll returns nullish
## Expected Behavior

`formInteractionTracking`'s `MutationObserver` callback should tolerate a `querySelectorAll` that returns a nullish value, and a single problematic added node should not prevent the remaining nodes in the same mutation batch from being registered.

## Current Behavior

In `packages/analytics-browser/src/plugins/form-interaction-tracking.ts`, the observer callback guards that `querySelectorAll` **exists and is callable**, but not that its **return value** is iterable:

```ts
mutation.addedNodes.forEach((node) => {
  if (node.nodeName === 'FORM') {
    addFormInteractionListener(node as HTMLFormElement);
  }
  if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
    Array.from(node.querySelectorAll('form') as HTMLFormElement[]).map(addFormInteractionListener);
  }
});
```

When `node.querySelectorAll('form')` returns `undefined`, `Array.from` throws. Because the throw happens inside nested `forEach` callbacks, it aborts iteration of both `mutation.addedNodes` and the outer `mutations` array, then escapes the `MutationObserver` callback as an **unhandled** error.

Two consequences:

1. **Silent, permanent data loss.** Every form in that mutation batch after the offending node is never registered, so `Form Started` / `Form Submitted` never fire for those forms. The observer survives for later batches, so the loss is partial and invisible — it looks like users simply didn't interact with those forms.
2. **An unhandled `TypeError` in the host page**, attributed to the bundle the SDK is compiled into.

Note the `as HTMLFormElement[]` cast is also unsound — `querySelectorAll` returns a `NodeList`, not an array — which is why the compiler doesn't flag the nullish case.

### The same throw, worded differently per engine

This surfaces as two distinct error messages, which makes it easy to mistake for two unrelated bugs:

- **V8** (Chrome/Edge/Samsung): `TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
- **JavaScriptCore** (Safari/iOS): `TypeError: Array.from requires an array-like object - not null or undefined`

Representative production stack (V8), against a bundle containing `analytics-browser`:

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
  at Array.from @ <anonymous>
  at <anonymous> @ /_next/static/chunks/<hash>.js      <- Array.from(node.querySelectorAll('form'))
  at NodeList.forEach @ <anonymous>                     <- mutation.addedNodes.forEach
  at <anonymous> @ /_next/static/chunks/<hash>.js
  at Array.forEach @ <anonymous>                        <- mutations.forEach
  at MutationObserver.<anonymous> @ /_next/static/chunks/<hash>.js
```

## Steps to Reproduce

The nullish return is environment-dependent, so here is a deterministic reproduction of the robustness gap. This is the plugin's callback copied verbatim, with one added node whose `querySelectorAll` returns `undefined`. Paste into any browser console:

```js
function runScenario(poison) {
  const registered = [];
  const addFormInteractionListener = (f) => registered.push(f.id || '(anon)');

  // verbatim from form-interaction-tracking.ts
  const observer = new MutationObserver((mutations) => {
    mutations.forEach((mutation) => {
      mutation.addedNodes.forEach((node) => {
        if (node.nodeName === 'FORM') {
          addFormInteractionListener(node);
        }
        if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
          Array.from(node.querySelectorAll('form')).map(addFormInteractionListener);
        }
      });
    });
  });
  observer.observe(document.body, { subtree: true, childList: true });

  const mk = (id) => {
    const d = document.createElement('div');
    const f = document.createElement('form');
    f.id = id;
    d.appendChild(f);
    return d;
  };

  const w1 = mk('early');
  const bad = document.createElement('div');
  if (poison) bad.querySelectorAll = function () { return undefined; };
  const w2 = mk('late');

  document.body.append(w1, bad, w2);   // one mutation batch

  return new Promise((resolve) => setTimeout(() => {
    observer.disconnect();
    [w1, bad, w2].forEach((n) => n.remove());
    resolve(registered);
  }, 50));
}

console.log('control:', await runScenario(false));
console.log('poisoned:', await runScenario(true));
```

Observed:

```
control:  ["early", "late"]
poisoned: ["early"]            // "late" silently never registered
Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

`late` is a perfectly ordinary form, unrelated to the problematic node, and it is lost purely because it was appended later in the same batch.

## Field data

From production RUM on a site running `analytics-browser` with `defaultTracking: true`: **642 occurrences over 30 days**, from real user sessions (bots excluded), across **21 browser/OS combinations**. It is not engine- or platform-specific:

| Browser | OS | Count |
|---|---|---|
| Mobile Safari | iOS | 270 |
| Chrome | Windows | 66 |
| Chrome Mobile | Android | 62 |
| Chrome Mobile iOS | iOS | 54 |
| Safari | macOS | 40 |
| Mobile Safari UI/WKWebView | iOS | 34 |
| Chrome | macOS | 32 |
| Samsung Internet | Android | 14 |
| Chrome | ChromeOS | 10 |
| Chrome Mobile WebView | Android | 10 |
| Edge | Windows | 10 |
| _(+10 more, 2–6 each)_ | | |

All events are `handling: unhandled`. The oldest occurrence we retain dates to 2024-10-24, so this has been present for a long time.

**On the underlying cause:** per spec `querySelectorAll` returns a `NodeList` and never `null`, so something in these environments is patching or shadowing it. Two of our affected sessions are suggestive — one in an Instagram in-app webview, one behind a Zscaler corporate proxy, both environments that inject or rewrite page scripts. We can't prove that's the mechanism, and it may vary. But the SDK doesn't need to know the cause to be resilient to it: it already accepts that the DOM may be hostile enough to warrant the `'querySelectorAll' in node && typeof … === 'function'` guard, and validating the return value is the same class of defense.

## Possible Solution

Guard the return value, and drop the unsound cast:

```ts
if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
  const forms = node.querySelectorAll('form');
  if (forms) {
    Array.from(forms).forEach(addFormInteractionListener);
  }
}
```

Optionally also wrap the per-node body in a `try/catch` that logs via `config.loggerProvider`, so no single malformed node can abort registration for the rest of the batch — that addresses the silent-data-loss half independently of this specific nullish case.

For reference, the adjacent `observer.observe(document.body, …)` line in this same `setup()` had a comparable robustness bug (#678, fixed in #764). This line appears to have been missed by that pass.

## Environment

- JS SDK Version: reproduced by inspection in **2.2.3** and **2.45.5** (latest at time of writing) — the line is byte-for-byte identical in both, so all versions in between are presumably affected
- Installation Method: npm/pnpm
- Browser and Version: engine-independent; confirmed in Chrome (V8) and Safari/iOS (JSC) — see table above

Workaround for anyone hitting this: disable the plugin via `defaultTracking: { formInteractions: false }` (or `autocapture: { formInteractions: false }` on newer versions), at the cost of losing `Form Started` / `Form Submitted`.

Mercy811:
### Summary

Fixes amplitude/Amplitude-TypeScript#1918

`formInteractionTracking` and `fileDownloadTracking` guard that an added node's `querySelectorAll` **exists and is callable**, but not that its **return value** is iterable:

```ts
if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
  Array.from(node.querySelectorAll('form') as HTMLFormElement[]).map(addFormInteractionListener);
}
```

When a patched `querySelectorAll` returns a nullish value, `Array.from` throws. Because the throw happens inside nested `forEach` callbacks, it aborts iteration of both `mutation.addedNodes` and the outer `mutations` array, then escapes the `MutationObserver` callback as an **unhandled** error. Two consequences:

1. **Silent, partial data loss.** Every form/link appended after the offending node in the same mutation batch is never registered, so `Form Started` / `Form Submitted` / `File Downloaded` never fire for those elements. The observer survives for later batches, so the loss is invisible — it looks like users simply didn't interact.
2. **An unhandled** `TypeError` attributed to the host page's bundle.

Per amplitude/Amplitude-TypeScript#1918, production RUM shows **642 occurrences over 30 days across 21 browser/OS combinations**, all `handling: unhandled`, with retained occurrences dating to 2024-10-24. The `as HTMLFormElement[]` cast is what kept the compiler from flagging the nullish case — `querySelectorAll` returns a `NodeList`, not an array.

`file-download-tracking.ts` had the byte-identical pattern for anchor tags and failed the same way, so both are fixed here.

### Changes

Per node, in both plugins:

* **Guard the return value** before iterating.
* **Drop the unsound casts** (`as HTMLFormElement[]` / `as HTMLAnchorElement[]`). The node is narrowed to `Element` so `querySelectorAll('form')` resolves to `NodeListOf<HTMLFormElement>` via the tag-name overload, and the result is annotated `| null | undefined` — a sound widening that admits the type lies at runtime, rather than a narrowing cast that hides it. No `any` is reintroduced, which matters because `no-unsafe-*` is active in `src/`.
* **Wrap the per-node body in** `try`**/**`catch` that logs via `config.loggerProvider.warn`, so one malformed node cannot abort registration for the rest of the batch. This addresses the silent-data-loss half independently of this specific nullish case — e.g. a hostile node whose `querySelectorAll` throws outright.
* `Array.from` is no longer needed: `querySelectorAll` returns a static `NodeList`, so there is nothing to snapshot, and `NodeList.forEach` is already relied on for `mutation.addedNodes` in these same callbacks.

Also corrects a copy-paste comment in the form plugin ("anchor tags" → "form tags") on the block being rewritten.

### Tests

Four regression tests, two per plugin, in the existing suites:

* **nullish return** — a node whose `querySelectorAll` returns `undefined`, appended *between* two wrapped elements in a single mutation batch; asserts the later element is still registered and tracked.
* **batch continuation on throw** — a node whose `querySelectorAll` throws; asserts the warning is logged and the element appended after it is still registered.

Verified these are genuine regression tests: with the `src` changes stashed, exactly these **4 fail** (with the production error, `TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))` at `Array.from`) and the 24 pre-existing tests pass. With the fix, all 28 pass, and both changed files stay at **100% statements/branches/functions/lines** — no new `istanbul ignore` needed.

### Out of scope

`querySelectUniqueElements` in `plugin-autocapture-browser/src/helpers.ts` has the same guard-existence-but-not-return-value gap, but it is **unreachable**: no production module imports it (only its own test does) and it is not re-exported from that package's `index.ts`. Fixing it would mean a second package release with no user-facing effect; the more appropriate follow-up is deleting it, which `pnpm lint:deps` / knip territory covers better than this PR.

### Checklist

- [X] Does your PR title have the correct [title format](<https://github.com/amplitude/Amplitude-TypeScript/blob/main/CONTRIBUTING.md#pr-commit-title-conventions>)?
- Does your PR have a breaking change?: No — behavior is unchanged for conforming DOM nodes; previously-throwing nodes are now skipped with a warning.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

---

> < />!NOTE< /> **Low Risk**  <br>Defensive bugfix in analytics DOM observers; behavior is unchanged for normal nodes and only adds resilience plus warning logs.
>
> **Overview**  <br>Fixes silent data loss in `fileDownloadTracking` and `formInteractionTracking` when a dynamically added node's `querySelectorAll` returns nullish or throws.
>
> Previously, `Array.from(node.querySelectorAll(...))` inside nested `forEach` callbacks aborted the entire mutation batch on failure, so later links/forms were never registered. Per-node processing is now wrapped in `try`/`catch` (with a warn log), nullish returns are guarded via optional chaining, and unsound casts are dropped. Regression tests cover both the nullish-return and throw cases.
>
> Reviewed by [Cursor Bugbot](<https://cursor.com/bugbot>) for commit 35eadf0345ab59ac9864e5a79664082a95c51051. Configure [here](<https://www.cursor.com/dashboard/bugbot>).

### DIFF
--- a/packages/analytics-browser/src/plugins/file-download-tracking.ts
+++ b/packages/analytics-browser/src/plugins/file-download-tracking.ts
@@ -86,11 +86,21 @@ export const fileDownloadTracking = (): EnrichmentPlugin => {
         observer = new MutationObserver((mutations) => {
           mutations.forEach((mutation) => {
             mutation.addedNodes.forEach((node) => {
-              if (node.nodeName === 'A') {
-                addFileDownloadListener(node as HTMLAnchorElement);
-              }
-              if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
-                Array.from(node.querySelectorAll('a') as HTMLAnchorElement[]).map(addFileDownloadListener);
+              // Registration of one node must not abort registration of the rest of the batch, otherwise
+              // links appended after a malformed node are silently never tracked.
+              try {
+                if (node.nodeName === 'A') {
+                  addFileDownloadListener(node as HTMLAnchorElement);
+                }
+                if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
+                  // Some environments patch querySelectorAll to return a nullish value, so the return type
+                  // is widened to reflect what may actually arrive at runtime.
+                  const queryableNode = node as Element;
+                  const links: NodeListOf<HTMLAnchorElement> | null | undefined = queryableNode.querySelectorAll('a');
+                  links?.forEach(addFileDownloadListener);
+                }
+              } catch (error) {
+                config.loggerProvider.warn(`Failed to track file downloads for an added node: ${String(error)}`);
               }
             });
           });
--- a/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
+++ b/packages/analytics-browser/src/plugins/form-interaction-tracking.ts
@@ -131,17 +131,27 @@ export const formInteractionTracking = (): EnrichmentPlugin => {
       const forms = Array.from(document.getElementsByTagName('form'));
       forms.forEach(addFormInteractionListener);
 
-      // Adds listener to anchor tags added after initial load
+      // Adds listener to form tags added after initial load
       /* istanbul ignore else */
       if (typeof MutationObserver !== 'undefined') {
         observer = new MutationObserver((mutations) => {
           mutations.forEach((mutation) => {
             mutation.addedNodes.forEach((node) => {
-              if (node.nodeName === 'FORM') {
-                addFormInteractionListener(node as HTMLFormElement);
-              }
-              if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
-                Array.from(node.querySelectorAll('form') as HTMLFormElement[]).map(addFormInteractionListener);
+              // Registration of one node must not abort registration of the rest of the batch, otherwise
+              // forms appended after a malformed node are silently never tracked.
+              try {
+                if (node.nodeName === 'FORM') {
+                  addFormInteractionListener(node as HTMLFormElement);
+                }
+                if ('querySelectorAll' in node && typeof node.querySelectorAll === 'function') {
+                  // Some environments patch querySelectorAll to return a nullish value, so the return type
+                  // is widened to reflect what may actually arrive at runtime.
+                  const queryableNode = node as Element;
+                  const forms: NodeListOf<HTMLFormElement> | null | undefined = queryableNode.querySelectorAll('form');
+                  forms?.forEach(addFormInteractionListener);
+                }
+              } catch (error) {
+                config.loggerProvider.warn(`Failed to track form interactions for an added node: ${String(error)}`);
               }
             });
           });
--- a/packages/analytics-browser/test/plugins/file-download-tracking.test.ts
+++ b/packages/analytics-browser/test/plugins/file-download-tracking.test.ts
@@ -142,6 +142,110 @@ describe('fileDownloadTracking', () => {
     });
   });
 
+  describe('malformed added nodes', () => {
+    const createLinkInWrapper = (id: string) => {
+      const link = document.createElement('a');
+      link.setAttribute('id', id);
+      link.setAttribute('href', `https://analytics.amplitude.com/files/${id}.pdf`);
+      link.text = `${id}-text`;
+
+      const wrapper = document.createElement('div');
+      wrapper.appendChild(link);
+
+      return { link, wrapper };
+    };
+
+    const expectedEventProperties = (id: string) => ({
+      [FILE_EXTENSION]: 'pdf',
+      [FILE_NAME]: `/files/${id}.pdf`,
+      [LINK_ID]: id,
+      [LINK_TEXT]: `${id}-text`,
+      [LINK_URL]: `https://analytics.amplitude.com/files/${id}.pdf`,
+    });
+
+    test('should track links added after a node whose querySelectorAll returns a nullish value', async () => {
+      // setup
+      const config = createConfigurationMock();
+      const plugin = fileDownloadTracking();
+      await plugin.setup?.(config, amplitude);
+      window.dispatchEvent(new Event('load'));
+
+      const early = createLinkInWrapper('early-link-id');
+      const late = createLinkInWrapper('late-link-id');
+
+      // simulate an environment that patches querySelectorAll to return a nullish value
+      const poisoned = document.createElement('div');
+      Object.defineProperty(poisoned, 'querySelectorAll', {
+        value: () => undefined,
+        configurable: true,
+      });
+
+      // append all three in a single mutation batch, with the poisoned node between the two links
+      document.body.append(early.wrapper, poisoned, late.wrapper);
+
+      // allow mutation observer to execute and event listeners to be attached
+      await new Promise((r) => r(undefined)); // basically, await next clock tick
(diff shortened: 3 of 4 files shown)


Mercy811:
bugbot run

daniel-graham-amplitude (approved):
Lgtm

PR #1922, opened 2026-08-10
feat(zoning): add support for shadow dom

cely404:
### Summary
Adds remotely gated Shadow DOM traversal and autocapture support across element selection, event enrichment, and observers.

Changes:

Adds depth-bounded Shadow DOM selector generation and resolution.
Extends autocapture, mutation, and exposure handling into open shadow roots.
Adds documentation, unit/E2E coverage, and manual performance guards.

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches hot paths (every click/mutation) and arbitrary customer DOM with new observer fan-out when shadow is enabled; defaults and kill-switch tests limit blast radius, but mis-targeted selectors or perf on shadow-heavy SPAs are plausible rollout risks.
> 
> **Overview**
> Adds **opt-in open shadow DOM support** for element selectors and browser autocapture, controlled independently from the v1 selector engine via remote config `shadowDomEnabled` and `maxShadowDomDepth` (defaults off, depth clamped 1–10).
> 
> **`@amplitude/element-selector`** generates composite paths joined by ` >>> ` when piercing is on, exports `resolveSelector` (required instead of `document.querySelector`), and keeps **byte-identical output when shadow is off**. The engine splits generation into per-tree segments with traversal caps and shadow-root anchoring for positional chains.
> 
> **`@amplitude/plugin-autocapture-browser`** wires a page-scoped **`ShadowGate`** that latches shadow mode on first enabling remote config; updates hierarchy, closest-ancestor, text masking, `composedPath` target resolution, and **mutation/exposure observers** that fan out into discovered open shadow roots. Remote-config handling now accepts explicit `shadowDomEnabled` (not only `enabled`), and event enrichment is wrapped in a try/catch so DOM failures do not crash the page.
> 
> Docs (`SHADOW-DOM.md`, element-selector README/schema), unit/scenario tests, Playwright e2e, and a manual perf spec are included. The **analytics-browser gzip size limit** moves 65kb → 66kb for the added bundle cost.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52019126d8dbdddaef329a6a52b81b92dd9f4d93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


daniel-graham-amplitude:
bugbot run

packages/plugin-autocapture-browser/src/helpers.ts
```diff
@@ -199,7 +216,91 @@ export const getClosestElement = (element: Element | null, selectors: string[]):
...
+export const querySelectorAllDeep = (
+  root: Element | Document,
+  selectorString: string,
+  shadow: ShadowMode = SHADOW_OFF,
+): Element[] =>
+  shadow.enabled
+    ? querySelectorAllDeepInShadow(root, selectorString, shadow.maxDepth)
+    : querySelectorAllDeepLight(root, selectorString);
+
+const querySelectorAllDeepLight = (root: Element | Document, selectorString: string): Element[] => {
```
daniel-graham-amplitude, line 250: Would it work if we made a return type like:

```
type ElementList = Element[] || NodeListOf<Element>;
```

And then just have this function return the result of `querySelectorAll` without having to coerce it into a JS Array?

packages/plugin-autocapture-browser/src/observables.ts
```diff
@@ -1,23 +1,148 @@
...
       observer.next(mutations);
+
+      // Mutations are emitted first, then shadow discovery runs. The whole
+      // branch is wrapped because this is a native callback: a throw here would
+      // surface on the host page's `window.onerror`.
+      try {
+        const shadow = readGate(shadowGate);
+        if (!shadow.enabled) {
+          return;
+        }
```
daniel-graham-amplitude, line 111: I noticed that you still observe for mutations even if `!shadow.enabled`. Is this because it's possible for shadow capture to be enabled later and we want the observer to be running when that happens?

packages/element-selector/src/helpers/shadow.ts
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/**
+ * Collect the OPEN shadow roots within `root`'s composed subtree (its own light
+ * children and any nested shadow trees), each paired with how many shadow
+ * boundary crossings deep it sits relative to `root` (1-based). Bounded by
+ * `maxShadowDepth` crossings and by {@link MAX_SHADOW_DOM_TRAVERSAL_NODES}
+ * total nodes. Closed roots are invisible (`el.shadowRoot` is null) and skipped.
+ */
+export function collectOpenShadowRoots(root: Element, maxShadowDepth: number): OpenShadowRootEntry[] {
```
daniel-graham-amplitude, line 226: This function is the one addition to the code that gives me concerns about performance, being a depth first search traversal, it could cause some lengthy main thread blocking and add jank to our customers website.

Do you think it would be a good idea, not for the scope of this PR, if we added something like a `shadowNodeCssSelector` that, when provided by the customer, it identifies which nodes are shadow nodes.

Mercy811 (approved):
LGTM! I thought some code changes are for zoning only and thought we could split this into a smaller PR for autocapture but looks like this is already the smallest PR. 

One question: is shadowDomEnabled remote config public or it's on dynconf only?

cely404:
> LGTM! I thought some code changes are for zoning only and thought we could split this into a smaller PR for autocapture but looks like this is already the smallest PR.
> 
> One question: is shadowDomEnabled remote config public or it's on dynconf only?

Would you prefer I use stacked prs for this? Happy to do so if that is the preference. And as for rollout @jxiwang can you advise? I was planning to roll this out in the same way that the selector engine changes were rolled out.

packages/plugin-autocapture-browser/src/observables.ts
cely404: Zoning added an exposure event as part of the initial launch months ago so we need to observe for mutations with or without shadow dom piercing. Mutations are still emitted we just skip attaching/discovering open shadow roots when the gate is off.

Rn we have a single observer lifecycle.
We could swap to separate observables when remote config lands, but that would mean tearing down and reconnecting observers. During the switch we may drop mutations. 

Wdyt?

packages/element-selector/src/helpers/shadow.ts
cely404: Yea that sounds like a good idea. 

My cursor agent also recommends a registry where we do a DFS scan once when remote config lands and then register new roots incrementally when addedNodes exposes them.

packages/plugin-autocapture-browser/SHADOW-DOM.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Improving steady-state cost (if needed)
+
+Prefer these over patching `Element.prototype.attachShadow`:
+
+- `ShadowRootRegistry` — centralize discovery outside the mutation callback hot path
+- Cheaper `addedNodes` checks (e.g. `node.shadowRoot`) before DFS
+- Batched discovery and more targeted exposure indexing
+
+**Do not patch `attachShadow` by default.** Session replay (`@amplitude/rrweb-record`) and other libraries already wrap the same API; a global monkey patch is a last resort and needs dual-plugin validation (autocapture + session replay) before consideration.
```
cely404, line 104: https://github.com/amplitude/rrweb/blob/master/packages/rrweb/src/record/shadow-dom-manager.ts#L118 

Probably okay to do if we verify things work as expected

jxiwang:
@cely404 I'd lean towards remote config settings for more flexibility in the future. Would this be something we enable by default in the future? I'm assuming not since there are some performance concerns.

packages/element-selector/src/helpers/shadow.ts
daniel-graham-amplitude: Okay, yeah let's leave it as a follow-up.

daniel-graham-amplitude (approved):

PR #1923, opened 2026-08-10
fix(analytics-node): add a request timeout to the HTTP transport

Mercy811:
Fixes [SDK-188](https://linear.app/amplitude/issue/SDK-188/memory-leak-in-analytics-node-due-to-missing-http-timeout).

## What's the problem?

`@amplitude/analytics-node` never gave up on a broken network request. If the connection to Amplitude's servers stalled, the SDK waited forever for a response. While it waited, it stopped sending anything else, but kept collecting new events in memory. Memory grew without bound until the process crashed.

Reproduced against the published SDK: 20,000 events into a server that never responds → 1 request sent, 0 events ever delivered, +118 MB of heap growth that never comes back down.

## How do we solve it?

We give every upload a **10-second time limit**, matching what our Java and Python SDKs already do. If a full response doesn't come back in time, we cancel the request and treat it as a failure instead of waiting forever. We fixed a couple of related "wait forever" cases the same way: an empty response body, and a response that gets cut off mid-way.

New optional config, for anyone who wants a different limit:

```ts
amplitude.init(API_KEY, { requestTimeoutMillis: 5000 });
```

## What happens now if a request stalls?

It **retries, it doesn't drop**. A timed-out or cut-off upload goes back into the in-memory queue and retries with backoff, up to 12 attempts by default (`flushMaxRetries`) — same as any other failure, before or after this fix. Verified live: with the retry limit set to 4, a stalled upload retried exactly 4 times before giving up.

One case is intentionally left as-is: a request that never reaches the server at all (connection refused, bad DNS) still drops immediately, same as before this PR. That's existing behavior and out of scope here.

## Anything risky?

Low risk. We didn't touch the existing retry/backoff logic — we only made sure a stalled request always reports a result instead of hanging silently, so it can enter that logic like any other failure would. All 38 tests pass (existing + new), and re-running the original repro shows memory stays flat and the queue drains to zero instead of growing forever.

## What's *not* fixed here

Unlike Java (`MAX_CACHED_EVENTS = 16000`) and Python (`MAX_BUFFER_CAPACITY = 20000`), the Node SDK still has **no cap** on the in-memory queue. If the server is down for a long time, events can still pile up unbounded — just slower now, since retries succeed once things recover instead of piling up instantly on the first try. That's shared logic used by the browser and React Native SDKs too, so it's a separate follow-up rather than part of this fix.

Tracked as [SDK-209](https://linear.app/amplitude/issue/SDK-209/cap-the-in-memory-event-queue-in-destination-analytics-core).

### DIFF
--- a/packages/analytics-node/src/transports/http.ts
+++ b/packages/analytics-node/src/transports/http.ts
@@ -3,8 +3,8 @@ import * as http from 'http';
 import * as https from 'https';
 
 /**
- * Node's HTTP client applies no socket or request timeout of its own, so without
- * this a stalled upload hangs forever. Matches the 10s network timeout used by
+ * Node's HTTP client applies no timeout of its own, so without this a stalled or
+ * slow-trickling upload hangs forever. Matches the 10s network timeout used by
  * the Amplitude Java and Python SDKs.
  */
 export const DEFAULT_REQUEST_TIMEOUT_MILLIS = 10000;
@@ -48,11 +48,15 @@ export class Http extends BaseTransport implements Transport {
       // leaves Destination.flushId set forever, which silently no-ops all later
       // flushes while the event queue keeps growing. See SDK-188.
       let settled = false;
+      // A plain object rather than a bare variable, so `deadline` can be assigned
+      // once req exists without ESLint flagging the let as a const candidate.
+      const timer: { deadline?: ReturnType<typeof setTimeout> } = {};
       const settle = (response: Response | null) => {
         if (settled) {
           return;
         }
         settled = true;
+        clearTimeout(timer.deadline);
         resolve(response);
       };
 
@@ -91,10 +95,17 @@ export class Http extends BaseTransport implements Transport {
       // Unlike the cases above, a request that never reached the server keeps its
       // long-standing drop-on-error behavior.
       req.on('error', () => settle(null));
-      req.setTimeout(this.requestTimeoutMillis, () => {
+
+      // An absolute deadline, not req.setTimeout()'s socket-inactivity timer:
+      // that resets on every byte of traffic, so a response trickled slower than
+      // requestTimeoutMillis but never finished would keep it from firing at all.
+      // unref() so it can't keep the event loop alive on its own.
+      timer.deadline = setTimeout(() => {
         req.destroy();
         settle(this.buildResponse({ code: REQUEST_TIMEOUT_STATUS_CODE }));
-      });
+      }, this.requestTimeoutMillis);
+      timer.deadline.unref();
+
       req.end(requestPayload);
     });
   }
--- a/packages/analytics-node/src/transports/http.ts
+++ b/packages/analytics-node/src/transports/http.ts
@@ -9,8 +9,11 @@ import * as https from 'https';
  */
 export const DEFAULT_REQUEST_TIMEOUT_MILLIS = 10000;
 
-// buildResponse() maps 408 to Status.Timeout, which Destination retries with backoff.
+// buildResponse() maps 408 to Status.Timeout and 0 to Status.Unknown. Destination
+// retries both with backoff, bounded by flushMaxRetries, rather than dropping the
+// batch the way it does for a null response.
 const REQUEST_TIMEOUT_STATUS_CODE = 408;
+const INCOMPLETE_RESPONSE_STATUS_CODE = 0;
 
 export class Http extends BaseTransport implements Transport {
   constructor(private readonly requestTimeoutMillis: number = DEFAULT_REQUEST_TIMEOUT_MILLIS) {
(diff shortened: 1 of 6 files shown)


packages/analytics-node/test/transport/http.test.ts
```diff
@@ -158,3 +164,99 @@ describe('http transport', () => {
...
+  const listen = async (handler: http.RequestListener) => {
+    server = http.createServer(handler);
+    await new Promise<void>((resolve) => server.listen(0, resolve));
+    url = `http://localhost:${(server.address() as AddressInfo).port}/2/httpapi`;
+  };
+
+  afterEach(async () => {
+    server.closeAllConnections();
+    await new Promise((resolve) => server.close(resolve));
+  });
```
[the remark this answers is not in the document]
Mercy811: Confirmed — running that test alone failed with a TypeError in afterEach. Fixed in 3ae99914a: server now resets to undefined after each test and the teardown helper checks server.listening before touching it, so every test in the block passes in isolation and as part of the suite.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mercy811:
bugbot run

packages/analytics-node/src/transports/http.ts
```diff
@@ -27,27 +44,57 @@ export class Http extends BaseTransport implements Transport {
...
+            settle(this.buildResponse({ code: res.statusCode }));
           }
         });
       });
-      req.on('error', () => resolve(null));
+
+      // Unlike the cases above, a request that never reached the server keeps its
+      // long-standing drop-on-error behavior.
+      req.on('error', () => settle(null));
+      req.setTimeout(this.requestTimeoutMillis, () => {
```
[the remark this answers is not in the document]
Mercy811: Confirmed and fixed in 4a210fe12. Replaced `req.setTimeout()` with an unref'd `setTimeout()` covering the whole upload, cleared in `settle()`. Verified against a server that trickles a byte every 500ms forever: the old timer never fired (request ran 20s+ unbounded); the new one fires at exactly the configured deadline (2000ms -> settles at 2007ms).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/analytics-node/src/transports/http.ts
```diff
@@ -27,27 +44,68 @@ export class Http extends BaseTransport implements Transport {
...
         res.setEncoding('utf8');
         let responsePayload = '';
         res.on('data', (chunk: string) => {
           responsePayload += chunk;
         });
+        res.on('aborted', settleAsIncomplete);
+        res.on('error', settleAsIncomplete);
 
         res.on('end', () => {
-          if (res.complete && responsePayload.length > 0) {
```
daniel-graham-amplitude, line 38: Iiuc, we don't check `responsePayload.length > 0` anymore, because that's already caught by the `try -> catch` below (`JSON.parse('')`)?

packages/analytics-node/src/transports/http.ts
```diff
@@ -27,27 +44,68 @@ export class Http extends BaseTransport implements Transport {
...
+
+      // An absolute deadline, not req.setTimeout()'s socket-inactivity timer:
+      // that resets on every byte of traffic, so a response trickled slower than
+      // requestTimeoutMillis but never finished would keep it from firing at all.
+      // unref() so it can't keep the event loop alive on its own.
+      timer.deadline = setTimeout(() => {
+        req.destroy();
+        settle(this.buildResponse({ code: REQUEST_TIMEOUT_STATUS_CODE }));
+      }, this.requestTimeoutMillis);
+      timer.deadline.unref();
```
daniel-graham-amplitude, line 109: Nitpick: Instead of defining `const timer: { deadline?: ReturnType<typeof setTimeout> } = {};` on line 53 and then this on line 103, could we just do this?

```
const timer = {
  deadline: setTimeout(() => {
    req.destroy();
    settle(this.buildResponse({ code: REQUEST_TIMEOUT_STATUS_CODE }));
  }, this.requestTimeoutMillis),
};
timer.deadline.unref();
```

It just saves on the cognitive effort to have them together.

packages/analytics-node/src/transports/http.ts
Mercy811: that's right. Currently if response body is empty, the block is skipped and the request hangs here forever.

packages/analytics-node/src/transports/http.ts
Mercy811: updated in https://github.com/amplitude/Amplitude-TypeScript/pull/1923/changes/73723d7554d493d90c18e1e01d8331115b910050

daniel-graham-amplitude (approved):

PR #1934, opened 2026-08-12
fix(plugin-autocapture-browser): flush Viewport Content Updated on pagehide/visibilitychange

jxiwang:
### Summary

`[Amplitude] Viewport Content Updated` was systematically under-delivered relative to `[Amplitude] Page Viewed`, especially for land-and-leave sessions. Two root causes:

1. **Delivery race on exit.** The event was fired inside a `beforeunload` handler, but `track()` only *schedules* a send on a timer (`flushIntervalMillis`, default 1s). The document is normally torn down before that timer runs, so the event was written to the `AMP_unsent_*` local storage queue and only replayed (with a stale timestamp) on the browser's next init — if the user ever returned. On `MemoryStorage` or for bounced users it was lost entirely.
2. **Missing exit signals.** `beforeunload` does not fire on mobile app-switch, tab discard, or bfcache eviction, so those page views produced no exit event at all.

This PR makes the exit path reliable **without changing the event's semantics** (still at most one page-end event per page view; mid-page 18k-char buffer flushes are unchanged):

- **Prefer `pagehide` over `beforeunload`** for the page-end trigger (feature-detected via a new `getPageEndEventName` helper; falls back to `beforeunload` when `pagehide` is unavailable). `pagehide` also fires on mobile teardown and bfcache eviction and, unlike `beforeunload`, doesn't make the page bfcache-ineligible. This mirrors the pattern already used by `session-replay-browser`.
- **Add a `visibilitychange` → `hidden` checkpoint.** This is the last signal reliably delivered before a tab is backgrounded and possibly discarded. It flushes as a mid-page checkpoint (`isPageEnd === false`) so page-level scroll/exposure state is preserved if the user returns; a real page end shortly after is collapsed by the existing 100ms dedupe window.
- **Flush on exit.** Both exit handlers call `amplitude.flush()` after firing, so the request starts (with `keepalive`, already set by the fetch transport for small bodies) while the page is still alive, instead of waiting on the interval timer.
- **Scope the page-end dedupe to page-end triggers only.** Previously any call (including mid-page buffer flushes) armed `pageViewEndFired`, which could suppress a genuine page-end event (and its state reset) that followed within 100ms. Now only `isPageEnd === true` triggers arm it.

Tests updated to dispatch `pagehide` and to cover the new `visibilitychange` (hidden and visible) paths and the flush-on-exit behavior; a unit test covers `getPageEndEventName` across all branches. Package tests pass at 100% coverage and lint is clean.

### Checklist

<div><a href="https://cursor.com/agents/bc-f033818c-821d-46c4-8717-6c5edf7d29a7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f033818c-821d-46c4-8717-6c5edf7d29a7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/plugin-autocapture-browser/src/autocapture-plugin.ts
+++ b/packages/plugin-autocapture-browser/src/autocapture-plugin.ts
@@ -18,6 +18,7 @@ import { VERSION } from './version';
 import * as constants from './constants';
 import {
   createShouldTrackEvent,
+  getPageEndEventName,
   type ElementBasedTimestampedEvent,
   type TimestampedEvent,
   type NavigateEvent,
@@ -141,7 +142,7 @@ export const autocapturePlugin = (
 
   const dataExtractor = new DataExtractor(options, context);
 
-  // Page-level state shared across trackers, emitted in a single Page View End event on beforeunload
+  // Page-level state shared across trackers, emitted in a single Page View End event on page exit
   // elementExposedForPage holds the total set of elements seen during the entire page view lifetime
   const elementExposedForPage = new Set<string>();
   // currentElementExposed only holds the set of elements that will be flushed during the next [Amplitude] Viewport Content Updated event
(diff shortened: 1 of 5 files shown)


jxiwang:
bugbot run

PR #1948, opened 2026-08-20
fix(autocapture): exposure snapshot

jxiwang:
## Summary

Fixes `[Amplitude] Viewport Content Updated` under-reporting on SPAs and duplicate/partial events caused by hydration and history API behavior.

**Problem**
- On SSR/hydrated sites (e.g. clubmed.fr), only header/footer elements were exposed during a session; main content was missed until page leave.
- Many sites (Google, React SPAs) fire same-URL `pushState`/`replaceState` during hydration, which triggered spurious page-end flushes and a second snapshot cycle — producing two events where the second was a superset of the first.

**Solution**
1. **Initial exposure snapshot** — after DOM mutations settle, rescan allowlisted elements and flush one `Viewport Content Updated` event per page view.
2. **`DocumentFragment` support** — observe elements committed via React hydration fragments, not just direct `Element` nodes.
3. **Same-URL navigation guard** — only treat history changes as page end when the normalized URL actually changes.
4. **Sent-event deduplication** — elements already included in a flushed event are not re-sent on subsequent flushes within the same page view.

## How the initial snapshot works

---

> [!NOTE]
> **Medium Risk**
> Touches global `history.pushState`/`replaceState` proxies and changes when viewport events fire and what they contain, which can shift analytics semantics for SPAs and hydration-heavy sites.
> 
> **Overview**
> Fixes **`[Amplitude] Viewport Content Updated`** under-reporting on hydrated SPAs and noisy/duplicate flushes from hydration and history API churn.
> 
> Adds an **initial exposure snapshot** (`createInitialExposureSnapshotScheduler`) that waits for a quiet DOM window (or max wait), **rescans** allowlisted elements via a new exposure observable `rescan` hook, then flushes viewport content after exposure timers complete. The scheduler **resets** on real page end.
> 
> **SPA / history behavior** now compares **normalized page URLs** (query stripped, hash kept) so same-URL `pushState`/`replaceState`, query-only Navigation API updates, and unchanged `popstate` do not end the page view. Real navigations flush the **previous** page before applying history so **Page URL** stays correct; `replaceState` is proxied like `pushState`.
> 
> **Event payload fixes:** `[Amplitude] Element Exposed` sends only **new** paths since the last flush (`elementExposedInSentEvents`); page-end cleanup also resets `lastScroll` to avoid false scroll deltas. Exposure tracking clears **pending timers** on re-intersection after rescan; mutations observe **`DocumentFragment`** children for React hydration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c90daffa454aa77a3fca7fcf89e568b5cf4a9f2d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jxiwang:
bugbot run

packages/plugin-autocapture-browser/src/autocapture/schedule-initial-exposure-snapshot.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    quietTimer = setTimeout(takeSnapshot, quietMs);
+  };
+
+  const arm = () => {
+    clearTimers();
+    removeDomContentLoadedListener();
+    snapshotTaken = false;
+    scheduleQuietTimer();
+    maxWaitTimer = setTimeout(takeSnapshot, maxWaitMs);
```
jpollock-ampl: If `document.readyState === 'loading'` when `maxWaitTimer` fires, `takeSnapshot` bails into the `DOMContentLoaded` listener branch instead of snapshotting at `maxWaitMs`. That means on a slow-loading page, `EXPOSURE_SNAPSHOT_MAX_WAIT_MS` isn't really an upper bound, the snapshot just gets deferred until `DOMContentLoaded`, which could fire well after 4s.

Is that the intended tradeoff (favor correctness over the time bound), or should the max-wait timer force a snapshot regardless of `readyState`?

packages/plugin-autocapture-browser/src/autocapture-plugin.ts
```diff
@@ -398,13 +466,20 @@ export const autocapturePlugin = (
       const navigateObservable = allObservables[ObservablesEnum.NavigateObservable];
       if (navigateObservable) {
         subscriptions.push(
-          navigateObservable.subscribe(() => {
+          navigateObservable.subscribe((timestampedEvent) => {
+            const navigateEvent = timestampedEvent.event;
+            const nextPageUrl = getDecodeURI(navigateEvent.destination.url.split('?')[0]);
```
jpollock-ampl: Is there a reason the Navigation API path can't reuse `getNormalizedPageUrl` (`helpers.ts:288`) instead of duplicating its normalization logic here?

We now have three places computing "normalized current/next URL": `getNormalizedPageUrl`, `resolveHistoryNavigationUrl`, and this inline one. Worth consolidating so the normalization rules (e.g. if we ever need to also strip hash fragments) can't drift out of sync between them.

packages/plugin-autocapture-browser/src/constants.ts
```diff
@@ -59,3 +59,12 @@ export const MAX_ATTRIBUTE_LENGTH = 128;
...
 export const MAX_ELEMENT_EXPOSED_STR_LENGTH = 18_000;
+
+/** DOM must be quiet for this long before the initial exposure snapshot runs. */
+export const EXPOSURE_SNAPSHOT_QUIET_MS = 750;
+
+/** Upper bound to wait for DOM hydration before taking the initial exposure snapshot. */
+export const EXPOSURE_SNAPSHOT_MAX_WAIT_MS = 4_000;
+
+/** Extra buffer after rescan before flushing, so exposure timers can complete. */
+export const EXPOSURE_SNAPSHOT_FLUSH_BUFFER_MS = 50;
```
jpollock-ampl: Worth checking whether 50ms is enough buffer here. `IntersectionObserver` callbacks fire asynchronously (typically next paint), and the `exposureDuration` timer in `trackExposure` doesn't start until `entry.isIntersecting` fires. So the real gap between rescan and "all exposure timers have fired" is closer to one observer tick plus `exposureDuration`, not just `exposureDuration` + buffer.

On a slow device or a page with many allowlisted elements, that margin could run out before some elements' timers complete, causing the initial flush to under-report. Have we considered making this configurable, or adding a test that simulates a slow `IntersectionObserver` callback to confirm the flush waits long enough?

jxiwang:
bugbot run

packages/plugin-autocapture-browser/src/autocapture/schedule-initial-exposure-snapshot.ts
```diff
@@ -0,0 +1,131 @@
...
+    snapshotTaken = true;
+    clearTimers();
+    removeDomContentLoadedListener();
+
+    onRescan();
+
+    exposureWaitTimer = setTimeout(() => {
+      exposureWaitTimer = undefined;
+      onFlush();
+    }, exposureDuration + constants.EXPOSURE_SNAPSHOT_FLUSH_BUFFER_MS);
```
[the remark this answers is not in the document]
jxiwang: We can revisit if this is an issue.

packages/plugin-autocapture-browser/src/constants.ts
```diff
@@ -59,3 +59,12 @@ export const MAX_ATTRIBUTE_LENGTH = 128;
 export const PAGE_VIEW_SESSION_STORAGE_KEY = 'AMP_PAGE_VIEW';
 
 export const MAX_ELEMENT_EXPOSED_STR_LENGTH = 18_000;
+
+/** DOM must be quiet for this long before the initial exposure snapshot runs. Matches DEFAULT_EXPOSURE_DURATION. */
+export const EXPOSURE_SNAPSHOT_QUIET_MS = 150;
+
+/** Upper bound to wait for DOM hydration before taking the initial exposure snapshot. */
+export const EXPOSURE_SNAPSHOT_MAX_WAIT_MS = 4_000;
```
jxiwang, line 67: Just chose a value for this, do not think it would reach this.

daniel-graham-amplitude (approved):
Lgtm, everything looks good. But I'd want other code reviews on this too.

PR #1949, opened 2026-08-21
feat(session-replay-browser): expose start and stop recording APIs

jxiwang:
### Summary

Exposes `start()` and `stop()` on the Session Replay browser SDK so customers can pause and resume capture without tearing down the SDK.

- **Standalone** (`@amplitude/session-replay-browser`): `sessionReplay.stop()` / `sessionReplay.start()`
- **Plugin** (`@amplitude/plugin-session-replay-browser`): methods on the plugin instance
- **Unified**: available via `client.sessionReplay().start()` / `.stop()` because that client already returns `AmplitudeSessionReplay`

`stop()` cancels the live rrweb recorder (it does not keep observing the DOM), flushes already-captured events, and keeps session id, config, and page listeners in place. `start()` begins a new rrweb recording. Sampling, targeting, opt-out, and remote `captureEnabled` still apply. Focus and targeting paths will not auto-restart recording after `stop()`.

In-flight `_recordEvents()` that already passed `getShouldRecord()` is abandoned after `stop()` / `shutdown()`, so a paused session cannot resume from a stale `record()` call.

In coordinated cross-origin child iframes, `init()` still waits for a parent start signal, but a customer `start()` self-starts capture so resume does not hang if the parent is already recording.

This matches the start/stop API already present on the React Native Session Replay SDK. `shutdown()` remains the way to fully discontinue collection and remove listeners.

Coverage:
- Unit tests for start/stop, including that `stop()` invokes the cancel callback returned by rrweb `record()`, and that `stop()` during an in-flight `recordEvents()` does not start rrweb
- Child iframe `start()` after `stop()` self-starts without a parent signal
- Playwright e2e in `packages/session-replay-browser/e2e/start-stop.spec.ts` that mutates the DOM against a live recorder and asserts mutations after `stop()` never reach the track API

### Checklist

<div><a href="https://cursor.com/agents/bc-ab2755ff-823f-4ccf-bc67-d9cfbb8eddae?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ab2755ff-823f-4ccf-bc67-d9cfbb8eddae&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/session-replay-browser/src/session-replay.ts
+++ b/packages/session-replay-browser/src/session-replay.ts
@@ -128,6 +128,11 @@ export class SessionReplay implements AmplitudeSessionReplay {
    * passed `getShouldRecord()` cannot start rrweb after capture was paused.
    */
   private recordingGeneration = 0;
+  /**
+   * Set by customer `start()` so coordinated child iframes self-start instead of
+   * waiting for a parent signal that may never be resent.
+   */
+  private startChildRecordingOnSetup = false;
   private pendingEmitEvents: Array<{ event: eventWithTime; sessionId: string | number }> = [];
 
   /** Current page URL, kept in sync with SPA navigations for URL-based masking */
@@ -962,6 +967,8 @@ export class SessionReplay implements AmplitudeSessionReplay {
     const config = this.config;
     const shouldRecord = this.getShouldRecord();
     const sessionId = this.identifiers?.sessionId;
+    const selfStartChild = this.startChildRecordingOnSetup;
+    this.startChildRecordingOnSetup = false;
     if (!shouldRecord || !sessionId || !config) {
       return;
     }
@@ -1040,6 +1047,11 @@ export class SessionReplay implements AmplitudeSessionReplay {
             }
           },
         });
+        // Customer start() must not wait for a parent signal: the parent may already
+        // be recording and will not send another start. Init still waits.
+        if (selfStartChild) {
+          this._recordEventsInChildMode(recordFunction, sessionId, config, hooks);
+        }
         return;
       }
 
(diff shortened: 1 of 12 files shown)


packages/session-replay-browser/src/session-replay.ts
```diff
@@ -1207,6 +1225,21 @@ export class SessionReplay implements AmplitudeSessionReplay {
...
+  }
+
+  private async _start() {
+    this.recordingEnabled = true;
+    await this.recordEvents();
+  }
+
+  stop() {
+    this.recordingEnabled = false;
+    this.stopRecordingEvents();
```
jpollock-ampl: What's the thinking behind `stop()` not going through `recordEvents()`'s in-flight guard (`recordEventsInFlight`) at all? If a `_recordEvents()` call is mid-flight (say, awaiting `getRecordFunction()` or `initializeNetworkObservers()`) when `stop()` fires, `stop()`'s own `stopRecordingEvents()` call is a no-op since `recordCancelCallback` hasn't been set yet, and then the in-flight call finishes and sets it anyway.

This is really the same root issue as the Bugbot-flagged race, just from the `stop()` side rather than the `_recordEvents()` side - happy to fold this into whatever fix addresses that one.

bravecod (approved):

PR #1956, opened 2026-08-26
fix(browser): normalize empty instance names

Mercy811:
### Summary

- Add a shared Analytics Core normalizer so `undefined` and exactly `''` resolve to `$default_instance`.
- Make connector lookup defensive and retain the normalized value in `BrowserConfig`.
- Route Browser Analytics initialization and identity updates through the normalized connector.
- Let the Browser Experiment plugin inherit the Analytics instance name unless explicitly overridden.
- Add regression coverage for default, empty, custom, whitespace, and explicit-override cases.

### Why

Browser Analytics could bind identity and its event receiver to an empty-string connector while Experiment resolved the same input to `$default_instance`. The SDKs would initialize successfully but silently lose identity synchronization and Experiment exposure-event forwarding.

PR #1944 fixed the React Native occurrence that exposed this issue. This PR applies the corresponding Browser/Core fix tracked by SDKW-59.

### Safety and storage impact

- A full `packages/analytics-browser` search found that the configured instance name only routes in-memory connector activity.
- The similarly named snippet `createInstance(instanceName)` registry is separate and is not changed.
- The Analytics Connector stores identity, listeners, its receiver, and its bounded event queue only in memory under `globalThis.analyticsConnectorInstances`.
- Browser identity/session, unsent-event, and attribution storage keys are based on the API key, not `instanceName`; no persistent storage migration is involved.
- Non-empty custom names remain isolated and unchanged. Whitespace-only names are intentionally not trimmed.

### Validation

- Analytics Core: 884 tests passed, 100% coverage.
- Browser Analytics: 509 tests passed, 100% coverage.
- Browser Experiment plugin: 12 tests passed, 100% coverage.
- Typecheck and lint passed for all three affected packages.

Ticket: https://linear.app/amplitude/issue/SDKW-59/normalize-empty-instancename-across-browser-analytics-and-experiment

### DIFF
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -304,7 +304,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
     // Set up the analytics connector to integrate with the experiment SDK.
     // Send events from the experiment SDK and forward identifies to the
     // identity store.
-    const connector = getAnalyticsConnector(options.instanceName);
+    const connector = getAnalyticsConnector(this.config.instanceName);
     connector.identityStore.setIdentity({
       userId: this.config.userId,
       deviceId: this.config.deviceId,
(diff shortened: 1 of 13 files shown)


Mercy811:
@codex review

daniel-graham-amplitude (approved):

PR #1961, opened 2026-08-27
fix(session-replay-react-native): make setDeviceId non-nullable

aliaksandr-kazarez:
### Summary

Make the public `setDeviceId` API on `@amplitude/session-replay-react-native` a non-nullable `string` (including the TurboModule spec and the JS typed facade). This matches Android and Flutter public APIs.

SDKRN-59 found that `setDeviceId(null)` was documented as clearing the device ID, but that clear was never implemented: native persists `""` (Android coerces with `?: ""`). This PR removes the documented-null contract instead of adding iOS nil-clearing. Config `deviceId?: string | null` is unchanged (init still may send `null`).

The Segment plugin only calls `setDeviceId` when the event has a device ID or anonymous ID, because it types through the standalone setter.

Native E2E was not re-run for this diff.

### Test plan

- [x] `pnpm --filter @amplitude/session-replay-react-native run test`
- [x] `pnpm --filter @amplitude/session-replay-react-native run typecheck`
- [x] `pnpm --filter @amplitude/session-replay-react-native run lint`
- [x] `pnpm --filter @amplitude/segment-session-replay-plugin-react-native exec jest test/segment-session-replay-plugin.test.ts`
- [x] `pnpm --filter @amplitude/segment-session-replay-plugin-react-native run typecheck`
- [ ] Confirm callers no longer pass `null` into `setDeviceId`

### Checklist

* Does your PR have a breaking change?: Yes — `setDeviceId` no longer accepts `null`

---
*Posted automatically with Cursor. LLMs make mistakes — please verify before acting.*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ yarn-error.log*
 *.p12
 *.pfx
 secrets.*
+sdk59.local.ts
 
 lerna-debug.log
 
--- a/packages/segment-session-replay-plugin-react-native/src/segment-session-replay-plugin.ts
+++ b/packages/segment-session-replay-plugin-react-native/src/segment-session-replay-plugin.ts
@@ -83,7 +83,9 @@ export class SegmentSessionReplayPlugin extends Plugin {
     const deviceId = getDeviceId(event);
 
     await setSessionId(sessionId);
-    await setDeviceId(deviceId);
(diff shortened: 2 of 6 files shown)


aliaksandr-kazarez:
nullability is something we want to support, and having empty string on Android considered a better move!

PR #1965, opened 2026-09-02
feat(plugin-web-vitals-browser): add reportSoftNav option for soft navigation reporting

daniel-graham-amplitude:
### Summary

Adds a `reportSoftNav` option to a new `WebVitalsOptions` type, and reports web vitals per soft navigation when it is enabled.

`autocapture.webVitals` now accepts an options object in addition to a boolean:

```ts
amplitude.init(API_KEY, {
  autocapture: {
    webVitals: { reportSoftNav: true },
  },
});
```

and the plugin takes the same options directly: `webVitalsPlugin({ reportSoftNav: true })`.

**Default behavior is unchanged.** `reportSoftNav` defaults to `false`, and `webVitals: true` keeps producing exactly the same single `[Amplitude] Web Vitals` event with the same properties as before.

#### What changes when it's on

The plugin passes `reportSoftNavs: true` to each of `onLCP`/`onFCP`/`onINP`/`onCLS`/`onTTFB`, so web-vitals measures LCP, FCP, INP, CLS and TTFB per [soft navigation](https://github.com/WICG/soft-navigations) instead of only for the initial page load. The plugin then:

- Buckets metrics by `metric.navigationId` and sends **one `[Amplitude] Web Vitals` event per navigation**.
- Takes the page properties from `metric.navigationURL` rather than the current URL — a navigation's metrics can be reported after the next navigation has already begun, so the current URL is the wrong one.
- Reports `navigationStart` relative to the navigation being measured (`performance.timeOrigin + metric.navigationStartTime`) and includes `navigationId` on each metric. Metrics for a soft navigation carry `navigationType: 'soft-navigation'`, so they're distinguishable downstream.
- Sends a navigation's event once a newer navigation starts reporting metrics, deferred by 1s (`SOFT_NAV_FLUSH_DELAY_MS`) so metrics finalized just after the navigation boundary still land in the right event. The current navigation is still sent on `visibilitychange` → `hidden`.
- Keeps the `visibilitychange` listener installed (today it removes itself after firing once), since the page can be hidden and shown again with more navigations to report. Events with no metrics are not sent.

In browsers without the Soft Navigations API, every metric shares the hard navigation's `navigationId`, so behavior falls back to a single event — same as today.

### Dependency bump

`web-vitals` 5.1.0 → 6.2.1. Soft navigation support (`ReportOpts.reportSoftNavs`, `Metric.navigationId` / `navigationURL` / `navigationStartTime`) landed in v6; there is no way to do this on 5.x. Same package layout and same call signatures for the five `onX` functions we use.

One wrinkle worth a look: web-vitals@6's Soft Navigations types reference DOM types that don't exist in the TypeScript 4.9 DOM lib (`NavigationType`), so `skipLibCheck` is set for this package (there's precedent in the react-native packages). The alternative is a repo-wide TypeScript upgrade, which felt out of scope here.

### Open questions for review

- **Flush heuristic.** The 1s grace period before sending a superseded navigation is a heuristic. Alternatives: send immediately, or hold everything until `visibilitychange`. Happy to change it.
- **`navigationId` on the metric payload.** Currently only added when `reportSoftNav` is on, to keep the default event shape byte-identical. Worth always including?
- **Remote config.** `AutocaptureOptionsRemoteConfig extends AutocaptureOptions`, so the object form flows through remote config without extra work. No transform was added — let me know if this should be gated differently server-side.

### Testing

- `plugin-web-vitals-browser`: 20 tests, 100% statements/branches/functions/lines. Covers the unchanged default path, opting in per metric, per-navigation events, late-arriving metrics for a superseded navigation, several soft navigations in quick succession, repeated hides, teardown with a pending flush, and unparseable navigation URLs.
- `analytics-browser`: added coverage for `isWebVitalsEnabled` with the object form and for the new `getWebVitalsConfig`, plus a `browser-client` test asserting the options reach the plugin. Full suite passes at 100% coverage.
- `analytics-core`: full suite passes at 100% coverage.
- Manual: added `test-server/browser-sdk/web-vitals-soft-nav.html`, which pushes history entries and paints new content so soft navigations are actually detected. Needs Chromium 151+ to exercise the new path — I have not yet run it against a browser with Soft Navigations support, which is part of why this is a draft.

### Checklist

* Does your PR have a breaking change?: No — `autocapture.webVitals` widens from `boolean` to `boolean | WebVitalsOptions`, and the default event is unchanged. The `web-vitals` major bump is an internal dependency of the plugin.

---

> [!NOTE]
> **Medium Risk**
> Default web vitals behavior is unchanged, but the plugin’s event timing/shape changes when `reportSoftNav` is enabled and it depends on a major `web-vitals` bump.
> 
> **Overview**
> Adds **`WebVitalsOptions`** (starting with **`reportSoftNav`**) and widens **`autocapture.webVitals`** to `boolean | WebVitalsOptions`. The browser SDK forwards object config via **`getWebVitalsConfig`** into **`webVitalsPlugin(options)`**; **`webVitals: true`** still passes no options and behavior stays the same.
> 
> With **`reportSoftNav: true`**, the plugin upgrades **`web-vitals`** to v6, opts into soft-navigation measurement, buckets metrics by navigation, attaches page properties from **`navigationURL`**, and emits **one `[Amplitude] Web Vitals` event per navigation** (superseded navigations flush after a 1s delay; the active navigation still flushes on hide). The visibility listener stays registered for repeated hide/show cycles when soft nav is on.
> 
> Docs, unit/integration tests, and a manual **`web-vitals-soft-nav.html`** test page are included; the plugin enables **`skipLibCheck`** for v6 DOM types on TS 4.9.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a239830e68a92d9eb689da169a6d7a9ecb7b5f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -2,9 +2,10 @@ const limits = [
   {
     // analytics-browser bundle
     // Bumped 65kb → 66kb for shadow DOM support in plugin-autocapture-browser
-    // (SR-4788). Current actual: ~65.0kb gzipped.
+    // (SR-4788), then 66kb → 68kb for soft navigation support in
+    // plugin-web-vitals-browser (web-vitals v6). Current actual: ~65.7kb gzipped.
     path: './packages/analytics-browser/lib/scripts/amplitude-min.js.gz',
-    limit: '66kb',
+    limit: '68kb',
     brotli: false,
   },
   {
@@ -15,8 +16,10 @@ const limits = [
   },
   {
     // unified SDK bundle
+    // Bumped 225kb → 228kb for soft navigation support in
+    // plugin-web-vitals-browser (web-vitals v6). Current actual: ~220.1kb gzipped.
     path: './packages/unified/lib/scripts/amplitude-min.umd.js.gz',
-    limit: '225kb',
+    limit: '228kb',
     brotli: false,
   },
   {
--- a/packages/analytics-browser/src/browser-client.ts
+++ b/packages/analytics-browser/src/browser-client.ts
@@ -46,6 +46,7 @@ import {
   isPageViewTrackingEnabled,
   isNetworkTrackingEnabled,
   isWebVitalsEnabled,
+  getWebVitalsConfig,
   isFrustrationInteractionsEnabled,
   getFrustrationInteractionsConfig,
   isPerformanceTrackingEnabled,
@@ -374,7 +375,7 @@ export class AmplitudeBrowser extends AmplitudeCore implements BrowserClient, An
 
     if (isWebVitalsEnabled(this.config.autocapture)) {
       this.config.loggerProvider.debug('Adding web vitals plugin');
-      await this.add(webVitalsPlugin()).promise;
+      await this.add(webVitalsPlugin(getWebVitalsConfig(this.config))).promise;
     }
 
     if (isPerformanceTrackingEnabled(this.config.autocapture)) {
--- a/packages/analytics-browser/src/default-tracking.ts
+++ b/packages/analytics-browser/src/default-tracking.ts
@@ -11,6 +11,7 @@ import {
   FrustrationInteractionsOptions,
   CustomEnrichmentOptions,
   PerformanceTrackingOptions,
+  WebVitalsOptions,
   isChromeExtension,
   normalizeNetworkCaptureRules,
 } from '@amplitude/analytics-core';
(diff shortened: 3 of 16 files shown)


daniel-graham-amplitude:
bugbot run

Mercy811:
@codex review

Mercy811 (commented):
overall lgtm, but the two codex comments make sense to me. @daniel-graham-amplitude could you fix them? Thanks!

daniel-graham-amplitude:
Fixes made thanks @Mercy811

Mercy811:
@codex review

Mercy811 (approved):

daniel-graham-amplitude:
@codex review

daniel-graham-amplitude:
@codex review

PR #1972, opened 2026-09-04
fix(autocapture): expose zones when the mid-height line is viewed

jxiwang:
### Summary

A zone is exposed when its **mid-height pixel line has been displayed for at least 150ms**.

This is vertical depth scrolled through, not the share of the element's area on screen at one moment. The difference matters for zones taller than the viewport: less than half their area can ever be visible, but their midpoint is still scrolled through.

### Implementation

- The shared `IntersectionObserver` uses `threshold: 0` purely to keep a small candidate set of elements touching the viewport.
- The shared scroll observable listens in the capture phase, so non-bubbling scroll events from overflow containers are included alongside window scrolling.
- `trackExposure` evaluates candidates on intersection changes and scroll via `getBoundingClientRect()`.
- Midpoint visibility is clipped against the viewport and each overflow-clipping visual ancestor, including assigned slots and open shadow-root hosts. A midpoint hidden outside a scroll panel does not count.
- A 150ms timer starts when the element's mid-height line is visible, and is cancelled if that line leaves first, so scrolling straight past a zone does not expose it.
- Repeated evaluations do not restart a pending timer, so slow scrolling across the midpoint still satisfies the dwell.
- `reset()` clears only page-view state (reported zones and in-flight dwells) and re-evaluates what is still on screen. It deliberately keeps the intersecting candidate set: `IntersectionObserver` never re-notifies an element that never stopped intersecting, so clearing it silently stopped scroll-driven checks for everything currently visible after any page-view end or SPA navigation. Teardown clears everything.
- The line check allows 1px of tolerance. Engines round fractional scroll offsets differently: after an identical scroll to a zone's exact midpoint, Chromium put the line 0.41px inside the viewport and WebKit 0.59px outside. A sub-pixel rounding artifact should not decide the metric.

### Tests

Unit tests cover sub-pixel boundaries, clipping ancestors, slots, open shadow roots, capture of non-bubbling element scroll events, and the reset regression (both that scroll keeps working afterwards and that a partly elapsed dwell does not carry into the next page view).

Real browser (Chromium + WebKit) coverage in the plugin e2e directory includes 49% depth (not exposed), exactly the mid-height line (exposed), 60% and full depth (exposed), crossing the midpoint then leaving before the dwell (not exposed), an oversized zone (exposed once its midpoint is viewed, with under half its area on screen), a nested overflow scroller whose clipped midpoint must not expose until the panel is scrolled, a sticky site header that is reported on the first event and stays in view after scroll, and the emitted payload shape.

### Manual verification

From the repository root:

```bash
pnpm install
pnpm build
npx vite dev
```

Open http://localhost:5173/autocapture/viewport-exposure.html and simply scroll. Each zone shows a dashed 50% line, turns amber while that line is in view, and green once the SDK reports it. The top bar is a sticky site header that stays in the viewport, so it should appear in an exposed event on load. The HUD lists every zone with its max depth viewed, share on screen, and state, and the summary names any zone it is still waiting on together with the reason — clipped by a scroll panel, passed without dwelling, or not reached yet. `?exposureDuration=` stretches the dwell if you want to watch it happen. The purple panel covers the nested-scroller case, and the jump buttons cover the precise 30 / 49 / 50 / 60 / 100% cases.

The right-hand panel shows the actual event objects the SDK sends, newest first, read off an enrichment plugin so it is exactly what gets uploaded — `[Amplitude] Element Exposed` alongside the page URL, viewport size, and max scroll that ship with it. Flushes that only advance scroll depth appear with an empty `Element Exposed` array, which makes the batching visible:

Harness with the event payload panel showing an Element Exposed array and the rest of the event properties

Partway down a normal scroll, zones expose one after another as each midpoint is passed:

Harness partway down a scroll: four zones exposed, later zones still pending

The 250vh zone exposes with only 40% of it on screen, which an area-based threshold could never do. Note also `#zone-5` and `#exposure-target`, which a jump scrolled straight past: their line was reached but never dwelled on, so they are correctly unexposed.

Oversized zone exposed at its midpoint with 40% on screen; skipped zones remain unexposed

To run the automated spec while that server is up:

```bash
npx playwright test packages/plugin-autocapture-browser/e2e/viewport-exposure.spec.ts
```

### Checklist

* Does your PR have a breaking change?: No, but exposure semantics change intentionally — from requiring the whole element visible to midpoint viewing, so more elements (including ones taller than the viewport) will now be reported.

<div><a href="https://cursor.com/agents/bc-7c4ec441-e00f-42a2-ae74-96cf241790f3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7c4ec441-e00f-42a2-ae74-96cf241790f3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
+++ b/packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
@@ -5,31 +5,14 @@ import { EXPOSURE_INTERSECTION_THRESHOLD } from '../constants';
 import { DataExtractor } from '../data-extractor';
 
 // Sub-pixel rounding can report a ratio a hair below the threshold on the very callback that
-// crossed it, which would strand an element that stops scrolling right at the boundary.
+// crossed it, which would otherwise strand an element that stops scrolling right at the boundary.
 const INTERSECTION_RATIO_TOLERANCE = 0.001;
 const MIN_INTERSECTION_RATIO = EXPOSURE_INTERSECTION_THRESHOLD - INTERSECTION_RATIO_TOLERANCE;
 
-// An element larger than the viewport can never reach EXPOSURE_INTERSECTION_THRESHOLD of its own
-// area, so it instead counts as seen once its visible part fills that fraction of the viewport.
-const fillsViewport = (entry: IntersectionObserverEntry): boolean => {
-  const rootBounds = entry.rootBounds;
-  if (!rootBounds) {
-    return false;
-  }
-  const rootArea = rootBounds.width * rootBounds.height;
-  if (rootArea <= 0) {
-    return false;
-  }
-  const visibleArea = entry.intersectionRect.width * entry.intersectionRect.height;
-  return visibleArea / rootArea >= MIN_INTERSECTION_RATIO;
-};
-
-const isSeen = (entry: IntersectionObserverEntry): boolean => {
-  if (!entry.isIntersecting) {
-    return false;
-  }
-  return entry.intersectionRatio >= MIN_INTERSECTION_RATIO || fillsViewport(entry);
-};
+// `isIntersecting` alone is not enough: the spec defines it as any overlap with the root, so the
+// ratio is what actually holds the element to EXPOSURE_INTERSECTION_THRESHOLD.
+const isSeen = (entry: IntersectionObserverEntry): boolean =>
+  entry.isIntersecting && entry.intersectionRatio >= MIN_INTERSECTION_RATIO;
 
 export function trackExposure({
   allObservables,
--- a/packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
+++ b/packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
@@ -1,7 +1,35 @@
 /* eslint-disable no-restricted-globals */
 import { DEFAULT_EXPOSURE_DURATION } from '@amplitude/analytics-core';
 import { AllWindowObservables } from '../autocapture-plugin';
+import { EXPOSURE_INTERSECTION_THRESHOLD } from '../constants';
 import { DataExtractor } from '../data-extractor';
+
+// Sub-pixel rounding can report a ratio a hair below the threshold on the very callback that
+// crossed it, which would strand an element that stops scrolling right at the boundary.
+const INTERSECTION_RATIO_TOLERANCE = 0.001;
+const MIN_INTERSECTION_RATIO = EXPOSURE_INTERSECTION_THRESHOLD - INTERSECTION_RATIO_TOLERANCE;
+
+// An element larger than the viewport can never reach EXPOSURE_INTERSECTION_THRESHOLD of its own
+// area, so it instead counts as seen once its visible part fills that fraction of the viewport.
+const fillsViewport = (entry: IntersectionObserverEntry): boolean => {
+  const rootBounds = entry.rootBounds;
+  if (!rootBounds) {
+    return false;
+  }
+  const rootArea = rootBounds.width * rootBounds.height;
+  if (rootArea <= 0) {
+    return false;
+  }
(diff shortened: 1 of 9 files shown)


packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
```diff
@@ -1,8 +1,19 @@
...
 
+// Sub-pixel rounding can report a ratio a hair below the threshold on the very callback that
+// crossed it, which would otherwise strand an element that stops scrolling right at the boundary.
+const INTERSECTION_RATIO_TOLERANCE = 0.001;
+const MIN_INTERSECTION_RATIO = EXPOSURE_INTERSECTION_THRESHOLD - INTERSECTION_RATIO_TOLERANCE;
+
+// `isIntersecting` alone is not enough: the spec defines it as any overlap with the root, so the
+// ratio is what actually holds the element to EXPOSURE_INTERSECTION_THRESHOLD.
+const isSeen = (entry: IntersectionObserverEntry): boolean =>
+  entry.isIntersecting && entry.intersectionRatio >= MIN_INTERSECTION_RATIO;
```
jpollock-ampl: Since the observer is constructed with a single `threshold: EXPOSURE_INTERSECTION_THRESHOLD` value (no array), I believe the callback only fires on crossing 0.5 in either direction, plus the initial observe. If that's right, `INTERSECTION_RATIO_TOLERANCE` is mainly compensating for floating-point rounding at that one crossing point, not a general "hovering near threshold" scenario. Might be worth a one-line comment saying so, since a future reader could otherwise assume the observer fires continuously near the threshold.

packages/plugin-autocapture-browser/test/autocapture-plugin/track-exposure.test.ts
```diff
@@ -208,7 +208,7 @@ describe('trackExposure', () => {
     triggerExposure({
       isIntersecting: false,
       target: element,
-      intersectionRatio: 0.5, // < 1.0
+      intersectionRatio: 0,
```
jpollock-ampl: Could we add a test case at the boundary itself, e.g. ratio ~0.4995 (inside the tolerance band, should still expose)? The new tests cover 0.4 vs 0.6, which confirms the half-visible cutoff generally, but nothing exercises `INTERSECTION_RATIO_TOLERANCE` directly, so a regression there wouldn't be caught.

daniel-graham-amplitude (approved):

jxiwang:
bugbot run

packages/plugin-autocapture-browser/src/autocapture/track-exposure.ts
```diff
@@ -20,55 +101,85 @@ export function trackExposure({
...
+      cancelExposure(element);
+    }
+  });
 
-          // Clear the timer reference
-          exposureTimerMap.set(element, null);
-        }, exposureDuration);
+  const scrollSubscription = scrollObservable.subscribe(() => {
+    intersectingElements.forEach(updateExposure);
+  });
```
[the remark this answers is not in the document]
jxiwang: Will follow up with this since it's a very narrow case with performance implications that depends on the user never interacting again with the page.

PR #1991, opened 2026-09-15
ci(release): stream lerna version output

Mercy811:
## Summary
- stream lerna version output live during stable releases
- preserve the underlying command exit code with PIPESTATUS
- retain the no-changed-packages guard using the captured log file

This makes runner shutdowns and canceled release attempts diagnosable. In https://github.com/amplitude/Amplitude-TypeScript/actions/runs/35007246638/job/104550910088, command substitution hid all lerna progress before the process received SIGTERM (exit 143).

## Validation
- focused shell test for streamed output and exit-code preservation
- YAML parse
- git diff --check
- pnpm docs:check
- pnpm lint
- pnpm lint:deps

Local environment limitations:
- pnpm build reached the final Rollup tasks but one existing package process did not exit
- pnpm test encountered Watchman socket permission failures
- pnpm test:examples reports the existing missing jest.setup.examples.js file

---

> [!NOTE]
> **Low Risk**
> Release workflow shell-only change; behavior for success, failure, and empty version sets stays the same with better observability.
> 
> **Overview**
> The **Create release version** step in `publish-v2.yml` no longer buffers `pnpm deploy:version` output in command substitution. It now pipes stdout/stderr through `tee` into a temp log so Lerna progress shows live in the Actions log when a run is canceled or the runner shuts down.
> 
> Exit handling is unchanged in spirit: failures still fail the job, and the **no changed packages** guard still runs via `grep` on that log file. Exit code comes from **`PIPESTATUS[0]`** so a successful `tee` does not mask a failing Lerna command. The redundant dump of captured output after the command was removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02cf199c5a042684c9242b127ee50f1548e1fbea. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Mercy811:
bugbot run

jxiwang (approved):

daniel-graham-amplitude (approved):

PR #1994, opened 2026-09-16
fix(autocapture): reliably apply selector config per SDK instance

jxiwang:
## Summary

Fix two causes of intermittent element-selector activation:

1. Remote-config subscribers shared one mutable response object. Browser config normalization removed `elementSelector.enabled` before later subscribers and cache persistence could read it.
2. Every browser SDK instance on a page shared one module-level selector engine and shadow latch. Config from one API key therefore changed selector behavior for every other SDK instance.

The first issue made activation depend on fetch timing. The second made multiple SDK instances last-delivery-wins for `enabled`, while `shadowDomEnabled` became first-enable-wins because its global gate latches on.

## Design

`BrowserConfig` is now the selector-runtime ownership boundary:

- Each `DataExtractor` starts with a private dormant runtime, preserving standalone usage before plugin setup.
- At the beginning of plugin `setup(config)`, the extractor obtains the runtime for that exact `BrowserConfig` from a `WeakMap`.
- Autocapture and frustration plugins installed by one `BrowserClient` receive the same config object, so they share an engine and shadow gate.
- Separate browser SDK instances receive different config objects, even if they reuse an API key or `instanceName`, so their selector state is isolated.
- Weak keys allow runtime state to be collected with the SDK config; there is no string-keyed global registry or explicit lifecycle cleanup.
- Both engine state and shadow-gate state live in the runtime. Isolating only the engine would leave cross-instance shadow piercing behind.

```mermaid
flowchart TB
  RC_A[Remote config client A] --> SUB_A1[Autocapture subscription A]
  RC_A --> SUB_A2[Frustration subscription A]
  CFG_A[BrowserConfig A] --> MAP[WeakMap&lt;BrowserConfig, SelectorRuntime&gt;]
  SUB_A1 --> RT_A[SelectorRuntime A]
  SUB_A2 --> RT_A
  RT_A --> ENG_A[SelectorEngine A]
  RT_A --> GATE_A[ShadowGate A]
  ENG_A --> EVENTS_A[Events sent by SDK A]
  GATE_A --> EVENTS_A

  RC_B[Remote config client B] --> SUB_B1[Autocapture subscription B]
  RC_B --> SUB_B2[Frustration subscription B]
  CFG_B[BrowserConfig B] --> MAP
  SUB_B1 --> RT_B[SelectorRuntime B]
  SUB_B2 --> RT_B
  RT_B --> ENG_B[SelectorEngine B]
  RT_B --> GATE_B[ShadowGate B]
  ENG_B --> EVENTS_B[Events sent by SDK B]
  GATE_B --> EVENTS_B

  RT_A -. isolated .- RT_B
```

For `A: enabled=true, shadowDomEnabled=true` and `B: enabled=false, shadowDomEnabled=false`, A emits v1/piercing selectors and B emits legacy/light-DOM selectors. Updating either project cannot change the other's runtime.

## Remote-config delivery fix

`RemoteConfigClient.sendCallback` previously handed every subscriber a live reference into the fetched object. The browser SDK subscribes to `configs.analyticsSDK.browserSDK` and runs `translateRemoteConfigToLocal`, which normalizes in place by deleting `enabled` from every `{ enabled: true, ... }` node—including `autocapture.elementSelector`.

Consequences:

- A later `…autocapture.elementSelector` subscriber could receive a payload without `enabled` and ignore it.
- `subscribeAll` persisted the already-mutated object, so later page loads could not arm from cache.
- A failed fetch persisted `remoteConfig: null`, erasing the last good cache.

The client now gives each subscriber its own copy and skips persistence when a fetch produced no config.

## Live verification

Verified against `https://pgatour-uat.pgatour.com/leaderboard`. Both API keys initialized by the page currently return `autocapture.elementSelector.enabled: true`. Event uploads were intercepted and stubbed so no probe events reached either project.

A probe `<button id="amp-probe-12345">` distinguishes the algorithms: legacy `cssPath` keeps the id, while v1 suppresses it because the project's configured autogenerated-id patterns include `\d{4,}`.

| Scenario | Stock 2.45.8 | Build with delivery fix |
| --- | --- | --- |
| Healthy load | engine selector | engine selector |
| Cached payload | `enabled` stripped | `enabled: true` preserved |
| Reload while config endpoint fails | legacy for whole page | engine, armed from cache |
| Cache after failure | overwritten with `remoteConfig: null` | last good config preserved |

## Validating this in the repo

`test-server/element-selector-multi-instance.html` reproduces the reported setup on the dev server and checks itself, so the behavior can be validated by hand with `pnpm dev` (open the page from the index listing) or in CI through `packages/plugin-autocapture-browser/e2e/element-selector-multi-instance.spec.ts`.

The page initializes two SDK instances with different API keys against a new mock remote-config endpoint (`/api/remote-config/<apiKey>`, added to the dev server's mock API, which derives `elementSelector.enabled` from the key). One click on a probe button with an autogenerated-looking id (`amp-probe-12345`) is captured by both instances, and each instance's `[Amplitude] Element Path` is read from its own event pipeline through an enrichment plugin. Four checks must hold:

| Check | Fails without |
| --- | --- |
| Instance A (`enabled: true`) drops the autogenerated id | delivery isolation, then runtime scoping |
| Instance B (`enabled: false`) keeps the legacy `button#amp-probe-12345` | — |
| Instance A caches `{"enabled":true}` | delivery isolation |
| Instance B caches `{"enabled":false}` | delivery isolation |

Both fixes are load-bearing, confirmed by reverting each one against this page:

- Without the delivery clone, A never arms (its cached and delivered payloads are the normalized booleans `true` / `false`), so both instances emit the legacy selector and both cache checks fail.
- With the delivery clone but a shared runtime, A arms and B's `enabled: false` immediately disables the shared engine, so both instances again emit the legacy selector.

Element selector multi-instance validation page reporting pass

## Test-page fallout from runtime scoping

The shadow-DOM test pages (`shadow-dom-test.html`, `shadow-dom-perf.html`, `shadow-dom-stress-test.html`) armed shadow mode by constructing a throwaway `DataExtractor` before `init`, which only worked because the engine and gate were module-level singletons. That made the shadow-DOM e2e specs fail on the previous revision of this PR. They now reach their own instance's runtime through plugin setup (`test-server/helpers/selector-config.js`), which is the same per-config runtime remote config arms in production.

## Tests

Remote-config regression coverage verifies:

- subscriber mutation cannot reach another subscriber or the source object
- subscriber mutation is not persisted
- failed fetches retain the prior cache
- browser normalization cannot strip `enabled` from the nested selector subscriber
- unclonable payloads preserve existing best-effort delivery behavior

Selector-runtime coverage verifies:

- extractors bound to one SDK config share selector-engine state
- extractors bound to different SDK configs remain isolated
- same-config extractors share one shadow latch
- different-config extractors have isolated shadow latches

Validation for this revision:

- full Playwright suite against the built test server (`pnpm build:vite` + `pnpm start`): 64 tests passing on chromium and webkit, including the new multi-instance spec and the repaired shadow-DOM specs
- same specs re-run against `pnpm dev`
- `pnpm lint:deps`

Validation completed before this revision:

- `pnpm --filter @amplitude/plugin-autocapture-browser test` — 536 tests, 100% coverage
- `pnpm --filter @amplitude/plugin-autocapture-browser build`
- `pnpm --filter @amplitude/plugin-autocapture-browser lint`
- Earlier delivery-fix revision: `pnpm build`, `pnpm test`, `pnpm lint`, `pnpm docs:check`, `pnpm lint:deps`, and `npx size-limit`

`pnpm test:examples` is currently broken on `main`: `jest.setup.examples.js` was removed when example tests were consolidated into `pnpm test`.

## Scope note

Visual-tagging messenger ownership is unchanged. The page opens one tagging overlay and the existing page-level idempotency guard still prevents duplicate handlers. This change only scopes selector generation and shadow mode for SDK event pipelines.

<div><a href="https://cursor.com/agents/bc-bd84f9ac-b396-45c9-9dd3-10ac8c59ec9e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bd84f9ac-b396-45c9-9dd3-10ac8c59ec9e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Mercy811:
@codex review

packages/analytics-core/src/remote-config/remote-config.ts
```diff
@@ -162,6 +162,32 @@ export interface IRemoteConfigClient {
...
+ */
+function cloneConfig(config: RemoteConfig | null): RemoteConfig | null {
+  if (config === null) {
+    return null;
+  }
+  try {
+    return JSON.parse(JSON.stringify(config)) as RemoteConfig;
+  } catch {
+    return config;
+  }
```
Mercy811, line 188: Can we add a log here? 
```suggestion
  } catch {
    this.logger.warn('Remote config client failed to clone config, fallback to returning the original object');
    return config;
  }
```

packages/analytics-core/src/remote-config/remote-config.ts
```diff
@@ -240,12 +266,30 @@ export class RemoteConfigClient implements IRemoteConfigClient {
...
+   * good copy, so the next page load has no cache to fall back on while the
+   * remote fetch is in flight — and stays on default behavior for the whole page
+   * view if that fetch fails too. Keeping the previous copy makes a failed fetch
+   * a no-op instead.
+   */
+  persistConfig(remoteConfigInfo: RemoteConfigInfo) {
+    if (remoteConfigInfo.remoteConfig === null) {
+      this.logger.debug('Remote config client skipping storage update: Fetch returned no config');
+      return;
+    }
```
Mercy811, line 289: Does this also include no remote config? What will happen if I first set some remote configs and then disable all of them? Will this change fail to invalid remote config because null is no-op now?

Mercy811 (changes requested):

PR #1997, opened 2026-09-17
feat(session-replay): share DOM privacy policy across recorders

crleona:
### Summary

Extract the browser Session Replay DOM privacy implementation into `@amplitude/session-replay-dom-privacy`, and make browser replay its first consumer. Also recognize `data-amp-block`, `data-amp-mask`, and `data-amp-unmask` alongside the existing classes, without a configuration flag.

This is a **draft for SR team architecture and behavior approval**. The intended next consumer is the JavaScript recorder injected into iOS WKWebViews. This PR changes the TypeScript side only; it does not yet integrate or release the package on mobile.

### Problem and intended outcome

Browser replay and the iOS WebView recorder currently have separate copies of DOM masking policy. The copies have diverged: the browser implementation includes URL-specific mask levels and attribute-value masking, while the iOS script recognizes `data-amp-*` annotations. In addition, iOS does not currently pass the browser remote privacy policy into its injected recorder.

The goal is to let the same website receive the same DOM privacy treatment in browser replay and an embedded WebView, using one maintained policy implementation. These settings control replay capture; they do not block website navigation or introduce a URL/domain capture denylist.

### Architecture

```text
Browser local + remote config             Future iOS remote config
             |                                     |
Browser-owned config resolution          Native-to-JavaScript bridge
             |                                     |
             +---- shared DOM privacy package -----+
                                  |
                     rrweb privacy recorder options
                                  |
                       Host-owned rrweb recorder
                                  |
                       Host-owned event delivery
```

The shared package owns:

- `PrivacyConfig`, `MaskLevel`, and defaults.
- DOM masking/unmasking precedence, input classification, whitespace-preserving replacement, and configured HTML attribute-value masking.
- URL glob matching and effective mask-level evaluation.
- CSS-selector validation and a strict parser for remote-only consumers.
- Construction of privacy-related rrweb options through `createPrivacyRecorderOptions`.

The host continues to own remote fetching, local/remote config resolution, navigation observation, recorder start/stop and snapshot lifecycle, rrweb version selection, event bridging, storage, and uploads. This is a utility package with an rrweb option adapter, **not a wrapper around the complete recorder or browser SDK**. Its only runtime dependency is `tslib`; it publishes ESM/CommonJS and type declarations and declares `sideEffects: false`.

Browser now uses the shared implementation for its recorder options. Existing internal helper/type paths re-export the moved definitions where needed. Browser local/remote merging remains in browser code. A future mobile adapter can consume a remote policy directly without importing that merger or the browser lifecycle.

### Shared schema

```ts
type PrivacyConfig = {
  defaultMaskLevel?: 'light' | 'medium' | 'conservative';
  blockSelector?: string | string[];
  maskSelector?: string[];
  unmaskSelector?: string[];
  maskAttributes?: string[];
  urlMaskLevels?: Array<{
    match: string;
    maskLevel: 'light' | 'medium' | 'conservative';
  }>;
};
```

The extraction preserves browser policy semantics: medium is the default; explicit masking beats unmasking; URL rules are first-match-wins; and configured attribute names are processed through masking policy rather than unconditionally redacted. `style` remains excluded from attribute-value masking because rrweb has a separate style-mutation path.

`parsePrivacyConfig(unknown, validateSelector)` rejects malformed known fields and invalid selectors, ignores unknown fields for forward compatibility, and returns a copy. It requires a CSS validator from the target DOM. Callers must handle rejection before starting capture. Missing remote config and recorder restart behavior remain host responsibilities.

Browser keeps its existing warning-and-drop handling for invalid selectors through an explicit validation callback. It does not silently switch to the new strict remote-policy boundary in this PR.

### Observable browser behavior change: data attributes

These work without initialization options:

```html
<section data-amp-block>Excluded from replay</section>
<p data-amp-mask>Masked text</p>
<input data-amp-unmask />
```

- Existing `amp-block`, `amp-mask`, and `amp-unmask` classes continue to work.
- Attribute **presence** activates the annotation: `data-amp-unmask="false"` still unmasks.
- Mask annotations on an element or ancestor take precedence over unmask annotations. Unmasking does not override a blocked subtree.
- Sites already containing these attributes will now have them interpreted by browser replay. In particular, existing `data-amp-unmask` markup can make content visible under a masking default. This behavior change should be included in release notes.

Mask/unmask checks use combined class/attribute selectors rather than a second `closest` traversal. rrweb's block/text selectors also include the attribute forms. There is additional selector-matching work; this PR does **not** establish a performance result. We are proposing unconditional support and will optimize if measurements identify a problem.

### Future mobile integration — not included

1. Confirm delivery of browser `sr_privacy_config` in the native remote-config response.
2. Bundle this package into the injected WebView script and pass the remote policy through the native bridge; no mobile local/remote merger initially.
3. Apply/validate policy before the first snapshot and preserve it across document navigation and reinjection.
4. Supply the current document URL for URL rules, including SPA navigation, and define safe policy-replacement/restart behavior with a fresh baseline. Previously emitted content cannot be retroactively redacted.
5. Align mobile recorder compatibility with the browser-led rrweb implementation and compare capture output in WKWebView. This PR does not change either platform's recorder version.

### Review requested

- Is this package boundary appropriate for browser and native WebView consumers?
- Approve always-on class/attribute equivalence, including presence semantics and mask-over-unmask precedence.
- Review the strict remote-only parsing contract while browser retains existing invalid-selector behavior.
- Confirm package naming/publication and the browser-led recorder compatibility approach before mobile adopts it.

### Validation

Passed:

- 484 focused tests across shared privacy, browser helpers, config resolution, recorder behavior, and actual rrweb snapshot/mutation integration (coverage enforcement disabled for the focused run).
- Shared package ESM/CommonJS compilation and browser TypeScript check.
- Focused ESLint and `git diff --check`.
- `pnpm docs:check`.
- `pnpm lint:deps`.
- A Rollup dependency inspection of the shared package included only its modules and `tslib`, with no browser SDK, recorder, storage, or transport dependency.

**Not merge-ready / outstanding validation:**

- The shared package's default test command runs all 13 current tests successfully but **fails the repository's 100% coverage gate**: 85.63% statements, 76.61% branches, 74.07% functions, and 85.5% lines. Additional shared-package coverage is required before merge; existing browser tests still exercise much of the extracted logic through re-exports but do not count toward this package's isolated coverage report.
- `pnpm test:examples` fails before executing tests because `jest.setup.examples.js` is missing from the checkout.
- The prior full workspace build emitted browser bundles but did not exit after Rollup output and was terminated. Focused package compilation/type-check succeeded; a clean full workspace build is not claimed.
- No performance benchmark, WKWebView runtime test, remote-config delivery verification, or mobile recorder compatibility validation has been performed.

- Full `pnpm test` and `pnpm lint` were also attempted; each exceeded a 180-second local budget and was stopped. They are not reported as passing. CI still needs to establish the complete workspace result.

### Checklist

- [x] PR title follows the conventional-commit format.
- Breaking API change: no intentional removal of customer-facing browser configuration. There is an intentional **capture behavior change** for existing `data-amp-*` markup, described above.
- [ ] Complete shared-package coverage and full workspace validation before marking ready for merge.
- [ ] SR team approval of the shared package boundary and always-on data-attribute semantics.


daniel-graham-amplitude:
@codex review

